PR #3308, opened 2022-07-14
Add more styled examples

devongovett:
Adds three new styled examples:

* A select with macOS style popup positioning, where the popover is positioned such that the selected option is placed directly above the text in the button.
* An alert dialog, animated with react-transition-group.
* A simple menu.


snowystinger (approved):
Hah, that positioning one is cool

PR #3326, opened 2022-07-21
Export types from all packages

### LINKED ISSUE #3087: Why are interfaces e.g. `TableProps`, part of the top-level external API, not exported?
Trying to define a component with an external API e.g.:

```
import * as React from 'react';
import { useTable } from 'react-aria';
import { useTableState } from 'react-stately';

export type ContentTableProps = TableProps & {

}

export const ContentTable: React.FC<ContentTableProps> = props => {
  return <div>

  </div>;
};
```

The necessary `TableProps` here is used in the parameters of the `useTable` signature, so it should be exported so I can define a new type based on it. Or just so I can create a type-safe object of it. Is there a reason it's not exported?

`TableProps` here is just an example, from what I can see there are lots of exported APIs which are defined against non-exported interfaces.

FWIW Microsoft makes a TypeScript analysis tool called API Extractor; it's Number 2 Headline Feature is designed to help detect and correct this situation: https://api-extractor.com/
> Missing exports: Suppose the awesome-widgets package exports an API function AwesomeButton.draw() that requires a parameter of type DrawStyle, but you forgot to export this enum. Things seem fine at first, but when a developer tries to call that function, they discover that there’s no way to specify the DrawStyle. How to avoid these oversights?

### LINKED ISSUE #3240: react-aria and react-stately should re-export all types
# 🙋 Feature Request

If you use `react-aria` and/or `react-stately` with TypeScript you still have to manage other dependencies  to be able to access all needed types (e.g. `import type { AriaListBoxOptions } from '@react-aria/listbox';` or `import type { Node } from '@react-types/shared';`). For a better developer experience both libs should re-export all the types which are used.

## 🤔 Expected Behavior

```
import type { AriaListBoxOptions , Node } from 'react-aria';
```

## 😯 Current Behavior

```
import type { AriaListBoxOptions } from '@react-aria/listbox';
import type { Node } from '@react-types/shared';
```

## 🔦 Context

I use pnpm and want to enforce people to only import direct dependencies (and not dependencies of dependencies). That's why I need to manage libs like `@react-aria/listbox` _on top of_ `react-aria`.

devongovett:
Closes #3240. Closes #3087.

We were inconsistent about which types were exported from React Aria and React Stately packages. Mostly it was just if we needed it elsewhere, then it was exported. In addition, developers often needed to install `@react-types` packages in addition to the code packages to get the corresponding type definitions, leading to annoyance.

This ensures that all types that are the inputs or outputs of hooks or components (e.g. prop and state objects) are exported. In addition, it re-exports types that come from `@react-types` packages in the corresponding `@react-stately`, `@react-aria`, and `@react-spectrum` packages. Finally, it re-exports all public types from the corresponding mono packages. This means developers usually only need to install one package to get both the code and types they need.

Where they were not already exported (and therefore public API), and the naming was inconsistent, I also tried to address this before they become public. We will need to be more mindful of naming conventions for types going forward.

This was mostly automated using a script, but with manual review of all changes.


snowystinger (approved):
Nothing jumps out at me as wrong, I see the backwards-compat one for the TimePicker/TimeField.
I'm assuming nothing changed in the @types because that's all already public, and what we want to export already is.

reidbarber (approved):

packages/@react-aria/grid/src/index.ts
```diff
@@ -23,3 +18,12 @@ export {useGridCell} from './useGridCell';
 export {useGridSelectionCheckbox} from './useGridSelectionCheckbox';
 export {useHighlightSelectionDescription} from './useHighlightSelectionDescription';
 export {useGridSelectionAnnouncement} from './useGridSelectionAnnouncement';
+
+export type {GridProps, GridAria} from './useGrid';
+export type {GridCellAria, GridCellProps} from './useGridCell';
+export type {GridRowGroupAria} from './useGridRowGroup';
+export type {GridRowProps, GridRowAria} from './useGridRow';
```
LFDanLu, line 25: I noticed that the aria component hook options now follow "AriaCOMPONENTOptions" or "AriaCOMPONENTProps". At the risk of being overly pedantic, should the grid hook props follow this same convention?

Just trying to clarify the naming convention for the future

LFDanLu (approved):
LGTM, one question about the naming convention

packages/@react-aria/grid/src/index.ts
devongovett: Yes ideally, but, I couldn't change some of them that were already exported...

I think the convention is:

* Aria___Props if it's props. Most of these come from `@react-types`
* Aria___Options if we extended the props to add additional options only for the hook (e.g. additional refs)
* ___Aria for the return types

LFDanLu: Ah right, GridProps were already exported, I'm fine keeping the other grid props naming in line with that then. Thanks for the clarification

PR #3327, opened 2022-07-21
Use mono-packages for docs examples

devongovett:
This updates the React Aria and React Stately examples to use the mono-packages rather than the individual packages, similar to how we have been doing it for React Spectrum. This simplifies the examples and reduces the number of packages people need to install.

Questions:

- How should we handle the `version` shown at the top of the page? In RSP, people have been confused since this doesn't match the mono package version.
- How should we expose what individual package a hook is part of?
- Why did we only do this in the production docs rather than PR/local builds before? For now I turned that off so I could test, but not sure what the reason was.

### DIFF
--- a/packages/dev/parcel-transformer-mdx-docs/MDXTransformer.js
+++ b/packages/dev/parcel-transformer-mdx-docs/MDXTransformer.js
@@ -407,32 +407,42 @@ function transformExample(node, preRelease) {
   /* Replace individual package imports in the code
    * with monorepo imports if building for production and not a pre-release
    */
-  if (process.env.DOCS_ENV === 'production' && !preRelease && node.value.includes('@react-spectrum')) {
-    let specifiers = [];
-    let last;
+  if (!preRelease && /@react-spectrum|@react-aria|@react-stately/.test(node.value)) {
+    let specifiers = {};
+    const recast = require('recast');
     const traverse = require('@babel/traverse').default;
     const {parse} = require('@babel/parser');
-    const generate = require('@babel/generator').default;
-    let ast = parse(node.value, {
-      sourceType: 'module',
-      plugins: ['jsx', 'typescript']
+    let ast = recast.parse(node.value, {
+      parser: {
+        parse() {
+          return parse(node.value, {
+            sourceType: 'module',
+            plugins: ['jsx', 'typescript'],
+            tokens: true
+          });
+        }
+      }
     });
 
     traverse(ast, {
       ImportDeclaration(path) {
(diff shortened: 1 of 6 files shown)


packages/dev/parcel-transformer-mdx-docs/MDXTransformer.js
```diff
@@ -441,26 +451,25 @@ function transformExample(node, preRelease) {
...
+
+              path.unshiftContainer('body', [decl]);
+            }
           }
         }
       }
     });
 
-    node.value = generate(ast).code.replace(/(<WRAPPER>(?:.|\n)*<\/WRAPPER>)/g, '\n($1)');
+    node.value = recast.print(ast, {objectCurlySpacing: false, quote: 'single'}).code.replace(/(<WRAPPER>(?:.|\n)*<\/WRAPPER>)/g, '\n($1)');
```
devongovett, line 472: Recast does a better job of code generation, because it preserves the original formatting for anything that wasn't modified.

snowystinger:
> Why did we only do this in the production docs rather than PR/local builds before? For now I turned that off so I could test, but not sure what the reason was

I think we did this so that we could develop docs for components/hooks that weren't released at all yet

devongovett:
hmm yeah, but there's a separate check for pre-releases also, which uses individual packages...

snowystinger:
I just saw that, maybe that was added later?

devongovett:
Ah yeah, I think you're right https://github.com/adobe/react-spectrum/pull/712/files#diff-8eac6389915347070f55b54438fc38057f8933d3d7d4ce54ce8f24e25aac995fR317

snowystinger:
is it enough to include both sets of information? maybe some better names than I used for some things, was trying to figure out short words/phrases for it.
Screen Shot 2022-07-21 at 11 13 49 AM

devongovett:
I think it might just be confusing. I think for now we can change the version to just be the mono package version. The "View package" card below links to the individual package, where you can easily find the version.

@dannify also wanted to change "version" to "added" or "since", to show the version a component was added to the mono package instead of the current version. Might be more useful. That will require some more manual work (to be done separately).

snowystinger:
> @dannify also wanted to change "version" to "added" or "since", to show the version a component was added to the mono package instead of the current version. Might be more useful. That will require some more manual work (to be done separately).

I think this would help, we get people confused about when things were added pretty consistently.

If we kept an old version of docs for older versions of the mono packages, I think we could also solve it that way

dannify (approved):

snowystinger (approved):

PR #3335, opened 2022-07-23
Fix re-export of useValueEffect

devongovett:
Works around a parcel issue until we upgrade. Also this extra file was kinda unnecessary anyway. 😄 

https://github.com/parcel-bundler/parcel/pull/8299


LFDanLu (approved):
verdaccio build works again

PR #3341, opened 2022-07-25
Fix menu trigger longpress

devongovett:
Fixes a regression found in #3264. MenuTrigger with menuTrigger=longpress would open both when long pressing and pressing normally. This was due to onPress being added in useOverlayTrigger. So now we override that at the useMenuTrigger level.


snowystinger (approved):

LFDanLu (approved):
LGTM, tested on desktop and mobile, confirmed that menu only opens on long press or option/alt + up/down arrow

PR #3361, opened 2022-07-29
Workaround for tree shaking issue with Parcel

devongovett:
The intl-messageformat dependency should be tree shaken unless `useMessageFormatter` is used, however, Parcel does not do this. This is because it doesn't track that because this function is unused, the imported dependency is also unused.

A workaround to this is to build `useMessageFormatter` into a separate file rather than bundling it with the rest of the package. Then, re-export that file along with the real entry. Then Parcel is able to see that nothing in that file is unused, and due to `"sideEffects": false` in the package.json, exclude it.

Unfortunately it's not terribly clean because this involves building that file manually using Babel. But I verified locally that it solves the problem.


packages/@react-aria/i18n/src/index.ts
```diff
@@ -11,14 +11,13 @@
...
 export {I18nProvider, useLocale} from './context';
-export {useMessageFormatter} from './useMessageFormatter';
 export {useLocalizedStringFormatter} from './useLocalizedStringFormatter';
 export {useDateFormatter} from './useDateFormatter';
 export {useNumberFormatter} from './useNumberFormatter';
 export {useCollator} from './useCollator';
 export {useFilter} from './useFilter';
 
-export type {FormatMessage} from './useMessageFormatter';
+export type {useMessageFormatter, FormatMessage} from './useMessageFormatter';
```
devongovett, line 20: By only exporting the type for `useMessageFormatter` here, it stays in the generated TS definitions, but is excluded from the Parcel-built JS.

devongovett:
Had to delete the docs for this to work, but we were gonna do that anyway.

reidbarber (approved):

dannify (approved):
📦

PR #3380, opened 2022-08-04
Stop propagation when pressing Enter on a native link

### LINKED ISSUE #2599: TableView: space key on link within cell scrolls the TableView down a page
# 🐛 Bug Report 

Within TableView, pressing the space key on a link within a cell scrolls the TableView down a page. After scrolling the row containing the link out of view, the keyboard focus gets shifted to the TableView container element itself, yet pressing space again to scroll down does nothing.

This behavior can be reproduced using the [TableView async loading example in Storybook](https://reactspectrum.blob.core.windows.net/reactspectrum/0940d13d2f58c191ef01def65b6745dbdc6824e2/storybook/index.html?path=/story/tableview--async-loading).

## 🤔 Expected Behavior

The Space key event handler within TableView should `stopPropagation` to prevent the browser's default behavior, to scroll whichever scrolling region comes first or the document body.

Links are typically activated using the Enter key, so I don't necessarily expect Space to activate the link. Depending on the selection mode, pressing Space with focus on a link within a cell should either select the row or do nothing.

## 😯 Current Behavior

Pressing the space key on a link within a cell scrolls the TableView down a page. After scrolling the row containing the link out of view, the keyboard focus gets shifted to the TableView container element itself, yet pressing space again to scroll down does nothing.

## 💁 Possible Solution

`preventDefault` but don't `stopPropagation` of keyboard event using space key on link.

## 🔦 Context

This bug was identified during and accessibility review for a control using the Inventory component within quarry.

## 🌍 Your Environment

| react-spectrum   | @react-spectrum/table@3.0.1
| Browser          | Chrome 96.0.4664.55 (Official Build) (x86_64)
| Operating System | macOS 11.6 (20G165)

## 🧢 Your Company/Team

 Adobe/Accessibility

devongovett:
Closes #2599. Closes #2619.

Slightly different version of the fix in #2619 (test and partial fix reused). This fixes scrolling and selection when pressing Space when focused on a link in a TableView, while letting Enter trigger the link natively. Two fixes:

1. Use `currentTarget` rather than `target` to check whether to handle a keyboard event. This ensures the check for link handling doesn't apply when the table row is handling an event that propagated up from a link (e.g. Space key). This is the same as in #2619.
2. Stop propagation on native HTML links during keydown, even when we don't otherwise handle the event, because the browser will handle it natively and we don't want TableView to also handle it. This is different from #2619 in that the link handles it rather than the table. I think we should stop propagation here because the event is handled (by the browser), so elements above it should not handle it as well.

## 📝 Test Instructions:
1. Open TableView async loading story.
2. Move focus to a link in the table.
3. Press Space key. Selection of the table row should be toggled, and it should not scroll.
4. Press Enter on the link. It should open in a new tab.

### DIFF
--- a/packages/@react-aria/interactions/src/usePress.ts
+++ b/packages/@react-aria/interactions/src/usePress.ts
@@ -228,7 +228,7 @@ export function usePress(props: PressHookProps): PressResult {
 
     let pressProps: DOMAttributes = {
       onKeyDown(e) {
-        if (isValidKeyboardEvent(e.nativeEvent) && e.currentTarget.contains(e.target as Element)) {
+        if (isValidKeyboardEvent(e.nativeEvent, e.currentTarget) && e.currentTarget.contains(e.target as Element)) {
           if (shouldPreventDefaultKeyboard(e.target as Element)) {
             e.preventDefault();
           }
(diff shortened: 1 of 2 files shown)


LFDanLu (approved):
LGTM

majornista (approved):

majornista:
Thanks for this. LGTM

dannify (approved):

PR #3390, opened 2022-08-08
StaticField

yihuiliao:
Closes 

## ✅ Pull Request Checklist:


packages/@react-aria/i18n/src/useListFormatter.tsx
```diff
@@ -0,0 +1,25 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+import {ListFormatOptions, ListFormatter} from '@internationalized/string';
+import {useLocale} from './context';
+import {useMemo} from 'react';
+
+/**
+ * Provides localized number formatting for the current locale. Automatically updates when the locale changes,
+ * and handles caching of the number formatter for performance.
```
LFDanLu, line 19: You'll need to update the description here since this references number formatting

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,180 @@
...
+            <FormattedDate value={value as DateTimeValue} formatOptions={formatOptions as Intl.DateTimeFormatOptions} />}
+
+          {value instanceof Date && 
+            <FormattedDate value={value as DateTimeValue} formatOptions={formatOptions as Intl.DateTimeFormatOptions} />}
+          
+          {value instanceof Time && 
+            <FormattedDate value={value as DateTimeValue} formatOptions={formatOptions as Intl.DateTimeFormatOptions} />}
+
+          {typeof value === 'string' && 
+            props.value}
```
LFDanLu, line 93: I know you are still working on some of the typescript at the moment, but I'm curious if we need all the explicit value casting here (e.g. `value as string[]`). I figured the conditionals would help typescript infer the true type of value. I think it will work after adding the extends stuff we talked about to StaticField<T>

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,180 @@
...
+            <FormattedNumber value={value as NumberValue} formatOptions={formatOptions as Intl.NumberFormatOptions} />}
+
+          {typeof value === 'object' && 'calendar' in (value as any) && 
+            <FormattedDate value={value as DateTimeValue} formatOptions={formatOptions as Intl.DateTimeFormatOptions} />}
+
+          {value instanceof Date && 
+            <FormattedDate value={value as DateTimeValue} formatOptions={formatOptions as Intl.DateTimeFormatOptions} />}
+          
+          {value instanceof Time && 
+            <FormattedDate value={value as DateTimeValue} formatOptions={formatOptions as Intl.DateTimeFormatOptions} />}
```
LFDanLu, line 90: whats the difference between these three? They result in the same FormattedDate component usage right?

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  if (value instanceof Time) {    
+    return <>{dateFormatter.format(convertValue(value).toDate(timeZone))}</>;
+  }
+
+  return <>{dateFormatter.format(value as Date)}</>;
+}
+
+function convertValue(value: TimeValue, date: DateValue = today(getLocalTimeZone())) {
```
LFDanLu, line 167: If we don't use the `date` parameter, then we can just get rid of it and initialize it inside the function instead

LFDanLu (commented):
Some initial comments, I'll review the behavior and labeling later since I know you are still working on that

packages/@react-spectrum/label/src/StaticField.tsx
snowystinger: I thought this too, but i haven't been able to get it to behave yet

packages/@react-spectrum/label/src/StaticField.tsx
yihuiliao: hmmm i think it's possible to combine them given the right conditional statement, but the thing with CalendarDate, CalendarDateTime, ZonedTimeDate, and Time is that they don't have one unique property that they all shared. for now, what i can come up with is `(typeof value === 'object' && ('calendar' in (value as any) || 'hour' in (value as any)) || (value instanceof Date)`. as for Date i wasn't too sure what to do about that so i just left it as` (value instanceof Date)`

packages/@react-spectrum/label/src/StaticField.tsx
yihuiliao: okay so i don't think i can get ride of things like value={value as string[]} without a typescript error but it is possible now to do `'calendar' in value` instead of `'calendar' in (value as any)` now

packages/@react-spectrum/label/test/StaticField.test.js
```diff
@@ -0,0 +1,255 @@
...
+  it('renders when no visible label is provided', function () {
+    let {getByTestId} = render(
+      <StaticField
+        data-testid="test-id"
+        value="test" />
+    );
+
+    let labelledField = getByTestId('test-id');
+    expect(labelledField).toBeInTheDocument();
+    expect(labelledField).not.toHaveAttribute('aria-describedby');
```
snowystinger, line 252: it would be good on each of these to also include the expected formatted text
```
expect(labelledField).toHaveTextContent('test');
```
note toHaveTextContent may not be a valid matcher, might need to do
```
expect(labelledField.textContent).toBe('test');
```

packages/@react-spectrum/label/chromatic/StaticField.chromatic.tsx
```diff
@@ -0,0 +1,88 @@
...
+import {StaticField} from '../';
+
+const meta: Meta<SpectrumStaticFieldProps<SpectrumStaticFieldTypes>> = {
+  title: 'StaticField',
+  component: StaticField
+};
+
+export default meta;
+
+const Template = (): Story<SpectrumStaticFieldProps<SpectrumStaticFieldTypes>> => (args) => (
```
snowystinger, line 26: if you have some time, it'd be great if we could write our stories in CSF 3.0, no worries if you don't have the time though
See https://github.com/adobe/react-spectrum/pull/3129/files for an example

packages/@react-spectrum/label/stories/StaticField.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+    width: {
+      control: {
+        type: 'radio',
+        options: [null, '300px', '600px']
+      }
+    }
+  }
+};
+
+storiesOf('StaticField', module)
```
snowystinger, line 54: these stories as well

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+  return (
+    <>{stringFormatter.format(props.value)}</>
+  );
+}
+
+function FormattedNumber<T extends NumberValue>(props: NumberProps<T>) {
+  let numberFormatter = useNumberFormatter(props.formatOptions); 
+  let value = props.value;
+
+  if (typeof value === 'object' && typeof (value as RangeValue<NumberValue>).start === 'number' && typeof (value as RangeValue<NumberValue>).end === 'number') {
```
devongovett, line 133: just checking if it's an object should be enough. I don't think you need to check whether start and end are numbers - that's implied from the type. If it's an object, we know it's an object containing numbers based on the `NumberValue` type definition.

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+    return <>{numberFormatter.formatRange(value.start, value.end)}</>;
+  }
+
+  return <>{numberFormatter.format(value as number)}</>;
```
devongovett, line 137: I don't think the type assertion should be necessary here

```suggestion
  return <>{numberFormatter.format(value)}</>;
```

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+import React, {RefObject} from 'react';
+import type {TimeValue} from '@react-types/datepicker';
+import {useDateFormatter, useListFormatter, useNumberFormatter} from '@react-aria/i18n';
+import {useStyleProps} from '@react-spectrum/utils';
+
+interface StaticFieldBaseProps extends DOMProps, StyleProps, Omit<SpectrumLabelableProps, 'necessityIndicator' | 'isRequired'>, AriaLabelingProps {}
+
+type NumberValue = number | RangeValue<number>;
+interface NumberProps<T extends NumberValue> {
+    value: T,
```
devongovett, line 28: fix indentation to 2 spaces on these types (and below)

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+        start = start.toDate(timeZone);
+        end = end.toDate(timeZone);
+      }
+    } 
+  }
+  
+  return <>{dateFormatter.formatRange(start, end)}</>;
+}
+
+function FormattedDate<T extends DateTimeValue>(props: DateProps<T>) {
```
devongovett, line 165: Is there a reason the Date and DateRange formatters are split into separate components? For numbers there is a single one that does both. Seems like there is a reasonable amount of the logic that could be shared. You could pull out some of the logic below into a reusable function that accepts a `DateTime` and returns a `Date`, and then reuse it for both the start and end of a range as well.

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+  let timeZone = dateFormatter.resolvedOptions().timeZone || getLocalTimeZone();
+  let start;
+  let end;
+
+  if (typeof value === 'object' && 'start' in value) {
+    start = value.start;
+    end = value.end;
+
+    if (!(start instanceof Date) && !(end instanceof Date)) {
+      if (start instanceof Time && end instanceof Time) {
```
devongovett, line 152: let's avoid instanceof checks on `Time` or other values from `@internationalized/date`. That way it avoids being included if unused. You could check if `'hour' in value` like you did in other spots.

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+        start = convertValue(start).toDate(timeZone);
+        end = convertValue(end).toDate(timeZone);
+      } else if (!(start instanceof Time) && !(end instanceof Time)) {
```
devongovett, line 155: Should these checks be more similar to the ones below for single date formatting (i.e. check if `timeZone` or `calendar` are present)?

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  if (value instanceof Time) {    
+    return <>{dateFormatter.format(convertValue(value).toDate(timeZone))}</>;
+  }
+
+  return <>{dateFormatter.format(value as Date)}</>;
+}
+
+function convertValue(value: TimeValue) {
```
devongovett, line 185: I think you only need this if the value is explicitly a `Time`, and not a `CalendarDateTime` or `ZonedDateTime`, which already have a `toDate` method. So the type here could be `Time`, and you could remove some of the logic below (e.g. `day in value`, and the check if it's null).

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+  
+  return <>{dateFormatter.formatRange(start, end)}</>;
+}
+
+function FormattedDate<T extends DateTimeValue>(props: DateProps<T>) {
+  let dateFormatter = useDateFormatter(props.formatOptions);
+  let value = props.value;
+  let timeZone = dateFormatter.resolvedOptions().timeZone || getLocalTimeZone();
+
+  if (typeof value === 'object' && 'timeZone' in value) {
```
devongovett, line 170: Date values are always objects so these checks aren't needed.

packages/@react-spectrum/label/src/StaticField.tsx
```diff
@@ -0,0 +1,200 @@
...
+}
+
+function FormattedDateRange<T extends DateTimeValue>(props: DateProps<T>) {
+  let dateFormatter = useDateFormatter(props.formatOptions);
+  let value = props.value;
+  let timeZone = dateFormatter.resolvedOptions().timeZone || getLocalTimeZone();
+  let start;
+  let end;
+
+  if (typeof value === 'object' && 'start' in value) {
```
devongovett, line 147: Date values are always objects so this check isn't needed.

packages/@react-spectrum/label/stories/StaticField.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+export let TimeType: StaticFieldStory = {
+  ...Default,
+  args: {...Default.args, value: new Time(9, 45), formatOptions: {timeStyle: 'short'}},
+  name: 'Time'
+};
+
+export let CalendarDateRange: StaticFieldStory = {
+  ...Default,
+  args: {...Default.args, value: {start: new CalendarDate(2019, 6, 5), end: new CalendarDate(2019, 7, 5)}, formatOptions: {dateStyle: 'medium'}},
```
devongovett, line 87: Might be an idea to eventually default the format options based on the value. For example if we have a `CalendarDate` then it doesn't make sense to include a time, whereas with a `CalendarDateTime` it would.

packages/@react-spectrum/label/stories/StaticField.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+  ...Default,
+  args: {...Default.args, value: 10},
+  name: 'Number'
+};
+
+export let NumberRange: StaticFieldStory = {
+  ...Default,
+  args: {...Default.args, value: {start: 10, end: 20}},
+  name: 'RangeValue<Number>'
+};
```
LFDanLu, line 125: For some reason this StaticField takes more horizontal room than it needs, something to note for a followup:
image

packages/@react-spectrum/label/test/StaticField.test.js
```diff
@@ -0,0 +1,273 @@
...
+import React from 'react';
+import {render} from '@react-spectrum/test-utils';
+import {StaticField} from '../src';
+
+describe('StaticField', function () {
+  it('renders correctly with string value', function () {
+    let {getByTestId} = render(
+      <StaticField
+        data-testid="test-id"
+        label="Field label"
```
LFDanLu, line 23: It would be good to have a test that checks that the label is actually rendered and that it is a span element.

packages/@react-spectrum/label/test/StaticField.test.js
```diff
@@ -0,0 +1,273 @@
...
+        value="test" />
+    );
+
+    let staticField = getByTestId('test-id');
+    expect(staticField).toBeInTheDocument();
+    expect(staticField).not.toHaveAttribute('aria-describedby');
```
LFDanLu, line 29: was this aria-describedby check a remnant from an older test? Just curious where it came from

packages/@react-spectrum/label/test/StaticField.test.js
```diff
@@ -0,0 +1,273 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {CalendarDate, CalendarDateTime, Time, ZonedDateTime} from '@internationalized/date';
+import React from 'react';
+import {render} from '@react-spectrum/test-utils';
+import {StaticField} from '../src';
+
+describe('StaticField', function () {
```
LFDanLu, line 18: It would be a good idea to include a test that checks that the user provided format options actually affect the rendered content

packages/@react-spectrum/label/test/StaticField.test.js
```diff
@@ -0,0 +1,273 @@
...
+    expect(staticField).not.toHaveAttribute('aria-describedby');
+    expect(staticField).toHaveTextContent('2/3/2020');
+  });
+
+  it('renders correctly with Date value', function () {
+    let {getByTestId} = render(
+      <StaticField
+        data-testid="test-id"
+        label="Field label"
+        value={new Date(2000, 5, 5)} />
```
LFDanLu, line 94: Lol, I always forget that the Date object is 0 indexed for the month

packages/@react-spectrum/label/test/StaticField.test.js
```diff
@@ -0,0 +1,273 @@
...
+    let {getByTestId} = render(
+      <StaticField
+        data-testid="test-id"
+        label="Field label"
+        ref={ref}
+        value="test" />
+    );
+
+    let staticField = getByTestId('test-id');
+    expect(ref.current).toBe(staticField);
```
LFDanLu, line 242: You may want to check that `ref.current` contains the StaticField label `span` or something to that effect just to make this test more robust.

packages/@react-spectrum/label/stories/StaticField.stories.tsx
snowystinger: It doesn't for me?

packages/@react-spectrum/label/test/StaticField.test.js
snowystinger: that's a bit of an implementation detail that it's a span, it's fine to check for the label in the dom, but lets not verify that it's a particular element

packages/@react-spectrum/label/test/StaticField.test.js
snowystinger: true, we do not really need this

packages/@react-spectrum/label/stories/StaticField.stories.tsx
yihuiliao: this seems to happen when `labelPosition = 'top'` and `labelAlign = 'end'`, but it doesn't seem to be a problem when labelPosition is not specified. it looks like `labelPosition = top` adds `spectrum-Field-positionTop` class on the outermost div which sets the width to some default size.

packages/@react-spectrum/label/package.json
```diff
@@ -43,7 +44,8 @@
     "@spectrum-icons/ui": "^3.3.1"
   },
   "devDependencies": {
-    "@adobe/spectrum-css-temp": "3.0.0-alpha.1"
+    "@adobe/spectrum-css-temp": "3.0.0-alpha.1",
+    "@react-types/datepicker": "^3.1.0"
```
snowystinger, line 48: unused? lets remove this

packages/@react-spectrum/label/test/StaticField.test.js
```diff
@@ -0,0 +1,328 @@
...
+      <StaticField
+        data-testid="test-id"
+        label="Field label"
+        value={new ZonedDateTime(2020, 3, 3, 'America/Los_Angeles', -28800000)}
+        formatOptions={{dateStyle: 'full', timeStyle: 'short'}} />
+    );
+
+    let staticField = getByTestId('test-id');
+    expect(staticField).toBeInTheDocument();
+    expect(staticField).toHaveTextContent('Tuesday, March 3, 2020 at 3:00 AM');
```
LFDanLu, line 136: Currently looking into why this fails on circleci, looks like the formatted date time is dependent on the local timezone of the user running the test which is what we expect since this is ZonedDateTime not CalendarDateTime. The assertion will have to be flexible to accommodate this behavior, thinking of how to do so short of grabbing the local time zone and doing a separate time format option to produce the same string

snowystinger: You can use https://jestjs.io/docs/expect#tomatchregexp--string to match the start of the string and ignore the time

LFDanLu: hm, can't think of an alternative than basically rehashing the logic present in StaticField to generate the expected result:
```
    let formatOptions = {dateStyle: 'full', timeStyle: 'short'};
    let zoned = new ZonedDateTime(2020, 3, 3, 'America/Los_Angeles', -28800000);
    let formatter = new DateFormatter('en-US', formatOptions);
    let expected =  formatter.format(zoned.toDate());
    let {getByTestId} = render(
      <StaticField
        data-testid="test-id"
        label="Field label"
        value={zoned}
        formatOptions={{dateStyle: 'full', timeStyle: 'short'}} />
    );

    let staticField = getByTestId('test-id');
    expect(staticField).toBeInTheDocument();
    expect(staticField).toHaveTextContent(expected);
```

LFDanLu (approved):
LGTM

devongovett (approved):
Nice work! 🥳

snowystinger (approved):
lgtm! congrats!

PR #3391, opened 2022-08-09
Drag and drop bug fixes

devongovett:
This fixes a few bugs found while writing drag and drop docs:

* When starting a drag with the keyboard, the nearest drop target (positionally) is focused by default rather than the first one on the page.
* When pressing a modifier key while dragging to change the drop operation, the cursor sometimes did not update correctly. Also, if `getDropOperation` returns an operation that isn't allowed, the drag is now canceled.
* Drop target state would sometimes get stuck because of the logic with counting the number of times onDragEnter and onDragLeave are called. Instead, we now use `relatedTarget` to determine if we're in a nested target, similar to how `useFocusWithin` is implemented.
* Moved `getDropOperationForPoint` into a `DragTargetDelegate` interface to make it reusable, and implemented a `ListDragTargetDelegate`. This uses the DOM for layout. A separate implementation is also provided in `ListLayout` for virtualizer uses. Also removes a hard coded condition in ListView for specific folder dnd behavior, and behaves better when only certain drop positions are supported.
* Added some states to the return types of hooks.

### DIFF
--- a/packages/@react-stately/layout/src/ListLayout.ts
+++ b/packages/@react-stately/layout/src/ListLayout.ts
@@ -11,7 +11,7 @@
  */
 
 import {Collection, DropTarget, DropTargetDelegate, KeyboardDelegate, Node} from '@react-types/shared';
-import {InvalidationContext, Layout, LayoutInfo, Rect, Size} from '@react-stately/virtualizer';
+import {InvalidationContext, Layout, LayoutInfo, Point, Rect, Size} from '@react-stately/virtualizer';
 import {Key} from 'react';
 // import { DragTarget, DropTarget, DropPosition } from '@react-types/shared';
 
@@ -489,11 +489,12 @@ export class ListLayout<T> extends Layout<Node<T>> implements KeyboardDelegate, 
     x += this.virtualizer.visibleRect.x;
     y += this.virtualizer.visibleRect.y;
 
-    let layoutInfo = this.getVisibleLayoutInfos(new Rect(x, y, 1, 1))[0];
-    if (!layoutInfo) {
+    let key = this.virtualizer.keyAtPoint(new Point(x, y));
+    if (key == null) {
       return;
     }
 
+    let layoutInfo = this.getLayoutInfo(key);
     let rect = layoutInfo.rect;
     let target: DropTarget = {
       type: 'item',
--- a/packages/@react-stately/layout/src/ListLayout.ts
+++ b/packages/@react-stately/layout/src/ListLayout.ts
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
-import {Collection, KeyboardDelegate, Node} from '@react-types/shared';
+import {Collection, DropTarget, DropTargetDelegate, KeyboardDelegate, Node} from '@react-types/shared';
 import {InvalidationContext, Layout, LayoutInfo, Rect, Size} from '@react-stately/virtualizer';
 import {Key} from 'react';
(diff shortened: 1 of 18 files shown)


packages/@react-stately/layout/src/ListLayout.ts
```diff
@@ -510,4 +484,38 @@ export class ListLayout<T> extends Layout<Node<T>> implements KeyboardDelegate {
     layoutInfo.transform = 'scale3d(0.8, 0.8, 0.8)';
     return layoutInfo;
   }
+
+  getDropTargetFromPoint(x: number, y: number, isValidDropTarget: (target: DropTarget) => boolean): DropTarget {
```
snowystinger, line 488: difference between this and listdroptargetdelegate? or reason for both?

devongovett: ListDropTargetDelegate uses the DOM for layout information, ListLayout uses the layout when virtualized

snowystinger:
I'm having trouble in Chrome with the reoder story, the second time I do a reorder, the indicator gets stuck.

packages/@react-stately/virtualizer/src/Virtualizer.ts
```diff
@@ -410,12 +410,15 @@ export class Virtualizer<T extends object, V, W> {
     let rect = new Rect(point.x, point.y, 1, 1);
     let layoutInfos = this.layout.getVisibleLayoutInfos(rect);
 
-    let layoutInfo = layoutInfos[0];
-    if (!layoutInfo) {
-      return null;
+    // Layout may return multiple layout infos in the case of
+    // persisted keys, so find the first one that actualy intersects.
```
snowystinger, line 414:
```suggestion
    // persisted keys, so find the first one that actually intersects.
```

snowystinger (commented):
Behaving well in the examples we currently have. With the persisted keys, though, is it possible that we'll get a drop target we can't see because it's been scrolled out of visible view and would be the closest? I can't tell if that's taken into account.

devongovett:
@snowystinger that should be handled here: https://github.com/adobe/react-spectrum/pull/3391/files#diff-f9405406c2dc605226f231a1af6ab2a05f64a6c2f83ab9eaf3f2b8c7dd8b7ae2R413-R418. It filters out any out of view items.

snowystinger (approved):
behaves well for me in Chrome

reidbarber (commented):
The 'Drag between lists (Root only)' story is really buggy in Safari when mouse dragging between lists. Not seeing this issue on main. Similar issue with the Reorder story.

snowystinger:
@reidbarber I see it too in Safari, i can get drop indicators to get stuck in the visible state, and they flicker a lot
Screen Shot 2022-08-10 at 2 32 09 PM

devongovett:
That was due to `relatedTarget` always being null in Safari: https://bugs.webkit.org/show_bug.cgi?id=66547.

But in Chrome, the dragleave event sometimes never fires for some elements. I tracked this to the drag indicators, which get removed from the DOM before dragleave fires. So I came up with the workaround above, which tracks the entered elements in a set and removes the ones that are no longer in the DOM on dragleave.

reidbarber (approved):
Looks good in Chrome, FF, and Safari! `getDropOperation` is still off in FF and Safari, but I think that is out of scope of this fix.

snowystinger (approved):
working well for me in chrome/ff/safari

PR #3415, opened 2022-08-19
Fix ComboBox left/right arrows so they move the input cursor when the menu is open

### LINKED ISSUE #3400: ComboBox: left/right arrow keys do not work after inputing a value while dropdown menu is down
# 🐛 Bug Report 

1. Have a ComboBox with options in it
2. Typing in a valid value in the input box of a ComboBox
3. The dropdown menu opens
4. When attempting to press the left or right arrow key on the keyboard, the text cursor does not move
5. While the menu is still open, if you move the text cursor anywhere on the string and press the left or right keys, it still does not move
6. when the menu closes (via selecting a option, inputing a invalid text, or manually closing it) the text cursor is movable again with the left and right keys

## 🤔 Expected Behavior

- user is able to move the text cursor left and right with the left and right keyboard keys

## 😯 Current Behavior

- user is **not** able to move the text cursor left and right with the left and right keyboard keys

## 💻 Code Sample

This issue was found in the Spectrum v3 storybook
https://react-spectrum.adobe.com/react-spectrum/ComboBox.html#example

https://user-images.githubusercontent.com/12926470/184436517-c51fefe6-0069-4519-be5b-a45ea6d92a62.mov

LFDanLu:
Closes #3400

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Go to any combobox story and do the following:
1. Type in something that will open the menu
2. Attempt to move the cursor left/right via the arrow keys. You should be able to now

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-stately/layout/src/ListLayout.ts
+++ b/packages/@react-stately/layout/src/ListLayout.ts
@@ -377,14 +377,6 @@ export class ListLayout<T> extends Layout<Node<T>> implements KeyboardDelegate, 
     }
   }
 
-  getKeyLeftOf(key: Key): Key {
-    return key;
-  }
-
-  getKeyRightOf(key: Key): Key {
-    return key;
-  }
-
   getKeyPageAbove(key: Key) {
     let layoutInfo = this.getLayoutInfo(key);
 


packages/@react-stately/layout/src/ListLayout.ts
```diff
@@ -377,14 +377,6 @@ export class ListLayout<T> extends Layout<Node<T>> implements KeyboardDelegate,
...
   }
 
-  getKeyLeftOf(key: Key): Key {
-    return key;
-  }
-
-  getKeyRightOf(key: Key): Key {
-    return key;
-  }
-
```
LFDanLu, line 387: Unnecessary now that useGridListItem handles left/right arrow keys presses in ListView rather than using useGridCell. Normal lists that use ListLayout won't have elements within their rows so they won't even use useGridCell

Would've written a test but didn't get accurate `selectionStart` returned by onKeyUp when triggering arrowleft/right. A updated version of user test library should fix this theoretically but that will require a greater refactor of our tests so I'll handle that later

snowystinger (approved):
Works as described in Chrome/Safari/Firefox

reidbarber (approved):
LGTM

PR #3421, opened 2022-08-19
Support onAction and drag and drop in listbox

devongovett:
While making the docs examples for drag and drop, I realized it wouldn't work with listbox because an explicit focusable drag affordance was required to activate accessible drag and drop mode and listbox options don't support focusable children. This is because the interactions conflict with selection and other actions on the items themselves. It would be useful to support drag and drop without a separate drag button in these cases.

This is accomplished by using Enter to trigger dragging, while reserving Space for selection. If row actions are also present, then it becomes Alt + Enter. For touch screen readers, a long press is used to trigger dragging (this actually fires the native onDragStart event). When actions are present, the first long press enters selection mode as before, and long pressing again drags (same behavior as without a screen reader on touch).

A new `hasDragButton` option is added to useDrag to opt-into a separate drag button as before, which keeps the interactions off the main draggable element and only activates accessible dnd through the drag button. This is used in ListView.


packages/@react-aria/utils/src/useId.ts
```diff
@@ -89,7 +89,7 @@ export function useSlotId(depArray: ReadonlyArray<any> = []): string {
     setResolvedId(function *() {
       yield id;
 
-      yield document.getElementById(id) ? id : null;
+      yield document.getElementById(id) ? id : undefined;
```
devongovett, line 92: These changes from null to undefined are to better accommodate mergeProps, which treats undefined differently from null. If null, this will override the previous value, whereas undefined will not.

packages/@react-aria/listbox/src/useOption.ts
```diff
@@ -127,15 +121,16 @@ export function useOption<T>(props: AriaOptionProps, state: ListState<T>, ref: R
...
   }
 
-  let {itemProps, isPressed} = useSelectableItem({
+  let {itemProps, isPressed, hasAction, allowsSelection} = useSelectableItem({
     selectionManager: state.selectionManager,
     key,
     ref,
     shouldSelectOnPressUp,
-    allowsDifferentPressOrigin: shouldSelectOnPressUp,
+    allowsDifferentPressOrigin: shouldSelectOnPressUp && shouldFocusOnHover,
```
devongovett, line 129: standalone list boxes should not allow different press origin, only ones within overlays like picker, combobox, etc.

packages/@react-aria/dnd/src/useDraggableItem.ts
```diff
@@ -48,19 +80,63 @@ export function useDraggableItem(props: DraggableItemProps, state: DraggableColl
...
   }
 
   return {
-    dragProps,
+    dragProps: isDisabled ? {} : dragProps,
     dragButtonProps: {
       ...dragButtonProps,
-      'aria-label': message
+      isDisabled,
+      'aria-label': dragButtonLabel
```
LFDanLu, line 139: Won't disabled items erroneously get the `dragItem`/`dragSelectedItems` messaging here?

LFDanLu (commented):
Tested on iPad, Android, and desktop browsers. New announcements and long press to start dragging seem to work well. Future thinking for when we add DnD to ListBox: we may want to still add a visual indicator to a ListBox item to indicate that it is draggable, but maybe something different than the drag indicator handle to distinguish it from things like ListView? Just worried about keyboard DnD discoverability.

I did notice that the default Drag and Drop story was a bit broken when using TalkBack/VoiceOver. In TalkBack, long pressing didn't really seem to fully shift the story into a dragging state, couldn't perform a drop but re-focusing the draggable button announced "double tap to cancel drag". For VoiceOver, I was able to get into a dragging state, but couldn't get to the last drop target.

packages/@react-aria/dnd/src/useDraggableItem.ts
LFDanLu: Ah, I guess the user would/should not render the drag handle for the disabled drag item

packages/@react-aria/dnd/src/useDrag.ts
```diff
@@ -195,12 +210,67 @@ export function useDrag(options: DragOptions): DragResult {
...
+    // If there's no separate button to trigger accessible drag and drop mode,
+    // then add event handlers to the draggable element itself to start dragging.
+    // For keyboard, we use the Enter key in a capturing listener to prevent other
+    // events such as selection from also occurring. We attempt to infer whether a
+    // pointer event (e.g. long press) came from a touch screen reader, and then initiate
+    // dragging in the native onDragStart listener above.
+
+    interactions = {
+      ...descriptionProps,
+      onPointerDown({nativeEvent: e}) {
```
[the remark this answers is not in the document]
devongovett: By default, unless you pass `hasDragButton` and merge the resulting `dragButtonProps` through `useButton` onto the element. Then it would just be click.

packages/@react-aria/dnd/src/useDraggableItem.ts
devongovett: And even so, it would read as disabled so I think it's ok

LFDanLu (approved):
Approving so we can get this in for testing, would be good to see if other see the behavior described [here](https://github.com/adobe/react-spectrum/pull/3421#pullrequestreview-1087512017) on VoiceOver/Talkback

devongovett:
Yeah it seems VO on iOS has a bug where setting aria-hidden on an option in a listbox doesn't cause the a11y tree to be updated properly so it thinks the items are still there when you swipe but they aren't. If you tap outside the list and then go back in it starts working. I'm not exactly sure if we can work around this or not but we can look into it separately.

As for TalkBack, it seems to work fine on my device so not sure.

dannify (approved):
Slightly concerned about the change from null to undefined for description and useId but it's probably ok.

PR #3427, opened 2022-08-22
Scale illustrations to fill width

reidbarber:
Illustrations should scale to fit the width of their slot. 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Check [ListView -> Thumbnail](https://reactspectrum.blob.core.windows.net/reactspectrum/584fdc61daa131747ee8def12a5413b2345d2c4c/storybook/index.html?path=/story/listview--thumbnails&providerSwitcher-toastPosition=bottom) story at all scales

Illustration should fill width of it's container on all scales.

### DIFF
--- a/packages/@react-spectrum/icon/src/Illustration.tsx
+++ b/packages/@react-spectrum/icon/src/Illustration.tsx
@@ -57,6 +57,9 @@ export function Illustration(props: IllustrationProps) {
     ariaHidden = undefined;
   }
 
+  let width = children.props.width || '32';
+  let height = children.props.height || '32';
+
   return React.cloneElement(children, {
     ...filterDOMProps(otherProps),
     ...styleProps,
@@ -64,6 +67,7 @@ export function Illustration(props: IllustrationProps) {
     'aria-label': ariaLabel,
     'aria-labelledby': ariaLabelledby,
     'aria-hidden': ariaHidden,
-    role: hasLabel ? 'img' : undefined
+    role: hasLabel ? 'img' : undefined,
+    viewBox: `0 0 ${width} ${height}`
   });
 }
--- a/packages/@react-spectrum/icon/src/Illustration.tsx
+++ b/packages/@react-spectrum/icon/src/Illustration.tsx
@@ -57,9 +57,6 @@ export function Illustration(props: IllustrationProps) {
     ariaHidden = undefined;
   }
 
-  let width = children.props.width || '32';
-  let height = children.props.height || '32';
-
   return React.cloneElement(children, {
     ...filterDOMProps(otherProps),
     ...styleProps,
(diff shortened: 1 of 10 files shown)


packages/@react-spectrum/list/src/styles.css
```diff
@@ -337,7 +337,7 @@
     align-items: center;
     justify-content: center;
 
-    > img, > svg {
```
reidbarber, line 340: this wasn't being used because the svg isn't a child of `--thumbnail`, it has the class itself

packages/@react-spectrum/icon/src/Illustration.tsx
```diff
@@ -57,13 +57,17 @@ export function Illustration(props: IllustrationProps) {
...
   return React.cloneElement(children, {
     ...filterDOMProps(otherProps),
     ...styleProps,
     focusable: 'false',
     'aria-label': ariaLabel,
     'aria-labelledby': ariaLabelledby,
     'aria-hidden': ariaHidden,
-    role: hasLabel ? 'img' : undefined
+    role: hasLabel ? 'img' : undefined,
+    viewBox: `0 0 ${width} ${height}`
```
reidbarber, line 71: I believe svgs need a viewBox to fill the width of their parent container. This is the only way I could figure out how to get it working correctly without breaking existing illustrations (IllustratedMessage).

devongovett: This technically means all of the illustrations now scale, even within an illustrated message. Is that what we want or do we want to keep this more limited to File/Folder for now? If so, you could just apply the width/height/viewBox attributes within those specific SVGs.

reidbarber: I think it may be safer to set height/width/viewBox on just those for now, like you said. Let's go that route.

devongovett (dismissed):

packages/@spectrum-icons/illustrations/src/File.tsx
```diff
@@ -1,10 +1,11 @@
-import {Illustration, IllustrationPropsWithoutChildren} from "@react-spectrum/icon";
-import React from "react";
+import {Illustration, IllustrationPropsWithoutChildren} from '@react-spectrum/icon';
+import React from 'react';
 
 export default function File(props: IllustrationPropsWithoutChildren) {
+  console.log(props);
```
LFDanLu, line 5: stray console log

reidbarber: ahh good catch

snowystinger (approved):

devongovett (approved):

PR #3446, opened 2022-08-25
Fix usePress with HTML input elements

devongovett:
This makes usePress work better with HTML `<input>` elements, which have default keyboard behavior. For example, checkboxes toggle when the space key is pressed. Previously this was [somewhat broken](https://codesandbox.io/s/musing-architecture-r1v5dx?file=/src/App.js), in that onPressStart would only be triggered on key up due to the browser firing a virtual onClick event.

In #3380, I changed usePress to look at the currentTarget rather than target when determining whether to handle an event. This caused TableView's select all checkbox to stop working, because the event would bubble to the cell and get handled there (with preventDefault).

Now we allow the space key for checkboxes and other input elements to trigger onPressStart on key down, but do not call `preventDefault`, allowing the browser to toggle the checkbox.

### DIFF
--- a/packages/@react-aria/interactions/src/usePress.ts
+++ b/packages/@react-aria/interactions/src/usePress.ts
@@ -229,7 +229,7 @@ export function usePress(props: PressHookProps): PressResult {
     let pressProps: DOMAttributes = {
       onKeyDown(e) {
         if (isValidKeyboardEvent(e.nativeEvent, e.currentTarget) && e.currentTarget.contains(e.target as Element)) {
-          if (shouldPreventDefaultKeyboard(e.target as Element)) {
+          if (shouldPreventDefaultKeyboard(e.target as Element, e.key)) {
             e.preventDefault();
           }
           e.stopPropagation();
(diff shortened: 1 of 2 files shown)


devongovett:
One slight change is that the Enter key on checkboxes will now bubble since the checkbox doesn't handle it, so table row selection will toggle in that case as well. I think this is ok?

packages/@react-aria/interactions/src/usePress.ts
```diff
@@ -774,8 +773,35 @@ function shouldPreventDefault(target: Element) {
...
+  'checkbox',
+  'radio',
+  'range',
+  'color',
+  'file',
+  'image',
+  'button',
+  'submit',
+  'reset'
+]);
```
LFDanLu, line 798: How was this list generated? I assume by going through a list of input types and adding any that wasn't a text input?

LFDanLu (approved):
Did some basic testing, confirmed table select all checkbox is working now. Other components seem to work still (form submit/reset buttons, checkboxes, etc), deeper sweep to come tmrw

packages/@react-aria/interactions/src/usePress.ts
devongovett: yeah https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

reidbarber (approved):

PR #3479, opened 2022-09-02
Handle drag and drop modifier keys more consistently across browsers

devongovett:
Browsers are very inconsistent about how they handle modifier keys during drag and drop. Only Chrome on macOS works fully as expected. Holding down Alt or Control allows switching between link, copy, and move operations, and it updates `effectAllowed` accordingly. Safari also updates `effectAllowed` when pressing a modifier key, but doesn't propagate the value set during the `dragstart` event and always sets it to `copyMove` in subsequent events. Firefox on macOS, and all Windows browsers don't update on modifier key presses at all. See code comment for full explanation and links to browser bugs.

The solution is to do some filtering ourselves based on modifier key presses, and also propagate the allowed operations from dragstart to other events ourselves when needed. This makes things much more consistent. The only cases that don't fully work are `link` operations in Safari and Android Chrome, but hopefully this is ok.


reidbarber (approved):
Working as expected in Chrome/FF Mac and Chrome/FF/Edge Windows. Link not working in Safari as mentioned in PR

LFDanLu (approved):
LGTM, tested on Chrome, FF, Safari, and Android, will do more in depth tests tmrw on Windows/other browsers. Looking forward to merging this with the other global dnd state stuff so Android DnD fully works with drag end as well

Issue #3047, opened 2022-04-19
Interactive elements hidden by the VisuallyHidden component don't receive Talkback virtual clicks on certain devices


LFDanLu:
# 🐛 Bug Report 

Elements like a Menu's DismissButton and the ListView's drag handle don't receive a click event when focused and triggered by Android Talkback. This seems to be device specific since it works on my Samsung Galaxy S9+ but not on other Android devices that have the same Talkback and Chrome version (e.g. Android 7.1.1 on HTC Nexus 9 using Talkback 12.1 with Chrome 100.0.4896.88). The root cause seems to be due to `clip-path` and/or the size of the element itself.

Base reproduction with a possible fix: https://codepen.io/majornista/pen/Exorwzq?editors=1111
bug filed: https://issuetracker.google.com/issues/229646660

## 🤔 Expected Behavior

Talkback should be able to trigger a click on a visually hidden element

## 😯 Current Behavior

Talkback isn't able to trigger a click on a visually hidden element on certain devices

## 💁 Possible Solution

Increasing the height/width of the visually hidden element seems to help, see code pen 

## 🧢 Your Company/Team

RSP

PR #3491, opened 2022-09-08
Fix dragging in Android Talkback

devongovett:
TalkBack fires pointer events on the center of the `currentTarget`, not `target`. In the case of the default dnd story, there is a nested span.

### DIFF
--- a/packages/@react-aria/dnd/src/useDrag.ts
+++ b/packages/@react-aria/dnd/src/useDrag.ts
@@ -237,17 +237,19 @@ export function useDrag(options: DragOptions): DragResult {
 
     interactions = {
       ...descriptionProps,
-      onPointerDown({nativeEvent: e}) {
+      onPointerDown(e) {
         // Try to detect virtual drags.
         if (e.width < 1 && e.height < 1) {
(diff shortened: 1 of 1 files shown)


LFDanLu (approved):
LGTM, confirmed the default drag and drop story now works in TalkBack now. The other stories seem to still work as well.

snowystinger (approved):

majornista (commented):
Is this meant to fix #3047, or is it something else? Testing on my Android devices, the default example seems to work, but examples with collections and or the list view do not work as expected. 

With collections drag and drop examples, while I seem to be able to start dragging and navigate to various hidden drop targets in the listbox, when I double tap to drop, the item being dragged disappears, but it does not show up within the listbox containing the drop target.

With the ListView Drag and Drop examples, when I try to double tap to start dragging, the ListViewItem gets selected, but dragging does not start. This is the same behavior described in #3047.

devongovett:
No, unrelated to #3047. This was to fix the default story, which were broken by https://github.com/adobe/react-spectrum/pull/3421.

PR #3492, opened 2022-09-08
Hide dragging ListView row from Talkback during virtual drag

LFDanLu:
If a listview was only draggable, the dragged row element and drag handle was focusable via TalkBack since we only removed the tabindexes from the rows if the list was droppable. Now we do the same if the list is draggable

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Go to the "drag out of list" story in ListView drag and drop.
2. Begin a drag via TalkBack and ensure that the row containing the dragged row can't be focused during the drag operation

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-spectrum/list/src/ListViewItem.tsx
+++ b/packages/@react-spectrum/list/src/ListViewItem.tsx
@@ -129,6 +129,7 @@ export function ListViewItem<T>(props: ListViewItemProps<T>) {
   let {visuallyHiddenProps} = useVisuallyHidden();
 
   let dropProps = isDroppable ? droppableItem?.dropProps : {'aria-hidden': droppableItem?.dropProps['aria-hidden']};
+  let isVirtualDragging = dropHooks?.isVirtualDragging() || dragHooks?.isVirtualDragging();
   const mergedProps = mergeProps(
     rowProps,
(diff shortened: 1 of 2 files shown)


packages/@react-spectrum/list/src/ListViewItem.tsx
```diff
@@ -129,6 +129,7 @@ export function ListViewItem<T>(props: ListViewItemProps<T>) {
   let {visuallyHiddenProps} = useVisuallyHidden();
 
   let dropProps = isDroppable ? droppableItem?.dropProps : {'aria-hidden': droppableItem?.dropProps['aria-hidden']};
+  let isVirtualDragging = dropHooks?.isVirtualDragging() || dragHooks?.isVirtualDragging();
```
devongovett, line 132: So I guess once your other PR goes in, this will end up being merged anyway 😂

devongovett (approved):

packages/@react-spectrum/list/src/ListViewItem.tsx
LFDanLu: yep :)

reidbarber (approved):

PR #3495, opened 2022-09-08
Update chromatic button stories to have wrapping Text elements

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Look at chromatic

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-spectrum/button/chromatic/Button.chromatic.tsx
+++ b/packages/@react-spectrum/button/chromatic/Button.chromatic.tsx
@@ -58,17 +58,17 @@ storiesOf('Button', module)
       <Flex gap="size-200">
         <Button
           variant="primary">
-          {0} Dogs
+          <Text>{0} Dogs</Text>
         </Button>
         <Button
           isDisabled
           variant="primary">
-          {0} Dogs
+          <Text>{0} Dogs</Text>
         </Button>
         <Button
           isQuiet
           variant="primary">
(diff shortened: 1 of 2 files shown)


packages/@react-spectrum/button/chromatic/Button.chromatic.tsx
```diff
@@ -58,17 +58,17 @@ storiesOf('Button', module)
       <Flex gap="size-200">
         <Button
           variant="primary">
-          {0} Dogs
+          <Text>{0} Dogs</Text>
```
LFDanLu, line 61: Note: people who did something similar in their Button components will experience the unset `min-width` applied by the new `iconOnly` class. I figure this is ok since correct usage of Button is to wrap the contents in a `Text` element if it isn't a plain string

devongovett (approved):

devongovett:
May not be needed with this change: https://github.com/adobe/react-spectrum/pull/3499#discussion_r967368874

packages/@react-spectrum/button/chromatic/Button.chromatic.tsx
jluyau: I added an extra conditional for iconOnly styling in #3503 so that it doesn't break these implementations

LFDanLu:
Ah, yep that fix should suffice as well

PR #3497, opened 2022-09-09
Fix sticky drop indicator when exiting list

devongovett:
This fixes the flaky drop indicator behavior where sometimes it sticks around even when exiting a list. This was due to drag events firing faster than react re-renders, thereby breaking our equality check. Need to store the target in a ref in addition to state so we always have the updated one.


reidbarber (approved):
LGTM, tested in Chrome/FF/Safari and not able to reproduce issue

ktabors (approved):
LGTM, the cleanup of the drop spot is much quicker. I tried for a while and couldn't reproduce it.

snowystinger (approved):

PR #3499, opened 2022-09-09
Add icon only docs for Button

jluyau:
Closes 

Add icon-only entry for Button in docs 

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/button/docs/Button.mdx
+++ b/packages/@react-spectrum/button/docs/Button.mdx
@@ -18,6 +18,7 @@ import packageData from '@react-spectrum/button/package.json';
 import {Button} from '@react-spectrum/button';
 import Bell from '@spectrum-icons/workflow/Bell';
 import {View} from '@react-spectrum/view';
+import {Flex} from '@react-spectrum/layout';
 ```
 
 ---
@@ -160,4 +161,15 @@ function Example() {
 
 ```tsx example
 <Button variant="cta" isDisabled>Save</Button>
+```
+
+### Icon only
+[View guidelines](https://spectrum.adobe.com/page/button/#Label-and-icon)
+
+```tsx example
+<Flex direction="row" gap={8}>
+  <Button variant="cta"><Bell /></Button>
+  <Button variant="primary"><Bell /></Button>
+  <Button variant="secondary"><Bell /></Button>
+</Flex>
 ```
--- a/packages/@react-spectrum/button/docs/Button.mdx
+++ b/packages/@react-spectrum/button/docs/Button.mdx
@@ -168,8 +168,8 @@ function Example() {
 
 ```tsx example
 <Flex direction="row" gap={8}>
-  <Button variant="cta"><Bell /></Button>
-  <Button variant="primary"><Bell /></Button>
-  <Button variant="secondary"><Bell /></Button>
+  <Button variant="cta" aria-label="Ring for service"><Bell /></Button>
+  <Button variant="primary" aria-label="Ring for service"><Bell /></Button>
+  <Button variant="secondary" aria-label="Ring for service"><Bell /></Button>
 </Flex>
 ```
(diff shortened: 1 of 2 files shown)


packages/@react-spectrum/button/docs/Button.mdx
```diff
@@ -161,3 +162,14 @@ function Example() {
 ```tsx example
 <Button variant="cta" isDisabled>Save</Button>
 ```
+
+### Icon only
+[View guidelines](https://spectrum.adobe.com/page/button/#Label-and-icon)
+
+```tsx example
+<Flex direction="row" gap={8}>
+  <Button variant="cta"><Bell /></Button>
```
devongovett, line 171: The buttons need `aria-label` props.

devongovett:
Noticed that the padding is somehow off on the Events example now: https://reactspectrum.blob.core.windows.net/reactspectrum/eab25a55e7e4b7b5895d0670e92202c95d6b5057/docs/react-spectrum/Button.html#events

packages/@react-spectrum/button/src/Button.tsx
```diff
@@ -51,6 +51,7 @@ function Button<T extends ElementType = 'button'>(props: SpectrumButtonProps<T>,
   let {hoverProps, isHovered} = useHover({isDisabled});
   let {styleProps} = useStyleProps(otherProps);
   let hasLabel = useHasChild(`.${styles['spectrum-Button-label']}`, domRef);
+  let hasIcon = useHasChild(`.${styles['spectrum-Button-icon']}`, domRef);
```
jluyau, line 54: Add this conditional back in to make the icon-only scenario more specific because of the [Button Events example](https://react-spectrum.adobe.com/react-spectrum/Button.html#events).

In that case, React is splitting the children into a array -- [number, string], so the Button code will render that since it doesn't qualify `typeof children === 'string'`. Since it renders as is, the `.spectrum-Button-label` doesn't get applied, and with the previous code, it would apply the icon only styling, thus removing the min-width/padding.

devongovett: Seems fine. Unfortunately, you'll have to move this code change to a separate pr so we can merge it prior to release. The docs have to wait until after release.

ktabors (approved):
Confirming that some buttons look weird, but will be fixed once #3503 is merged.

Example of weird is the "Events" example button in the docs
image

ktabors (commented):
Do we need to be explicate about "If the label is hidden, an icon is required, and the label will appear in a tooltip." or add an example of this in the Tooltip (or another section) docs?

LFDanLu:
+1 to what @ktabors mentioned. I think a specific callout to the "If the label is hidden, an icon is required, and the label will appear in a tooltip" portion of the Spectrum guidelines is appropriate here in case people don't click on that guideline link. It would also be good to have the examples follow this practice as well.

devongovett:
> and the label will appear in a tooltip

this is not true of our implementation however.

LFDanLu:
right, we currently rely on the user to provide the tooltip... Is the tooltip mandatory for icon only buttons? If so, we could wrap it in a TooltipTrigger for the user but we would need them to provide a tooltip text value through a different prop. If it isn't mandatory but strongly recommended IMO we should have the example display that. Otherwise we can keep it as is

packages/@react-spectrum/button/docs/Button.mdx
majornista: We should also provide examples of buttons with both an icon and a label, where the icon has meaning that is separate from the label.

majornista: For example:

```html
<Button variant="primary">
  <Add  aria-label="Add" />
  <Text>Event</Text>
</Button>
```

or

```html
<Button variant="primary">
  <Text>Add</Text>
  <Comment  aria-label="Comment" />
</Button>

<Button variant="primary">
  <Text>Remove</Text>
  <Comment  aria-label="Comment" />
</Button>
```

dannify:
I wouldn't add Tooltip to Button. We could add an example though.

LFDanLu (approved):

PR #3504, opened 2022-09-09
Removing sticky from table checkboxes to prevent chrome crash

LFDanLu:
https://bugs.chromium.org/p/chromium/issues/detail?id=1359961

Closes 

## ✅ Pull Request Checklist:


packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -266,7 +268,7 @@ export class TableLayout<T> extends ListLayout<T> {
     let {height, isEstimated} = this.getEstimatedHeight(node, width, this.rowHeight, this.estimatedRowHeight);
     let rect = new Rect(x, y, width, height);
     let layoutInfo = new LayoutInfo(node.type, node.key, rect);
-    layoutInfo.isSticky = node.props?.isSelectionCell;
+    layoutInfo.isSticky = !this.disableSticky && node.props?.isSelectionCell;
```
LFDanLu, line 271: Note that the column header checkbox is still sticky, didn't seem to crash when testing so I opted to keep it in

LFDanLu:
NOTE: do NOT merge main in, this PR is only for testing but the patch will be done from the branch

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -439,4 +441,22 @@ export class TableLayout<T> extends ListLayout<T> {
 
     return res;
   }
+
+  // Checks if Chrome version is 105 or greater
+  private checkChrome105() {
```
ktabors, line 446: This feels like it should belong in util, not that anything else will use this.

LFDanLu: It's pretty specific and temporary so I kept it in here. Also this means we don't have to pull in another dep if this were in aria/utils

ktabors (approved):

PR #3454, opened 2022-08-26
Repositioning the popover and arrow position when the trigger is at the window edge 

LFDanLu:
This reverts commit 1dbd3aa3a1c5505239c8ea83eba6a06d62fd4ee6.

Closes 

## ✅ Pull Request Checklist:

- [ ] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- [ ] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [ ] Filled out test instructions.
- [ ] Updated documentation (if it already exists for this component).
- [ ] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/TR/wai-aria-practices-1.1/)

PR #3518, opened 2022-09-14
Popover edge position take two

### LINKED ISSUE #2258: Popover arrow missplaced in certain conditions
# 🐛 Bug Report 

When the trigger button is close to the bottom edge of the page, the popup will not render correctly. It starts collapsing and also tip gets miss-aligned 

## 🤔 Expected Behavior

Tip should not be miss-aligned and should be stick to popover dialog

## 😯 Current Behavior

Getting miss-aligned

## 🌍 Your Environment

https://codesandbox.io/s/dawn-breeze-3d6er

| react-spectrum   | 3.10.0
| Browser          | Chrome, Safari, Firefox
| Operating System | Mac

## 🧢 Your Company/Team

Working with Adobe Color Team

ktabors:
Closes #2258

refactored the previous work
- moved the logic from react-aria to react spectrum
- added horizontal scrolling to the story to improve testing

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Added a new story to test buttons at the top, bottom, start, and end of the page. There are three popover states to test, with the default padding of 12px, padding of 6px, and with 6px padding and no arrow.

Added chromatic stories instead of tests to avoid mocking since dom tests don't have dimensions. Only added four to not have too many stories, i.e. one of each state at each border.

You can also compare the storybook for this PR to the storybook of the old PR #3454 .

## 🧢 Your Project:
RSP


packages/@react-spectrum/dialog/chromatic/Dialog.chromatic.tsx
```diff
@@ -157,6 +158,105 @@ storiesOf('Dialog', module)
...
+              <ActionButton>T</ActionButton>
+              <Dialog><Content>Placement Top</Content></Dialog>
+            </DialogTrigger>
+          </View>
+        </Grid>
+      </View>
+    )
+  )
+  .add(
+    'position bottom adjusted 6px padding',
```
ktabors, line 239: This story is a little wonky. The bottom case was the hardest to get right so I wanted this example. Ant it uses `window.innerHeight` instead of `document.body.getBoundingClientRect()`. Due to the story parent min-height of 1000px the button and popover are forced out of the easily viewable area and don't really line up in the snapshot. Running Chromatic Storybook locally the popover shows the adjusted 6px padding, but in their snapshot you can't see that. The arrow is gone, maybe that is enough?

I tried a height of 100% instead of the minHeight=1000px and it still doesn't work how I'd like in the snapshot. Might just do a top story instead of this bottom one.

packages/@react-spectrum/dialog/test/DialogTrigger.test.js
```diff
@@ -136,6 +136,84 @@ describe('DialogTrigger', function () {
     expect(popover).toHaveAttribute('style');
   });
 
+  describe('popover arrows', function () {
```
ktabors, line 139: These tests work, but do we still want them if Chromatic covers these cases? The concern is the mocking of the data and test dom elements not really having dimensions.

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
```diff
@@ -144,13 +144,48 @@ export {_DialogTrigger as DialogTrigger};
 
 function PopoverTrigger({state, targetRef, trigger, content, hideArrow, isKeyboardDismissDisabled, ...props}) {
   let triggerRef = useRef<HTMLElement>();
-
   let overlayRef = useRef<DOMRefValue<HTMLDivElement>>();
+  let isCloserPadding = false;
```
snowystinger, line 148: what is isCloserPadding?

does it mean `isNextToContainerBoundary`? or `shouldUseLessPadding`?

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
```diff
@@ -144,13 +144,48 @@ export {_DialogTrigger as DialogTrigger};
...
+  // move closer to their trigger.
+  if (triggerRef?.current) {
+    // The popover ref is not defined here, we don't know the arrow size to factor into the below.
+    let triggerRect = triggerRef.current.getBoundingClientRect();
+    let documentRect = document.body.getBoundingClientRect();
+    // 'bottom' is the default when no value is provided
+    let [axisPlacement] = props.placement?.split(' ') || ['bottom']; // Flip and RTL don't matter
+
+    // The following is the math that gets the desired beavhior in both scales, but isn't as logical
+    // as desired. For example, would love to not multiple the default padding by 2.
```
snowystinger, line 160: what does this comment mean? why do we not want to multiply the default padding by 2? why are we multiplying the default padding by 2?

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
```diff
@@ -144,13 +144,48 @@ export {_DialogTrigger as DialogTrigger};
...
+    if (axisPlacement === 'start' || axisPlacement === 'end' || axisPlacement === 'right' || axisPlacement === 'left') {
+      // top edge start/end math
+      if ((DEFAULT_MODAL_PADDING * 2) - (triggerRect.height / 2) > triggerRect.top) {
+        isCloserPadding = true;
+      }
+      // bottom edge start/end math
+      // The documentRect doesn't work for this which is why we're using the window.innerHright.
```
snowystinger, line 167:
```suggestion
      // The documentRect doesn't work for this which is why we're using the window.innerHeight.
```

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
```diff
@@ -144,13 +144,48 @@ export {_DialogTrigger as DialogTrigger};
...
+  // For popover triggers close to the edge, decrease the container padding allowing popovers to
+  // move closer to their trigger.
+  if (triggerRef?.current) {
+    // The popover ref is not defined here, we don't know the arrow size to factor into the below.
+    let triggerRect = triggerRef.current.getBoundingClientRect();
+    let documentRect = document.body.getBoundingClientRect();
+    // 'bottom' is the default when no value is provided
+    let [axisPlacement] = props.placement?.split(' ') || ['bottom']; // Flip and RTL don't matter
+
+    // The following is the math that gets the desired beavhior in both scales, but isn't as logical
```
snowystinger, line 159:
```suggestion
    // The following is the math that gets the desired behavior in both scales but isn't as logical
```
what is the desired behavior? I think this goes along with my question about the multiplying by 2

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
snowystinger: why doesn't the document.body.getBoundingClientRect work? and why does the width work? I'm assuming this has to do with scrollable pages? we could have a horizontally scrolling page as well though, should we use innerWidth?

Do we actually want clientHeight? https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight#usage_notes

packages/@react-spectrum/dialog/test/DialogTrigger.test.js
```diff
@@ -136,6 +136,84 @@ describe('DialogTrigger', function () {
...
+    const originalOffsetHeight = Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'offsetHeight');
+    const originalOffsetWidth = Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'offsetWidth');
+
+    beforeAll(() => {
+      Object.defineProperty(HTMLElement.prototype, 'offsetHeight', {configurable: true, value: 500});
```
snowystinger, line 144: I think you want to use jest.spyOn
This way they will be known by jest as mocks and will get restored automatically when the test is done

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -102,6 +102,49 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
   let {modalProps} = useModal({
     isDisabled: isNonModal
   });
+  let [isHideArrow, setIsHideArrow] = useState(hideArrow);
+
+  useLayoutEffect(() => {
+    // check position of arrow and popover proximety to boundry edge
```
snowystinger, line 108:
```suggestion
    // check arrow and popover proximity to the boundary edge
```

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -102,6 +102,49 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
...
+    if (ref.current && arrowProps) {
+      let propsHideArrow = hideArrow;
+
+      // figure out the location of the offset and which popover dimension to use
+      let offset = arrowProps.style.left ? 'left' : 'top';
+      let overlayCrossSize = offset === 'left' ? ref.current.offsetWidth : ref.current.offsetHeight;
+
+      // for the start and top, no change is made to arrowPosition not caught in a condiational below
```
snowystinger, line 116: what does this mean?

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -102,6 +102,49 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
...
     isDisabled: isNonModal
   });
+  let [isHideArrow, setIsHideArrow] = useState(hideArrow);
+
+  useLayoutEffect(() => {
+    // check position of arrow and popover proximety to boundry edge
+    if (ref.current && arrowProps) {
+      let propsHideArrow = hideArrow;
+
+      // figure out the location of the offset and which popover dimension to use
```
snowystinger, line 112:
```suggestion
      // figure out the main (or is it the coss?) axis of the popover
```
which popover dimension to use for what?

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
```diff
@@ -144,13 +144,48 @@ export {_DialogTrigger as DialogTrigger};
 
 function PopoverTrigger({state, targetRef, trigger, content, hideArrow, isKeyboardDismissDisabled, ...props}) {
   let triggerRef = useRef<HTMLElement>();
-
   let overlayRef = useRef<DOMRefValue<HTMLDivElement>>();
+  let isCloserPadding = false;
+
+  // For popover triggers close to the edge, decrease the container padding allowing popovers to
+  // move closer to their trigger.
+  if (triggerRef?.current) {
```
snowystinger, line 152: this is all during a render? that seems wrong, especially on first render. `useLayoutEffect` is for when you need to make adjustments based on dom layout, your ref will be defined at that time

calculations can be performed during render, based on state, but we shouldn't be reading from the dom in render. See `Arrow` or your work in `PopoverWrapper`

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -102,6 +102,49 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
...
+      let offset = arrowProps.style.left ? 'left' : 'top';
+      let overlayCrossSize = offset === 'left' ? ref.current.offsetWidth : ref.current.offsetHeight;
+
+      // for the start and top, no change is made to arrowPosition not caught in a condiational below
+      if (DEFAULT_MODAL_PADDING > arrowProps.style[offset] && arrowProps.style[offset] > 0) {
+        // using the default padding for proper arrow placement
+        let arrowStyle = {...arrowProps.style};
+        arrowStyle[offset] = DEFAULT_MODAL_PADDING;
+        arrowProps.style = arrowStyle;
```
snowystinger, line 121: why make a copy and then overwrite it?
```suggestion
        arrowProps.style[offset] = DEFAULT_MODAL_PADDING;
```

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -102,6 +102,49 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
...
+        // isOpen can cause us to hide the arrow when we really want it to show, this reverts the margin removal when the arrow was hidden
```
snowystinger, line 122: where does isOpen cause us to hide the arrow?

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -102,6 +102,49 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
...
+        if (ref.current.style.margin === '0px') {
+          ref.current.style.removeProperty('margin');
+        }
```
snowystinger, line 125: what set this? can we change it there? i'd prefer not to manipulate the dom directly

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
ktabors: I remember that the triggerRef.current.getBoundingClientRect() and document.body.getBoundingClientRect() height and bottom values moved in relation to each other with a fixed difference instead of changing as the button is moved in and out of the viewable area.

packages/@react-spectrum/overlays/src/Popover.tsx
ktabors: Line 138 sets it to '0' on an early render. Maybe this logic should go into a useLayoutEffect?

packages/@react-spectrum/overlays/src/Popover.tsx
ktabors: See comment below about line 138.

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
ktabors: Maybe we need to talk. When I moved this logic into a `useLayoutEffect` (and a `useEffect`) it only runs on page load. It needs to run on trigger which is what happens without the useLayoutEffect.

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
ktabors: I don't remember why we are multiplying by two, it feels like we shouldn't. I tried a couple of different approaches to rework the calculations and what we have is the most accurate for both scales. The desired behavior is keeping the default padding as long as possible with the popover arrow moving to the popover edge, without going off, before being hidden.

packages/@react-spectrum/dialog/src/DialogTrigger.tsx
ktabors: Two notes:

1 - clientHeight is the height of the viewable area. We are currently calculating where the button is relative to that edge. innerHeight is the total height of the page.

2 - The two bottom values are always different by 24px. See the images below. This 24px is 4px (36px height - 32px button height) of height in the container and 20px of bottom margin.

image
image

packages/@react-spectrum/overlays/src/Popover.tsx
ktabors: Doing the suggested change doesn't update the value. The above code was the solution I found that updates the value. It's been too long since I wrote this for me to remember the reason, sorry.

ktabors:
current plan is to merge and do above improvements as followup

ktabors:
The hide arrow logic is no longer working correctly post merge conflict fix. Seems to be related to the overlay refactor, but don't see a change that would have caused this.

ktabors:
Fixed by #4220

PR #3520, opened 2022-09-15
TableView overlay auto tab stop Firefox fix

ktabors:
Closes

Found during testing, but pre-existing issue. In Firefox, an element with overflow auto or scroll is a tab stop. This caused an empty table to be tabbed too twice.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

In FireFox, go to any TableView story with no content (renderEmptyState, isLoading, etc.) and tab into the TableView. It should only stop once on the table body during keyboard tab navigation.

## 🧢 Your Project:
RSP

### DIFF
--- a/packages/@react-spectrum/table/src/TableView.tsx
+++ b/packages/@react-spectrum/table/src/TableView.tsx
@@ -449,6 +449,7 @@ function TableVirtualizer({layout, collection, focusedKey, renderView, renderWra
               }
             )
           }
+          tabIndex={-1}
           style={{flex: 1}}
           innerStyle={{overflow: 'visible', transition: state.isAnimating ? `none ${state.virtualizer.transitionDuration}ms` : undefined}}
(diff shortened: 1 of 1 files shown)


LFDanLu (approved):
Seem to work well in FF, Chrome, and Safari, TalkBack, and VO. Since this `tabindex=-1` is on the ScrollView it shouldn't interfere with the "restore focus back to the collection" code that useVirtualizer/useSelectableCollection has.

One interesting note is that this adds an additional "focus" target that TalkBack can access now because of the new tabindex but I think this is a minor point IMO

snowystinger (approved):
Seems fine, didn't notice any issues with the focus changes we recently made to collections. Cases such as deleting everything in ActionGroup stories.
BTW, we should link this in the wiki https://bugzilla.mozilla.org/show_bug.cgi?id=616594

PR #3533, opened 2022-09-19
Update simplified DnD handler api and memoize useDnDHooks

LFDanLu:
Followup review items from https://github.com/adobe/react-spectrum/pull/3266

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/@react-aria/dnd/src/useDroppableCollection.ts
```diff
@@ -59,8 +59,8 @@ export function useDroppableCollection(props: DroppableCollectionOptions, state:
       shouldAcceptItemDrop
     } = localState.props;
 
-    let {draggingCollectionRef, draggingKeys} = globalDndState;
-    let isInternalDrop = draggingCollectionRef?.current != null && draggingCollectionRef?.current === ref?.current;
+    let {draggingKeys} = globalDndState;
+    let isInternal = isInternalDropOperation(ref);
```
LFDanLu, line 63: A bunch of these `isInternal` checks use `ref` instead of relying on the global state's `dropCollectionRef` because I felt it was safer this way (global state is spooky in general). Since these are all within `useDroppableCollection`, the ref should be equal to dropCollectionRef for the most part (some exceptions exist where we want to calculate isInternal before we even set dropCollectionRef)

packages/@react-aria/dnd/src/utils.ts
```diff
@@ -336,6 +336,13 @@ export function setGlobalDnDState(state: DnDState) {
   globalDndState = state;
 }
 
+// Util function to check if the current dragging collection ref is the same as the current targeted droppable collection ref.
+// Allows a droppable ref arg in case the global drop collection ref hasn't been set
+export function isInternalDropOperation(ref?: RefObject<HTMLElement>) {
```
LFDanLu, line 341: `isInternal`? `isInternalDrop`? `isInternalTarget`?

packages/@react-aria/dnd/package.json
```diff
@@ -26,7 +26,6 @@
     "@react-aria/utils": "^3.13.3",
     "@react-aria/visually-hidden": "^3.4.1",
     "@react-stately/dnd": "3.0.0-alpha.10",
```
LFDanLu, line 28: Was doing some auditing and have a clarifying question: should things like `@react-stately/dnd`/`@react-types/*` be in devDeps instead if they are only used to import types/interfaces? I was under the impression that we could do a `import type` for things like that and thus avoid needing to include them within the dependencies because they only matter during dev (aka typechecking)

packages/@react-aria/interactions/src/utils.ts
```diff
@@ -33,7 +33,7 @@ export function isVirtualClick(event: MouseEvent | PointerEvent): boolean {
   // Android TalkBack's detail value varies depending on the event listener providing the event so we have specific logic here instead
   // If pointerType is defined, event is from a click listener. For events from mousedown listener, detail === 0 is a sufficient check
   // to detect TalkBack virtual clicks.
-  if (isAndroid() && (event as PointerEvent).pointerType != null) {
+  if (isAndroid() && (event as PointerEvent).pointerType) {
```
LFDanLu, line 36: click events triggered by `.click()` will have a `pointerType = ""` and `detail === 0` so we can avoid this check and fallback to the `detail === 0` check below

packages/@react-spectrum/dnd/src/useDnDHooks.ts
```diff
@@ -68,45 +68,46 @@ export interface DnDOptions extends Omit<DraggableCollectionProps, 'preview' | '
 }
 
 export function useDnDHooks(options: DnDOptions): DnDHooks {
-  let {
```
LFDanLu, line 71: This diff is really just me moving everything into a single useMemo

packages/@react-types/shared/src/dnd.d.ts
```diff
@@ -10,6 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
+import type {DirectorySymbolType} from '@react-aria/dnd';
```
LFDanLu, line 13: urgh, hopefully this is ok to have in the types package, otherwise I'm a bit at a loss at how to define the `directorySymbol` in such a way that the `@react-types/shared` package doesn't depend on the `@react-aria/dnd` package. 

Alternatively I could make `acceptedDragTypes` accept generic `symbol` instead or move `DroppableCollectionProps` definition out of types and into the `react-aria/dnd` package

LFDanLu: Annnd circular dependency

packages/@react-aria/dnd/src/utils.ts
```diff
@@ -159,12 +161,12 @@ export class DragTypes implements IDragTypes {
     this.includesUnknownTypes = !hasFiles && dataTransfer.types.includes('Files');
   }
 
-  has(type: string) {
-    if (this.includesUnknownTypes || (type === 'directory' && this.types.has(GENERIC_TYPE))) {
+  has(type: string | DirectorySymbolType) {
```
devongovett, line 164: does `string | symbol` work here? Might be good to keep this a little more vague rather than only allowing `directorySymbol`, in case we add more special symbol types in the future.

packages/@react-aria/dnd/src/utils.ts
```diff
@@ -18,6 +18,8 @@ import {Key, RefObject} from 'react';
 import {useId} from '@react-aria/utils';
 
 const droppableCollectionIds = new WeakMap<DroppableCollectionState, string>();
+export const directorySymbol = Symbol();
```
devongovett, line 21: We haven't really done something like this before, but do we want to use upper case for constants? I think that's a fairly common convention. Something like `DIRECTORY_DRAG_TYPE`? If not, I still think naming it with drag type in the name rather than "symbol" makes sense.

packages/@react-aria/dnd/src/utils.ts
devongovett: I think what you have is good

packages/@react-spectrum/dnd/src/useDnDHooks.ts
```diff
@@ -68,45 +68,46 @@ export interface DnDOptions extends Omit<DraggableCollectionProps, 'preview' | '
...
-  let mergedHooks = {
-    ...(isDraggable ? dragHooks : {}),
-    ...(isDroppable ? dropHooks : {}),
-    ...(isDraggable || isDroppable ? {isVirtualDragging} : {})
-  };
+    return {
+      ...(isDraggable ? dragHooks : {}),
+      ...(isDroppable ? dropHooks : {}),
+      ...(isDraggable || isDroppable ? {isVirtualDragging} : {})
+    };
```
devongovett, line 107: Instead of creating the objects unconditionally and then merging them, you could assign properties from within an if statement? Maybe a little more efficient.

```js
let hooks = {};
if (isDraggable) {
  hooks.useDraggableCollectionState = /* ... */
}

// etc.
```

packages/@react-aria/dnd/src/useDroppableCollection.ts
```diff
@@ -70,9 +70,9 @@ export function useDroppableCollection(props: DroppableCollectionOptions, state:
     let filteredItems = items;
     if (acceptedDragTypes) {
```
devongovett, line 71: I guess this check isn't needed anymore? Or maybe it should be `if (acceptedDragTypes !== 'all' || shouldAcceptItemDrop)` as an optimization?

LFDanLu: ah yep, thanks for catching this

packages/@react-types/list/src/index.d.ts
```diff
@@ -21,7 +21,6 @@ import {
   SpectrumSelectionProps,
```
LFDanLu, line 21: I would love to delete the `@react-types/list` package now that aria/spectrum define and export the types but a bit worried about if people are using it since it have been released.

devongovett: would it be weird to change this package to re-export the types from aria/spectrum instead? only for backward compatibility I guess. not sure how important tho

devongovett (approved):

packages/@react-types/list/src/index.d.ts
LFDanLu: feels a bit weird to have the dependencies go the other way, but at least it wouldn't be circular since nothing else should depend on this types package. I'll go ahead and update

devongovett (approved):

snowystinger (approved):

PR #3546, opened 2022-09-22
Fix clicking on DateField and TimeField label with contextual help

devongovett:
Small bug found in https://github.com/adobe/react-spectrum/pull/3521: when clicking on the label for DateField and TimeField, it would focus the contextual help button rather than the first date segment. This was due to using FocusManager around the entire field, rather than only the segments.


snowystinger (approved):
It's working as described 👍🏻

reidbarber (approved):
LGTM

PR #3549, opened 2022-09-23
DnD: Focus an item after it has been dropped on

reidbarber:
## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test DnD stories: When dropping on an item (i.e. ListView with folder), focus should then go to that item, instead of going to the list itself.


snowystinger (approved):
Working as expected in chrome. Verified old behavior as well to compare.

LFDanLu (approved):
Seems to work well, bit worried about the old `if (state.selectionManager.focusedKey === focusedKey) {` check being removed since we will now be moving focus even if focus shifted since the last drop but I'm not sure what purpose it had previously. Lets test it

PR #3569, opened 2022-09-28
Fix root drop interaction for iOS VO

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Test the root drop ListView DnD stories. You should be able to swipe onto the root drop indicator via iOS VO

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-spectrum/list/src/RootDropIndicator.tsx
+++ b/packages/@react-spectrum/list/src/RootDropIndicator.tsx
@@ -19,9 +19,8 @@ export default function RootDropIndicator() {
     <div role="row" aria-hidden={dropIndicatorProps['aria-hidden']}>
       <div
         role="gridcell"
-        aria-selected="false"
-        {...visuallyHiddenProps}>
-        <div role="button" {...dropIndicatorProps} ref={ref} />
+        aria-selected="false">
+        <div role="button" {...visuallyHiddenProps} {...dropIndicatorProps} ref={ref} />
       </div>
     </div>
   );


devongovett:
I still see scrollbars when dragging at the bottom of this story: https://reactspectrum.blob.core.windows.net/reactspectrum/294b68f2bba1fbed137cd9cec4cbdd5c79869feb/storybook/index.html?path=/story/listview-drag-and-drop--drag-within-list-reorder&providerSwitcher-toastPosition=bottom

LFDanLu:
@devongovett I wasn't able to get rid of the additional space that it the drop indicator takes up, lemme try an alternative to the blue border we currently use

snowystinger (dismissed):

LFDanLu:
Note for posterity/other reviewers: the scroll bar showing up in FF won't be handled in this PR, an issue in the board has been made to track that work along with a suggested fix

devongovett:
So without that, what is this PR fixing?

LFDanLu:
Fixes the focusing the root drop indicator for iOS VO and stops the scroll position from shifting when focusing the last insertion indicator via keyboard (aka no more wiggle). The scrollbar will still show up so the user can scroll down via touchpad/mouse and see the full indicator but that can be fixed in a future iteration

packages/@react-spectrum/list/src/ListView.tsx
```diff
@@ -258,7 +267,8 @@ function ListView<T extends object>(props: SpectrumListViewProps<T>, ref: DOMRef
                       <InsertionIndicator
                         key={`${item.key}-after`}
                         target={{key: item.key, type: 'item', dropPosition: 'after'}}
-                        isPresentationOnly={collection.getKeyAfter(item.key) != null} />
+                        isPresentationOnly={item.key !== lastKey}
+                        isLastRow={lastKey === item.key} />
```
devongovett, line 271: One problem is I'm not sure if this will update when a new row is added at the end, whereas it would have before via context.

devongovett (approved):

dannify (approved):

PR #3577, opened 2022-09-30
Rename useDnDHooks and ListView dndHooks prop

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

- [ ] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- [ ] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [ ] Filled out test instructions.
- [ ] Updated documentation (if it already exists for this component).
- [ ] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/TR/wai-aria-practices-1.1/)

PR #3579, opened 2022-09-30
Updating ListView dnd docs for new api

LFDanLu:
Closes 

This relies on https://github.com/adobe/react-spectrum/pull/3548 and https://github.com/adobe/react-spectrum/pull/3577
Only merge after release goes out

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


snowystinger (dismissed):
My only real comment is that some of the comments say
"similar configuration to the previous example" and then the previous example says
"same configuration as some other example"
could we just point both to the actual base? otherwise it's a lot to search through

devongovett (commented):
Sorry for more comments! happy to discuss, or lmk if you want any help updating anything

packages/@react-spectrum/list/docs/ListView.mdx
```diff
@@ -393,70 +392,64 @@ function DragIntoListExample() {
...
+      } = e;
+      let processedItems = await Promise.all(
+        items.map(async item => JSON.parse(await item.getText(acceptedDragTypes[0])))
+      );
+      targetList.append(...processedItems);
+    },
+    acceptedDragTypes
   });
-
+  /*- end highlight -*/
```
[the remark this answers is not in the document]
LFDanLu: Added, I think I pushed the change already while you were reviewing haha

packages/@react-spectrum/list/docs/ListView.mdx
```diff
@@ -919,7 +976,152 @@ function DragBetweenListsExample() {
...
+      ///- begin collapse -///
+      if (target.dropPosition === 'before') {
+        targetList.insertBefore(target.key, ...processedItems);
+      } else if (target.dropPosition === 'after') {
+        targetList.insertAfter(target.key, ...processedItems);
+      }
+      ///- end collapse -///
+    },
+    onRootDrop: async (e) => {
+      // Similar onRootDrop as DragIntoListExample...
```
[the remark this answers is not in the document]
LFDanLu: Since I've kept the examples code hidden in the details, what do you think about just removing the collapse portions within the code? Maybe the highlights are good enough and this way people can just copy paste the example code block without having to fill in the gaps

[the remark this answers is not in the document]
devongovett: Yeah let's try that.

packages/@react-spectrum/list/docs/ListView.mdx
```diff
@@ -361,24 +362,124 @@ This behavior is slightly different in the highlight selection style, where sing
...
 <details>
   <summary style={{fontWeight: 'bold'}}><ChevronRight size="S" /> Show code</summary>
 
 ```tsx example export=true render=false
-import {useDnDHooks} from '@react-spectrum/dnd';
+import {useDragAndDrop} from '@react-spectrum/dnd';
 
-function DragIntoListExample() {
+function DraggableListView(props) {
```
LFDanLu, line 382: Each example currently has its own implementation of a DraggableListView and a DroppableListView. I could instead make a single generic ListView that accepts drag/drop hook options via props and have the examples all reuse that component but I felt it would be clearer to have the drag/drop hook options provided inline to the `useDragAndDrop` hook. The downside to this is that there is a lot of duplicate code but it also means the user can copy paste from which ever example's code block they are interested in and not have to worry about finding the base component implementation hidden somewhere above

Lemme know what you think. If we wanna go with the single generic ListView component that accepts the DnDOptions then the first example's setup can be used as a starting point

packages/@react-spectrum/list/docs/ListView.mdx
```diff
@@ -393,122 +494,160 @@ function DragIntoListExample() {
...
+### Handling folder drops
+
+The example below replicates the previous example but demonstrates how to handle on item drops.
 
 <details>
   <summary style={{fontWeight: 'bold'}}><ChevronRight size="S" /> Show code</summary>
 
 ```tsx example export=true render=false
-function DragIntoListFilesOnly() {
+function DraggableListViewFolder(props) {
```
LFDanLu, line 524: I only highlighted the areas of code that differed heavily from the previous example or were the key dnd option/change for the current example. This means there really isn't too much highlighted, lemme know if we'd rather highlight all the DnD related code in each example

devongovett: Yep I think this is good!

packages/@react-spectrum/list/docs/ListView.mdx
devongovett: Yeah I think it's ok since they are collapsed. We can see what people think.

dannify (approved):

PR #3634, opened 2022-10-12
Expose `props` to `Section` collection item

luisadame:
The `Item` component [does this already](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/collections/src/Item.ts#L34), and I believe this is an inconsistency.

Although, we can get around this issue by doing this:
```
function Section() {
  return null
}

Section.getCollectionNode = function* (props) {
  const node = DefaultSection.getCollectionNode(props).next()
  yield { ...node.value, props }
}
```
I feel this behavior should be consistent across the collection components.


snowystinger (approved):
seems reasonable, thanks for the PR

just curious, what're you sending through?

luisadame:
> seems reasonable, thanks for the PR
> 
> just curious, what're you sending through?

I'm sending an `slice` prop which will try to slice the number of items within a section and add an expanding menu item so that multiple sections can be glanced quicker, instead of scrolling through a very first large section of items to choose an option from the second section.

A quick doodle:

Untitled-2021-02-09-2210

I'm not sure if I will be able to achieve that, but there may be more use cases for having props shared.

reidbarber (approved):

PR #3687, opened 2022-10-26
Remove `aria-checked` attribute from useCheckbox and useSwitch

### LINKED ISSUE #3130: aria-checked is not valid on an input type="checkbox"
# 🐛 Bug Report 

Using aria-checked on a checkbox is now a MUST NOT in ARIA-in-HTML
https://w3c.github.io/html-aria/#att-checked

## 😯 Current Behavior

spectrum/blob/main/packages/%40react-aria/checkbox/src/useCheckbox.ts#L48)

## 💁 Possible Solution

Remove all references to aria-checked. If spreading these props on an input type=checkbox they are not necessary

## 💻 Code Sample

https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/checkbox/src/useCheckbox.ts#L48

## 🌍 Your Environment

| react-spectrum   |
| Browser          |
| Operating System |

tywayne:
Removes the `aria-checked` attribute from `inputProps` returned from `useCheckbox` and `useSwitch`.

Tests for the `CardView` component were using this attribute to validate selection changes so those tests were failing. Went ahead and also updated those tests to checked the `checked` attribute instead.

Closes #3130 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:
View `Checkbox` and `Switch` components in storybook, toggle their states, view that the `aria-checked` attribute is no longer applied to the `<input />`


devongovett (approved):

devongovett:
GET_BUILD

joshsalverda:
This also removed `aria-checked="mixed"` which looks like it should have been replaced with the `indeterminate` attribute instead?

tywayne:
The `indeterminate` property is set directly on the input ref with the `useEffect` before the return. It is a property, not an attribute, and can't be set via the props returned by the hook.

PR #3692, opened 2022-10-27
TagGroup beta API updates

tiffany-sy-lin:
- remove isDisabled and disabledKeys
- In onRemove, pass tag id instead of content
- remove event arg from onRemove
- Rename isRemovable to follow props guidelines: 'allowsX'
- Remove 'with announcing' story
- Confirm focus is moved to prev/next item after removal

## ✅ Pull Request Checklist:


packages/@react-aria/tag/src/useTag.ts
```diff
@@ -56,14 +54,14 @@ export function useTag(props: TagProps<any>, state: GridState<any, any>): TagAri
     focusMode: 'cell'
   }, state, tagRef);
 
-  function onKeyDown(e: KeyboardEvent<Element>) {
+  function onKeyDown(e: KeyboardEvent) {
     if (e.key === 'Delete' || e.key === 'Backspace' || e.key === ' ') {
-      onRemove(children, e);
+      onRemove(gridCellProps['data-key']);
```
devongovett, line 59: You should use `item.key` here instead. Data attributes will be converted to strings, but keys can also be numbers. Same below.

packages/@react-spectrum/tag/src/Tag.tsx
```diff
@@ -59,10 +57,9 @@ export function Tag<T>(props: SpectrumTagProps<T>) {
           styles,
           'spectrum-Tags-item',
           {
-            'is-disabled': isDisabled,
             'focus-ring': isFocusVisible,
             'is-focused': isFocused,
-            'not-removable': !isRemovable,
+            'not-removable': !allowsRemoving,
```
reidbarber, line 62: Don't think this class gets used anywhere, so we can actually remove this line

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -208,35 +158,10 @@ describe('TagGroup', function () {
...
-    let inputs = getAllByRole('textbox');
-    expect(document.activeElement).toBe(inputs[0]);
-    userEvent.tab();
-    expect(document.activeElement).toBe(inputs[1]);
-    userEvent.tab({shift: true});
-    expect(document.activeElement).toBe(inputs[0]);
-  });
-
-  it('not disabled if extraneous keys are provided', function () {
+  it('not disabled if keys are provided', function () {
```
reidbarber, line 161: IMO we can remove this test since disableKeys won't be in the types

tiffany-sy-lin: I left it as a test to confirm disabled tests don't work anymore, but can remove if you don't think that's necessary

reidbarber (approved):
LGTM

dannify:
I found 2 issues with focus (which I think we already found?) but they are not blockers for beta or this PR. Please add them to an upcoming sprint.
1. Delete a Tag. Cannot use left/right arrows to navigate right after.
2. Deleting the last tag loses focus (should go to the previous tag)

dannify (approved):

devongovett (approved):

PR #3697, opened 2022-10-29
Wait until after a layout effect to prevent scroll in popovers

devongovett:
This fixes an issue found in tested where extra padding was added to the page to account for scrollbars when unnecessary. This was due to measuring before the max-height was added to the popover, thus overflowing the page. Now we wait for a layout effect after positioning to enable preventing scroll.


reidbarber (approved):
LGTM

snowystinger (approved):
can no longer bring up extra scroll bar in dialog story where content scrolls
lgtm

Issue #865, opened 2020-07-30
Add TypeScript Examples


matamatanot:
# 🙋 Documentation Request
I'd like examples of TypeScript. Please consider adding.

For examples, [React Hook Form](https://react-hook-form.com/api/) api docs offer JS and TS examples.
You can switch the displayed code snippets with a button.

matamatanot:
@devongovett 
Do you have any thoughts?
If we implement a mechanism to display TS examples in the same way as the React Hook Form, Many users can contribute. 

Of course I would like to contribute to the addition of examples.

dannify:
Hey @matamatanot. I think this is something we would be happy to have in the docs but the main question would be how would it work? I think it would require some investigation to determine if it is possible with our current documentation builder/website. If you had any thoughts on this, we can discuss the options here. If it ends up proving to be a challenge, perhaps an RFC would be needed.

aidan-keay:
Has there been any update on having TS examples?

dannify:
This is not something the core team is able to take on any time soon. Happy to discuss contributions!

aidan-keay:
> This is not something the core team is able to take on any time soon. Happy to discuss contributions!

I'd be happy to contribute the ts examples although how easy would it be to implement?

dannify:
I think the biggest hurdle to start with will be setting up the docs site to display and build ts examples. We have not done any investigations into this yet so I am not too sure how easy it will be.

alveshelio:
How about, while we wait to know whether or not it's feasible in the site, to add some examples in Typescript to this repo so at least people looking for Typescript examples can find something.
This would simplify the process.

xxKeefer:
Would Love to contribute some examples, I don't want to have to figure out how to integrate with your docs site however. I love typescript and react-aria however i have to say the current DX trying to get them to play nice together has been woeful. Would love to see the adobe team provide an outlet for TS dev looking to contribute TS examples

donaldpipowitch:
Until types in the examples are ready I have a really hard time to wrap my head around the usage of `props` in the existing examples. Is there some guidance on this?

Example:

```
function CheckboxGroup(props) {
  let { children, label, description, errorMessage, validationState } = props;
  let state = useCheckboxGroupState(props);
  let { groupProps, labelProps, descriptionProps, errorMessageProps } =
    useCheckboxGroup(props, state);
```

[...]

PR #3710, opened 2022-11-03
Type check docs examples

devongovett:
Part of #865

This adds a script that extracts all of the examples from the docs MDX files into TSX files, and then runs TypeScript over them. This will run in CI. It found a number of issues, such as unused variables, missing props, etc. So far, I have only fixed the existing type errors, adding annotations where needed. Eventually, we can add type annotations to all examples.

We may want to eventually have a toggle to switch between TS and JS. For now, only TS is displayed.


packages/@internationalized/date/src/CalendarDate.ts
```diff
@@ -42,6 +42,7 @@ export class CalendarDate {
   // This prevents TypeScript from allowing other types with the same fields to match.
   // i.e. a ZonedDateTime should not be be passable to a parameter that expects CalendarDate.
   // If that behavior is desired, use the AnyCalendarDate interface instead.
+  // @ts-ignore
```
devongovett, line 45: This ignores the "unused variable" error which comes up due to the tsconfig for the docs checker.

packages/@react-aria/dnd/docs/useDroppableCollection.mdx
```diff
@@ -277,53 +277,8 @@ function DropIndicator(props) {
 
 Now that the `DropIndicator` component is implemented, we can render an instance between each item in the list. This uses the `before` drop position by default, except for after the last item in the list.
 
-```tsx example render=false
-///- begin collapse -///
-function ListBox(props) {
```
devongovett, line 282: This was unused. The real implementation is in the below example.

packages/@react-aria/dnd/src/index.ts
```diff
@@ -19,7 +19,14 @@ export type {DragPreviewProps} from './DragPreview';
...
 export type {DropOptions, DropResult} from './useDrop';
 export type {ClipboardProps, ClipboardResult} from './useClipboard';
-export type {DropTargetDelegate} from '@react-types/shared';
+export type {
+  DropTargetDelegate,
+  DragItem,
+  DropItem,
+  TextItem,
+  FileItem,
+  DirectoryItem
```
devongovett, line 28: Should we rename these types to e.g. `TextDropItem` so it's more clear they are related to dnd? Especially when we add to the monopackage.

packages/@react-spectrum/tabs/docs/Tabs.mdx
```diff
@@ -80,21 +80,22 @@ function Example() {
...
+  type Tab = typeof tabs[0];
+  let [tabId, setTabId] = React.useState<React.Key>(1);
 
   return (
     <>
       <p>Current tab id: {tabId}</p>
       <Tabs aria-label="History of Ancient Rome" items={tabs} onSelectionChange={setTabId}>
         <TabList>
-          {item => (
+          {(item: Tab) => (
```
devongovett, line 91: These are annoying. Because we designed the API with `items`  on `Tabs` rather than `TabList` and `TabPanels`, this will always be `unknown` by default without an annotation.

packages/@react-aria/listbox/docs/useListBox.mdx
```diff
@@ -235,8 +236,10 @@ Use `defaultSelectedKeys` to provide a default set of selected items (uncontroll
 See the `react-stately` [Selection docs](../react-stately/selection.html) for more details.
 
 ```tsx example
+import type {Selection} from 'react-stately';
+
 function Example() {
-  let [selected, setSelected] = React.useState(new Set(['cheese']));
+  let [selected, setSelected] = React.useState<Selection>(new Set(['cheese']));
```
devongovett, line 242: These are annoying, and caused by `onSelectionChange` which may emit `"all" | Set<Key>`, but the type of this variable is only inferred to be `Set<string>` so it always requires an annotation.

LFDanLu (approved):
LGTM, verified locally that running `make check-examples` was catching typescript issues in the docs. Spot checked a couple of the docs changes and verified that they rendered just fine in the docs.

snowystinger (approved):
lgtm, also spot checked the docs

PR #3711, opened 2022-11-03
POC of automated Storybook accessibility/interaction testing  

LFDanLu:
Pros:
- can replay the interaction test step by step
- catches more accessibility issues since it runs in browser 

Cons:
- needs storybook running 
- reporting from the automated test is kinda hard to read
- some finicky behavior from interaction panel

TODO: followup
- figure out if can run accessibility tests with different themes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Try running yarn test-storybook
2. Try going to the "With state" HelpText story and the "Default"/"Sections"/"Keyboard tab focus" Picker story and play around with the interaction addon panel.

## 🧢 Your Project:

RSP


LFDanLu:
This PR will need to wait for all of our stories to be converted to CSF 3, for now will serve as an example of additional storybook accessiblity/in browser interaction testing + example of using axe-core in our tests

LFDanLu:
Running the accessibility test yields the following results:
```
Test Suites: 32 failed, 84 passed, 116 total
Tests:       202 failed, 959 passed, 1161 total
```
Spot checked the failures against the accessibility addon in storybook and looked to be about the same. Will go through and try to hide the ones that we know are false positives

package.json
```diff
@@ -21,6 +21,7 @@
     "build:docs": "DOCS_ENV=staging parcel build 'packages/@react-{spectrum,aria,stately}/*/docs/*.mdx' 'packages/react-aria-components/docs/*.mdx' 'packages/@internationalized/*/docs/*.mdx' 'packages/dev/docs/pages/**/*.mdx'",
     "test": "yarn jest",
     "test-strict": "cross-env STRICT_MODE=1 yarn jest",
+    "test-storybook": "test-storybook --url http://localhost:9003  --browsers chromium --no-cache",
```
reidbarber, line 24: Could this start storybook first (if not running), then run the test runner? Seems like we could, but also I don't see anything in their docs about it.

reidbarber (commented):
Awesome!

Have you looked into the chromatic interactions integration? Maybe that's something we could use: https://www.chromatic.com/docs/interactions

package.json
LFDanLu: I believe we could do something like https://github.com/adobe/react-spectrum/blob/ba727bdc0c4a57626131e84d9c9b661d0b65b754/scripts/verdaccio.sh#L45-L48 where we wait until storybook outputs "Built in" and then run `test-storybook`.

LFDanLu:
@reidbarber 

> Have you looked into the chromatic interactions integration? Maybe that's something we could use: https://www.chromatic.com/docs/interactions

Didn't realize that chromatic supported interactions, thanks for the link! Looks like we'd just have to write play functions in our chromatic stories and they will be run and snap shotted by Chromatic automatically. Def something we can explore using if we have a use case, but I'd like to get people's opinion of writing interaction tests via Storybook first (aka is it worth it or do we prefer the current react testing library tests?) IMO there are some use cases that are easier to test in browser, especially for cases that would require mocking of things like clientHeight/raf/etc

reidbarber:
>IMO there are some use cases that are easier to test in browser, especially for cases that would require mocking of things like clientHeight/raf/etc

@LFDanLu I like this idea a lot. I think it is better coverage and better DX. Makes sense for any interactions we have that come up similar to DnD and table column resizing.

PR #3723, opened 2022-11-07
Add preventDefault to onDragEnter and onDragLeave for iPad

devongovett:
Previously, I noticed that sometimes dropping items on iPad would cause Safari to do a google search for the dropped data. I have not been able to reproduce this after adding `preventDefault` in the onDragEnter and onDragLeave events.


LFDanLu (approved):
Tested in Safari iPad, Chrome Android, and desktop browsers. Functionality in the DnD and ListView DnD stories seems to be intact and I can't reproduce the weirdness where dragging + dropping opens a google search

reidbarber (approved):
LGTM, could not reproduce the issue on iPad

PR #3724, opened 2022-11-08
Fix useListData so that reorder operations are correctly inserted at the correct index

LFDanLu:
Specifically fix it for cases where the user is reordering a set of sequential items (3 items or more) and is moving it to a drop position one below the last item being moved. Addresses the comment  made here https://github.com/adobe/react-spectrum/pull/3423#discussion_r982907881

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Go to the ListView DnD reorder stories and select the top three items. Drag them after the fourth item and verify that the items are reordered correctly. 
To contrast, here is a broken example: https://reactspectrum.blob.core.windows.net/reactspectrum/842a0ed0ace6e17adee425286056456d0061d9a1/storybook/index.html?path=/story/listview-drag-and-drop-util-handlers--drag-within-list-reorder&providerSwitcher-toastPosition=bottom

Selecting and moving the top three items down one item actually moves them down two items (e.g. "Utilities" folder becomes the 2nd item in the list when it should have just moved the top three items after "Adobe inDesign")

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-stately/data/src/useListData.ts
+++ b/packages/@react-stately/data/src/useListData.ts
@@ -326,7 +326,7 @@ function insert<T>(state: ListState<T>, index: number, ...values: T[]): ListStat
 
 function move<T>(state: ListState<T>, indices: number[], toIndex: number): ListState<T> {
   // Shift the target down by the number of items being moved from before the target
-  toIndex = toIndex - indices.filter(index => index < toIndex).length;
+  toIndex -= indices.filter(index => index < toIndex).length;
 
   let moves = indices.map(from => ({
     from,
--- a/packages/@react-stately/data/src/useListData.ts
+++ b/packages/@react-stately/data/src/useListData.ts
@@ -326,11 +326,7 @@ function insert<T>(state: ListState<T>, index: number, ...values: T[]): ListStat
 
 function move<T>(state: ListState<T>, indices: number[], toIndex: number): ListState<T> {
   // Shift the target down by the number of items being moved from before the target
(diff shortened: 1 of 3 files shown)


packages/@react-stately/data/src/useListData.ts
```diff
@@ -326,11 +326,7 @@ function insert<T>(state: ListState<T>, index: number, ...values: T[]): ListStat
 
 function move<T>(state: ListState<T>, indices: number[], toIndex: number): ListState<T> {
   // Shift the target down by the number of items being moved from before the target
-  for (let index of indices) {
-    if (index < toIndex) {
-      toIndex--;
-    }
-  }
```
LFDanLu, line 333: Since the values in `indicies` are in ascending order, it is possible for `toIndex` to be decremented to the point where the final index in `indicies` is the same value as `toIndex`

packages/@react-stately/data/src/useListData.ts
```diff
@@ -326,11 +326,7 @@ function insert<T>(state: ListState<T>, index: number, ...values: T[]): ListStat
...
+  toIndex = toIndex - indices.filter(index => index < toIndex).length;
```
devongovett, line 329:
```suggestion
  toIndex -= indices.filter(index => index < toIndex).length;
```

This allocates an extra array, but probably fine...

devongovett (approved):
minor nit

ktabors (approved):

PR #3732, opened 2022-11-09
Docs updates for release

devongovett:
* Discourages use of `isKeyboardDismissDisabled` in `usePopover`
* Updates button docs for new variants
* Adds docs for `colorVersion`

### DIFF
--- a/packages/dev/docs/pages/react-spectrum/styling.mdx
+++ b/packages/dev/docs/pages/react-spectrum/styling.mdx
@@ -174,7 +174,12 @@ All of the properties supported by `View` are listed below. All style properties
 
 Style props that accept colors, such as backgrounds and borders, only accept Spectrum defined color values.
 In addition to ensuring consistency across products, Spectrum colors automatically adapt to color
-scheme changes, e.g. dark mode.
+scheme changes, e.g. dark mode. Learn more about color on the [Spectrum website](https://spectrum.adobe.com/page/color-fundamentals/).
+
+There are two versions of Spectrum colors available: v5, and v6. v6 offers an expanded color palette, and we
+recommend using it for new projects. However, v5 is still the default for backward compatibility. You can
+opt into v6 by setting the `colorVersion` prop on the [View](View.html) component to `6`. A [migration guide](https://spectrum.adobe.com/static/misc/Colors_v6.0_Migration-Guide.xd)
+is also available on the Spectrum website.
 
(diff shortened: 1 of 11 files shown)


packages/dev/docs/pages/react-spectrum/styling.mdx
```diff
@@ -174,26 +174,41 @@ All of the properties supported by `View` are listed below. All style properties
 
 Style props that accept colors, such as backgrounds and borders, only accept Spectrum defined color values.
 In addition to ensuring consistency across products, Spectrum colors automatically adapt to color
-scheme changes, e.g. dark mode.
+scheme changes, e.g. dark mode. Learn more about color on the [Spectrum website](https://spectrum.adobe.com/page/color-fundamentals/).
+
+There are two versions of Spectrum colors available: v5, and v6. v6 offers an expanded color palette, and we
+recommend using it for new projects. However, v5 is still the default for backward compatibility. You can
+opt into v6 by setting the `colorVersion` prop on the [View](View.html) component to `6`. A [migration guide](https://spectrum.adobe.com/static/misc/Colors_v6.0_Migration-Guide.xd)
```
LFDanLu, line 181: So for my own knowledge, this is saying that users need to wrap their app/components in a View that has `colorVersion={6}` in order to get the new colors? Or am I misunderstanding this and they would just need to have a View with `colorVersion={6}`? Also what do you think about linking back to this page from View's `colorVersion` prop description?

LFDanLu (commented):
Docs looks good, just one question

packages/dev/docs/pages/react-spectrum/styling.mdx
devongovett: No, they have to put `colorVersion={6}` on the `View` component for the `backgroundColor`/`borderColor` etc. props to be v6. It doesn't affect children.

LFDanLu: Ahhh, ok makes sense

ktabors: Devon's answer to Daniel doesn't match how I read that paragraph. His question and my confusion suggestion this should be rewritten to avoid confusion. Maybe my above suggestion is enough.

snowystinger (approved):

dannify (approved):

PR #3741, opened 2022-11-14
Fix button design issues

devongovett:
This fixes a few issues with buttons:

* Static color buttons and action buttons did not pass contrast requirements in their hover/active states over certain backgrounds such as ActionBar. This has been fixed by reducing the opacity of the white background on the buttons, and reverting the background color of ActionBar to to the old static color temporarily. I also updated our examples to use a passing background color. Spectrum design will be working on a more robust solution.
* In dark themes, the action button hover state and selected state were the same leading to buttons feeling like nothing happened when you clicked them. This has been fixed by inverting the color in the selected states in dark themes, as we do for light themes, matching checkbox/switch/radio.
* Filled buttons should not have a border, but in static color variants they appeared like they did because of the transparency of the border overlapping with the background. Now the border is always fully transparent to avoid this.
* Non-secondary outlined static white buttons should have darker borders to match static black. This was wrong in the XD file.


snowystinger (dismissed):

snowystinger (approved):

dannify (approved):

LFDanLu (approved):
LGTM

PR #3742, opened 2022-11-14
Do not close overlays on scroll

devongovett:
On iPad, tapping on a button to open a popover did not work. It immediately closed due to page scrolling caused by usePreventScroll. We don't need this anymore, so disable it by setting `onClose` to `null`.


snowystinger (approved):
I can open all the popovers in docs on my iPhone now

LFDanLu (approved):
Verified that I can now open spectrum/aria examples/components w/ popovers in iPad. Things still look fine in desktop and Android Chrome.

PR #3759, opened 2022-11-16
TagGroup: add `maxRows` prop to limit initially shown tags, and include 'Show all' button

reidbarber:
Closes https://github.com/adobe/react-spectrum/issues/3660

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Check maxRows stories under TagGroup. Resize container and confirm behavior.

## 🧢 Your Project:

RSP


packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -299,6 +299,85 @@ describe('TagGroup', function () {
...
+          <Item key="7" aria-label="Tag 7">Tag 7</Item>
+        </TagGroup>
+      </Provider>
+    );
+
+    let tags = getAllByRole('gridcell');
+    expect(tags.length).toBe(4);
+
+    let button = getByRole('button');
+    expect(button).toHaveTextContent('Show all (7)');
```
snowystinger, line 367: should the number refer to the hidden tags only? so this would be "Show all (3)"

snowystinger (commented):
We need a FocusScope around the TagGroup. See https://github.com/adobe/react-spectrum/blob/main/packages/%40react-spectrum/actiongroup/src/ActionGroup.tsx#L219 for an example

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -70,6 +72,78 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
+    setVisibleTagCount(function *() {
+      // Update to show all items.
+      yield gridCollection.size;
+
+      // Measure, and update to show the items until maxRows is reached.
+      let newVisibleTagCount = computeVisibleTagCount();
+      yield newVisibleTagCount;
+
+      // Remove tags until there is space for the collapse button on the last row.
+      let tags = [...domRef.current.children].filter(el => el.tagName !== 'BUTTON');
```
snowystinger, line 116: if the domRef can be null at run time above, isn't that true here as well?

packages/@react-spectrum/tag/test/TagGroup.test.js
reidbarber: Spectrum designs show the full count, so we can go with that for now.

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -70,6 +72,78 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
+
+  let updateVisibleTagCount = useCallback(() => {
+    let computeVisibleTagCount = () => {
+      // Refs can be null at runtime.
+      let currDomRef: HTMLDivElement | null = domRef.current;
+      if (!currDomRef) {
+        return;
+      }
+
+      let tags = [...currDomRef.children].filter(el => el.tagName !== 'BUTTON') as HTMLDivElement[];
```
snowystinger, line 86: if you track the width of each tag and also get the width of the button here, you can remove the loop that removes tags one by one, thereby reducing the number of renders from `n+1` to `2`

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -70,6 +72,78 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
+      for (let tag of tags) {
+        let {y} = tag.getBoundingClientRect();
+
+        if (y !== currY) {
+          currY = y;
+          rowCount++;
+        }
+
+        if (rowCount > maxRows) {
+          break;
```
snowystinger, line 100: to do the above comment, you'll have to not `break` since you need the widths of all the remaining tags

packages/@react-spectrum/tag/src/TagGroup.tsx
reidbarber: The collapse button doesn't exist yet here, but I've updated it to include a separate function for making space for the collapse button (and passing in the widths there), so number of renders should be 3.

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -361,7 +361,7 @@ describe('TagGroup', function () {
     );
 
     let tags = getAllByRole('gridcell');
-    expect(tags.length).toBe(4);
+    expect(tags.length).toBe(1);
```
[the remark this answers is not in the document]
reidbarber: This change was from adding additional `getBoundingClientRect` calls to the code. Test should still accurately test show/collapse behavior though.

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -76,62 +77,74 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
+
+        // Measure, and update to show the items until maxRows is reached.
+        let {index, tagWidths} = computeVisibleTagCount();
+        yield index;
+      
+        // Remove tags until there is space for the collapse button on the last row.
+        if (index < gridCollection.size) {
+          let collapseButtonWidth = [...domRef.current.children].find(el => el.tagName === 'BUTTON')?.getBoundingClientRect().width;
+          yield clearSpaceForCollapseButton({index, tagWidths, collapseButtonWidth});
+        }
```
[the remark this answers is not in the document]
reidbarber: Good idea, I've updated it to use this method.

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -70,33 +73,116 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
+          }
+          return acc;
+        }, [[], null]);
+        let currY = -Infinity;
+        let rowCount = 0;
+        let index = 0;
+        let tagWidths = [];
+        // Count rows and show tags until we hit the maxRows.
+        for (let tag of tags) {
+          let {y} = tag.getBoundingClientRect();
```
snowystinger, line 102:
```suggestion
          let {width, y} = tag.getBoundingClientRect();
```
then you can delete the getBoundingClientRect call below on 112

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -299,6 +299,85 @@ describe('TagGroup', function () {
...
+          <Item key="4" aria-label="Tag 4">Tag 4</Item>
+          <Item key="5" aria-label="Tag 5">Tag 5</Item>
+          <Item key="6" aria-label="Tag 6">Tag 6</Item>
+          <Item key="7" aria-label="Tag 7">Tag 7</Item>
+        </TagGroup>
+      </Provider>
+    );
+
+    let tags = getAllByRole('gridcell');
+    expect(tags.length).toBe(1);
```
snowystinger, line 364: if you log out these in the TagGroup code
```
        console.log('boundingClient tag', tag.outerHTML, width, y);
        console.log('boundingClient button', button.outerHTML, buttonWidth);
        console.log('boundingClient container', containerEnd);
        console.log('boundingClient last tag end', lastTagEnd);
```
you'll see what you need to change in the mock implementation
you're missing 'right'

snowystinger (commented):
it's getting pretty close :)

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -299,6 +299,48 @@ describe('TagGroup', function () {
     expect(onRemoveSpy).toHaveBeenCalledWith('1');
   });
 
+  it('maxRows should limit the number of tags shown', function () {
+    let offsetWidth = jest.spyOn(HTMLElement.prototype, 'getBoundingClientRect')
```
snowystinger, line 303: I assume these values were lifted from the browser? they are a bit hard to reason through as a human. I can't tell if they check for any edge cases. This could be easier if whole numbers were chose as well as rounded to numbers like 5 or 10 with specific instances where we don't do that to show we are testing an edge case.

snowystinger (dismissed):
Looks good to me, one comment on the test, but we don't need to address it now. Glad you were able to get the test working! :)

packages/@react-spectrum/tag/test/TagGroup.test.js
reidbarber: Good point, updated

snowystinger (dismissed):

reidbarber:
reidbot: https://reactspectrum.blob.core.windows.net/reactspectrum/f6204de4ba86c7132fcce0a400166cea81c6e244/storybook/index.html?path=/story/taggroup--maxrows-2

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -70,33 +73,116 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
+  let updateVisibleTagCount = useCallback(() => {
+    if (maxRows > 0) {
+      let computeVisibleTagCount = () => {
+        // Refs can be null at runtime.
+        let currDomRef: HTMLDivElement | null = domRef.current;
+        if (!currDomRef) {
+          return;
+        }
+
+        let [tags, button] = [...currDomRef.children].reduce((acc, el) => {
```
ktabors, line 88: What is acc short for?

packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
```diff
@@ -91,7 +91,19 @@ storiesOf('TagGroup', module)
         }
       </TagGroup>
     )
-  );
+  )
+  .add('maxRows: 2', () => (
```
ktabors, line 95: I want controls to test this with truncation (I know, narrow enough single column they truncate) and onRemove. Waiting for CSF 3 refactor. :)

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -299,6 +299,48 @@ describe('TagGroup', function () {
...
+      .mockImplementationOnce(() => ({x: 275, y: 300, width: 110, height: 32, top: 300, right: 385, bottom: 335, left: 275}))
+      .mockImplementationOnce(() => ({x: 200, y: 335, width: 65, height: 32, top: 335, right: 265, bottom: 370, left: 200}))
+      .mockImplementationOnce(() => ({x: 265, y: 335, width: 75, height: 32, top: 335, right: 345, bottom: 370, left: 265}))
+      .mockImplementationOnce(() => ({x: 200, y: 370, width: 120, height: 32, top: 370, right: 320, bottom: 400, left: 200}))
+      .mockImplementationOnce(() => ({x: 200, y: 400, width: 95, height: 32, top: 400, right: 290, bottom: 435, left: 200}))
+      .mockImplementationOnce(() => ({x: 200, y: 300, width: 200, height: 128, top: 300, right: 400, bottom: 435, left: 200}))
+      .mockImplementationOnce(() => ({x: 265, y: 335, width: 75, height: 32, top: 335, right: 345, bottom: 370, left: 265}));
+    let {getAllByRole, getByRole} = render(
+      <Provider theme={theme}>
+        <TagGroup maxRows={2} aria-label="tag group">
```
ktabors, line 314: Can we use a different maxRows number since storybook also uses 2?

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -299,6 +299,48 @@ describe('TagGroup', function () {
     expect(onRemoveSpy).toHaveBeenCalledWith('1');
   });
 
+  it('maxRows should limit the number of tags shown', function () {
```
ktabors, line 302: Shouldn't we test more cases? Right now I'm only thinking of one tag and three rows and there is no button.

ktabors: Is there a case where the user has max row one and a narrow width, in this case visibleTagCount is zero and showCollapseButton is true? Only the button shows, no tags show. I think so, but is that weird?

packages/@react-spectrum/tag/src/TagGroup.tsx
snowystinger: it's standard shorthand for accumulator in a reduce function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#grouping_objects_by_a_property

packages/@react-spectrum/tag/test/TagGroup.test.js
snowystinger: I think this is fine to keep, your other comment about more tests makes sense too

packages/@react-spectrum/tag/test/TagGroup.test.js
reidbarber: And I think always showing at least one tag make the most sense (which is the current behavior).

packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
reidbarber: Yeah, I think it might be worth going ahead and updating this to CSF3, and also adding some chromatic stories in this PR

reidbarber:
reidbot: https://reactspectrum.blob.core.windows.net/reactspectrum/740a6b0cb2d9538cb3fd07c771b1ca1bf64b777d/storybook/index.html?path=/story/taggroup--max-rows

snowystinger (approved):

ktabors (commented):
Thanks for switching the stories to CSF 3!

There seems to be some weird behavior with `maxRows`. I set the max to 5, that works, but as I continue to narrow it goes down to 4 rows like in this image.
image

Also, when max rows set to 1 and it gets too narrow it spills to 2 rows.
image

Should we set a minWidth on this so that at minimum it displays some of 1 tag and the show button?

This happens when resizing so I don't think it's a real use case. With many rows story there is a "ghost render" when it narrows where the button shows on a third row for a few milliseconds. Only mentioning in case this actually means there is a bug or issue this reveals.

reidbarber:
@ktabors Good catch with that first edge case, I just fixed this. With the second, I think that's desired behavior. I don't think we ever want to hide all tags and just show the button, and if someone is limiting the height to one row, I don't think they will also be restricting horizontal space.

snowystinger (commented):
Can we have a story that combines onRemove and maxRows? Just want to make sure the Tab behavior still works there.

reidbarber:
Reidbot: https://reactspectrum.blob.core.windows.net/reactspectrum/aa360a67f99409bc5ed32ef9ec9cacad2a4424a6/storybook/index.html?path=/story/taggroup--max-rows&providerSwitcher-express=false&providerSwitcher-toastPosition=bottom

snowystinger (commented):
In general, this looks good. One last thing I found while testing this, is that VO doesn't announce when you tab to the Show All button.
And it will announce EVERY tag that is shown when clicking show all, which could be extremely verbose if there are > ~3 tags hidden. I don't think it should announce any of them?
It also doesn't announce the hide all.

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -10,52 +10,154 @@
...
+            allowsRemoving={allowsRemoving}
+            onRemove={onRemove}>
+            {item.rendered}
+          </Tag>
+        ))}
+        {tagState.showCollapseButton &&
+          <ActionButton isQuiet onPress={handlePressCollapse}>
+            {isCollapsed ? `Show all (${state.collection.size})` : 'Show less '}
+          </ActionButton>
+        }
```
devongovett, line 158: I think this will be invalid aria because it's a button directly inside a `role="grid"` without any `row` or `gridcell` wrappers. Also, since you tab to these rather than arrow key navigating, maybe they should actually be outside the grid? Would that break the layout?

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -10,52 +10,154 @@
...
+            item={item}
+            state={state}
+            allowsRemoving={allowsRemoving}
+            onRemove={onRemove}>
+            {item.rendered}
+          </Tag>
+        ))}
+        {tagState.showCollapseButton &&
+          <ActionButton isQuiet onPress={handlePressCollapse}>
+            {isCollapsed ? `Show all (${state.collection.size})` : 'Show less '}
```
devongovett, line 156: Do we want to show the size of the collection here? What if the app hasn't actually loaded all the items yet. Then we'd be showing an incomplete number...

packages/@react-aria/tag/src/useTagGroup.ts
```diff
@@ -24,6 +24,18 @@ export interface TagGroupAria {
...
+  /**
+   * An optional keyboard delegate to handle arrow key navigation,
+   * to override the default.
+   */
+  keyboardDelegate?: TagKeyboardDelegate<T>,
+  /**
+   * Whether navigation through tab key is enabled.
+   */
+  allowsTabNavigation?: boolean
+}
```
devongovett, line 37: Do we need to expose these in the public API? Maybe if we move the buttons out of the grid `allowsTabNavigation` wouldn't be necessary. But also maybe you could just always set it to true? Not sure if we would ever have tabbable elements inside tags.

devongovett:
I think the button might need some custom styles to shrink its height since we don't have small size action buttons yet.

packages/@react-spectrum/tag/src/TagGroup.tsx
reidbarber: Yeah, I wanted to include it outside grid, but couldn't figure out how to get it on the last row of tags. I'll keep looking.

packages/@react-aria/tag/src/useTagGroup.ts
reidbarber: Agreed, I'll remove this if I move the button outside of the grid and fix the layout. If we can't do that, I can just set this to true.

reidbarber:
Reidbot: https://reactspectrum.blob.core.windows.net/reactspectrum/d3a0f99d0112068bc50a42079996b67eaf4b9b9a/storybook/index.html?path=/story/taggroup--max-rows&providerSwitcher-express=false&providerSwitcher-toastPosition=bottom

packages/@react-spectrum/tag/src/TagGroup.tsx
reidbarber: Managed to get it outside of the grid and keep the layout correct.

packages/@react-aria/tag/src/useTagGroup.ts
reidbarber: Removed this since the button is now outside of the grid.

reidbarber:
> Seems like when you make the width bigger, it doesn't display more tags now?

@devongovett If you refresh, it behaves correctly. Looking into why it's different on first load....

devongovett:
Hmm I tried refreshing and it still happened for me. Actually when I make the element smaller, it also wraps to show more than 2 lines: 
image

reidbarber:
@devongovett Yeah, looks like the new inline styles I introduced broke the overflow logic. Working on it now.

reidbarber:
Should be fixed now. One issue was with font loading, and I needed to set the resize observer to watch the new container. Also went ahead and added in translation strings for the button.

devongovett (approved):
Working well for me in Safari.

snowystinger (approved):
Working well for me in Chrome and on iPhone Safari

PR #2767, opened 2022-01-21
fix(#2766): Virtualizer does not scroll items into view within the browser viewport

majornista:
Add IntersectionObserver to detect whether or not a focused item is visible within the browser window, and if not scroll it into view.

Closes #2766 

## ✅ Pull Request Checklist:

- [x] Included link to corresponding [React Spectrum GitHub Issue #2766.
- [x] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [x] Filled out test instructions.
- [ ] Updated documentation (if it already exists for this component).
- [x] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/TR/wai-aria-practices-1.1/)

## 📝 Test Instructions:

1. Open https://reactspectrum.blob.core.windows.net/reactspectrum/d5eb632b807c8dda8faea50e3cbcc5995528d6bf/storybook/index.html?path=/story/listbox--async-loading&id=listbox--async-loading&providerSwitcher-toastPosition=bottom&viewMode=story
2. Reduce the inner height of the browser window to be less than the height of the ListBox within the story, ~200px should be good.
3. <kbd>Tab</kbd> to set keyboard focus to the first Item within the ListBox.
4. Use <kbd>ArrowDown</kbd> to scroll until you focus an Item within the ListBox that is outside of the browser viewport.
5. Notice that the browser will now automatically scroll the `document.body` to bring the focused Item into view.

Other stories to test: 

- https://reactspectrum.blob.core.windows.net/reactspectrum/d5eb632b807c8dda8faea50e3cbcc5995528d6bf/storybook/index.html?path=/story/tableview--async-loading&id=tableview--async-loading&providerSwitcher-toastPosition=bottom&viewMode=story
- https://reactspectrum.blob.core.windows.net/reactspectrum/d5eb632b807c8dda8faea50e3cbcc5995528d6bf/storybook/index.html?path=/story/useselectablelist--static-ul-static-sub-ul&id=useselectablelist--static-ul-static-sub-ul&providerSwitcher-toastPosition=bottom&viewMode=story
- https://reactspectrum.blob.core.windows.net/reactspectrum/d5eb632b807c8dda8faea50e3cbcc5995528d6bf/storybook/index.html?path=/story/focusscope--focus-safely&id=focusscope--focus-safely&providerSwitcher-toastPosition=bottom&viewMode=story

## 🧢 Your Project:
Adobe/Accessibility Adobe/Admin_Console Jira issues A11Y-5009 and TRON-8092

PR #3763, opened 2022-11-17
Fully scroll collection elements into view when keyboard navigating

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

NOTE: components in the storybook aren't immediately in view due to the additional scrolling, try pressing tab to bring them into view
Test that components that have keyboard navigation are scrolled into view upon keyboard navigating (e.g. TableView, ListView, Calendar, Tabs, etc).

**Build with "block: nearest, inline: nearest" as the main scroll strategy:** 
https://reactspectrum.blob.core.windows.net/reactspectrum/37d80d8801684badd21942d62c9e6ab7a66ebee0/storybook/index.html

**Build with "block: center, inline: center" as the main scroll strategy:** https://reactspectrum.blob.core.windows.net/reactspectrum/eee367b74cdb09810e36ba05248374831fa4c66a/storybook/index.html

If testing scrolling via keyboard on mobile/chrome emulator for VisualViewport calc, make sure to open the story outside the storybook iframe

Some cases to consider ():
EDIT: The team decided to that it would be good to have all interactable elements scroll into view upon keyboard interactions. This is to be addressed in a followup PR.
- Should sliders/ColorArea/TableView resizing/other slider components scroll into view when using arrow keys to shift their position
- Should popover trigger buttons be scrolled into view before opening their popover?
- Should other keyboard interactions scroll items into view? For example Picker's typeselect or using left/right arrow to cycle through the Picker's options. Same thing for NumberField/DateField + up/down arrows. What about components that 
- Should things always scroll into view via `block: nearest` or should we scroll things to the center of the screen? Native input elements are brought into the center of the screen if they are out of view and the user types/uses the left/right arrow key
   -  Team decided to try to scroll element to the center of the screen, try the build linked for a demo. I think doing "block: nearest" makes sense in most cases, "block: center" feels a bit odd when lists are partially in view 

## 🧢 Your Project:

RSP


LFDanLu:
Still testing performance + edge cases + intersectionObserver for visibility check.

packages/@react-aria/overlays/src/usePreventScroll.ts
```diff
@@ -47,13 +49,17 @@ export function usePreventScroll(options: PreventScrollOptions = {}) {
     }
 
     if (isIOS()) {
-      return preventScrollMobileSafari();
+      return chain(preventScrollMobileSafari(), setScrollPrevented());
```
snowystinger, line 52: this won't chain the return values

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -135,14 +136,18 @@ export function useGridCell<T, C extends GridCollection<T>>(props: GridCellProps
           e.preventDefault();
           e.stopPropagation();
           if (focusMode === 'cell' && direction === 'rtl') {
-            focusSafely(ref.current);
+            if (document.activeElement !== ref.current) {
+              focusSafely(ref.current);
+              scrollIntoViewFully(ref.current);
```
LFDanLu, line 141: This pattern of `focusSafely` -> `scrollIntoViewFully` repeats in a couple of other places, musing if I should move the `scrollIntoViewFully` call into `focusSafely`. My concern is that it is possible to be in keyboard modality and have `focusSafely` be triggered NOT by keyboard navigation (e.g. component w/ `autoFocus` renders)

packages/@react-aria/selection/src/useSelectableItem.ts
```diff
@@ -143,6 +143,13 @@ export function useSelectableItem(options: SelectableItemOptions): SelectableIte
       } else {
         focusSafely(ref.current);
       }
+
+      // TODO: is this the appropriate place to put this? We need the focused DOM node
+      // Also keyboard modality != keyboard navigation. If the manager.focusedKey changes programatically and not
+      // in response to keypress but the last interaction modality was keyboard then we'd perform the scroll
+      if (getInteractionModality() === 'keyboard') {
+        scrollIntoViewFully(ref.current);
+      }
```
LFDanLu, line 152: Still a bit wary about this, digging to see if there is anywhere in the code that actually will update the focusedKey in response to a non-keyboard input + when in keyboard modality

packages/@react-aria/slider/src/useSliderThumb.ts
```diff
@@ -121,7 +124,12 @@ export function useSliderThumb(
           setThumbValue(index, getThumbMaxValue(index));
           break;
       }
+
       setThumbDragging(index, false);
+      // Perform scroll after the thumb is fully positioned
+      requestAnimationFrame(() => {
+        scrollIntoViewFully(thumbRef.current, {block: 'center', inline: 'center'});
```
LFDanLu, line 131: the input of the slider thumb is actually visually hidden, so the IntersectionObserver will always think it isn't in the viewport. Added the `thumbRef` to get around this.

Also changed the scroll strategy to be  `block: 'center', inline: 'center'` which moves it to the center of the screen when using the end/page up/ etc to move the slider. I felt that using `block: 'nearest'` didn't really bring the slider into view enough and it may not be clear to the user that they just moved the slider thumb. Open to opinions here, try comparing the scrolling that happens when using the arrow keys to adjust the slider vs using page up/down/home/end

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +71,65 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+
+  // If scrolling is not currently prevented then we aren’t in a overlay nor is a overlay open, just use element.scrollIntoView to bring the element into view
+  if (!isScrollPrevented) {
+    target?.scrollIntoView?.(scrollOptions);
+  } else {
+    // If scrolling is prevented, we don't want to scroll the body since it will break the open overlay's positioning.
+    while (target && scrollParent && target !== root && scrollParent !== root) {
+      // TODO: adapt scrollIntoView so that it can work with the provided ScrollIntoViewOptions instead of mimicking scrollIntoView({block: 'nearest'}) only
+      // do this if we decide we actually want the scrollOption flexibility
+      scrollIntoView(scrollParent as HTMLElement, target as HTMLElement);
```
LFDanLu, line 130: Open to opinions here, do we feel it is valuable to allow `ScrollIntoViewOptions` to be passed to `scrollIntoViewFully`? IMO I like it for the slider experience commented above. 

This is perhaps a good argument to use something like https://scroll-into-view-if-needed.netlify.app/ to polyfill `scrollIntoViewIfNeeded` so that we can simplify all of this code since it allows us to customize the scroll options AND define scroll boundaries/propagation (aka prevent document body from scrolling if usePreventScroll is active).

packages/@react-aria/slider/src/useSliderThumb.ts
LFDanLu: Also we have other components that will scroll into view on keyboard nav (e.g. ActionGroup) via `.focus` which scrolls the newly focused element to the center of the screen. Might be nice to unify this behavior?

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +71,66 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+// Perhaps add a parameter to this func to customize the root so we can say how far up we want to actually adjust the scroll?
+// scrollIntoView is exported and available from aria/utils so is it a breaking change to replace it with this func?
+// TODO: test with zoom/pinch zoom
+function scrollIntoViewHelper(target: Element, scrollOptions: ScrollIntoViewOptions) {
+  let root = document.scrollingElement || document.documentElement;
+  let scrollParent = getScrollParent(target);
+
+  // If scrolling is not currently prevented then we aren’t in a overlay nor is a overlay open, just use element.scrollIntoView to bring the element into view
+  if (!isScrollPrevented) {
+    target?.scrollIntoView?.(scrollOptions);
```
LFDanLu, line 125: Note to self and for posterity: I tried changing TableView [here](https://github.com/adobe/react-spectrum/blob/b35d5c02fe900badccd0cf1a8f23bb593419f238/packages/%40react-spectrum/table/src/TableView.tsx#L428) so that it used scrollIntoViewFully with the options of `{{block: center, inline: center}}` + a ref to the resizer div itself instead of the visually hidden input (which has problem with the IntersectionObserver) and it turned out a bit jankey. It worked better when the resizer was completely out of view since it would scroll more of the table into view thus giving the user more context but it would also continually shift the table's position when moving the resizer to past right edge of the table via keyboard.

Additionally there seems to be some behavior differences when doing `document.activeElement?.scrollIntoView?.({block: 'nearest', inline: 'nearest'});`  vs `scrollIntoViewFully(document.activeElement, {block: 'nearest', inline: 'nearest'});` even though they result in an basically identical call of `element.scrollIntoView` with the same args. Not entirely sure why, but will drop that in favor of the approach in https://github.com/adobe/react-spectrum/pull/3786/files instead

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +70,64 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+  }
+
+  if (shouldScrollIntoView(target)) {
+    scrollIntoViewHelper(target, scrollOptions);
+  }
+}
+
+// Determines if a target element is partially/fully out of view and thus should be scrolled into view if possible.
+// Recurses through every scrollable parent of the target and returns true the target is not visible in any of the scrollable parents.
+function shouldScrollIntoView(target: Element, originalTarget?: Element) {
```
LFDanLu, line 89: Some problematic cases that I ran into:
- returning shouldScrollIntoView=true for partially out of view elements. Table row is always partially out of view if it is wider than the body
- table resizer/absolutely positioned elements. The resizer doesn't have a scrollable parent other than the root but can still be scrolled out of view since it is absolutely positioned and the table body is the scrollable element rather than the table header. If I switch the logic to look at each parentElement rather than the scrollParents, the resizer is always partially out of view of its parent element

packages/@react-aria/utils/src/scrollIntoView.ts
LFDanLu: Update: polyfill doesn't work with RTL so I'll work on modifying scrollIntoView in a followup PR when making all elements scroll into view on keyboard interaction if necessary

majornista:
I prefer `block: nearest, inline: nearest` in most cases, but testing using the build that does this, https://reactspectrum.blob.core.windows.net/reactspectrum/1d2e825a33bb4fe8761799e28673403626de61ec/storybook/index.html, there seem cases where the item within the collection does not scroll fully into view. I wonder if this has something to do with the fixed height of the component being taller than the height of the container, `300vh`.

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +70,56 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+  let {bottom: scrollParentBottom, top: scrollParentTop, left: scrollParentLeft, right: scrollParentRight} = scrollParent.getBoundingClientRect();
+
+  if (scrollParent === root) {
+    let viewportHeight = window.visualViewport?.height ?? window.innerHeight;
+    let viewportWidth = window.visualViewport?.width ?? window.innerWidth;
+    return (targetBottom < 0 || targetTop > viewportHeight || targetLeft > viewportWidth || targetRight < 0);
+  }
+
+  return targetBottom <= scrollParentTop || targetTop >= scrollParentBottom || targetLeft >= scrollParentRight || targetRight <= scrollParentLeft || shouldScrollIntoView(scrollParent, targetToCompare);
+}
```
majornista, line 109: I'm curious why this is preferable using `IntersectionObserver` as in https://github.com/adobe/react-spectrum/pull/2767/files#diff-2814ab64465ca7396b113c75472ed75d06046b918d1cc620e8225a7531c35c04R17-R47? In the ListBox example using `block: nearest` it seems like ListItems on the bottom of the viewport are not scrolling fully into view.

LFDanLu: There were concerns brought up about IntersectionObserver's performance impact vs just calculating if an element is in view. It felt like the IntersectionObserver was a bit overkill since it felt more geared towards things like infinite scrolling/observing the visibility of the element during its lifespan. I'll definitely do some testing for the performance between the two though.

As for the some ListBox items not scrolling into view, thanks for catching that, I'll take a look. It might be returning that those items are barely in view or something

LFDanLu: I also forgot to mention that IntersectionObserver will always return that a visually hidden element is not in view. This is expected but causes issues when we focus visually hidden elements like a slider's hidden input and perform keyboard operation on them. To a user, it seems as if the thumb of the slider is focused and thus is in view but to the scrollIntoView code it thinks we need to scroll the hidden input into view even though it is in the viewport. In the current PR this isn't an issue since I only scroll collection elements into view now but it would be an issue when we expand this scroll into view behavior to more components in followup work

devongovett:
Did we consider just calling `element.focus()` without the `preventScroll` option and letting the browser handle all the scrolling into view?

packages/@react-aria/calendar/src/useCalendarCell.ts
```diff
@@ -288,10 +288,8 @@ export function useCalendarCell(props: AriaCalendarCellProps, state: CalendarSta
 
       // Scroll into view if navigating with a keyboard, otherwise
       // try not to shift the view under the user's mouse/finger.
-      // Only scroll the direct scroll parent, not the whole page, so
-      // we don't scroll to the bottom when opening date picker popover.
-      if (getInteractionModality() !== 'pointer') {
-        scrollIntoView(getScrollParent(ref.current) as HTMLElement, ref.current);
+      if (getInteractionModality() === 'keyboard') {
```
devongovett, line 291: Does changing this if statement affect screen reader usage (where modality would be virtual). Also does changing it to fully scroll into view break the comment that was removed?

packages/@react-aria/selection/src/useSelectableItem.ts
```diff
@@ -143,6 +143,16 @@ export function useSelectableItem(options: SelectableItemOptions): SelectableIte
...
         focusSafely(ref.current);
       }
+
+      // TODO: perhaps move this into the keydown handler of useSelectableCollection so that we guarentee scrolling
+      // only happens on keyboard interaction. Do this when making followup PR to have useKeyboard handle scrolling
+      if (getInteractionModality() === 'keyboard') {
+        // Wait for virtualizer to scroll the item into view so we don't double scroll. Also fixes the jank behavior w/ non-virtualized lists
+        // and scrollOption block: center where the whole page scrolls to move the newly focused item to the center of the page.
+        requestAnimationFrame(() => {
+          scrollIntoViewFully(ref.current);
```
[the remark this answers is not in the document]
LFDanLu: just to clarify, is the suggestion here to also have some code in Virtualizer that calls scrollIntoViewFully after it scrolls the item into view?

LFDanLu:
> Did we consider just calling `element.focus()` without the `preventScroll` option and letting the browser handle all the scrolling into view?

@devongovett I leaned away from `.focus` because `element.scrollIntoView` allows us to customize the exact scroll behavior in various situations to a degree (aka `block: nearest` vs `block/inline: center) while `.focus` behaves like `block/inline: center` always, causing large jumps in scroll position

packages/@react-aria/calendar/src/useCalendarCell.ts
LFDanLu: Double checked via TalkBack/VoiceOver on iPad/Android/desktop, it doesn't seem to affect screen reader usage, cells still scroll into view when focused

> Also does changing it to fully scroll into view break the comment that was removed?

So `scrollIntoViewFully` won't cause scrolling on the body if `preventScroll` is active (aka overlay is open) so it doesn't break that comment.

devongovett:
Isn't it weird though that nearest ends up looking like this rather than scrolling the whole table into view?

image

If you just call `.focus()`, it centers it the same way as if you had tabbed to it from out of view.

LFDanLu:
I'd argue that using `.focus` introduces weirder behavior due to the large scroll position jumps that happen to center the item in view. Here is a recording of the scroll jumps that happen:

https://user-images.githubusercontent.com/9661127/211108433-ba74eea8-0e76-4849-8e02-14a1489e48df.mov
Feels like the user wouldn't expect such a large shift in scroll position, feels quite jarring

Since `.focus` only scrolls an item into view if it wasn't focused yet, I replaced the `focusSafely` with `.focus` [here](https://github.com/adobe/react-spectrum/blob/73d8be4dd7ff1713084cfe3f602f11aef1cecb07/packages/%40react-aria/grid/src/useGridCell.ts#L79-L85) (and of course got rid of the scrollIntoViewFully)

devongovett:
Is that because the entire table doesn't fit in the page viewport? When you make your browser window big enough, it seems ok.

LFDanLu:
Not necessarily an issue with fitting the table in the viewport, it will happen so long as the item you are keyboard navigating to is out of view -> the resulting scroll position shift causes the next item that you will keyboard navigate to be out of view  


https://user-images.githubusercontent.com/9661127/211112675-61a8b0a0-cc36-471c-874a-6312b178286f.mov

packages/@react-aria/datepicker/src/useDateSegment.ts
```diff
@@ -258,7 +258,7 @@ export function useDateSegment(segment: DateSegment, state: DateFieldState, ref:
 
   let onFocus = () => {
     enteredKeys.current = '';
-    scrollIntoView(getScrollParent(ref.current) as HTMLElement, ref.current);
+    scrollIntoViewFully(ref.current, {block: 'center', inline: 'center'});
```
LFDanLu, line 261: Right now this is the only non `block: nearest` scrollIntoViewFully call, but it brings the experience of keyboard navigating through DateSegments in Safari in line with other browsers. If the original intention was to have this behave like `block: nearest` (which I suspect to be the case with how `scrollIntoView` works), then I'd be happy to adjust. I do think the experience with `{block: 'center', inline: 'center'}` here is nice though since the whole DateField is brought into view and the component is small enough that there won't be weird scroll jumps like in Table when using block: center

LFDanLu:
Note to self: have people test the latest commit and compare it to the one before it on Monday

Issue #3070, opened 2022-04-26
Refactor TagGroup to focus row instead 


LFDanLu:
# 🙋 Feature Request

Discussed in testing session, would be nice if the Tags followed the same pattern as https://github.com/adobe/react-spectrum/pull/3000, aka focusing the row instead of the cell. If possible, move away from using grid hooks and replace with list hooks/state since Tags allow for focusable children. Perhaps try a FocusScope/FocusManager approach instead to navigate between the Tags instead.

Keep in mind that some of the grid specific code might be handling announcements (probably non-issue since we don't have a notion of selected items w/ TagGroup)

After the refactor, revisit the announcement made when focusing a Tag:
Removable tags will read "Cool Tag 2, Remove Cool Tag 2" but don't really indicate how I should remove the tag. Should we add something like "space or delete to remove tag"? (RS)

## 🧢 Your Company/Team

RSP

Issue #3074, opened 2022-04-26
TagGroup accessibility followup


LFDanLu:
# 🐛 Bug Report 
Decide what the behavior should be for the below

- (VO) Stories with disabled tags will read off "tag group table, 3 rows" but because one tag is disabled, I can only access two of them (RS)
  - Is the above announcement confusing enough that we should make disabled rows keyboard focusable? Would we expect VO users to crawl the DOM instead using VO shortcuts and thus they are able to focus the disabled tag in that fashion?
- VO announces nothing when I click the remove button on a Tag, it doesn't move focus to the TagGroup either (RS). 
  - What should the behavior be here? Where should focus go when a Tag is removed? Should clicking on a Tag move focus to the Tag (by extension, should clicking on a non-actionable TableView row move focus to that row? They currently share the same grid hooks hence the same behavior)

## 🧢 Your Company/Team

RSP

reidbarber:
Closing, as these are all either fixed or covered in other issues.

PR #3798, opened 2022-11-29
TagGroup: Add useTagGroupState and fix focus issues

reidbarber:
Closes https://github.com/adobe/react-spectrum/issues/3070, https://github.com/adobe/react-spectrum/issues/3074

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Open the TagGroup -> onRemove story. After removing a tag, focus should go to the next tag (or previous tag if no next tag) and you should still be able to keyboard navigate between tags using arrow keys.


packages/@react-stately/tag/src/useTagGroupState.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  let keyToRestoreOnRemove = useRef(null);
+
+  useEffect(() => {
+    // If the focused key is removed, restore focus to the tag before, or tag after if no tag before.
+    if (!listState.collection.getItem(state.selectionManager.focusedKey)) {
+      state.selectionManager.setFocusedKey(keyToRestoreOnRemove.current);
+    }
+    keyToRestoreOnRemove.current = listState.collection.getKeyBefore(state.selectionManager.focusedKey) || listState.collection.getKeyAfter(state.selectionManager.focusedKey);
+  }, [listState.collection, state.collection, state.selectionManager]);
```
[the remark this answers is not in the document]
LFDanLu: Also since it seemed like focus was already making it onto another tag, I'm surprised that useSelectableItem wasn't updating the selectionManager's focused key via https://github.com/adobe/react-spectrum/blob/b35d5c02fe900badccd0cf1a8f23bb593419f238/packages/%40react-aria/selection/src/useSelectableItem.ts#L158-L161. Perhaps that doesn't fire/isn't supposed to fire, didn't dig too deeply.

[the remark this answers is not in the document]
reidbarber: I think the focused key was being set, but the key was 'row-N' instead of 'N'. I'll take another look.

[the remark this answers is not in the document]
LFDanLu: If that is the case, that sounds similar to an issue we used to have in ListView when we were still using useGridRow + useGridCell for ListView if I recall correctly (I think it was the due to having the grid collection and the list collection but they contained different bits of information). 

Thinking out loud, maybe we could use useGridList and useGridListItem here instead and tear out the keydown handler from useGridList so that the TagKeyboadDelegate still works? No need to implement/change anything here, I haven't thought this one through just yet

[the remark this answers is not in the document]
reidbarber: > maybe we could use useGridList and useGridListItem here instead

This works, except for useGridListItem keydown taking Left/Right arrow keys. Maybe I can just update useTag to look similar to useGridListItem but remove that keydown handling and anything else not needed.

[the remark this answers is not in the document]
LFDanLu: Bleh, was deleting the keydown handler returned from useGridListItem insufficient? If that's the case perhaps evaluate what stuff would need to be moved into useTag from useGridListItem and we can see what approach is easiest/appropriate.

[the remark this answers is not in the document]
reidbarber: Ahh, yeah that does work. I thought I tried that, but I probably had something missing elsewhere. I'll go forward with trying this.

packages/@react-aria/tag/src/useTag.ts
```diff
@@ -28,26 +28,26 @@ export interface TagAria {
   clearButtonProps: ButtonHTMLAttributes<HTMLButtonElement>
 }
 
-export function useTag(props: TagProps<any>, state: GridState<any, any>): TagAria {
-  let {isFocused} = props;
-  const {
+export function useTag<T extends GridCollection<T>>(props: TagProps<T>, state: GridState<T, T>): TagAria {
```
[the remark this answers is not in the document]
reidbarber: Updated

packages/@react-spectrum/tag/src/Tag.tsx
```diff
@@ -21,7 +22,7 @@ import {useFocusRing} from '@react-aria/focus';
 import {useHover} from '@react-aria/interactions';
 import {useTag} from '@react-aria/tag';
 
-export function Tag<T>(props: SpectrumTagProps<T>) {
+export function Tag<T extends GridCollection<T>>(props: SpectrumTagProps<T>) {
```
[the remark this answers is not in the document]
reidbarber: Updated

packages/@react-stately/tag/src/useTagGroupState.ts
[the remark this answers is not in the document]
reidbarber: Updated to use `useGridList`. Should be working how we now. Also nows focuses the next item after deletion instead of previous, which seems more natural.

packages/@react-spectrum/tag/src/Tag.tsx
```diff
@@ -35,43 +35,43 @@ export function Tag<T>(props: SpectrumTagProps<T>) {
...
           slots={{
             icon: {UNSAFE_className: classNames(styles, 'spectrum-Tag-icon'), size: 'XS'},
-            text: {UNSAFE_className: classNames(styles, 'spectrum-Tag-content', {'tags-removable': allowsRemoving}), ...labelProps}
+            text: {UNSAFE_className: classNames(styles, 'spectrum-Tag-content'), ...labelProps}
           }}>
-
           {typeof children === 'string' ? <Text>{children}</Text> : children}
-          {allowsRemoving && <TagRemoveButton item={item} {...clearButtonProps} UNSAFE_className={classNames(styles, 'spectrum-Tag-action')} />}
+          <ClearSlots>
+            {allowsRemoving && <TagRemoveButton item={item} {...clearButtonProps} UNSAFE_className={classNames(styles, 'spectrum-Tag-action')} />}
```
LFDanLu, line 73: Minor clean up nit: since this is no longer using slots to pass props to the `TagRemoveButton` (and in fact doesn't seem like it was before), mind removing the `useSlotProps` and `clearBtnRef` (not sure what that one is for) in `TagRemoveButton`

packages/@react-stately/tag/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "type": "git",
+    "url": "https://github.com/adobe/react-spectrum"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.6.2",
+    "@react-stately/grid": "^3.4.1",
+    "@react-stately/list": "^3.6.0",
+    "@react-stately/utils": "^3.0.0",
+    "@react-types/tag": "3.0.0-beta.0",
+    "react-stately": "^3.19.0"
```
LFDanLu, line 30: Any reason to have the `react-stately` monopackage as a dep here? Also don't think you need `"@react-stately/grid"` and  `"@react-stately/utils": "^3.0.0",` anymore

```suggestion
    "@react-stately/list": "^3.6.0",
    "@react-types/tag": "3.0.0-beta.0",
```

packages/@react-types/tag/package.json
```diff
@@ -10,7 +10,9 @@
   },
   "dependencies": {
     "@react-stately/grid": "^3.4.1",
-    "@react-types/shared": "^3.16.0"
+    "@react-types/grid": "^3.1.5",
+    "@react-types/shared": "^3.16.0",
+    "react-stately": "^3.19.0"
```
LFDanLu, line 15:
```suggestion
    "@react-types/shared": "^3.16.0",
    "react-stately": "^3.19.0"
```
Also maybe the IDE autofilled the the `react-stately` here, prob can point directly to `react-stately/list`

LFDanLu (commented):
The behavior looks good to me now, tested on VoiceOver on iPad and TalkBack on Android. Switching to useGridList/useGridListItem seems to have made the cancel button focusable by TalkBack which was a nice bonus since we previously couldn't and thus couldn't delete a tag. 

There is now some iffy behavior that can be optimized/changed with the Talkback experience, namely how it will announce "press space/delete to delete the tag" (we are in Talkback so users probably don't need to hear that?), and how you can focus the row -> then the label -> then the clear button which results in some repetitive announcements. These we can address/figure out if we can fix in a separate PR/issue.

packages/@react-aria/tag/package.json
```diff
@@ -24,10 +24,14 @@
   "dependencies": {
     "@babel/runtime": "^7.6.2",
     "@react-aria/grid": "^3.5.1",
+    "@react-aria/gridlist": "^3.1.1",
     "@react-aria/i18n": "^3.6.2",
     "@react-aria/interactions": "^3.13.0",
+    "@react-aria/selection": "^3.12.0",
```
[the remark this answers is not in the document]
reidbarber: Thanks for catching these dependency issues. Updated these.

packages/@adobe/spectrum-css-temp/components/tags/index.css
```diff
@@ -58,28 +58,37 @@ governing permissions and limitations under the License.
...
-  outline: none;
-}
+  .spectrum-Tag-icon {
+    grid-area: icon;
+    margin-inline-end: var(--spectrum-global-dimension-size-100);
+  }
+
+  .spectrum-Tag-content {
+    grid-area: content;
+    block-size: 100%;
```
snowystinger, line 76: this height 100% is doing nothing because its parent doesn't have a height
you may not need a height at all on this, but if you do, then you need to add height 100% to this class' parent so it'll calculate based on the item which does have a height

packages/@react-aria/tag/src/TagKeyboardDelegate.ts
```diff
@@ -10,25 +10,25 @@
...
+  getFirstKey() {
+    return this.collection.getFirstKey();
   }
 
   getLastKey() {
-    let key = this.collection.getLastKey();
-    let item = this.collection.getItem(key);
-
-    return [...item.childNodes][0].key;
+    return this.collection.getLastKey();
```
snowystinger, line 29: this is likely to behave oddly when in the collapsed mode
just something to keep an eye out for, ActionGroup home/end don't do anything

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -296,44 +317,74 @@ describe('TagGroup', function () {
...
+        <TagGroup aria-label="tag group" allowsRemoving onRemove={onRemoveSpy}>
+          <Item key="1" aria-label="Tag 1">Tag 1</Item>
+          <Item key="2" aria-label="Tag 2">Tag 2</Item>
+          <Item key="3" aria-label="Tag 3">Tag 3</Item>
+        </TagGroup>
+      </Provider>
+    );
+
+    let tags = getAllByRole('row');
+    fireEvent.click(tags[0]);
```
snowystinger, line 336: best to use triggerPress
fireEvent will only fire the 'click' event, it won't do any of the pointer/mouse down

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -296,44 +317,74 @@ describe('TagGroup', function () {
...
+    expect(onRemoveSpy).not.toHaveBeenCalled();
+
+    let removeButton = within(tags[0]).getByRole('button');
+    fireEvent.click(removeButton);
```
snowystinger, line 340: same thing

packages/@react-spectrum/tag/test/TagGroup.test.js
```diff
@@ -296,44 +317,74 @@ describe('TagGroup', function () {
...
-  //
-  //   debug();
-  //
-  //   let tag = getByRole('row');
-  //   expect(tag).toHaveAttribute('aria-invalid', 'true');
-  // });
+    let tags = getAllByRole('row');
+    userEvent.tab();
+    expect(document.activeElement).toBe(tags[0]);
+    fireEvent.keyDown(document.activeElement, {key: props.keyPress});
```
snowystinger, line 382: pair with keyUp

packages/@react-stately/tag/src/useTagGroupState.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Provides state management for a TagGroup component.
+ */
+export function useTagGroupState<T extends object>(props: TagGroupProps<T>): TagGroupState<T> {
+  let state = useListState(props);
+
+  let keyToRestoreOnRemove = useRef(null);
+
+  useEffect(() => {
+    // If the focused key is removed, restore focus to the tag after, or tag before if no tag after.
+    if (!state.collection.getItem(state.selectionManager.focusedKey)) {
```
snowystinger, line 27: I don't think this is quite right, this will happen for reasons other than the Tag remove button being called

I think useTagGroupState should have a callback 'onRemove' that the aria implementation can call when a tag is removed. Then here we can manage focus at that point in time. It'll be before the next render that way as well, instead of being a render behind as it is now

if a tag is removed some other way, then it is likely that focus isn't on the taggroup at that time, so it's ok I think for now if focus is lost in those cases to the collection of tags

snowystinger (commented):
Noticed that the class `.spectrum-Tags-item` has 
```
  display: grid;
  grid-template-columns: 1fr auto;
  grid-template-areas: "icon content action";
```
but it should have a third column defined
```
  display: grid;
  grid-template-columns: auto 1fr auto;
  grid-template-areas: "icon content action";
```

packages/@react-stately/tag/src/useTagGroupState.ts
reidbarber: I thought that might be the case. Just updated based on this recommendation, thanks.

snowystinger (dismissed):

snowystinger (approved):

ktabors (dismissed):
The testing says "previous" tag the code in stately says "next", unless there is no "next" then it is "previous".

packages/@react-aria/tag/src/useTag.ts
```diff
@@ -10,77 +10,78 @@
...
-  // Don't want the row to be focusable or accessible via keyboard
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  let {tabIndex, ...otherRowProps} = rowProps;
 
-  let {gridCellProps} = useGridCell({
-    node: [...item.childNodes][0],
-    focusMode: 'cell'
-  }, state, tagRef);
+  // We want the group to handle keyboard navigation between tags.
+  delete rowProps.onKeyDownCapture;
```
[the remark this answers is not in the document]
reidbarber: We could set it to undefined down there, but it would do the same thing. And I think we want `onKeyDown` here, we're just removing that handler since it would handle focusing within a cell, which we don't want.

packages/@adobe/spectrum-css-temp/components/tags/index.css
```diff
@@ -58,28 +58,36 @@ governing permissions and limitations under the License.
...
-  outline: none;
-}
+  .spectrum-Tag-icon {
+    grid-area: icon;
+    margin-inline-end: var(--spectrum-global-dimension-size-100);
+  }
+
+  .spectrum-Tag-content {
+    grid-area: content;
+    line-height: calc(var(--spectrum-tag-height) - calc(var(--spectrum-tag-border-size) * 2));
```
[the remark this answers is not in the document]
reidbarber: This style was pre-existing. I think we can improve styles in a follow-up closer to the release of this component if we want. I'm not following the onRemove part.

majornista (approved):

snowystinger (dismissed):

packages/@react-types/list/src/index.d.ts
```diff
@@ -11,4 +11,5 @@
  */
 
 export type {AriaGridListProps, GridListProps} from '@react-aria/gridlist';
+export type {ListState} from '@react-stately/list';
```
LFDanLu, line 14: The exports here for `AriaGridListProps`, `GridListProps`, and `SpectrumListViewProps` were kept in here to preserve back compat, but these props are now exported from their specific packages. IMO we can go ahead and just export ListState from `@react-stately/list` if needed, where did you need this type specifically?

packages/@react-stately/tag/src/useTagGroupState.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Provides state management for a TagGroup component.
+ */
+export function useTagGroupState<T extends object>(props: TagGroupProps<T>): TagGroupState<T> {
+  let state = useListState(props);
+
+  let onRemove = (key) => {
+    // If a tag is removed, restore focus to the tag after, or tag before if no tag after.
+    let restoreKey = state.collection.getKeyAfter(key) || state.collection.getKeyBefore(key);
+    state.selectionManager.setFocusedKey(restoreKey);
```
LFDanLu, line 30: Can be follow up, but I just noticed that focus seems to be lost when deleting a tag via TalkBack, not sure why that is.

LFDanLu (approved):
Just a couple of things of small things I noticed on this final pass but def can handle in a followup instead

packages/@react-types/list/src/index.d.ts
reidbarber: Good point, I went ahead and removed this export

packages/@react-stately/tag/src/useTagGroupState.ts
reidbarber: We can investigate in a follow-up before release.

snowystinger (approved):

LFDanLu (approved):
LGTM

PR #3809, opened 2022-12-01
fix(#3804): usePopover: non-modal popover allows page and popover to scroll when target has position: fixed

### LINKED ISSUE #3804: usePopover: non-modal popover allows page and popover to scroll when target has position: fixed
# 🐛 Bug Report 

A non-modal popover allows page and popover to scroll even when the target is fixed.

The new documentation search box feature in development includes a SearchAutocomplete combo box in a page header that has `position: fixed`

1. Open https://reactspectrum.blob.core.windows.net/reactspectrum/d5aed9e2838b40d54ba7e7ba7a7a77f8fec212c6/docs/react-spectrum/index.html
2. Enter some characters to search into the SearchAutocomplete, which should open a popover displaying the results.
3. With the popover open, use the mouse wheel to scroll the page.

## 🤔 Expected Behavior

The popover should not scroll with the underlying page, but should remain positioned relative to the SearchAutocomplete.

## 😯 Current Behavior

The popover scrolls with the page on top of the SearchAutocomplete contained in the page header with `position: fixed`.

## 💁 Possible Solution

`isNonModal` should not disable `usePreventScroll` at:

https://github.com/adobe/react-spectrum/blob/b35d5c02fe900badccd0cf1a8f23bb593419f238/packages/%40react-aria/overlays/src/usePopover.ts#L99-L107

## 🌍 Your Environment

| react-spectrum   |  @react-aria/overlays@3.12.0
| Browser          |  All
| Operating System | macOS

## 🧢 Your Company/Team

Adobe/Accessibility

#3383

majornista:
Closes #3804

## ✅ Pull Request Checklist:

- [x] Included link to corresponding [React Spectrum GitHub Issue #3804](https://github.com/adobe/react-spectrum/issue/3804).

## 🧢 Your Project:

Adobe/Accessibility

### DIFF
--- a/packages/@react-spectrum/combobox/stories/ComboBox.stories.tsx
+++ b/packages/@react-spectrum/combobox/stories/ComboBox.stories.tsx
@@ -509,6 +509,40 @@ storiesOf('ComboBox', module)
         {renderRow({validationState: 'invalid', isQuiet: true})}
       </Flex>
     )
+  )
+  .add(
+    'Fixed positioning',
+    () => (
+      <>
+        <div
+          style={{
+            position: 'fixed',
+            top: 'calc(50vh - var(--spectrum-global-dimension-size-400) - 1.3 * var(--spectrum-global-dimension-size-150)'
+          }}>
+          <AsyncLoadingExample />
+        </div>
+        <div
+          style={{
+            position: 'relative',
+            height: '200vh'
+          }}>
+          <div
+            style={{
+              position: 'absolute',
+              top: '0'
+            }}>
+            Top
+          </div>
+          <div
+            style={{
+              position: 'absolute',
+              bottom: '0'
+            }}>
+            Bottom
+          </div>
+        </div>
+      </>
+    )
   );
 
 


LFDanLu (commented):
Tested in VO and Talkback, confirmed that I was still able to interact with the ComboBox even with the dropdown open with these changes. However, I was no longer able to scroll the list of options on iPad [here](https://reactspectrum.blob.core.windows.net/reactspectrum/0ee71a531fcce40bbb58a4e0954f8902a23b117e/docs/react-spectrum/ComboBox.html), most likely due to how usePreventScroll prevents default on certain touch events. Additionally, due to the scroll position handling that happens, the dropdown extends past the edge of the screen on iPad (though we have a different issue without this change due to the virtual keyboard that appears)

I'll need to think about what we can do here instead or if we need to allow position: fixed to be provided to a overlay

majornista:
@LFDanLu I agree that we will probably need to account for targets that have or inherit `position: fixed` by making the overlay also have `position: fixed` and by taking that into account when positioning the overlay.

LFDanLu:
Hm, definitely weird that the storybook example works on iPad but it doesn't in the docs (maybe something to do with the storybook not being scrollable?). Maybe indicative of a bug in the usePreventScroll code since it should allow scrolling in nested scroll regions now that I'm looking at the code again

packages/@react-aria/overlays/src/usePopover.ts
```diff
@@ -92,8 +92,18 @@ export function usePopover(props: AriaPopoverProps, state: OverlayTriggerState):
...
+  // Delay preventing scroll until popover is positioned to avoid extra scroll padding.
+  // This requires a layout effect so that positioning has been committed to the DOM
+  // by the time usePreventScroll measures the element.
+  let [isPositioned, setPositioned] = useState(false);
+  useLayoutEffect(() => {
+    if (placement) {
+      setPositioned(true);
+    }
+  }, [isNonModal, placement]);
+
```
ktabors, line 104:
```suggestion
```
Don't think we need the useLayoutEffect.

packages/@react-aria/overlays/src/usePopover.ts
```diff
@@ -92,8 +92,18 @@ export function usePopover(props: AriaPopoverProps, state: OverlayTriggerState):
...
   usePreventScroll({
-    isDisabled: isNonModal
+    isDisabled: !isPositioned
```
ktabors, line 106:
```suggestion
```
We'd like to test this PR with the useLayoutEffect removed and usePreventScroll() always set. Looking at related PRs we think the issues are related to iPad async example.

devongovett:
Looks like the original issue was closed by something else.

PR #3892, opened 2023-01-06
Support landmark navigation and toasts across iframes

devongovett:
Depends on #3868

This adds the necessary APIs to implement landmark navigation across iframe boundaries. In addition, this enables toasts to pop out of iframes and for the user to navigate between the toasts and iframe content.

It works by adding a custom DOM event named `react-aria-landmark-navigation`, which can be canceled by calling `preventDefault` to override the default behavior. This event is triggered during landmark navigation when we reach the end of the landmark sequence in either direction. By default, we wrap around to the first/last landmark again, but if `preventDefault` is called on this event we do not. This enables the iframe page to fire a message out of the iframe, where the F6 key is re-dispatched so that the landmark navigation on the outer page is applied. The iframe itself is made a landmark as well, and an additional `focus` option is supported so that when landmark navigation reaches the iframe itself, focus can be redirected back inside. This makes it behave as is there were a single landmark navigation sequence across the whole page.

This PR also fixes a couple bugs:

1. Landmarks inside `aria-hidden` trees were not skipped.
2. When adding an element while `ariaHideOutside` is active, it could hide `data-react-aria-top-layer` elements (toasts). Now we apply the tree walker to added elements to ensure this is not the case.


snowystinger (dismissed):

LFDanLu:
Danielbot: https://reactspectrum.blob.core.windows.net/reactspectrum/8607ba1547e60214f5ff0c595dcdabf00d3de2b6/storybook/index.html?path=/story/accordion--default&providerSwitcher-express=false&providerSwitcher-toastPosition=bottom

LFDanLu:
Retested toasts in Android Talkback and Windows FF NVDA. Confirmed that Talkback properly announces the Toasts as they appear but Windows FF NVDA still has intermittent behavior where it will suddenly stop announcing the Toasts. Doesn't seem like the ariaHideOutside is blocking it there.

devongovett:
I can look into the NVDA thing separately.

packages/@react-spectrum/toast/stories/Toast.stories.tsx
```diff
@@ -146,3 +153,99 @@ function MultipleInner() {
...
+    // When the iframe is re-focused, restore focus back inside where it was before.
+    window.addEventListener('focus', () => {
+      if (prevFocusedElement) {
+        prevFocusedElement.focus();
+        prevFocusedElement = null;
+      }
+    });
+
+    // Move focus to first or last landmark when we receive a message from the parent page.
+    window.addEventListener('message', e => {
```
LFDanLu, line 199: Just a story thing I noticed: if you move out of the dialog in the iframe and onto the toast via F6, you can't get back inside the iframe. Maybe this is because the `main` is aria-hidden due to the dialog being open?

LFDanLu (approved):
LGTM for the most part, verified that aria-hide outside, landmarks, and toast seems to be working as expected. Just one thing I noticed with the iframe story, but that is story implementation specific and be looked at later with the shell work

packages/@react-aria/landmark/test/useLandmark.test.tsx
```diff
@@ -1156,4 +1156,122 @@ describe('LandmarkManager', function () {
...
+
+    expect(onLandmarkNavigation).toHaveBeenCalledTimes(1);
+    expect(onLandmarkNavigation.mock.calls[0][0].detail).toEqual({
+      direction: 'backward'
+    });
+
+    window.removeEventListener('react-aria-landmark-navigation', onLandmarkNavigation);
+  });
+
+  it('goes skips over aria-hidden landmarks', function () {
```
reidbarber, line 1234: nit:
```suggestion
  it('skips over aria-hidden landmarks', function () {
```

reidbarber (approved):

packages/@react-spectrum/toast/stories/Toast.stories.tsx
devongovett: Yep that's exactly it. There are no landmarks visible in the iframe anymore so landmark navigation skips over them.

Issue #2766, opened 2022-01-21
Collections do not scroll items into view within the browser viewport with keyboard


majornista:
# 🐛 Bug Report 

When the bounds of a virtualized collection extend beyond the browser viewport, and setting focus to an item within the collection that is out of view, the browser will not scroll the focused item into view.

## 🤔 Expected Behavior

An item receiving focus in a virtualized collection should not only scroll the item into view within the collection, but it should scroll the item into view within the browser as well.

## 😯 Current Behavior

1. Open https://reactspectrum.blob.core.windows.net/reactspectrum/0661785f30971bdbcbd291282b26836cf272c3f8/storybook/iframe.html?id=listbox--async-loading&viewMode=story
2. Reduce the inner height of the browser window to be less than the height of the ListBox within the story,  ~200px should be good.
3. Tab to set keyboard focus to the first Item within the ListBox.
4. Use <kbd>ArrowDown</kbd> to scroll until you focus an Item within the ListBox that is outside of the browser viewport.
5. Notice that the browser does not automatically scroll the `document.body` to bring the focused Item into view.

## 💁 Possible Solution

Add the following to `useVirtualizer` method in `@react-aria/virtualizer/src/Virtualizer.tsx`:

```tsx
  // Define an IntersectionObserver to evaluate whether the browser should scroll the element receiving focus into view.
  let intersectionObserver = useMemo(() => {
    const intersectionObserverOptions:IntersectionObserverInit = {
      root: undefined,
      rootMargin: '0px', 
      threshold: 1
    };

    const scrollIntoViewOptions:ScrollIntoViewOptions = {
      behavior: 'auto',
      block: 'nearest',
      inline: 'nearest'
    };

    const intersectionObserverCallback = (entries:Array<IntersectionObserverEntry>, observer:IntersectionObserver) => {
      entries.forEach(entry => {
        if (!entry.isIntersecting) {
          entry.target.scrollIntoView(scrollIntoViewOptions);
        }
        observer.unobserve(entry.target);
      });
    };

    return new IntersectionObserver(intersectionObserverCallback, intersectionObserverOptions);
  }, []);
```
Then at the end of the `onFocus` method, add the following after updating `isFocusWithin.current`:

```tsx
    // Evaluate whether the browser should scroll to bring element receiving focus into view.
    if (isFocusWithin.current) {
      intersectionObserver.observe(e.target);
    }
```

## 🔦 Context

[...]

PR #3899, opened 2023-01-10
Fully scroll collection elements into view when keyboard navigating

LFDanLu:
Closes https://github.com/adobe/react-spectrum/issues/2766

EDIT: overzealous when thinking this fixed the other scroll into view issues, seems like it still happens (albeit rarely locally) 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test that components that have keyboard navigation are scrolled into view upon keyboard navigating e.g. TableView, ListView, Calendar (standalone and in overlay) , Tabs, non-virtualized collections (useSelectableList stories, useTab stories). Use the "scrolling" checkbox to enable scroll in the stories (you'll have to tab to the component to bring it into view).

Also test popovers in general since we've updated how the positioning coords are applied to the popover. 

Note: behavior at varying pinch zoom levels/on mobile emulation is a bit finicky, digging

## 🧢 Your Project:

RSP


packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +71,33 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+    let {left: originalLeft, top: originalTop} = targetElement.getBoundingClientRect();
+
+    // TODO: For some reason in the useSelectableList stories, originalTop and newTop differ by 1px, causing us to unecessarily
+    // scroll the containing element even though it may already be fully in view
+    // This is because the scrollable parent (ul) has a border of 1 that partially obscures the focused item even after scrollIntoView is called...
+    // Perhaps modify scrollIntoView calculations to take into account the borders, bleh...
+
+    // use scrollIntoView({block: 'nearest'}) instead of .focus to check if the element is fully in view or not since .focus()
+    // won't cause a scroll if the element is already focused and doesn't behave consistently when an element is partially out of view horizontally vs vertically
+    targetElement?.scrollIntoView?.({block: 'nearest'});
```
devongovett, line 87: Cool that this works. I had assumed that `scrollIntoView` would always align the element, but looks like "nearest" doesn't scroll if it is already in view?

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +71,33 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+
+export function scrollIntoViewport(targetElement: Element, containingElement?: Element) {
+  // If scrolling is not currently prevented then we aren’t in a overlay nor is a overlay open, just use element.scrollIntoView to bring the element into view
+  if (!isScrollPrevented) {
+    let {left: originalLeft, top: originalTop} = targetElement.getBoundingClientRect();
+
+    // TODO: For some reason in the useSelectableList stories, originalTop and newTop differ by 1px, causing us to unecessarily
+    // scroll the containing element even though it may already be fully in view
+    // This is because the scrollable parent (ul) has a border of 1 that partially obscures the focused item even after scrollIntoView is called...
+    // Perhaps modify scrollIntoView calculations to take into account the borders, bleh...
```
devongovett, line 83: Does changing `box-sizing` to `border-box` have any effect?

packages/@react-aria/utils/src/scrollIntoView.ts
LFDanLu: Right, from my testing `scrollIntoView({block: 'nearest'})` seems to behave like `scrollIntoViewIfNeeded`

packages/@react-aria/utils/src/scrollIntoView.ts
LFDanLu: Mm, good idea, I tried applying `box-sizing: border-box` to the story's `ul` but unfortunately it still reproduces the issue. Lemme poke around some more

LFDanLu: Using `let maxY = y + scrollView.clientHeight;` in `scrollIntoView` gets us pretty close but takes into account both top/bottom borders which borks the calculations a bit

packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -360,9 +360,10 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
       let element = scrollRef.current.querySelector(`[data-key="${manager.focusedKey}"]`) as HTMLElement;
       if (element) {
         scrollIntoView(scrollRef.current, element);
+        scrollIntoViewport(element, ref.current);
```
snowystinger, line 363: both and opposite signatures?

packages/@react-aria/virtualizer/src/Virtualizer.tsx
```diff
@@ -141,10 +141,19 @@ export function useVirtualizer<T extends object, V, W>(props: VirtualizerOptions
       } else {
         virtualizer.scrollToItem(focusedKey, {duration: 0});
       }
+
+      if (modality === 'keyboard' && ref.current.contains(document.activeElement)) {
+        // Wait till virtualizer scrolls the focused key into view before checking if we need to scroll it into the viewport
+        // Specifically need to do this for the column headers since they aren't part of the scrollable table body and their position
+        // syncing is a bit delayed compared to when scrollToItem is fired.
+        requestAnimationFrame(() => {
```
snowystinger, line 149: don't forget to cleanup the raf in case we unmount mid raf
I think there were a couple others in the other files, make sure they are all cleaned up

packages/@react-aria/selection/src/useSelectableCollection.ts
LFDanLu: yeaaah. The ideal (?) flow is as follows:
1. Scroll the newly focused element into view with respect to the scrollView. Limit this strictly to the first scrollParent (pre-existing behavior aka `scrollIntoView(scrollRef.current, element);`). This means the element may still be out of view for a variety of reasons (nested scroll regions, collection is extends past the viewport, something is covering the element) 
2. Scroll everything in such a way that the element is actually visible to the user (aka scrollIntoViewport(element, ref.current);`. If `preventScroll` is on, we only scroll up to the body so that we don't accidentally shift any overlays that maybe open 

The opposite signatures is a bit unfortunate, the second arg to `scrollIntoViewport` is the `containerElement` and is optional hence why it is the second arg (e.g. if we were scrolling a button into view, that case isn't actually implemented in this PR). I'll think a bit about the api here

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +76,29 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+  // If scrolling is not currently prevented then we aren’t in a overlay nor is a overlay open, just use element.scrollIntoView to bring the element into view
+  if (!isScrollPrevented) {
+    let {left: originalLeft, top: originalTop} = targetElement.getBoundingClientRect();
+
+    // use scrollIntoView({block: 'nearest'}) instead of .focus to check if the element is fully in view or not since .focus()
+    // won't cause a scroll if the element is already focused and doesn't behave consistently when an element is partially out of view horizontally vs vertically
+    targetElement?.scrollIntoView?.({block: 'nearest'});
+    let {left: newLeft, top: newTop} = targetElement.getBoundingClientRect();
+    // Account for sub pixel differences from rounding
+    if ((Math.abs(originalLeft - newLeft) > 1) || (Math.abs(originalTop - newTop) > 1)) {
```
LFDanLu, line 90: Note that the original positioning and the new positioning of the target element won't differ at all if the target was "out of view" from pinch zooming. It will still be scrolled into view due to `targetElement?.scrollIntoView?.({block: 'nearest'});` but the position w/ respect to the layout viewport is still the same.

packages/@react-aria/utils/src/getScrollParent.ts
```diff
@@ -11,14 +11,18 @@
  */
 
 export function getScrollParent(node: Element): Element {
+  if (isScrollable(node)) {
+    node = node.parentElement;
+  }
```
LFDanLu, line 16: Modified this so we don't just return the same node as its scroll parent if it happens to be scrollable. Caused a infinite loop [here](https://github.com/adobe/react-spectrum/blob/3039b89b78730e7f825ef5af36ed01e1fd8010e7/packages/%40react-aria/utils/src/scrollIntoView.ts#L98-L102) otherwise

.storybook/preview.js
```diff
@@ -25,11 +25,7 @@ export const parameters = {
 };
 
 export const decorators = [
```
LFDanLu, line 27: @snowystinger Just a heads up, I know you had previously suggested looking into `layout: centered` as per https://storybook.js.org/docs/react/configure/story-layout. Tried it out, it unfortunately means our stories won't take up the full canvas, hence we would still need a decorator to make the story full screen (aka: adapt to the viewports height + width etc).

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -22,20 +30,25 @@ export function scrollIntoView(scrollView: HTMLElement, element: HTMLElement) {
   let height = element.offsetHeight;
   let x = scrollView.scrollLeft;
   let y = scrollView.scrollTop;
-  let maxX = x + scrollView.offsetWidth;
-  let maxY = y + scrollView.offsetHeight;
+
+  // Account for top/left border offsetting the scroll top/Left
+  let {borderTopWidth, borderLeftWidth} = window.getComputedStyle(scrollView);
```
snowystinger, line 35: getComputedStyle is safe enough, we're already forcing layout with offsetHeight a couple lines up

packages/@react-aria/virtualizer/src/Virtualizer.tsx
```diff
@@ -141,10 +142,25 @@ export function useVirtualizer<T extends object, V, W>(props: VirtualizerOptions
       } else {
         virtualizer.scrollToItem(focusedKey, {duration: 0});
       }
+
+      if (modality === 'keyboard' && ref.current.contains(document.activeElement)) {
+        // Wait till virtualizer scrolls the focused key into view before checking if we need to scroll it into the viewport
+        // Specifically need to do this for the column headers since they aren't part of the scrollable table body and their position
+        // syncing is a bit delayed compared to when scrollToItem is fired.
+        raf = requestAnimationFrame(() => {
```
snowystinger, line 150: same as above comment

packages/@react-aria/calendar/src/useCalendarCell.ts
```diff
@@ -283,17 +283,27 @@ export function useCalendarCell(props: AriaCalendarCellProps, state: CalendarSta
...
       // try not to shift the view under the user's mouse/finger.
-      // Only scroll the direct scroll parent, not the whole page, so
-      // we don't scroll to the bottom when opening date picker popover.
+      // If in a overlay, scrollIntoViewport will only cause scrolling
+      // up to the overlay scroll body to prevent overlay shifting
       if (getInteractionModality() !== 'pointer') {
-        scrollIntoView(getScrollParent(ref.current) as HTMLElement, ref.current);
+        // If this calendar is in a popover, we need to wait till it is fully positioned before attempting to
+        // scroll it into view, otherwise it won't be brought into view properly
+        raf = requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
LFDanLu: I don't believe this runs into the same problem that the FocusScope bug was running into right? This uses a variable initialized in the useEffect instead of setting it to a ref outside of the useEffect. The new raf is only created if any of the dependencies change and the cleanup will cancel the pre-existing raf before the new one is then created.

packages/@react-aria/selection/src/useSelectableCollection.ts
LFDanLu: Just to clarify, the suggestion here is to flip `scrollIntoView`'s arguments to match `scrollIntoViewport` arg order? Or is it to make `scrollIntoView`'s signature the following -> `scrollIntoView(objectArg, element)` where objectArg is `{element: Element, scrollRef: Element}`? Or is it to modify `scrollIntoViewport`'s signature?

packages/@react-aria/calendar/src/useCalendarCell.ts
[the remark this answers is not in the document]
snowystinger: o, yep, you are right. I didn't notice it was inside the useEffect

packages/@react-aria/selection/src/useSelectableCollection.ts
snowystinger: make both of them into
`functionName(objectArg)` where objectArg is what you outlined
you'd need to maintain the current signature for scrollIntoView as well, which if the args aren't optional, then it's pretty easy to distinguish, one arg would be the new signature and 2 would be the old, you'd use TS overload syntax

snowystinger (approved):
it's working well for me in chrome, both storybook and docs

majornista (dismissed):

packages/@react-aria/selection/src/useSelectableCollection.ts
LFDanLu: Hm, I see. I'm a bit hesitant to combine them into a single function since they fulfill slightly different purposes. `scrollIntoView` scrolls a element into view with respect to its scroll parent (not necessarily scrolling it visually into the viewport) and `scrollIntoViewport` brings a element into the viewport by centering its containing element (not necessarily the scroll parent or anything) in the viewport. It also uses `scrollIntoView` in the case where preventScroll is active so that function needs to exist as a utility

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -10,32 +10,67 @@
...
  * Similar to `element.scrollIntoView({block: 'nearest'})` (not supported in Edge),
  * but doesn't affect parents above `scrollView`.
  */
-export function scrollIntoView(scrollView: HTMLElement, element: HTMLElement) {
-  let offsetX = relativeOffset(scrollView, element, 'left');
-  let offsetY = relativeOffset(scrollView, element, 'top');
-  let width = element.offsetWidth;
-  let height = element.offsetHeight;
+export function scrollIntoView(element: HTMLElement, opts: ScrollOpts);
+export function scrollIntoView(element: HTMLElement, opts: HTMLElement | ScrollOpts) {
```
LFDanLu, line 38: See discussion [here](https://github.com/adobe/react-spectrum/pull/3899#discussion_r1066458214) for background on this change. This will also give us some wiggle room in the future if/when we expand the scroll behavior to mimic things like `block: center` instead

snowystinger (dismissed):

packages/@react-aria/utils/src/scrollIntoView.ts
snowystinger: shouldn't need the override since you've specified that the second arg can be either

snowystinger: I think the overrides is overkill, it's simple enough to just allow either of them, and we can use the one that is more readable in our code. the other can just be a shortcut 🤷🏻 
less confusion about which one is deprecated too

LFDanLu: fair enough, pushed the update

reidbarber (approved):
Behavior looks good, just a few small comments.

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -10,32 +10,59 @@
...
+}
+
+interface ScrollOpts {
+  scrollView: HTMLElement
+}
+
 /**
  * Scrolls `scrollView` so that `element` is visible.
  * Similar to `element.scrollIntoView({block: 'nearest'})` (not supported in Edge),
  * but doesn't affect parents above `scrollView`.
```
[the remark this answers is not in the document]
LFDanLu: Mm, yeah I can see how it would be confusing now with `opts: HTMLElement | ScrollOpts` as the signature. I kinda want to bring back the overloads that were removed [here](https://github.com/adobe/react-spectrum/pull/3899/commits/371eede4b6222aa2d0de5c10fa0fb6655bebc48c) since changing this description to be "Scrolls opts.scrollView so that element is visible" feels a bit unclear, especially since `opts` is actually the `element` in the old signature that this still supports.

@snowystinger how do you feel about the above?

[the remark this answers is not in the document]
snowystinger:
```
/**
 * Scrolls `scrollView` in ScrollOptions so that `element` is visible.
 * Similar to `element.scrollIntoView({block: 'nearest'})` (not supported in Edge),
 * but doesn't affect parents above `scrollView`.
 * `scollView can be passed directly in place of the ScrollOptions.`
```
and whenever we add any options to the ScrollOpts, we can add to the last sentence. ', as a shorthand for *whatever default options we want to specify*'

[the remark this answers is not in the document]
snowystinger: does that feel better?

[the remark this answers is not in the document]
LFDanLu: Well, thats the thing, `scrollView` doesn't actually substitute as `opts` if passed directly since the old signature was the following:
`scrollIntoView(scrollView: HTMLElement, element: HTMLElement)` 
versus the new signature
`scrollIntoView(element: HTMLElement, opts: HTMLElement | ScrollOpts)`

This means `element` is actually mapped to `scrollView` and `opts` is mapped to `element` if the user is providing a `HTMLElement` to opts (aka they are using the old signature). This feels pretty confusing and hard to specify clearly in a single description :/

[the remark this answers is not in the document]
snowystinger: ahhhhh, sorry, i got mixed up then. please keep the override syntax

packages/@react-aria/virtualizer/src/Virtualizer.tsx
snowystinger, line 150: I know ScrollView has an onScrollEnd, is it possible to use that here instead

packages/@react-aria/calendar/src/useCalendarCell.ts
snowystinger, line 297: I think we technically need to be delayed by a render cycle, not a full ~16ms
Could see set some state here instead, then in response to that state change call the scrollIntoViewport?

snowystinger (dismissed):
Approving; works well. I have two questions about the rafs to see if we can get rid of them. But if you've already tried everything, then this is fine.

packages/@react-aria/virtualizer/src/Virtualizer.tsx
LFDanLu: Ah I hadn't considered this, definitely an intriguing idea since it reduces the randomness of when this raf executes w/ regards to the Virtualizer/ScrollView's scrolling. I tried it out but this raf actually needs to trigger even if the ScrollView doesn't perform a scroll, specifically for the case where the table itself is partially out of view.  

Ideally, what this would do is 
1. focusedKey changes
2. wait for the scrollview to scroll. If it doesn't need to scroll, then immediately proceed to the next step
3. check if the document.activeElement needs to be scrolled into view

Still thinking about a way to do step 2, maybe something in virtualizer state can give me that info

packages/@react-aria/calendar/src/useCalendarCell.ts
devongovett: This is also pretty specific to calendars in a popover so it's is a bit weird to put in here. I guess this would technically be the same for _any_ component inside a popover that called scroll into view. Is there a way to disable scroll into view until popovers are positioned more generally?

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -10,32 +10,68 @@
  * governing permissions and limitations under the License.
  */
 
+import {getScrollParent} from './';
+
+let isScrollPrevented = false;
+
+export function setScrollPrevented(value: boolean) {
```
devongovett, line 17: A bit hesitant about exposing this global state

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +99,39 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+    let {left: newLeft, top: newTop} = targetElement.getBoundingClientRect();
+    // Account for sub pixel differences from rounding
+    if ((Math.abs(originalLeft - newLeft) > 1) || (Math.abs(originalTop - newTop) > 1)) {
+      opts?.containingElement?.scrollIntoView?.({block: 'center', inline: 'center'});
+      targetElement.scrollIntoView?.({block: 'nearest'});
+    }
+  } else {
+    let root = document.scrollingElement || document.documentElement;
+    let scrollParent = getScrollParent(targetElement);
+    // If scrolling is prevented, we don't want to scroll the body since it might move the overlay partially offscreen and the user can't scroll it back into view.
```
devongovett, line 130: What's the case where this is needed? Don't we set `overflow: hidden` on the body, so it wouldn't be scrollable anyway? What happens if you just call `element.scrollIntoView`?

packages/@react-aria/virtualizer/src/Virtualizer.tsx
LFDanLu: Tried, using `virtualizer.getView` to check if the focused key is in view within the virtualizer, but it always returns as a non-null value inside the useEffect hrm...

packages/@react-aria/utils/src/scrollIntoView.ts
LFDanLu: Even if `overflow: hidden` is set on the body, `element.scrollIntoView` will can still scroll the element into view. 

https://user-images.githubusercontent.com/9661127/213599990-a81f6c77-6cbb-42f0-8a17-5a98e2630ab9.mov

The current case where this is needed is if the Datepicker popover is opened off screen, causing the entire page to scroll to bring it into view if we just use `element.scrollIntoView`, something that maybe I can mitigate using a long timeout or something. However, this may become a problem down the line if there ever is a case where usePreventScroll is active and the user performs some interaction that calls scrollIntoViewport

packages/@react-aria/calendar/src/useCalendarCell.ts
LFDanLu: Well, the information for if a popover has finished positioning is [here](https://github.com/adobe/react-spectrum/blob/38a57d3360268fb0cb55c6b42b9a5f6f13bb57d6/packages/%40react-aria/overlays/src/usePopover.ts#L98), so I'd need to have scrollIntoViewport somehow receive that information + know that the target element is within a popover.

packages/@react-aria/virtualizer/src/Virtualizer.tsx
LFDanLu: Some supplementary information: `onScrollStart` occur after this useEffect happens, meaning I cannot use onScrollStart as a way to set the global `delayScrolling` to delay scrolling until the column headers position themselves. Additionally, `virtualizer.visibleViews` includes more columns than are visually visible so `virtualizer.getView` won't help here. Gonna try using the visible rect of the virtualizer vs the layoutInfo of the focused key, might be promising

LFDanLu: Implemented in https://github.com/adobe/react-spectrum/pull/3943, but ran into some issues

Problematic cases:
- using the focusedKey's view's layoutInfo rect to check if it is partially out of view of the virtualizer's visible rect breaks if the focusedKey is the row itself.  The row might be very long and thus will always think that focusing it will need to cause a scroll in ScrollView
  -  Might be able to get around this by changing the intersection logic a bit, will need to think on this (perhaps if the width/height of the focused rect is greater than the visible rect, then we go with a block: nearest strategy and deem it only in need of scrolling if the top left corner isn't in view)

- relying on ScrollView's onScrollEnd is a bit delayed (can be seen by partially scrolling the bottom of the table out of view and then pressing arrow down to bring a row that isn't in view into view). Will need to find the earliest time possible where the column header position has finished scrolling to match the ScrollView's new position

packages/@react-aria/calendar/src/useCalendarCell.ts
LFDanLu: Updated, had to create another piece of global state so the popover could inform the calendar cell when it was finished positioning and that it should let scrollIntoViewport happen now. I don't feel that great about that global state but didn't feel that just delaying it by a render cycle was really appropriate (feels more brittle and reliant on implementation), lemme know if you have any suggestions on getting rid of the global state

packages/@react-aria/virtualizer/src/Virtualizer.tsx
LFDanLu: Pushed the change from above, lemme know what you think

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -10,32 +10,83 @@
...
+  scrollView: HTMLElement
+}
+
+interface ScrollIntoViewportOpts {
+  /** The optional containing element of the target to be centered in the viewport. */
+  containingElement?: Element
+}
+
+let isScrollPrevented = false;
+let delayScrolling = false;
```
LFDanLu, line 26: Open to any ideas about getting rid of these globals

packages/@react-aria/overlays/src/usePopover.ts
```diff
@@ -112,6 +113,14 @@ export function usePopover(props: AriaPopoverProps, state: OverlayTriggerState):
     }
   }, [isNonModal, state.isOpen, popoverRef]);
 
+  useLayoutEffect(() => {
+    if (state.isOpen && !isPositioned) {
+      setDelayScrolling(true);
+    } else {
+      setDelayScrolling(false);
+    }
```
snowystinger, line 121: if it unmounts while open, we probably need to do cleanup?

packages/@react-aria/utils/src/scrollIntoView.ts
snowystinger: Have any ideas? We can't invert it so that the popover state of isPositioned is used, the dependency already is popover depends on utils, not utils depending on popover.

Can't use context because it's the aria hooks that are calling these and it seems worse to pass this into the hooks.

If we watched for attributes in the dom, we'd be too delayed since those are async fired.

We could try some different names that are less about the state we're setting and more about the action that's happening? popoverIsPositioning? not sure what else to do about scrollingPrevented though...

LFDanLu: Thanks for the brainstormed alternatives, I actually hadn't considered the DOM attributes one.

> We could try some different names that are less about the state we're setting and more about the action that's happening? popoverIsPositioning? not sure what else to do about scrollingPrevented though...

IMO, I'd prefer keeping it about the state we are setting since I'm uncertain what else might call `usePreventScroll` in the future.

packages/@react-aria/overlays/src/usePopover.ts
LFDanLu: Ah, good catch, I suppose this could render but unmount before positioning finishes.

packages/@react-aria/utils/src/scrollIntoView.ts
devongovett: could you check `isScrollable(document.scrollingElement)`?

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -10,32 +10,83 @@
...
  * but doesn't affect parents above `scrollView`.
+ * @deprecated - provide `scrollView` via opts instead.
  */
-export function scrollIntoView(scrollView: HTMLElement, element: HTMLElement) {
-  let offsetX = relativeOffset(scrollView, element, 'left');
-  let offsetY = relativeOffset(scrollView, element, 'top');
-  let width = element.offsetWidth;
-  let height = element.offsetHeight;
+export function scrollIntoView(scrollView: HTMLElement, element: HTMLElement);
+export function scrollIntoView(element: HTMLElement, opts: HTMLElement | ScrollOpts) {
```
devongovett, line 54: Could you explain the API change here? Is it necessary?

LFDanLu: nope not necessary, more along the lines of making the `scrollIntoView` and `scrollIntoViewport` signatures more alike, see https://github.com/adobe/react-spectrum/pull/3899#discussion_r1066458214 for some context.

packages/@react-aria/utils/src/scrollIntoView.ts
LFDanLu: Unfortunately, `isScrollable(document.scrollingElement)` returns false even if the page is scrollable. The `scrollingElement` (aka <html>) could have `overflow: visible` but a child div/element might have `overflow: auto`/`overflow: scroll` (aka body/root div/etc) and thus be scrollable. 

I guess I could query if `document.scrollingElement/documentElement` has `overflow: hidden` applied on it instead then, hmm

LFDanLu: Pushed the update, checking for `overflow: hidden` works for our standard and mobile safari prevent scroll strategies. Note that there is still a reliance on delaying the `scrollIntoViewport` call until the popover is fully positioned (aka `usePreventScroll` is delayed in `usePopover` until positioning is finished)

packages/@react-aria/utils/src/scrollIntoView.ts
devongovett: hmm not quite sure why we need both `scrollIntoView` and `scrollIntoViewport` as you mentioned there - doesn't `scrollIntoViewport` first scroll the container and then the element? what happens if you don't call `scrollIntoView` first?

Either way, I'd leave this function as it was (it's confusing to have two signatures that are opposites of each other), and maybe make the new one match this one, or just leave it. The options object isn't really needed if there's only one option (which is required?), unless we think there will be a lot more in the future.

Should `containerElement` in the other one even be optional? It seems to be provided in every call to the function so maybe that doesn't need to be an options object either?

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -442,22 +442,63 @@ function TableVirtualizer({layout, collection, focusedKey, renderView, renderWra
...
+        }
+      } else if (
+        focusedKeyRect.x < visibleRect.x ||
+        focusedKeyRect.maxX > visibleRect.maxX ||
+        focusedKeyRect.y < visibleRect.y ||
+        focusedKeyRect.maxY > visibleRect.maxY
+      ) {
+        syncScrollPos = true;
+      }
+    }
```
devongovett, line 477: Won't `scrollIntoViewport` handle scrolling the cell into view if needed? Maybe I'm not understanding what this extra logic is for...

devongovett (commented):
Yay no more globals!!

packages/@react-aria/utils/src/scrollIntoView.ts
LFDanLu: > hmm not quite sure why we need both scrollIntoView and scrollIntoViewport as you mentioned there - doesn't scrollIntoViewport first scroll the container and then the element? what happens if you don't call scrollIntoView first?

If we don't call `scrollIntoView` first and only rely on calling `scrollIntoViewport` then we run into the case where we shift the container's position needlessly when scrolling an item into view. See the below:

https://user-images.githubusercontent.com/9661127/214989391-9a4b32e5-f24f-4fe0-bf3d-37f1a6eb06b6.mov

The items being scrolled into view is "out of view" (aka not visible) and thus `targetElement?.scrollIntoView?.({block: 'nearest'});` will cause a position change. Since there was a position change, the `scrollintoViewport` [logic](https://github.com/adobe/react-spectrum/blob/0d06149c05549653cd27feba22dff6bbf4ac8cdc/packages/%40react-aria/utils/src/scrollIntoView.ts#L117-L121) will think we need to center the container. This logic is meant for a scenario like this:

https://user-images.githubusercontent.com/9661127/214991325-6e34384c-0178-4ead-adda-4b96c8faca29.mov

where we decided that it would good for the user to get the full context of where their focus moved and thus opted to scroll the container to the center of the screen. Therefore the ideal flow that I could think of was the two step process mentioned https://github.com/adobe/react-spectrum/pull/3899#discussion_r1066458214, hence the need to call `scrollIntoView` and then `scrollIntoViewport`.

> Should containerElement in the other one even be optional? It seems to be provided in every call to the function so maybe that doesn't need to be an options object either?

This was a thought for the future for when we would scroll any standalone component into view upon keyboard interaction (e.g. hitting enter on a MenuTrigger button that was scrolled off screen). I figured `options` would accept a `scrollStrategy` (i.e. 'nearest', 'center', etc) so that we could have the element be scrolled into the center of the screen instead of just the edge. In the case of these standalone components, the `containerElement` may not exist and thus is optional

> Either way, I'd leave this function as it was (it's confusing to have two signatures that are opposites of each other), and maybe make the new one match this one, or just leave it. The options object isn't really needed if there's only one option (which is required?), unless we think there will be a lot more in the future.

 I'll go ahead and revert `scrollIntoView` signature to what it used to be then. Happy to discuss `scrollIntoViewport`'s signature further

packages/@react-aria/overlays/src/useOverlayPosition.ts
```diff
@@ -115,20 +115,27 @@ export function useOverlayPosition(props: AriaPositionProps): PositionAria {
...
+      boundaryElement,
+      offset,
+      crossOffset,
+      maxHeight
+    });
+
+    // Modify overlay styles directly so positioning happens immediately without the need of a second render
+    // This is so we don't have to delay autoFocus scrolling or delay applying preventScroll for popovers
+    let positionPx = {maxHeight: position.maxHeight != null ? position.maxHeight + 'px' : undefined};
+    Object.keys(position.position).forEach(key => positionPx[key] = position.position[key] + 'px');
```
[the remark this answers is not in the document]
LFDanLu: heh, when I tried doing `Object.assign(overlayRef.current.style, {...position.position, maxHeight: position.maxHeight})`  it actually didn't work because the values in `position.position` are all numbers and not "px" values and thus the styles weren't actually applied at all.

packages/@react-spectrum/table/src/TableView.tsx
LFDanLu: This is solving for a similar issue as described in my [comment here](https://github.com/adobe/react-spectrum/pull/3899#discussion_r1088508307). When 
```
virtualizer.scrollToItem(key, {
      duration: 0,
...
```
is called to scroll a table cell into view, the table column header's scroll position isn't actually synced immediately, resulting in something that looks like this:
image

In the image above, I moved focus from "Column 5" to "Column 6" (which is slightly out of view) and you can see that the body of the table has scrolled to reflect the new position. The table column header row only updates its scroll position via https://github.com/adobe/react-spectrum/blob/0d06149c05549653cd27feba22dff6bbf4ac8cdc/packages/%40react-spectrum/table/src/TableView.tsx#L507-L520, with `onScroll` occurring before the useEffect and being the call that syncs the header row's position with the body. However, that means we trigger `scrollIntoViewport` before fully repositioning the table column header row scroll position, resulting in https://github.com/adobe/react-spectrum/blob/0d06149c05549653cd27feba22dff6bbf4ac8cdc/packages/%40react-aria/utils/src/scrollIntoView.ts#L120-L122 being triggered and shifting the table unnecessarily. To solve this, I simply update the scroll position of the header row early, and only if I determine that the column would have been scrolled into view already.

I tried previously moving my scrollIntoViewport call out into a useLayoutEffect/useEffect that would trigger on `focusedKey` change instead but that ran into all kinds of issues where it was too early (both were before the `onScroll` call that would update the column header's scroll position)

packages/@react-aria/overlays/src/useOverlayPosition.ts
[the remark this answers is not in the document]
devongovett: Yeah I meant like

```js
Object.keys(position.position).forEach(key => overlayRef.current.style[key] = position.position[key] + 'px');
```

packages/@react-spectrum/table/src/TableView.tsx
devongovett: That part I get. Why is syncing the header scroll position conditional though? Does it hurt to always keep them in sync?

LFDanLu: Ah I see. I had initially wanted to limit the scope of the change but it doesn't seem to have any problematic effects, pushed

ktabors (commented):
Tabbed to the DateRange popover open button, scrolled away, pressed enter and the popover opened like in the following image. There was no page/viewport scrolling and we can't scroll because a popover is open. The popover opens partially within the viewport and not next to the button that opened it. If I close the popover the page has still not moved because there was no scroll. Note: if you do horizontal scroll instead of vertical the entire popover opens within the view. Bug?

image

packages/@react-aria/utils/src/scrollIntoView.ts
```diff
@@ -63,3 +75,37 @@ function relativeOffset(ancestor: HTMLElement, child: HTMLElement, axis: 'left'|
...
+      let {left: newLeft, top: newTop} = targetElement.getBoundingClientRect();
+      // Account for sub pixel differences from rounding
+      if ((Math.abs(originalLeft - newLeft) > 1) || (Math.abs(originalTop - newTop) > 1)) {
+        opts?.containingElement?.scrollIntoView?.({block: 'center', inline: 'center'});
+        targetElement.scrollIntoView?.({block: 'nearest'});
+      }
+    } else {
+      let scrollParent = getScrollParent(targetElement);
+      // If scrolling is prevented, we don't want to scroll the body since it might move the overlay partially offscreen and the user can't scroll it back into view.
+      while (targetElement && scrollParent && targetElement !== root && scrollParent !== root) {
```
ktabors, line 104: Why is this a `while` and not an `if`?

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -442,22 +442,35 @@ function TableVirtualizer({layout, collection, focusedKey, renderView, renderWra
     transitionDuration
   });
 
+  let scrollToItem = useCallback((key) => {
```
ktabors, line 445: I know this wasn't previous in the aria package, but is there a reason it isn't? I could see someone wanting this.

ktabors (commented):
I like the way it behaves.

LFDanLu:
@ktabors Did you scroll the entire datepicker out of view when you noticed that bug? If so, then that is a separate issue not handled by this PR. We've had some discussions about making all elements scroll into view upon keyboard interaction to fix such a case but punted for future PRs due to the number of edge cases/complexity that were discovered.

packages/@react-aria/utils/src/scrollIntoView.ts
LFDanLu: This is a while because we iterate through every scroll body and adjust the scroll positioning of the target in such a way that it is "visible" in its scrollable parent. The scenario this covers is if you had a bunch of nested scroll regions and each were scrolled in such a way that they obscured the target and thus just scrolling the original target and its parent scroll region wouldn't be sufficient.

packages/@react-spectrum/table/src/TableView.tsx
LFDanLu: This had some values/things specific to our tableview implementation (scroll duration of 0, and handling due to how we make the checkbox column sticky).

snowystinger (approved):

devongovett (approved):

andresgutgon:
Hi @LFDanLu I just came across this PR that I think is related with one think is happening to me:

I have a text input and on `ArrowDown` in that input I do:
``` javascript
const state = useListState<T>({
  items: options,
  children,
  selectionBehavior: 'toggle',
  selectionMode: 'multiple'
})
const onArrowDown = () => {
  state.selectionManager.setFocusedKey(state.collection.getFirstKey(), 'first')
  state.selectionManager.setFocused(true)
}
```
But the `useListBox` first element is scrolled this way:
image

## The question
Any idea why is `scrolling into view` the first item but in reality is scrolling out of view?

LFDanLu:
@andresgutgon Mind sharing a codesandbox for the above example? I assume this is a non-virtualized listbox which means it should be hitting [this portion of the code](https://github.com/adobe/react-spectrum/blob/d7cb26e27cac7b43f6fdff68ce4c10c1185d1bb3/packages/%40react-aria/selection/src/useSelectableCollection.ts#L358-L366). If that ListBox is in a overlay that prevents scrolling, we would be using [this](https://github.com/adobe/react-spectrum/blob/d7cb26e27cac7b43f6fdff68ce4c10c1185d1bb3/packages/%40react-aria/utils/src/scrollIntoView.ts#L25-L53) instead of the browser's `scrollIntoView` which means it doesn't take into account elements that might phyiscally cover the item

andresgutgon:
Hey @LFDanLu sorry for the delay. Your request of doing a codesandbox served me as rubber duck 🦆 . I found my problem.

In this sandbox I replicated part of my implementation. 
https://codesandbox.io/s/sleepy-keller-qjygke?file=/src/hooks/useKeyboarEvents.ts

The issue is not present in that sandbox but because I'm wrapping my component inside a [Radix popover](https://www.radix-ui.com/docs/primitives/components/popover) and for some reason inside it `useListBox` things it needs to scroll down pass first item as in the image. As you said is because [this code](https://github.com/adobe/react-spectrum/blob/d7cb26e27cac7b43f6fdff68ce4c10c1185d1bb3/packages/%40react-aria/selection/src/useSelectableCollection.ts#L358-L366)

So the fix was to `event.preventDefault` inside the [useKeyboard](https://react-spectrum.adobe.com/react-aria/useKeyboard.html) I'm using. 

Thanks! 🎉

PR #3938, opened 2023-01-19
TagGroup: Field support

reidbarber:
Adds label, description, error message, and contextual help support to TagGroup. 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test new story at TagGroup: 'with label, description, contextual help'. Use controls to test combinations of validation states.


snowystinger (commented):
Looks like this story broke? https://reactspectrum.blob.core.windows.net/reactspectrum/74a747396df9a7ee2e471c1189228b03a9d44d9c/storybook/index.html?path=/story/taggroup--with-action-and-max-rows&providerSwitcher-express=false
the tags won't expand to fill

packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
```diff
@@ -82,6 +119,7 @@ export const OnRemove: TagGroupStory = {
 };
 
 export const Wrapping: TagGroupStory = {
+  args: {width: '100%'},
```
snowystinger, line 122: I'm worried about needing this 'width' prop on everything
can we include it in the component somehow?

and I think I just realized why, this is because fields are default 192px isn't it?
could we put the width 100% inside our component though and pass that to Field? then if a prop with a specific width is passed in, we pass that through to Field instead?

Or, should all TagGroups be 192 width... unless someone specifies the width like you have?

reidbarber: This is just to demonstrate the wrapping behavior in storybook. These stories have a resizable container decorator, and this just ensures that it fills that container.

snowystinger: but it should just fill the resizable container on its own? Like ActionGroup does https://reactspectrum.blob.core.windows.net/reactspectrum/59f432c581d1ca0da5ae71b1436c0a41f87f8156/storybook/index.html?path=/story/actiongroup--overflow

reidbarber: You're right they do fill. Now that `width: auto` on the field is set, they no longer need width: '100%' to fill.

reidbarber: The 'maxRows + onRemove' story still won't fill unless it has `width: '100%'`. Looking into this.

ktabors (approved):
There is a conflict otherwise this looks good.

reidbarber:
Action button seems to immediately lose focus now. Not sure what's causing it, looking into it.

reidbarber:
> Action button seems to immediately lose focus now. Not sure what's causing it, looking into it.

Needed to update which ref was being passed into useTagGroup. Should be fixed now.

snowystinger:
Noticed that tags don't line up with other form elements.
I noticed that in the TagField contributions, there is always a border or line to show it's a Field, so it's ok that what is inside doesn't line up since the border does. I wonder if maybe TagGroup shouldn't be in Forms and instead TagField would be? or maybe we get rid of the leading margin, which will be hard for multiple rows since this isn't flex
Screenshot 2023-01-23 at 1 08 05 PM

packages/@react-aria/tag/src/useTagGroup.ts
```diff
@@ -44,9 +51,9 @@ export function useTagGroup<T>(props: AriaTagGroupProps<T>, state: TagGroupState
   let {direction} = useLocale();
   let keyboardDelegate = props.keyboardDelegate || new TagKeyboardDelegate(state.collection, direction);
   let {gridProps} = useGridList({...props, keyboardDelegate}, state, ref);
+  let {labelProps, fieldProps, descriptionProps, errorMessageProps} = useField(props);
```
LFDanLu, line 54: Noticed that `useGridList` is warning about missing a `aria-label`/`aria-labelledby` in the form stories. You'll want to call this `useField` hook before `useGridList` and supply the `fieldProps` (I think?) to `useGridList`. I wonder if there might be an issue with how `useGridList` uses its own generated `id` and not the user provided one though, don't quite remember why that is if there is even a reason for that at all...

packages/@react-spectrum/form/stories/Form.stories.tsx
```diff
@@ -30,6 +30,7 @@ import {SearchField} from '@react-spectrum/searchfield';
 import {StatusLight} from '@react-spectrum/statuslight';
 import {storiesOf} from '@storybook/react';
 import {Switch} from '@react-spectrum/switch';
+import {TagGroup} from '@react-spectrum/tag';
```
ktabors, line 33: In the Form story "isDisabled", it makes the TagGroup `description` greyed out like the rest of the form components. Should this Form prop also disable `allowsRemoving` if that prop is set? Also disable the action buttons?

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -12,21 +12,23 @@
...
 import styles from '@adobe/spectrum-css-temp/components/tags/vars.css';
 import {Tag} from './Tag';
+import {useFormProps} from '@react-spectrum/form';
 import {useId, useLayoutEffect, useResizeObserver, useValueEffect} from '@react-aria/utils';
 import {useLocale, useLocalizedStringFormatter} from '@react-aria/i18n';
 import {useProviderProps} from '@react-spectrum/provider';
 import {useTagGroupState} from '@react-stately/tag';
 
-export interface SpectrumTagGroupProps<T> extends AriaTagGroupProps<T>, StyleProps {
+export interface SpectrumTagGroupProps<T> extends AriaTagGroupProps<T>, StyleProps, SpectrumLabelableProps, Validation, SpectrumHelpTextProps {
```
LFDanLu, line 31: Think `Validation` should be a part of the `AriaTagGroupProps`?

packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
```diff
@@ -54,7 +56,52 @@ export default {
...
+    contextualHelp: {
+      table: {
+        disable: true
+      }
+    },
+    isRequired: {
+      control: 'boolean'
+    },
+    necessityIndicator: {
+      control: 'select',
```
ktabors, line 81: Why is this control a `select`? In other stories the control for this prop is a radio. We should pick one and use everywhere. Ditto for other controls here.

packages/@adobe/spectrum-css-temp/components/tags/index.css
```diff
@@ -112,3 +112,11 @@ governing permissions and limitations under the License.
     }
   }
 }
+
+.spectrum-Tags-fieldWrapper.spectrum-Tags-fieldWrapper {
+  width: unset;
+  
+  &:not(.spectrum-Tags-fieldWrapper--positionSide) {
+    display: grid;
```
LFDanLu, line 120: Out of curiosity, what case did this cover? Was Field's `inline-grid` not working?

ktabors (commented):
Are we going to add Chromatic stories for TagGroup in followup?

LFDanLu (changes requested):

packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
LFDanLu: Weighing in here, I like the select more since it takes less room, especially if there are a large number of controls.

packages/@adobe/spectrum-css-temp/components/tags/index.css
snowystinger: correct, inline-grid won't expand as a block level element would in a parent with no defined width, so if it collapses, it'd never un-collapse

packages/@react-aria/tag/src/useTagGroup.ts
LFDanLu: Did some more digging, don't think there is a good reason for `useGridList` to not accept a user provided id so you can probably go ahead and make https://github.com/adobe/react-spectrum/blob/c183944ce6a8ca1cf280a1c7b88d2ba393dd0252/packages/%40react-aria/gridlist/src/useGridList.ts#L84 be `let id = useId(props.id);` and pass `fieldProps` to `useGridList`

packages/@adobe/spectrum-css-temp/components/tags/index.css
```diff
@@ -112,3 +112,11 @@ governing permissions and limitations under the License.
     }
   }
 }
+
+.spectrum-Tags-fieldWrapper.spectrum-Tags-fieldWrapper {
+  width: unset;
```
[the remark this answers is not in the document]
reidbarber: See comment here: https://github.com/adobe/react-spectrum/pull/3938#discussion_r1084681497

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -146,56 +149,69 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
     <FocusScope>
-      <div
-        ref={containerRef}
-        {...styleProps}
-        className={classNames(styles, 'spectrum-Tags-container', styleProps.className)}>
+      <Field
+        {...props}
+        labelProps={labelProps}
+        descriptionProps={descriptionProps}
+        errorMessageProps={errorMessageProps}
```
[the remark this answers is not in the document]
reidbarber: These are due to those fields having `aria-invalid=true`, which isn't something this component will have.

packages/@react-aria/tag/src/useTagGroup.ts
```diff
@@ -56,10 +63,15 @@ export function useTagGroup<T>(props: AriaTagGroupProps<T>, state: TagGroupState
   let domProps = filterDOMProps(props);
   return {
     tagGroupProps: mergeProps(gridProps, domProps, {
+      role: state.collection.size ? 'grid' : undefined,
```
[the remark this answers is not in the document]
reidbarber: Good catch, it should be null so it doesn't get overridden by gridProps.role

packages/@react-spectrum/form/stories/Form.stories.tsx
```diff
@@ -264,6 +265,14 @@ function render(props: any = {}) {
           </ContextualHelp>
         )} />
       <TextField label="Zip code" description="Please enter a five-digit zip code." errorMessage="Please remove letters and special characters." />
+      <TagGroup label="Favorite tags" description="Select your favorite tags." errorMessage="Incorrect combination of tags.">
```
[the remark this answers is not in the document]
snowystinger: you need to turn on the control for "showErrorIcon" too

packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
```diff
@@ -159,6 +206,19 @@ export const WithActionAndMaxRows: TagGroupStory = {
   storyName: 'with action and maxRows'
 };
 
+export const WithLabelDescriptionContextualHelp: TagGroupStory = {
```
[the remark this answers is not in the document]
reidbarber: I can add the resizable container. Chromatic stories will be added closer to release since the UI may change.

packages/@react-spectrum/form/stories/Form.stories.tsx
reidbarber: > Should this Form prop also disable allowsRemoving if that prop is set?

I would say that seems a bit too opinionated at this point, since there isn't a disabled state for tags.

> Also disable the action buttons?

I would also say this seems a bit too opinionated.

___

I think maybe we should reconsider if TagGroup should even have Form support, since it isn't necessarily an input on its own.

snowystinger: Actually, this is already an issue, the buttons will pick up the isDisabled prop off the context, so they are already disabled, including the show more/show less. Here, I tried it out locally.
Screenshot 2023-01-24 at 3 09 40 PM

reidbarber: @snowystinger good catch, this has been fixed.

snowystinger (approved):
Looks good to me

LFDanLu (approved):
LGTM

PR #3942, opened 2023-01-20
Add docs for custom icons

reidbarber:
We've been linking to the FAQ lately, so this might be something we want in the docs instead.

## ✅ Pull Request Checklist:


snowystinger (dismissed):
looks good

packages/@react-spectrum/icon/docs/custom-icons.mdx
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## Coloring
+
+Icons support four Spectrum semantic colors: `negative`, `notice`, `positive`, and `informative`. Icons within React Spectrum components typically
+have their colors styled appropriately, but you can use the `color` prop to adjust the color of any standalone icons.
+
+```tsx example
+import Alert from '@spectrum-icons/workflow/Alert';
+
```
LFDanLu, line 89:
```suggestion
```

packages/@react-spectrum/icon/docs/custom-icons.mdx
```diff
@@ -0,0 +1,107 @@
...
+category: Icons
+---
+
+# Custom Icons
+
+Custom icons can be used by wrapping them in an Icon component.
+
+<HeaderInfo
+  packageData={packageData}
+  sourceData={[
```
LFDanLu, line 31:
```suggestion
  packageData={packageData}
  componentNames={['Icon']}
  sourceData={[
```

packages/@react-spectrum/icon/docs/custom-icons.mdx
```diff
@@ -0,0 +1,107 @@
...
+    {type: 'Spectrum', url: 'https://spectrum.adobe.com/page/iconography/'}
+  ]} />
+
+## Example
+
+```tsx example export=true
+import {Icon} from '@adobe/react-spectrum';
+
```
LFDanLu, line 39:
```suggestion
```

LFDanLu (approved):

packages/@react-spectrum/icon/docs/custom-icons.mdx
```diff
@@ -0,0 +1,105 @@
...
+  return(
+    <Icon {...props}>
+      <svg viewBox="0 0 36 36">
+        <path d="M18.477.593,22.8,12.029l12.212.578a.51.51,0,0,1,.3.908l-9.54,7.646,3.224,11.793a.51.51,0,0,1-.772.561L18,26.805,7.78,33.515a.51.51,0,0,1-.772-.561l3.224-11.793L.692,13.515a.51.51,0,0,1,.3-.908L13.2,12.029,17.523.593A.51.51,0,0,1,18.477.593Z" />
+      </svg>
+    </Icon>
+  );
+}
+
+<CustomIcon aria-label="Star" />
```
ktabors, line 50: Should the aria-label be part of the custom icon? `<Icon aria-label="Star" {...props}>` If we did that we'd have to specify an `aria-hidden="true"` in the button example.

packages/@react-spectrum/icon/docs/custom-icons.mdx
```diff
@@ -0,0 +1,105 @@
...
+import {Icon} from '@react-spectrum/icon';
+```
+
+---
+category: Icons
+---
+
+# Custom Icons
+
+Custom icons can be used by wrapping them in an Icon component.
```
ktabors, line 28: Maybe? `Custom icons are graphical metaphors or symbols that can be used to compliment existing experiences, which can be implemented by wrapping an SVG in an Icon component.`

packages/@react-spectrum/icon/docs/custom-icons.mdx
```diff
@@ -0,0 +1,105 @@
...
+</Flex>
+```
+
+## Labeling
+
+By default, icons are considered decorative, and are hidden from assistive technology. When used within a component like a button
+that has no label, an `aria-label` should be provided to the parent component. If used standalone, an `aria-label` can be provided
+to the icon itself.
+
+## Props
```
ktabors, line 103: I don't remember why we don't show props for Workflow Icons. It seems weird that we'd have them in this place and not both.

ktabors (commented):
Looks good!

packages/@react-spectrum/icon/docs/custom-icons.mdx
```diff
@@ -0,0 +1,105 @@
...
+  packageData={packageData}
+  componentNames={['Icon']}
+  sourceData={[
+    {type: 'Spectrum', url: 'https://spectrum.adobe.com/page/iconography/'}
+  ]} />
+
+## Example
+
+```tsx example export=true
+function CustomIcon(props) {
```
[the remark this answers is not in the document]
reidbarber: I went with `CustomIcon` to avoid confusion of importing an actual star icon from `@spectrum-icons/workflow`.

packages/@react-spectrum/icon/docs/custom-icons.mdx
reidbarber: I think the user could do that if they wanted to, but this seems more useful in case they were to re-use the icon multiple places with different labels. Also the labelling section mentions when you need to add a label. This also matches the `@spectrum-icons/workflow` icons.

packages/@react-spectrum/icon/docs/custom-icons.mdx
reidbarber: That description is specific to workflow icons, so I'm not sure it would apply to all custom icons.

packages/@react-spectrum/icon/docs/custom-icons.mdx
reidbarber: Not sure why either, I can add them in this PR.

snowystinger (approved):

PR #3964, opened 2023-01-26
TagGroup: add docs for help text, contextual help, and avatars

reidbarber:
## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


ktabors (dismissed):

LFDanLu (dismissed):
LGTM

devongovett:
Is it right that the label doesn't align with the first tag?

image

devongovett:
Also the label should not use a `<label>` element, because it isn't labeling an HTML form element. Should just be a `<span>`.

packages/@react-spectrum/tag/docs/TagGroup.mdx
```diff
@@ -154,3 +164,113 @@ function Example() {
   </Item>
 </TagGroup>
 ```
+
+### With avatars
+
+```tsx example
+<TagGroup aria-label="TagGroup with avatars example">
+  <Item>
+    <Avatar src="https://i.imgur.com/kJOwAdv.png" alt="Person 1 avatar" />
```
devongovett, line 173: I think these avatars should not have alt text (it defaults to "" in the Avatar component). Right now tags are read by screen readers as "Person 1 avatar Person 1", which is kinda unnecessarily verbose. The avatar is essentially decorative so I don't think it needs its own alt text.

reidbarber:
> Is it right that the label doesn't align with the first tag?

I agree, it does look weird. I can fix by changing the margin from 4px on each side, to 8px on end.

devongovett:
Or maybe the tags shouldn't have margins and we could use gap instead? Then the first/last tag wouldn't have a margin.

reidbarber:
@devongovett We're using `display: inline` on the tags so that the actions can be on the same line, while being outside of the tag group. Gap only applies to flex, grid, and columns. Looking into a way to do both.

devongovett:
Can you use `inline-flex` with `flex-wrap`?

reidbarber:
Looks like that might work if I also make the actions inline-flex. Double checking for any side effects.

reidbarber:
Pushed a fix in https://github.com/adobe/react-spectrum/pull/4102

devongovett:
FYI looks like the example in the "Content" section overflows horizontally on mobile causing page scrolling.

reidbarber:
@devongovett I can remove the Flex's used in these; don't think they're adding much.

Should I replace the Adobe products tags with News, Shopping, etc? May be closer to real life use case and have more realistic width for wrapping on mobile.

devongovett:
Sure 🤷

devongovett (approved):

packages/@react-spectrum/tag/docs/TagGroup.mdx
```diff
@@ -154,3 +147,104 @@ function Example() {
...
+    {id: 4, name: 'Shopping'}
+  ];
+
+  let [items, setItems] = React.useState(defaultItems);
+
+  let removeItem = (key) => {
+    setItems(prevItems => prevItems.filter((item) => key !== item.id));
+  };
+
+  let isValid = items.length <= 8;
```
LFDanLu, line 211:
```suggestion
  let isValid = items.length <= 3;
```

LFDanLu (changes requested):

LFDanLu (approved):

dannify (approved):

devongovett (approved):

PR #3966, opened 2023-01-26
Make LandmarkManager a global singleton

devongovett:
This ensures that there is only ever a single `LandmarkManager` instance on the page, even if there are multiple copies or different versions loaded. This is necessary so their global keyboard listeners do not conflict.

It is implemented by attaching the current instance to the `document` under a hidden symbol. If multiple different versions are loaded, the newest one replaces the older ones. We must ensure that the minimal API in the `LandmarkManagerApi` and `Landmark` interfaces never contain breaking changes so that registering landmarks from an older version of `useLandmark` against a newer version of `LandmarkManager` continues to work.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -123,7 +123,6 @@ module.exports = {
     'radix': [ERROR, 'always'],
     'react/jsx-uses-react': WARN,
     'eol-last': ERROR,
-    'arrow-body-style': [ERROR, 'as-needed'],
     'arrow-spacing': ERROR,
     'space-before-blocks': [ERROR, 'always'],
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -123,7 +123,6 @@ module.exports = {
     'radix': [ERROR, 'always'],
     'react/jsx-uses-react': WARN,
     'eol-last': ERROR,
-    'arrow-body-style': [ERROR, 'as-needed'],
```
devongovett, line 126: Discussed in slack.

snowystinger (approved):

reidbarber (approved):

PR #3982, opened 2023-01-31
Table column resizing aria docs

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Looks at the column resizing docs for useTableState and useTable. In addition, try the styled example out here: https://codesandbox.io/s/objective-cherry-g837on?file=/src/App.tsx

## 🧢 Your Project:

RSP


packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -242,7 +244,8 @@ function TableColumnHeader({column, state}) {
       style={{
         textAlign: column.colspan > 1 ? 'center' : 'left',
         padding: '5px 10px',
-        outline: isFocusVisible ? '2px solid orange' : 'none',
+        outline: 'none',
+        boxShadow: isFocusVisible ? 'inset 0 0 0 2px orange' : 'none',
```
LFDanLu, line 248: outline wasn't looking great in the resize example since it wasn't inset, replaced it with `boxShadow`. Did so for these pre-existing examples as well to reduce the diff between the this and the column resize examples

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,823 @@ function Checkbox(props) {
...
+the table cells in the body match the columns at all times.
+
+The various style changes below are to make the table itself scrollable and to support table body/column widths greater than the 400px applied to the table itself.
+
+<details>
+  <summary style={{fontWeight: 'bold'}}><ChevronRight size="S" /> Show code</summary>
+
+```tsx example export=true render=false
+/*- begin highlight -*/
+import {useTableColumnResizeState} from '@react-stately/table';
```
LFDanLu, line 526: Open to opinions. The changes made can be argued to be a bit extensive/excessive for an aria example but what I have here is what I felt to be a happy middle ground. Will be pointing out gotchas/places for alternative implementation here and there which I'd like opinions on

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,823 @@ function Checkbox(props) {
...
+  });
+
+  let ref = useRef();
+  let {collection} = state;
+  let {gridProps} = useTable(
+    {
+      ...props,
+      /*- begin highlight -*/
+      // The table itself is scrollable rather than just the body
+      scrollRef: ref
```
LFDanLu, line 551: Two possible strategies that I considered here. One is the current which is to make the table scrollable + sticky column headers (exactly how the aria example we have in the storybook is). Unfortunately this makes scrollIntoView iffy since it doesn't account for the sticky column headers taking room since it is floating on top of the scroll region.

Second strategy is to follow how we sync header scroll position in TableView. This felt a bit more complex for an aria example, but happy to implement if people think its ok

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,823 @@ function Checkbox(props) {
...
+    if (node.props.isSelectionCell) {
+      return 20;
+    }
+    return 75;
+  }, []);
+
+  let layoutState = useTableColumnResizeState({
+    getDefaultWidth,
+    getDefaultMinWidth,
+    tableWidth: 400
```
LFDanLu, line 578: this hard coded width  means the example is a bit limited when it comes to accepting table widths from the user such as `width: auto` or `maxWidth`/`minWidth`, etc. However the alternative is to use a resizeObserver which felt too much for the aria example (and had some finicky behavior will have to poke around)

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,823 @@ function Checkbox(props) {
...
+    </tr>
+  );
+}
+```
+
+</details>
+
+The `TableColumnHeader` is where we see the bulk of the changes required to support resizable columns. First of all, we need to accommodate a `Resizer` element in every resizable column that
+the user can drag or focus to perform a resize operation. To avoid interfering with the existing grid keyboard navigation, we only display this resizer when
+the column is hovered or if the column itself is being resized. This however introduces another issue: how will keyboard, screen reader, and touch users begin a resize operation if the resizer
```
LFDanLu, line 732: This is actually a bug (?) in our TableView, if you hover over the column headers and then use left/right to navigate through the columns, you'll actually land on the resizer and enter resizing mode. It also feels like a bit of a strange restriction to require the user to hide the resizer to avoid grid keyboard navigation issues

Did some thinking about this, maybe it would be nice to make an arbitrary element excluded from the FocusableTreewalker, open to ideas

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,823 @@ function Checkbox(props) {
...
+
+<details>
+  <summary style={{fontWeight: 'bold'}}><ChevronRight size="S" /> Show code</summary>
+
+```tsx example export=true render=false
+// Reuse the MenuButton from your component library. See below for details.
+import {MenuButton} from 'your-component-library';
+import {useHover} from '@react-aria/interactions';
+
+function ResizableTableColumnHeader({column, state, layoutState, onResizeStart, onResize, onResizeEnd}) {
```
LFDanLu, line 747: This is quite the change to the aria example's TableColumnHeader but IMO is good to have to keep interaction handling for all interaction types. Could be reduced if we wanna omit sorting handling but IMO its good to keep the previous functionality from the original aria example

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,823 @@ function Checkbox(props) {
...
+const MenuButton = React.forwardRef((props, ref) => {
+  // Create state based on the incoming props
+  let state = useMenuTriggerState(props);
+
+  // Get props for the button and menu elements
+  let {menuTriggerProps, menuProps} = useMenuTrigger({trigger: 'press'}, state, ref);
+  /*- begin highlight -*/
+  // Continue propagation of keydown event so that the left/right arrow key presses properly bubble to the table (useSelectableCollection)
+  let onKeyDown = (e) => e.continuePropagation();
+  /*- end highlight -*/
```
LFDanLu, line 948: another interesting bug (?) that I encountered. Our MenuTrigger avoids this problem since the `onKeyDown` from `useMenuTrigger` is sent via context and thus `useFocusable` won't automatically `stopPropagation` on it via `useKeyboard` https://github.com/adobe/react-spectrum/blob/8e57eaeda9112ea0783432409b8e6f9416d2fb28/packages/%40react-aria/focus/src/useFocusable.tsx#L74-L76

The aria implementation of a MenuButton runs into this problem if it doesn't do something similar. Maybe add `continuePropagation` to useMenuTrigger's onKeydown? Feels too specific though

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,823 @@ function Checkbox(props) {
...
+```
+
+</details>
+
+
+And with that, all necessary changes to the previous table implementation have been made and we now have a table that supports resizable columns!
+The example below supports resizing via mouse, keyboard, touch, and screen reader interactions. Previous behaviors such as selection and sorting are
+also still supported.
+
+```tsx example
```
LFDanLu, line 1280: Maybe I'll include more examples (for example with sort)? Don't think we wanna include a resizable columns table in every example in the Usage section though, open to opinions here

packages/@react-stately/table/docs/useTableColumnResizeState.mdx
```diff
@@ -0,0 +1,41 @@
...
+import docs from 'docs:@react-stately/table';
+import {ClassAPI, HeaderInfo, FunctionAPI, PageDescription} from '@react-spectrum/docs';
+import packageData from '@react-stately/table/package.json';
+
+---
+category: Collections
+keywords: [table, state, grid]
+---
+
+# useTableColumnResizeState
```
LFDanLu, line 22: Feels more discoverable to have this as a separate page. `useTableColumnResize` feels ok to have in `useTable` docs since it modifies the original implementation and thus needs context

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,826 @@ function Checkbox(props) {
...
+<ResizableColumnsTable aria-label="Table with selection and resizable columns" selectionMode="multiple">
```
LFDanLu, line 1284: Hrm, experience w/ the resizer is broken in Safari + iffy in VO and Talkback, digging

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,827 @@ function Checkbox(props) {
...
+    if (node.props.isSelectionCell) {
+      return 20;
+    }
+    return 75;
+  }, []);
+
+  let layoutState = useTableColumnResizeState({
+    getDefaultWidth,
+    getDefaultMinWidth,
+    tableWidth: 300
```
LFDanLu, line 578: this hard coded width means the example is a bit limited when it comes to accepting table widths from the user such as width: auto or maxWidth/minWidth, etc. However the alternative is to use a resizeObserver which felt too much for the aria example (and had some finicky behavior will have to poke around)

packages/@react-aria/table/docs/useTable.mdx
LFDanLu: IMO that would depend if we'd want to make that treewalker filtering/excluding public api or for internal use only. Maybe it could be a prop/callback like you said, maybe we provide it through a context since useGridCell is a bit deep sometimes, maybe we have a specific DOM attribute we look for and auto filter elements that have that attribute (heh)

packages/@react-aria/table/docs/useTable.mdx
LFDanLu: Sounds good, I'll go ahead and make the docs example simple and do a tailwind example with all the bells and whistles

packages/@react-aria/table/docs/useTable.mdx
LFDanLu: makes sense, that would conform with our philosophy of stopping propagation for events in general. Bit unfortunate that this is a "gotcha" for people implementing this in their tables. 

If we are amenable to easing up the propagation stuff, I could see an argument to continue propagation for keydowns we don't explicitly handle maybe?

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,827 @@ function Checkbox(props) {
...
+  return (
+    <div
+      role="presentation"
+      style={{
+        cursor: undefined,
+        width: '6px',
+        height: 'auto',
+        border: '2px',
+        borderStyle: 'none solid',
+        borderColor: layoutState.resizingColumn === column.key ? 'blue' : 'grey',
```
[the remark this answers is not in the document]
LFDanLu: orange sounds good, I'll go ahead with that

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,827 @@ function Checkbox(props) {
...
+  return (
+    <table
+      {...gridProps}
+      ref={ref}
+      style={{
+        borderCollapse: 'collapse',
+        /*- begin highlight -*/
+        width: '300px',
+        height: '200px',
+        display: 'block',
```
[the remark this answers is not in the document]
LFDanLu: I'll keep it as `display: block` and add a comment here so the changes between this and the previous implementation are as minimal as possible

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -488,12 +492,827 @@ function Checkbox(props) {
...
+    </Overlay>
+  );
+}
+```
+
+</details>
+
+### Button
+
+The `Button` component is used in the above example to toggle the menu. It is built using the [useButton](useButton.html) hook, and can be shared with many other components.
```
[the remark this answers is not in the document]
LFDanLu: Maybe I'll rename MenuButton to MenuTrigger, MenuButton was the naming we have in the useMenu docs heh

packages/@react-aria/table/docs/useTable.mdx
snowystinger: i think that would be fine, no reason to block keys we don't handle

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -800,6 +804,645 @@ let rows = [
...
+      }}
+      ref={ref}>
+      {cell.rendered}
+    </td>
+  );
+}
+```
+
+And with that, all necessary changes to the previous table implementation have been made and we now have a table that supports resizable columns!
+The example below supports resizing via mouse, keyboard, touch, and screen reader interactions. To see an example with sorting and selection, see the styled example!
```
LFDanLu, line 1412: Styled example in progress here: https://codesandbox.io/s/agitated-darwin-uki38i?file=/src/Table.tsx. I think the nightly is causing issues due to dupe packages, still need to investigate fully

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -800,6 +804,645 @@ let rows = [
...
+
+```tsx example
+<ResizableColumnsTable aria-label="Table with resizable columns">
```
reidbarber, line 1415: After pressing the 'Resize column' option in the menu, focus briefly goes back to the column header. Not a huge deal.

reidbarber: Also, when dragging via mouse, the dragged column header will have a focus ring for a bit.

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -800,6 +804,645 @@ let rows = [
...
+  );
+});
+```
+
+</details>
+
+### Menu
+
+The `Menu` is used to display the various actions available to the column, rendered with a `Popover` when the user presses the column header.
+This is built using the [useMenu](useMenu.html) and [useTreeState](/react-stately/useTreeState.html) hooks.
```
reidbarber, line 1190: Stately links don't work in the build. I think they will still work in prod, but maybe safer to use relative path like `[useTreeState](../react-stately/useTreeState.html)`?

packages/@react-aria/table/docs/useTable.mdx
LFDanLu: I believe this is because it is using the button's native focus ring. I can override it, but was a bit wary of going too far with the custom styling

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -800,6 +806,655 @@ let rows = [
...
+      {children}
+    </tr>
+  );
+}
+```
+
+The `TableColumnHeader` is where we see the bulk of the changes required to support resizable columns. First of all, we need to accommodate a `Resizer` element in every resizable column that
+the user can drag or focus to perform a resize operation. To avoid interfering with the existing grid keyboard navigation, we only display this resizer when
+the column is hovered or if the column itself is being resized. This however introduces another issue: how will keyboard, screen reader, and touch users begin a resize operation if the resizer
+is only visible on hover? We can't just have the user press on the column header since that action is reserved for sorting.
```
snowystinger, line 982: qq, we no longer support sorting here, can we actually do away with the menu entirely and just have clicking on the column header go into resize mode for this example?

the tailwind one supports sorting and shows how they can work together.

LFDanLu: I considered doing that but it felt rather non-standard and a bit divergent from the tailwind example. I'd have to add aria-labeling/message to let screenreader users know that pressing on the header would trigger resizing and demonstrate overriding handlers from `useTableColumnHeader` in the off chance sorting was enabled as well. 

Personally I like having the menu here since it serves as a nice step 1.5 for people to consider what to do if they wanted additional actions on the TableColumnHeader. Happy to discuss if we still feel that the menu is over complicating things here

snowystinger: I'm fine with it as is.
We should update the sentence this is commenting on since the column header in this example is not reserved for sorting though.

snowystinger (approved):

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -800,6 +806,696 @@ let rows = [
...
+### Table
+
+As mentioned previously, we first need to call <TypeLink links={statelyDocs.links} type={statelyDocs.exports.useTableColumnResizeState} /> to initialize the widths for our table's columns.
+We'll pass the state returned by <TypeLink links={statelyDocs.links} type={statelyDocs.exports.useTableColumnResizeState} /> along with any user defined `onResize` handlers
+to our `ResizableTableColumnHeaders` so it can be used by <TypeLink links={docs.links} type={docs.exports.useTableColumnResize} />. The `widths` from this state are provided to each table element so
+the table cells in the body match their parent column's widths at all times.
+
+The various style changes below are to make the table itself scrollable when the row content overflows and to support table body/column widths greater than the 300px applied to the table itself.
+
+<details>
```
LFDanLu, line 830: Flipped back to collapsing the whole code sections blocks, better experience on mobile IMO

reidbarber (approved):
LGTM

packages/@react-aria/table/docs/useTable.mdx
devongovett: I would expand them. The code is sorta the point of the docs I think. Most people will be looking at this on desktop as well.

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -800,6 +806,696 @@ let rows = [
...
+            : item.index % 2
+              ? 'var(--spectrum-alias-highlight-hover)'
+              : 'none',
+        color: isSelected ? 'white' : null,
+        outline: 'none',
+        boxShadow: isFocusVisible ? 'inset 0 0 0 2px orange' : 'none',
+        /*- begin highlight -*/
+        display: 'flex',
+        width: 'fit-content'
+        /*- end highlight -*/
```
devongovett, line 1400: hmm a lot of the changes we are highlighting are purely style changes. Wonder if there is a way of simplifying this. Also, why does the table have a flex layout rather than table?

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -800,6 +806,696 @@ let rows = [
...
+
+The `TableColumnHeader` is where we see the bulk of the changes required to support resizable columns. First of all, we need to accommodate a `Resizer` element in every resizable column that
+the user can drag or focus to perform a resize operation. To avoid interfering with the existing grid keyboard navigation, we only display this resizer when
+the column is hovered or if the column itself is being resized. This however introduces another issue: how will keyboard, screen reader, and touch users begin a resize operation if the resizer
+is only visible on hover? We could have the user press on the column header to trigger a resize operation, but that action is often reserved for sorting and may feel unexpected to a user.
+
+To resolve this, we need to render a menu button in the column header if the column is resizable, allowing non-mouse users to select from a list of available actions on the column. For resizing, selecting the corresponding option
+should cause focus to shift to the resizer itself, displaying the resizer for touch screen users to drag and letting keyboard and screen readers resize the column via keyboard arrows and screen reader operations
+accordingly.
+
```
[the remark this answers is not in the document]
snowystinger: The issue with this was grid navigation. When arrow key navigating, you couldn't move between column headers. Instead, it'd start resizing.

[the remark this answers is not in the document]
LFDanLu: https://github.com/adobe/react-spectrum/pull/3982#discussion_r1091336950 alludes to this issue. When using the arrow keys to navigate between columns, you'll actually move focus to the resizer instead which means you are now in resize mode.

packages/@react-aria/table/docs/useTable.mdx
snowystinger: display table has a lot of issues with respecting designated column sizes and cannot be horizontally scrolled.
For example, if you specify all columns with a min of 50px and then resize one column so it's huge and would cause horizontal scrolling, all the other columns will shrink to be smaller than the min specified.

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -800,6 +806,696 @@ let rows = [
...
+        flex: '0 0 auto',
+        boxSizing: 'border-box',
+        visibility: showResizer ? 'visible' : 'hidden',
+        marginLeft: '4px'
+      }}
+      {...resizerProps}>
+      <VisuallyHidden>
+        <input
+          ref={ref}
+          type="range"
```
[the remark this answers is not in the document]
LFDanLu: makes sense, I'll open a separate PR

packages/@react-aria/table/docs/useTable.mdx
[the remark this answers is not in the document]
devongovett: hmm maybe we should have a mode where only the up/down arrows change the size rather than left/right? the menu solution is cool, but kinda spectrum design specific. I could see this being a frequent request.

packages/@react-aria/table/docs/useTable.mdx
snowystinger: probably, but then you don't get the sticky headers if I recall correctly

packages/@react-aria/table/docs/useTable.mdx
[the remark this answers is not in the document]
snowystinger: that sounds reasonable, hadn't thought to divide up the actions like that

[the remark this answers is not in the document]
LFDanLu: I'll see if I can put a PR up for this as

snowystinger (approved):

packages/@react-aria/table/docs/useTable.mdx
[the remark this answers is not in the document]
LFDanLu: Ran into a snag trying to implement this. Since we `setKeyboardNavigationDisabled(true)` when the resizer is focused, I need to re-enable it when the user presses the left/right arrow keys. However, `useGridCell` `onKeyDownCapture` triggers before the `onKeyDownCapture` I setup in `useTableColumnResizing`, meaning we don't move focus back to the column header when on `ArrowLeft`.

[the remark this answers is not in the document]
LFDanLu: Similar issue happens if I get rid of `setKeyboardNavigationDisabled` entirely instead. Left/right arrow keys can be made to move between the resizer and the column but up/down arrows keys are then handled by `useGridCell` and results in focus moving from the resizer to the cell below.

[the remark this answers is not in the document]
snowystinger: ugh, yeah, that wouldn't be separable. keyboard grid navigation is either all or nothing... there's no way to distinguish between the two. we'd need to introduce a new prop or something and even then I'm not sure we could get ahead of it, as you mentioned you ran into a problem getting ahead of the capture events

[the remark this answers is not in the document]
snowystinger: could we change it from onFocus to on 'Return/Enter' pressed? maybe that could be the new prop?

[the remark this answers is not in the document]
LFDanLu: @devongovett any opinions on the above? Right now these are the options that I can see:
- make it so that "editMode" is enabled when the user presses Enter when focused on the resizer
  - perhaps non-intuitive?
- change the aria example so pressing enter on the column header enters resize mode
  - this doesn't solve grid navigation bit but it gets rid of the menu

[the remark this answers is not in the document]
devongovett: Either would be ok I think. Just need to make sure it's obvious somehow. Definitely would need an aria-label on the column header if that's how we go, but even for non-screen reader users it might not be obvious. Enter on the resizer itself is a little more I guess.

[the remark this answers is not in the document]
LFDanLu: pushed code changes for pressing enter on the column header to enter resize mode (note that I haven't updated the docs copy to reflect the new example code). I'll open a separate branch for the "editMode" updates since that will require changes to the hooks themselves

packages/@react-aria/table/stories/example-docs.tsx
```diff
@@ -0,0 +1,224 @@
...
+      className={classNames(ariaStyles, 'aria-table-row')}>
+      {children}
+    </tr>
+  );
+}
+
+function ResizableTableColumnHeader({column, state, layoutState, onResizeStart, onResize, onResizeEnd}) {
+  let {widths} = layoutState;
+  let ref = useRef(null);
+  let resizerRef = useRef(null);
```
LFDanLu, line 120: todo: don't forget to remove this

LFDanLu: also make resizer no longer a forward ref

LFDanLu:
TODO: wait till https://github.com/adobe/react-spectrum/pull/4077 is released before merging this. Code changes that exist in this pull are the same as in that pull

reidbarber (approved):
LGTM

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -817,6 +824,508 @@ let rows = [
...
+  let {widths} = layoutState;
+  let allowsResizing = column.props.allowsResizing;
+  let ref = useRef(null);
+  let {columnHeaderProps} = useTableColumnHeader({node: column}, state, ref);
+
+  return (
+    <th
+      {...columnHeaderProps}
+      className="aria-table-headerCell"
+      style={{width: widths.get(column.key)}}
```
devongovett, line 1050: wondering if we should even expose the `widths` directly from the hook. There is already a `getColumnWidth` method as well, which we could use here.

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -817,6 +824,508 @@ let rows = [
...
+
+```tsx example export=true render=false
+import {useTableColumnResize} from '@react-aria/table';
+
+function Resizer(props) {
+  let {column, layoutState, onResizeStart, onResize, onResizeEnd} = props;
+  let ref = useRef(null);
+  let {resizerProps, inputProps} = useTableColumnResize({
+    column,
+    label: 'Resizer',
```
devongovett, line 1135: hmm maybe this prop should be called 'aria-label' so it's clear what it's for?

devongovett (commented):
A few small suggestions

packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -817,6 +824,508 @@ let rows = [
...
+  let {columnHeaderProps} = useTableColumnHeader({node: column}, state, ref);
+
+  return (
+    <th
+      {...columnHeaderProps}
+      className="aria-table-headerCell"
+      style={{width: widths.get(column.key)}}
+      ref={ref}>
+      <div style={{display: 'flex', position: 'relative'}}>
+        <Button className="aria-table-headerTitle">
```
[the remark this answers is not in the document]
LFDanLu: It needs to be some kind of focusable element since focus will move to the [first focusable child](https://github.com/adobe/react-spectrum/blob/a16e8474ed620c8898b1c8133ad5ec796b25a773/packages/%40react-aria/table/src/useTableColumnHeader.ts#L48) due to `focusMode: child`. If I don't make the header content focusable, then focus will jump to the resizer. 

A button does feel odd since there isn't a press action but the other focusable elements don't feel appropriate either

packages/@react-aria/table/docs/useTable.mdx
LFDanLu: makes sense, I can give that a try

devongovett:
I also tried out an implementation that doesn't require overriding the table layout with flex and it removed a lot of the basic style overrides allowing us to reuse the components defined earlier. What do you think? https://github.com/adobe/react-spectrum/pull/4253

LFDanLu:
@devongovett urgh, I wanted to preserve the sticky column headers but I suppose most people should opt to follow the styled example's approach anyways since having sorting + resizing is probably the most common use case.

LFDanLu:
TODO: I'm updating the example to fit the changes we made to the resize hooks. Will update the codesandbox as well when the nightly gets bumped/we release the codechanges

LFDanLu:
TODO: update sandbox example so it uses input styles from useTableColumnResize hook instead of VisuallyHidden, need to wait for nightly/release for that

snowystinger (approved):

reidbarber (approved):

snowystinger (approved):

PR #4000, opened 2023-02-02
Fix SearchAutocomplete to call onClear

reidbarber:
Closes 

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-aria/autocomplete/src/useSearchAutocomplete.ts
+++ b/packages/@react-aria/autocomplete/src/useSearchAutocomplete.ts
@@ -55,7 +55,8 @@ export function useSearchAutocomplete<T>(props: AriaSearchAutocompleteProps<T>, 
     inputRef,
     listBoxRef,
     keyboardDelegate,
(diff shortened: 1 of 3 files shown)


snowystinger (approved):

LFDanLu (approved):
LGTM. As an aside I noticed that the field doesn't clear when hitting `Esc` but that seems to be pre-existing so can be handled later

reidbarber:
> LGTM. As an aside I noticed that the field doesn't clear when hitting `Esc` but that seems to be pre-existing so can be handled later

Weird, it does for me. Any other steps to reproduce?

LFDanLu:
Ah, I should've clarified: specifically after you've selected an option. I was expecting `Esc` to clear the selected key and empty the input but it seems to reset it back to the selected key's text. A bit odd since clicking the `'x'` clears the key so probably something isn't quite hooked up the same

reidbarber:
Ahh good catch. Yeah, we can handle in a follow-up.

Issue #2286, opened 2021-09-03
SearchAutocomplete UX issues


reidbarber:
# 🐛 Bug Report 

List of issues:

- On mobile with isQuiet, the search icon placement within the input is wrong.
- On mobile with isDisabled, the search icon is too dark (doesn't look disabled). Compare to SearchField disabled.
- The isReadOnly story needs a defaultValue or value to check that we can select the value or not.
- On mobile, the validation icon and clear icon are in different orders on the input and the tray.
- Add a long value story to see how it overlaps with the validation and clear icons.
- Pressing 'Esc' key while typing on mobile closes and re-opens tray. Should behave like ComboBox.
- Inputs on stories for size-3000 and size-6000 appear to be the same width.
- On the size-500 story, selecting a value like "Mechanical" shows the value overflowing outside of the input
- On mobile, the clear button is visible when the field is empty and validationState is set.

## 🧢 Your Company/Team

RSP

reidbarber:
Closing, as we shift focus to S2 and now have an Autocomplete component

PR #4001, opened 2023-02-02
SearchAutocomplete: fix to clear selected item if searchfield is cleared

reidbarber:
## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Type and select an item from the dropdown
2. Press Escape to clear the field

The searchfield should now clear.

### DIFF
--- a/packages/@react-aria/autocomplete/src/useSearchAutocomplete.ts
+++ b/packages/@react-aria/autocomplete/src/useSearchAutocomplete.ts
@@ -66,6 +66,7 @@ export function useSearchAutocomplete<T>(props: AriaSearchAutocompleteProps<T>, 
     autoComplete: 'off',
     onClear: () => {
       state.setInputValue('');
+      state.setSelectedKey(null);
       if (onClear) {
         onClear();
       }
--- a/packages/@react-aria/autocomplete/src/useSearchAutocomplete.ts
+++ b/packages/@react-aria/autocomplete/src/useSearchAutocomplete.ts
@@ -66,7 +66,7 @@ export function useSearchAutocomplete<T>(props: AriaSearchAutocompleteProps<T>, 
     autoComplete: 'off',
     onClear: () => {
       state.setInputValue('');
-      state.setSelectedKey(null);
+      state.selectionManager.setSelectedKeys(new Set());
       if (onClear) {
         onClear();
       }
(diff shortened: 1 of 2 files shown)


packages/@react-spectrum/autocomplete/test/SearchAutocomplete.test.js
```diff
@@ -816,6 +816,43 @@ describe('SearchAutocomplete', function () {
...
+      let items = within(listbox).getAllByRole('option');
+      expect(items).toHaveLength(1);
+      expect(searchAutocomplete).not.toHaveAttribute('aria-activedescendant');
+
+      act(() => {
+        fireEvent.keyDown(searchAutocomplete, {key: 'ArrowDown'});
+        fireEvent.keyUp(searchAutocomplete, {key: 'ArrowDown'});
+        fireEvent.keyDown(searchAutocomplete, {key: 'Enter'});
+        fireEvent.keyUp(searchAutocomplete, {key: 'Enter'});
+      });
```
snowystinger, line 840: do we need this extra act around the fireEvents?
same question for below with esc

reidbarber: I suppose not, just copied from a test above. Fixed.

snowystinger (approved):

packages/@react-spectrum/autocomplete/test/SearchAutocomplete.test.js
snowystinger: ah, yeah, we still have some of those floating around

packages/@react-aria/autocomplete/src/useSearchAutocomplete.ts
```diff
@@ -66,6 +66,7 @@ export function useSearchAutocomplete<T>(props: AriaSearchAutocompleteProps<T>,
     autoComplete: 'off',
     onClear: () => {
       state.setInputValue('');
+      state.selectionManager.setSelectedKeys(new Set());
```
LFDanLu, line 69: Noticed that `onInputChange` is being called multiple times with different values when using the Esc key:
image

This seems to be because of `useComboBox`'s Esc key handling which calls `state.revert`. Even if we clear the selected key here, that state update doesn't happen yet for `state.revert` which will cause the second onInputChange call. Ideally we'd drop the `Esc` key handling from useComboBox but keep the rest of it.

snowystinger: good catch, we should get a test for this as well

reidbarber:
Closing for now, added to https://github.com/adobe/react-spectrum/issues/2286

PR #4008, opened 2023-02-03
Tableview: Drag and drop support

reidbarber:
**From 02/27/23 testing session:**

Now

- [x] Would be nice to have horizontal scrolling stories with DnD (DL) ([3971c1c](https://github.com/adobe/react-spectrum/pull/4008/commits/3971c1c189c5429e2dd2f4175cce4c90def0f137))
- [x] The drag between tables example is missing textValue and reads "Insert between undefined and undefined" - "Drag between tables" story: Insertion indicators between rows are announced as "drop between undefined and undefined" (DL) (1f2aa79)
- [x] would be nice to have descriptions for some of the util handler stories like for "using getDropOperatations to determin default drop operation" (DL) ([b6693db](https://github.com/adobe/react-spectrum/pull/4008/commits/b6693dbc327a64cdb4268bdcb94a560c526f2efc))
- [x] There are two "Drag onto row"/"Drag between tables"/ and root drop stories stories (DL) ([c6b1110](https://github.com/adobe/react-spectrum/pull/4008/commits/c6b111006cdd5702b8027b4578cc96e270deb3ba))
- [x] Maybe the drop indicator element with role=gridcell shouldn't have aria-selected=false on it so it doesn't announce as "Not selected"? (DG) ([bc8bcfe](https://github.com/adobe/react-spectrum/pull/4008/commits/bc8bcfe5c308b6e211a9832fc3795e7dd79980f5))
- [x] Should the root drop indicator be inside the "rowgroup" for table body? (DG) ([50ab192](https://github.com/adobe/react-spectrum/pull/4008/commits/50ab192caa7747e6eb80e13348db27b879444ad2))
- [x] I think the focus ring around the drag handle is the default browser one? (DG) ([454e351](https://github.com/adobe/react-spectrum/pull/4008/commits/454e351454d33952f9d5a65492343216ea398d30))
- [x] After dropping an item, the cell that was dragged gets focus. Should probably be the row itself (RB) ([609c2ef](https://github.com/adobe/react-spectrum/pull/4008/commits/609c2efda29c511b58e3506a8f2d58b0fd6ac8a0))
- [x] The table row, table body, and table itself are included when swiping through valid drop targets, (DL) Handled for ListView row, look for PR. - "Drag between tables (root only)" There are two drop targets for each table that announce "Drop on Table x", probably the same issue as row 4 (DL) ([6609e1d](https://github.com/adobe/react-spectrum/pull/4008/commits/6609e1d5b3d40707ddd06f8fbcbb405d15c424af)) ([7770181](https://github.com/adobe/react-spectrum/pull/4008/commits/7770181fcc1fdd88730047adcd16efcc9d11c84a))
- [x] Dropping in second table in "complex drag between tables" example throws an error (DG) ([d51343e](https://github.com/adobe/react-spectrum/pull/4008/commits/d51343e41a28991d15e5e029825f4fb0aa14750e))
- [x] Can't drop on the folder 7 row in "Drag onto Row" story (DL) ([efaa9c2](https://github.com/adobe/react-spectrum/pull/4008/commits/efaa9c23485ef282718beb89dc5855483dc46370))
- [x] Don't show drag button if row is disabled (RB) ([ac3ca40](https://github.com/adobe/react-spectrum/pull/4008/commits/ac3ca402224769720db6018d41b4b074a03bef33))
- [x] "complex drag between tables": drop indicator above Pictures in the second table announces "drop between null and Pictures" (DL) [ca96e5a](https://github.com/adobe/react-spectrum/pull/4008/commits/ca96e5aa1c6942c58382dcb11e3a41f22ceea1b7#diff-aa7881d95c52e5c048c75e5167c2bc1e5d8e7a8d8baf1490066305c4cb3f5413)
- [x] "drop onto row/folder" Unable to navigate to any drop targets using VoiceOver (DL)

Later
- [ ] Drag within (Reorder), keyboard drag a row and arrow down quickly. Insertion indicator disappears if you go too fast (RB) - When dragging while scrolling, insertion indicators stop appearing. - "drag within table (reorder)" util handlers story: Unable to drop multiple items to a drop position that is out of view. When the table scrolls, the drop indicators don't appear anymore (DL) (RB)
- [ ] The drag within table story focuses the dropped row's drag handle, but the util handlers version does not - focus seems lost. (DG)
- [ ] Maybe we should construct the default textValue based on row headers? Also maybe warn if the row headers have non-string values and don't have textValue themselves. (DG) 

**From 03/17 testing session**:

- [x] In 'Drag between tables' story, random cells will get focused when dropping between tables (RB)
- [x] When 'Select all' checkbox is selected, and you've scrolled horizontally, you can see some of the column headers to the left of the checkbox. See screenshot 2 (RB)
- [x] Drop onto root story: Talkback still has two valid drop targets, the Table itself and the root drop indicator. Not sure if it is possible to get rid of the Table drop target (DL) 
- [ ] After dropping, the scrollbars flicker in and out a bunch on Windows (DG)

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test stories at TableView/Drag and Drop

## 🧢 Your Project:

RSP


packages/@react-spectrum/table/src/RootDropIndicator.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  if (!isDropTarget && dropIndicatorProps['aria-hidden']) {
+    return null;
+  }
+
+  return (
+    <div role="row" aria-hidden={dropIndicatorProps['aria-hidden']}>
+      <div
+        role="gridcell"
+        aria-selected="false">
```
majornista, line 22: I think this may need to include an `aria-colspan` equal to the number of columns in the table.

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -1048,7 +1212,9 @@ function TableCheckboxCell({cell}) {
               stylesOverrides,
               'react-spectrum-Table-cell'
             )
-          )}>
+          )
+        }>
+        {isTableDraggable && <DragButton />}
```
majornista, line 1217: Should we have a dedicated cell/column for the drag handle? It feels odd to navigate up or down from a row selection checkbox and to have focus go to the drag handle, rather than the next adjacent checkbox.

reidbarber: I think this is a good point. On the other hand, it may be strange to have a cell/column that is entirely empty unless the row is keyboard focused.

packages/@react-spectrum/table/src/DragPreview.tsx
```diff
@@ -0,0 +1,132 @@
...
+    density,
+    layout,
+    direction
+  } = props;
+  let {height, width} = layout.getLayoutInfo(item.key).rect;
+
+  let isDraggingMultiple = itemCount > 1;
+
+  let cells = [...item.childNodes];
+  let dragButtonCellWidth = cells[0].props.isDragButtonCell ? layout.getLayoutInfo(cells[0].key).rect.width : 0;
```
snowystinger, line 95: when is this one not the drag button cell?

packages/@react-spectrum/table/src/DragPreview.tsx
```diff
@@ -0,0 +1,132 @@
...
+  let selectionCellWidth = cells[1].props.isSelectionCell ? layout.getLayoutInfo(cells[1].key).rect.width : 0;
+
+  width = width - selectionCellWidth - dragButtonCellWidth;
```
snowystinger, line 98: Something is off about the Drag Preview position, if I drag from where the drag button would be, the preview is rendered floating off to the side

packages/@react-spectrum/table/src/InsertionIndicator.tsx
```diff
@@ -0,0 +1,50 @@
...
+  let {dropIndicatorProps} = dragAndDropHooks.useDropIndicator(props, dropState, ref);
+  let {visuallyHiddenProps} = useVisuallyHidden();
+
+  let isDropTarget = dropState.isDropTarget(target);
+
+  if (!isDropTarget && dropIndicatorProps['aria-hidden']) {
+    return null;
+  }
+
+  let style = {
```
snowystinger, line 27: I don't see this in react-spectrum/list/InsertionIndicator. Why is it needed here? Not that it's wrong, just surprised.

packages/@react-spectrum/table/src/DragPreview.tsx
```diff
@@ -0,0 +1,132 @@
...
+  layout: TableLayout<any>,
+  selectionCellWidth: number,
+  dragButtonCellWidth: number,
+  direction: Direction
+}
+
+export function TableCellPreview(props: TableCellPreviewProps) {
+  let {cell, layout, selectionCellWidth, dragButtonCellWidth, direction} = props;
+  let {align} = cell.column.props;
+  let {x: left, height, width} = layout.getLayoutInfo(cell.key).rect;
```
snowystinger, line 40: if some columns are yet to be rendered and you're having trouble getting the width, you should have all column widths available through the layout state
`state.getColumnWidth(cell.column.key)`

height you may have to find a cell that has been rendered and grab that height and apply it to all of them

you could probably ignore left if you laid out the cells in a Flex where every item has a width and the flex prop is set to `flex: 0 0 auto`. this should address your RTL support as well, remove the need to determine badge width, and make it easier to vertically center the badge.

packages/@react-spectrum/table/src/InsertionIndicator.tsx
```diff
@@ -0,0 +1,50 @@
...
+        role="gridcell"
+        aria-selected="false"
+        className={
+          classNames(
+            styles,
+            'react-spectrum-Table-InsertionIndicator',
+            {
+              'react-spectrum-Table-InsertionIndicator--dropTarget': isDropTarget
+            }
+        )}>
```
snowystinger, line 45: List renders the visually hidden div conditionally, why is that missing here? again, not wrong, just curious

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -113,28 +131,48 @@ export function useVirtualizerContext() {
...
+  let isTableDraggable = !!dragAndDropHooks?.useDraggableCollectionState;
+  let isTableDroppable = !!dragAndDropHooks?.useDroppableCollectionState;
+  let dragHooksProvided = useRef(isTableDraggable);
+  let dropHooksProvided = useRef(isTableDroppable);
+  if (dragHooksProvided.current !== isTableDraggable) {
+    console.warn('Drag hooks were provided during one render, but not another. This should be avoided as it may produce unexpected behavior.');
+  }
+  if (dropHooksProvided.current !== isTableDroppable) {
+    console.warn('Drop hooks were provided during one render, but not another. This should be avoided as it may produce unexpected behavior.');
+  }
```
snowystinger, line 150:
```suggestion
  useEffect(() => {
    if (dragHooksProvided.current !== isTableDraggable) {
      console.warn('Drag hooks were provided during one render, but not another. This should be avoided as it may produce unexpected behavior.');
    }
    if (dropHooksProvided.current !== isTableDroppable) {
      console.warn('Drop hooks were provided during one render, but not another. This should be avoided as it may produce unexpected behavior.');
    }
  }, [isTableDraggable, isTableDroppable]);
```

I know listview does this, it'll need to be corrected for React StrictMode

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -149,6 +187,7 @@ function TableView<T extends object>(props: SpectrumTableProps<T>, ref: DOMRef<H
   let state = useTableState({
     ...props,
     showSelectionCheckboxes,
+    showDragButtons: !!isTableDraggable,
```
snowystinger, line 190: this is already a boolean
```suggestion
    showDragButtons: isTableDraggable,
```

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -929,6 +1041,33 @@ function TableSelectAllCell({column}) {
...
 }
 
+function TableDragHeaderCell({column}) {
+  let ref = useRef();
+  let {state} = useTableContext();
+  let {columnHeaderProps} = useTableColumnHeader({
+    node: column,
+    isVirtualized: true
+  }, state, ref);
+  // TODO: How do we skip navigation to this cell?
```
snowystinger, line 1051: i think it's ok to have a blank table drag header cell with an aria-label that is navigable

packages/@react-spectrum/table/test/TableSizing.test.tsx
```diff
@@ -347,7 +347,8 @@ describe('TableViewSizing', function () {
         for (let row of bodyRows) {
           for (let [index, cell] of row.childNodes.entries()) {
             if (index === 0) {
-              expect((cell as HTMLElement).style.zIndex).toBe('2');
+              // TODO: This should be 2. Why did this change?
```
snowystinger, line 350: zindex across columns is decrementing so that earlier columns render *over* later columns. otherwise the resizer would be partially covered by columns that follow

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -124,7 +124,8 @@ export class TableLayout<T> extends ListLayout<T> {
     for (let column of this.collection.columns) {
       // The selection cell and any other sticky columns always need to be visible.
       // In addition, row headers need to be in the DOM for accessibility labeling.
-      if (column.props.isSelectionCell || this.collection.rowHeaderColumnKeys.has(column.key)) {
+      // TODO: how to support multiple sticky columns?
+      if (column.props.isDragButtonCell || this.collection.rowHeaderColumnKeys.has(column.key)) {
```
snowystinger, line 128: can't do this?
```
if (column.props.isSelectionCell || column.props.isDragButtonCell || this.collection.rowHeaderColumnKeys.has(column.key)) {
```

packages/@react-stately/table/src/TableCollection.ts
```diff
@@ -187,7 +209,7 @@ export class TableCollection<T> extends GridCollection<T> {
         }
       };
 
-      columns.unshift(rowHeaderColumn);
+      columns = [...columns, rowHeaderColumn];
```
snowystinger, line 212: unshift adds in place, you should be able to unshift twice to avoid creating a new array

snowystinger (commented):
What happens in tables with many columns? both if you have and have not horizontally scrolled (due to the lazy cell rendering we do now)
if it's hugely long, maybe we just add the key/id cell plus a couple next to it?

packages/@react-spectrum/table/src/InsertionIndicator.tsx
reidbarber: Because of how ListView items were laid out, they were stacked on top of each other in terms of z-index. So there wasn't a way to show the full insertion indicator below an item without it being cut off by the row below. So there is a fake (isPresentationOnly) indicator that renders above the item below, and we don't want to actually double up on drop targets.

In tableview, these positioned absolutely, they don't have the same issue, and don't need to double up with a fake one.

This answer also applies to your question in https://github.com/adobe/react-spectrum/pull/4008#discussion_r1102086646. I think in listview if we used `position: absolute` for positioning the insertion indicators like here, we wouldn't need the fake ones.

packages/@react-spectrum/table/src/InsertionIndicator.tsx
reidbarber: They were all rendering at the top of the container, so I positioned them absolute. I think becase the tableview rows themselves are positioned differently than listview items. See https://github.com/adobe/react-spectrum/pull/4008#discussion_r1109086970 for a related answer.

packages/@react-spectrum/table/src/DragPreview.tsx
reidbarber: should be fixed now

packages/@react-spectrum/table/src/DragPreview.tsx
reidbarber: Updated to use flex

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -983,31 +1162,94 @@ function TableRow({item, children, hasActions, ...otherProps}) {
...
-    </div>
+    <TableRowContext.Provider value={{dragButtonProps, dragButtonRef, isFocusVisibleWithin}}>
+      {isTableDroppable && isFirstRow &&
+        <InsertionIndicator
+          rowProps={props}
+          key={`${item.key}-before`}
+          target={{key: item.key, type: 'item', dropPosition: 'before'}} />
+      }
+      {isTableDroppable && !dropIndicator?.isHidden &&
+        <div role="button" {...visuallyHiddenProps} {...dropIndicator?.dropIndicatorProps} ref={dropIndicatorRef} />
```
devongovett, line 1215: FYI I think this can't be directly inside a grid/rowgroup, it'll need to be inside a gridcell within a row.

packages/@react-aria/dnd/src/useDropIndicator.ts
```diff
@@ -61,9 +61,16 @@ export function useDropIndicator(props: DropIndicatorProps, state: DroppableColl
...
+    let before = null;
+    if (target.dropPosition === 'before') {
+      let keyBefore = collection.getKeyBefore(target.key);
+      // TODO: move logic to TableView
+      if (keyBefore && !keyBefore.toString().startsWith('headerrow-')) {
+        before = keyBefore;
+      }
+    } else {
+      before = target.key;
+    }
```
reidbarber, line 73: This handles the case where it was announcing `"Insert between null and Item One"` instead of `"Insert before Item One"`. It was detecting the header row as the key before. Should probably handle this somewhere else if possible though.

LFDanLu: Thinking out loud, but perhaps we could add and additional check to the `before` and `after` logic here where we check the collection's first and last key via `getFirstKey` and `getLastKey`? If the first key matches the `target.key` then we can set `before` to null and do something similar for last key to guard against the case where we have nodes after what we consider is the "last" row.

The problem is that this relies on the collection's `getFirstKey` and `getLastKey` returning the row keys, which breaks in the case where the Table collection's bodyNodes are Sections instead... Perhaps `useDropIndicator` should accept some user defined options for determining `keyBefore` and `keyAfter` so we can customize this logic on a case by case basis? That would allow us to omit certain node types that we know don't make sense in a DnD context like the header row

packages/@react-aria/virtualizer/src/VirtualizerItem.tsx
```diff
@@ -65,7 +67,7 @@ export function layoutInfoToStyle(layoutInfo: LayoutInfo, dir: Direction, parent
     WebkitTransitionDuration: 'inherit',
     transitionDuration: 'inherit',
     width: layoutInfo.rect.width,
-    height: layoutInfo.rect.height,
+    height: layoutInfo.isSticky ? '100%' : layoutInfo.rect.height,
```
LFDanLu, line 70: weird that the checkbox shifts down without this, bit concerned that this would mean the computed layout height is no longer used

snowystinger: wonder if this is a content vs border box issue with box-sizing, the size of something could change when it gains the selected or focused border
do you know what the height is changing from/to?

reidbarber: The height of the checkbox cell doesn't actually change. I think the shift is due to showing the row focus indicator (which has `height: 100%`). The row has height `41`, and the checkbox cell has height `40`.

I think a better solution might just be to change the row focus indicator height to `calc(100% - 1px)`

reidbarber: This would cause the following change to the focus indicator, but I think that is okay, and maybe even better:

**Old:**
Old

---

**New:**
New

packages/@react-aria/dnd/src/useDropIndicator.ts
reidbarber: This seems like a good solution so I'll implement this for now!

packages/@react-spectrum/table/src/InsertionIndicator.tsx
```diff
@@ -0,0 +1,58 @@
...
+import styles from './table.css';
+import {useTableContext} from './TableView';
+import {useVisuallyHidden} from '@react-aria/visually-hidden';
+
+interface InsertionIndicatorProps {
+  target: ItemDropTarget,
+  rowProps: any
+}
+
+export default function InsertionIndicator(props: InsertionIndicatorProps) {
```
snowystinger, line 13: no default exports, only named

packages/@react-spectrum/table/src/InsertionIndicator.tsx
```diff
@@ -0,0 +1,58 @@
+import {classNames} from '@react-spectrum/utils';
+import {ItemDropTarget} from '@react-types/shared';
+import React, {useCallback, useRef} from 'react';
+import styles from './table.css';
+import {useTableContext} from './TableView';
+import {useVisuallyHidden} from '@react-aria/visually-hidden';
+
+interface InsertionIndicatorProps {
+  target: ItemDropTarget,
+  rowProps: any
```
snowystinger, line 10: can we restrict this at all?

packages/@react-spectrum/table/src/InsertionIndicator.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export default function InsertionIndicator(props: InsertionIndicatorProps) {
+  let {dropState, dragAndDropHooks, state} = useTableContext();
+  const {target, rowProps} = props;
+
+  let getKeyBefore = useCallback((key) => {
```
snowystinger, line 17: Something to discuss with team, better to have getKeyBefore and getKeyAfter to pass into the hook? or would it be better to pass a filter function (much like the dom tree walker has)

packages/@react-spectrum/table/src/RootDropIndicator.tsx
```diff
@@ -0,0 +1,28 @@
+import React, {useRef} from 'react';
+import {useTableContext} from './TableView';
+import {useVisuallyHidden} from '@react-aria/visually-hidden';
+
+export default function RootDropIndicator() {
```
snowystinger, line 5: no default exports

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -204,6 +245,35 @@ function TableView<T extends object>(props: SpectrumTableProps<T>, ref: DOMRef<H
...
+  let droppableCollection: DroppableCollectionResult;
+  let isRootDropTarget: boolean;
+  if (isTableDroppable) {
+    dropState = dragAndDropHooks.useDroppableCollectionState({
+      collection: state.collection,
+      selectionManager: state.selectionManager
+    });
+    droppableCollection = dragAndDropHooks.useDroppableCollection({
+      keyboardDelegate: layout,
+      // TODO: remove body from layout so drops go to rows
```
snowystinger, line 270: i think you mentioned in a standup this was done?

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -559,10 +668,16 @@ function TableVirtualizer({layout, collection, focusedKey, renderView, renderWra
...
+                classNames(
+                  stylesOverrides,
+                  {
+                    'react-spectrum-Table-body--dropTarget': !!isRootDropTarget
+                  }
+                )
               )
             }
-            tabIndex={-1}
+            tabIndex={isVirtualDragging ? null : -1}
```
snowystinger, line 680:
```suggestion
            tabIndex={isVirtualDragging ? undefined : -1}
```

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -943,15 +1094,51 @@ function TableRowGroup({children, ...otherProps}) {
...
+  return (
+    <FocusRing focusRingClass={classNames(stylesOverrides, 'focus-ring')}>
+      <div
+        {...dragButtonProps as React.HTMLAttributes<HTMLElement>}
+        className={
+        classNames(
+          stylesOverrides,
+          'spectrum-Table-row-draghandle-button'
+        )
+      }
```
snowystinger, line 1109: indent once more

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -1050,9 +1331,13 @@ function TableCheckboxCell({cell}) {
             },
             classNames(
               stylesOverrides,
-              'react-spectrum-Table-cell'
+              'react-spectrum-Table-cell',
+              {
+                'react-spectrum-Table-checkboxCell--isDraggable': isTableDraggable
```
snowystinger, line 1336: what's this for? i thought it would be so I could drag starting on a checkbox, but I don't see that as doing anything

packages/@react-spectrum/table/src/table.css
```diff
@@ -49,3 +64,148 @@
...
+}
+
+.react-spectrum-dragButtonHeadCell {
+  padding: 0;
+  height: inherit;
+}
+
+.react-spectrum-Table-row--dropTarget {
+  border: 1px var(--spectrum-table-dropindicator-border-color);
+  box-shadow: inset 2px 0 0 0 var(--spectrum-table-dropindicator-border-color),
```
snowystinger, line 192: what are all these 2 and -3 px values? can we use any variables for them? do they vary from medium to large?

packages/@react-spectrum/table/src/table.css
```diff
@@ -49,3 +64,148 @@
...
+    inset -2px 0 0 0 var(--spectrum-table-dropindicator-border-color),
+    inset 0 -3px 0 0 var(--spectrum-table-dropindicator-border-color),
+    inset 0 2px 0 0 var(--spectrum-table-dropindicator-border-color);
+
+  /* forced-color-adjust: none, so that box-shadow style will render */
+  forced-color-adjust: none;
+
+  /** Needed to style spectrum-Table-cellWrapper */
+  > * {
+    background-color: var(--spectrum-table-droptarget-background-color) !important;
```
snowystinger, line 202: odd it needs important, why is that? or what value is winning over it?

packages/@react-stately/table/src/TableCollection.ts
```diff
@@ -16,10 +16,12 @@ import {TableCollection as ITableCollection} from '@react-types/table';
 import {Key} from 'react';
 
 interface GridCollectionOptions {
-  showSelectionCheckboxes?: boolean
+  showSelectionCheckboxes?: boolean,
+  showDragButtons?: boolean
 }
 
 const ROW_HEADER_COLUMN_KEY = 'row-header-column-' + Math.random().toString(36).slice(2);
+const ROW_HEADER_COLUMN_KEY_DRAG = 'row-header-column-' + Math.random().toString(36).slice(2);
```
snowystinger, line 24: this and ROW_HEADER_COLUMN_KEY theoretically could be the same, we should check for a conflict and generate a new one if they do

packages/@react-stately/table/src/TableCollection.ts
```diff
@@ -197,7 +219,7 @@ export class TableCollection<T> extends GridCollection<T> implements ITableColle
         }
       };
 
-      columns.unshift(rowHeaderColumn);
+      columns.push(rowHeaderColumn);
```
snowystinger, line 222: why the change from unshift to push? that'll change its logical location in the table right?

packages/@react-stately/table/src/useTableState.ts
```diff
@@ -35,13 +37,16 @@ export interface TableState<T> extends GridState<T, ITableCollection<T>> {
...
   selectionMode: SelectionMode,
   columns: Node<T>[]
 }
 
 export interface TableStateProps<T> extends CollectionStateBase<T, ITableCollection<T>>, MultipleSelectionStateProps, Sortable {
   /** Whether the row selection checkboxes should be displayed. */
-  showSelectionCheckboxes?: boolean
+  showSelectionCheckboxes?: boolean,
+  /** Whether the row drag button should be displayed. */
+  showDragButtons: boolean
```
snowystinger, line 49: ListView doesn't have this, any chance we can drop this from the props?

snowystinger (commented):
Random observation that we'll need to get opinions on. Dropping on a table which is horizontally scrolled will cause the entire table to scroll to the first column.

In chrome, sometimes dragging a row on a tab causes the preview to fly in from the top in a weird arc.

packages/@react-spectrum/table/src/table.css
reidbarber: Updated to add logic to the renderWrapper that applies a class to the cell if the row is being dropped on, so this isn't needed anymore.

packages/@react-stately/table/src/TableCollection.ts
reidbarber: It just adds it after the drag button column. I supposed I could keep unshift, and just move the drag button column logic below it.

packages/@react-stately/table/src/useTableState.ts
reidbarber: I think this is necessary so that we know to include a cell for the drag handle when building the table collection. Open to suggestions if there is a way to go without it though.

packages/@react-spectrum/table/src/table.css
reidbarber: These are the same as ListView has. I don't think they are dependent on scale since it is basically a focus ring.

The reasoning behind the one -3px value, is because the bottom divider between rows will otherwise cover up 1px of the box shadow like so:

Screen Shot 2023-03-14 at 4 20 32 PM

snowystinger: ok, if it exists in listview, that's good enough for me. thanks for clearing that up

packages/@react-stately/table/src/useTableState.ts
snowystinger: ahhh

we just need the same logic that we have in TableView. we're already passing all the props through to useTableState

so just need this
```
    dragAndDropHooks
  } = props;
  let isTableDraggable = !!dragAndDropHooks?.useDraggableCollectionState;
```

then we shouldn't need a special prop

packages/@react-stately/table/src/useTableState.ts
```diff
@@ -55,10 +63,12 @@ const OPPOSITE_SORT_DIRECTION = {
  */
 export function useTableState<T extends object>(props: TableStateProps<T>): TableState<T> {
   let [isKeyboardNavigationDisabled, setKeyboardNavigationDisabled] = useState(false);
-  let {selectionMode = 'none'} = props;
+  let {selectionMode = 'none', dragAndDropHooks} = props;
+  let isTableDraggable = !!dragAndDropHooks?.useDraggableCollectionState;
```
snowystinger, line 67: i know it won't change over a components lifetime, but we should add this to the useMemo dep array like you had before on line 75

packages/@react-aria/dnd/src/useDropIndicator.ts
```diff
@@ -62,10 +70,10 @@ export function useDropIndicator(props: DropIndicatorProps, state: DroppableColl
     });
   } else {
     let before = target.dropPosition === 'before'
-      ? collection.getKeyBefore(target.key)
+      ? getKeyBefore(target.key)
```
devongovett, line 73: I wonder if `collection.getFirstKey() === target.key` would work? I think TableCollection returns the first row from `getFirstKey()` rather than the header row. Might be able to avoid adding new API that way.

LFDanLu: My concern with using `getFirstKey`/`getLastKey` is with Sections where the `collection.body.bodyNodes` which `getFirstKey` uses will be the sections themselves rather than the rows. We can go with `getFirstKey()` in this PR for now and I can look into changing that in my PR

devongovett: I think when we have sections we will still want `getFirstKey` to return the first real row rather than the section key. That way tabbing into the table would go to the first row. I think that's how ListCollection works with sections as well?

LFDanLu: Ya I can make that change in the TableCollection, I was currently handling the tabbing bit via the keyboard delegate logic instead (felt more appropriate to have it there to separate collection vs keyboard specific logic)

LFDanLu: It turns out that ListCollection actually returns the first section key from `getFirstKey`. I could either change it so all the collection key getters only return row keys (feels like a breaking change?) or add new getters that only return row keys specifically (e.g. getFirstRowKey. getRowKeyAfter). Alternatively, we could provide the `keyboardDelegate`/`layout` to `useDropIndicator` like we do to `useDroppableCollection`, which feels more correct to me.

devongovett: hmm yeah. would have to be optional though because otherwise that would also be a breaking change.

snowystinger (commented):
We need WHCM as well, are we doing that in a followup?

reidbarber:
> We need WHCM as well, are we doing that in a followup?

I can handle that here.

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -546,7 +549,7 @@ export class TableLayout<T> extends ListLayout<T> {
         let indices = this.persistedIndices.get(layoutInfo.parentKey);
         if (!indices) {
           // stickyColumnIndices are always persisted along with any cells from persistedKeys.
-          indices = collectionNode.type === 'cell' ? [...this.stickyColumnIndices] : [];
+          indices = collectionNode.type === 'cell' || collectionNode.type === 'column' ? [...this.stickyColumnIndices] : [];
```
reidbarber, line 552: Fixes the drag button header cell not being persisted

packages/@react-types/table/src/index.d.ts
```diff
@@ -63,7 +64,11 @@ export interface SpectrumTableProps<T> extends TableProps<T>, SpectrumSelectionP
    * Handler that is called after a user performs a column resize.
    * Can be used to store the widths of columns for another future session.
    */
-  onResizeEnd?: (widths: Map<Key, ColumnSize>) => void
+  onResizeEnd?: (widths: Map<Key, ColumnSize>) => void,
+  /**
+   * The drag and drop hooks returned by `useDragAndDrop` used to enable drag and drop behavior for the TableView.
+   */
+  dragAndDropHooks?: DragAndDropHooks['dragAndDropHooks']
```
LFDanLu, line 71: TODO: add `@alpha` to let ppl know

reidbarber: `@alpha` isn't valid JSDoc and fails lint.... I'll add `@verion alpha`

packages/@react-stately/table/src/useTableState.ts
```diff
@@ -55,14 +61,16 @@ const OPPOSITE_SORT_DIRECTION = {
...
 
   let context = useMemo(() => ({
     showSelectionCheckboxes: props.showSelectionCheckboxes && selectionMode !== 'none',
+    showDragButtons: isTableDraggable,
```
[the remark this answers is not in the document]
reidbarber: Added this back

packages/@react-types/table/package.json
```diff
@@ -9,6 +9,7 @@
     "url": "https://github.com/adobe/react-spectrum"
   },
   "dependencies": {
+    "@react-spectrum/dnd": "^3.0.1",
```
devongovett, line 12: Hmm not ideal to have a react-spectrum dependency in react-types... Wonder if we should move `SpectrumTableProps` into the `@react-spectrum/table` package also?

packages/@react-types/table/package.json
```diff
@@ -9,6 +9,7 @@
     "url": "https://github.com/adobe/react-spectrum"
   },
   "dependencies": {
+    "@react-spectrum/dnd": "^3.1.0",
```
devongovett, line 12:
```suggestion
```

devongovett (approved):

PR #4018, opened 2023-02-06
Center the table in the viewport if resizer is still out of view

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Use any of the table resize stories. Test that the table is only centered in the viewport if the resizer is still out of view after moving it with Left/Right arrows

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-spectrum/table/src/TableView.tsx
+++ b/packages/@react-spectrum/table/src/TableView.tsx
@@ -11,7 +11,7 @@
  */
 
 import ArrowDownSmall from '@spectrum-icons/ui/ArrowDownSmall';
-import {chain, mergeProps, scrollIntoViewport, useLayoutEffect} from '@react-aria/utils';
+import {chain, mergeProps, scrollIntoView, scrollIntoViewport, useLayoutEffect} from '@react-aria/utils';
 import {Checkbox} from '@react-spectrum/checkbox';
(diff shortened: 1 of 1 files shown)


packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -478,10 +478,11 @@ function TableVirtualizer({layout, collection, focusedKey, renderView, renderWra
   // header scroll position
   useEffect(() => {
     if (getInteractionModality() === 'keyboard' && headerRef.current.contains(document.activeElement)) {
-      document.activeElement?.scrollIntoView?.({block: 'nearest', inline: 'nearest'});
+      scrollIntoView(headerRef.current, document.activeElement as HTMLElement);
+      scrollIntoViewport(document.activeElement, {containingElement: domRef.current});
```
LFDanLu, line 482: Needs these two called in succession so we first scroll the resizer into view w/ respect to the column header, and then we check if we need to scroll the entire table into the viewport if it is still isn't visible. If we omit the first scrollIntoView call, the table will always get centered in the viewport when resizing the last column to be larger.

snowystinger (approved):

reidbarber (approved):
LGTM

PR #4039, opened 2023-02-10
Narrow combobox focus event handler types

joshuajaco:
The following code gives a typescript error, because the event target is typed as `Element` instead of `HTMLInputElement`, even though `ComboBox` forwards the `onFocus` prop to an `input` element.
```tsx
  <ComboBox onFocus={e => e.target.select()}>...</ComboBox>
```

The same goes for the `useComboBox` and `useComboBoxState` hooks. 
The `inputRef` prop is already enforced to contain a `HTMLInputElement`, so I think it is safe to assume that the focus props will always be forwarded to an `input` element.

```tsx
  useComboBox({ onFocus: e => e.target.select(), ...props });
  useComboBoxState({ onFocus: e => e.target.select(), ...props });
```

This PR makes the `FocusEvents` and `FocusableProps` types generic (similar to [`FocusEvent`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/750080df0d047c022fc1f802d02456623f621e47/types/react/index.d.ts#L1237-L1240)), which allows `ComboBoxProps` to override the event target type, while maintaining the old behaviour for other usages.
This pattern could be incrementally adopted by other components or other event handler types like `KeyboardEvents`.


devongovett:
GET_BUILD

devongovett (approved):
Thanks!

snowystinger (approved):

LFDanLu:
DanielBot: 
Storybook 18: https://reactspectrum.blob.core.windows.net/reactspectrum/92d699c6c26adab8c4097414f65c126e75a64e24/storybook/index.html?path=/story/accordion--default&providerSwitcher-express=false&providerSwitcher-toastPosition=bottom
Docs: https://reactspectrum.blob.core.windows.net/reactspectrum/92d699c6c26adab8c4097414f65c126e75a64e24/docs/react-spectrum/index.html

Branch build: https://github.com/adobe/react-spectrum/commits/joshuajaco-fix-combobox-focus-prop-types

LFDanLu (approved):
LGTM, confirmed the types in the docs

PR #4041, opened 2023-02-10
Fix TagGroup maxRows layout issues

reidbarber:
This fixes two issues:
- Previously, when calculating if there was enough space on the last row for the action buttons, we weren't factoring in margins. This lead to cases where the action buttons could end up on the next row.
- Some flickering was occurring when resizing due to the buttons flowing onto an extra row for one render. Now, we set a maxHeight when maxRows is provided, since the group's height shouldn't grow.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Test maxRows and onAction stories. Resize container and ensure that the action buttons never wrap to an extra line beyond the maxRows provided.


packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -120,8 +121,13 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
 
         // Measure, and update to show the items until maxRows is reached.
         yield computeVisibleTagCount();
+
+        // Set a fixed height on the container
+        setMaxHeight(containerRef.current?.getBoundingClientRect().height);
```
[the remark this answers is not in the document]
reidbarber: So to reproduce the original issue:
- In the maxRows story, grow the container, then shrink it again. As it shrinks, there is one render where the button will go to the next line (creating a flicker and potentially causing a layout shift)

The intent of this fix is to add a static height to the container (and overflow:hidden), so that no flickering or layout shifts will occur for that extra render while resizing. 

We could probably keep that static height value in a ref to avoid extra renders.

Ideally we would want to avoid that extra render in the first place, but I'm not sure how we would do that (open to ideas.)

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -166,6 +172,7 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
           )
         }>
         <div
+          style={maxRows != null && isCollapsed ? {maxHeight, overflow: 'hidden'} : {}}
```
snowystinger, line 175:
```suggestion
          style={maxRows != null && isCollapsed ? {maxHeight, overflow: 'hidden'} : undefined}
```

packages/@react-spectrum/tag/src/TagGroup.tsx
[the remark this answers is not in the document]
snowystinger: I think we could take the height+margins of a single tag and multiply it by the maxRows to determine this value earlier. It could be included in the returned yielded state.

It stops the flickering on a new line, but it hasn't quite stopped the flickering on the current line.
If we kept an identical copy of the tags in a visually hidden div, we could do our measurements on that, then just apply the result to the visible one. That would make it appear as though it just arrived at the correct spot immediately. We might still want/need the maxHeight.

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -100,15 +101,15 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
         let lastTagEnd = tags[index - 1]?.getBoundingClientRect()[end];
+        lastTagEnd += parseInt(window.getComputedStyle(tags[index - 1]).marginRight, 10);
         let availableWidth = containerEnd - lastTagEnd;
-        for (let tagWidth of tagWidths.reverse()) {
-          if (availableWidth >= buttonsWidth || index <= 1 || index >= state.collection.size) {
-            break;
-          }
-          availableWidth += tagWidth;
+
+        while (availableWidth < buttonsWidth && index > 1 && index < state.collection.size) {
```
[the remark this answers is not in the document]
reidbarber: This skips decrementing index in the case of all tags already being shown. If we remove this, we might end up in this state:

Screen Shot 2023-02-13 at 6 06 47 PM

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -166,6 +172,7 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
           )
         }>
         <div
+          style={maxRows != null && isCollapsed ? {maxHeight: tagState.maxHeight, overflow: 'hidden'} : undefined}
```
ktabors, line 175: image
I went to the "with-label-description-contextual-help" and set the maxRows to one. I narrow the container until the button disappears. I'm not able to reproduce this on main.

When I remove the `overflow: hidden` the button shows up where the `description` text is.

ktabors (changes requested):

packages/@react-spectrum/tag/src/TagGroup.tsx
reidbarber: Previously, it would always show at least one tag, but I've updated the logic to handle this edge case.

snowystinger:
Screenshot 2023-03-20 at 3 44 11 PM
I can get it into a situation where the show all/more button is hidden as shown in the screenshot

reidbarber:
> I can get it into a situation where the show all/more button is hidden as shown in the screenshot

@snowystinger  Looks like this is due to there not being enough width for the collapse button. I'll update this to still show the collapse button in this case.

snowystinger (approved):
looks better than before

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -100,31 +99,37 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
           index--;
         }
-        return {visibleTagCount: index, showCollapseButton: index < state.collection.size};
+        let tagStyle = window.getComputedStyle(tags[0]);
+        let maxHeight = (parseInt(tagStyle.height, 10) + parseInt(tagStyle.marginTop, 10) * 2) * maxRows;
+        return {
+          visibleTagCount: index,
+          showCollapseButton: index < state.collection.size,
+          maxHeight
+        };
```
LFDanLu, line 119: I was able to get it the taggroup into the following state where the "clear" button wasn't visible:
image
This was done via this [story](https://reactspectrum.blob.core.windows.net/reactspectrum/1506856214b1e7435876cd11fb62d193801ea221/storybook/index.html?path=/story/taggroup--with-action-and-max-rows&args=maxRows:3&providerSwitcher-express=false) by setting maxRows to 3. Do we need to take that button's existence into account as well?

reidbarber: Good catch. We shouldn't be setting the maxHeight in this case. I'll update the logic to reflect that.

devongovett:
I still see flickering of a third row when resizing... ([here](https://reactspectrum.blob.core.windows.net/reactspectrum/251d4b9f37994ccc9ef663b41c0531e610dac7c0/storybook/index.html?path=/story/taggroup--max-rows&providerSwitcher-express=false))

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -167,6 +172,7 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
           )
         }>
         <div
+          style={maxRows != null && tagState.showCollapseButton && isCollapsed ? {maxHeight: tagState.maxHeight, overflow: 'hidden'} : undefined}
```
devongovett, line 175: I think the flickering is when it transitions from not having an collapse button to having one for the first time, because there's no maxHeight defined when no collapse button.

devongovett (approved):
Let's merge this anyway since it is improved and follow up again with more after testing.

PR #1297, opened 2020-11-19
Fix incorrect positioning of useOverlayPosition

raon0211:
The positioning of `useOverlayPosition` is incorrect.
When placement is given as `'bottom left'`, the overlay is placed to the incorrect position as the following screenshot:

image

while it should be positioned to the correct position as:

image

This fixes the behavior by removing the min/max logic in `calculatePosition`.

## ✅ Pull Request Checklist:

- [ ] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- [x] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [x] Filled out test instructions.
- [x] Updated documentation (if it already exists for this component).
- [ ] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/TR/wai-aria-practices-1.1/)

## 📝 Test Instructions:

I added some storybook test cases. Without the fix, the overlays will be positioned in a wrong position.

PR #4097, opened 2023-02-22
@react-aria/overlays - Fix overlay position clamping and arrow position

filipw01:
Relates to https://github.com/adobe/react-spectrum/pull/1297 and https://github.com/adobe/react-spectrum/pull/3756

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Open newly created story in UseOverlayPosition story and play with different positions and offsets. More details about the issue that was fixed here https://github.com/adobe/react-spectrum/pull/1297

Open any TooltipTrigger story and play with offset controls to see how the and clamping arrow behaves

## 🧢 Your Project:

https://www.chilipiper.com/


filipw01:
The only thing left is to replace the `const ARROW_WIDTH = 8` with a prop value from `useOverlayPosition`

@LFDanLu You mentioned detecting arrow width variable from react-spectrum somewhere. Won't it be better to leave it decoupled and add use `arrowWidth` prop on `useOverlayPosition` wherever react-spectrum is using it?

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -88,6 +89,8 @@ const AXIS_SIZE = {
   left: 'width'
 };
 
+const ARROW_WIDTH = 8;
```
filipw01, line 92: This will be updated to use `useOverlayPosition` options object

filipw01:
I updated the react-spectrum components to use the new arrowCrossSize option

LFDanLu:
GET_BUILD

filipw01:
One think I noticed is that react-spectrum tooltips and popovers don't have enough overlap between arrow and overlay so they look a bit broken in these edge cases
Screenshot 2023-02-24 at 21 48 32

LFDanLu (commented):
Pushed a story to your fork for testing with Dialogs as well, hope you don't mind. Things look pretty good in preliminary testing with combinations of scrolling/crossOffset/positioning.

> You mentioned detecting arrow width variable from react-spectrum somewhere. Won't it be better to leave it decoupled and use arrowCrossSize option on useOverlayPosition wherever react-spectrum is using it?

Yeah, the concern here was mainly centered around how the arrow's existence is more of a Spectrum design thing and we wanna keep the overlayPosition hook/calculatePosition clear of those design system specifics. However, I feel having an arrow attached to a trigger + popover isn't a terribly uncommon thing so might not be too big of a deal to have that prop there.

> One think I noticed is that react-spectrum tooltips and popovers don't have enough overlap between arrow and overlay so they look a bit broken in these edge cases. How should we resolve this?

I'm leaning towards solving that outside the hooks/positioning logic since it is specific to Spectrum, maybe by modifying the arrow so it meshes better with the curved popover edge in that situation

LFDanLu:
GET_BUILD

filipw01:
> Pushed a story to your fork for testing with Dialogs as well, hope you don't mind

Of course not. I can work on fixing the arrows visual side on Monday, but if you want to close it before that feel free to do so

> However, I feel having an arrow attached to a trigger + popover isn't a terribly uncommon thing so might not be too big of a deal to have that prop there.

I agree this is a pretty common thing. Plus it already assumes there can be an arrow because it exposes arrowProps

> I'm leaning towards solving that outside the hooks/positioning logic since it is specific to Spectrum, maybe by modifying the arrow so it meshes better with the curved popover edge in that situation

Great 😄  As mentioned before, will take a look on Monday. I suspected that's the way to go here, but wanted to confirm before introducing even more changes

filipw01:
@LFDanLu I tried adjusting the arrows to look good, but it was a messy process, and the end effect wasn't good.

I approached it from another perspective, increasing the "arrowSize" so it never touches the rounded part of the overlay. I think this approach is much simpler and looks better. Let me know what you think.

LFDanLu:
GET_BUILD

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -75,13 +86,20 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
   } = props;
   let {styleProps} = useStyleProps(props);
 
+  let {size, borderWidth, arrowRef} = useArrowSize();
+  let borderDiagonal = borderWidth * ROOT_2;
+  let primary = size + borderDiagonal;
+  let secondary = primary * 2;
+  let arrowTotalCrossPadding = 4;
```
LFDanLu, line 93: We'll need to have a different value for medium vs large scale for positioning consistency. You can access the current scale via `useProvider` Not a blocker though

Medium:
image


Large:
image

LFDanLu (approved):
Calculations looks sound to me, just one small nitpick but that can be handled separately if need be

packages/@react-spectrum/overlays/src/Popover.tsx
snowystinger: should be equivalent to `--spectrum-popover-border-radius`, it might be a merged variable though, so we can add later, as @LFDanLu said, not a blocker
https://github.com/adobe/react-spectrum/blob/main/packages/%40adobe/spectrum-css-temp/components/popover/index.css#L47

packages/@react-spectrum/tooltip/src/TooltipTrigger.tsx
```diff
@@ -42,6 +42,8 @@ function TooltipTrigger(props: SpectrumTooltipTriggerProps) {
     trigger: triggerAction
   }, state, tooltipTriggerRef);
 
+  let arrowTotalCrossPadding = 8;
```
snowystinger, line 45: why 8 for this one?
this one should be equivalent to this I think?
`border-radius: var(--spectrum-border-radius);` from
https://github.com/adobe/react-spectrum/blob/main/packages/%40adobe/spectrum-css-temp/components/tooltip/index.css#L36

packages/@react-spectrum/tooltip/src/TooltipTrigger.tsx
```diff
@@ -50,7 +52,8 @@ function TooltipTrigger(props: SpectrumTooltipTriggerProps) {
     crossOffset,
     isOpen: state.isOpen,
     shouldFlip: props.shouldFlip,
-    containerPadding: props.containerPadding
+    containerPadding: props.containerPadding,
+    arrowCrossSize: 8 + arrowTotalCrossPadding
```
snowystinger, line 56: is this 2x? or why does it have a const 8 with a variable that is also 8?

snowystinger (dismissed):
Some small clarifications, otherwise it's looking good to me. It has fixed the issue we had with Table Column Resizing Menu being offset an odd amount.

packages/@react-spectrum/tooltip/src/TooltipTrigger.tsx
filipw01: You're right, it is `border-radius * 2`. 
I noticed it's not as simple as getting `--spectrum-border-radius` because what I actually see in the compiled CSS is
`border-radius: var(--spectrum-alias-border-radius-regular, var(--spectrum-global-dimension-size-50))`. For that reason I'm updating it to use the borderRadius property value directly instead

packages/@react-spectrum/overlays/src/Popover.tsx
filipw01: `--spectrum-popover-border-radius` is not defined (lost in some compilation step?) I used borderRadius directly from the overlay element to make it work

packages/@react-spectrum/tooltip/src/TooltipTrigger.tsx
filipw01: 8 here stands for the arrow width and arrowTotalCrossPadding is invisible space so visible part of the arrow doesn't overlap the part of overlay with rounder corner. I'm updating to get the arrow size dynamically now

LFDanLu:
GET_BUILD

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -75,13 +86,21 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
...
+  let primary = size + borderDiagonal;
+  let secondary = primary * 2;
+  let arrowTotalCrossPadding = borderRadius * 2;
+
   let {popoverProps, arrowProps, underlayProps, placement} = usePopover({
     ...props,
     popoverRef: ref,
-    maxHeight: null
+    maxHeight: null,
+    arrowCrossSize: secondary + arrowTotalCrossPadding
```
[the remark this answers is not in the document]
filipw01: Nice catch. I fixed it

snowystinger (approved):
thanks so much for sticking with this!

LFDanLu (commented):
LGTM, thanks for all your work on this! Provisional approval until we get this tested/reviewed by the rest of the team. Will also push a small fix to those DialogTrigger stories I added previously

snowystinger (approved):

LFDanLu:
GET_BUILD

LFDanLu:
@filipw01 Apologies, our testing session today revealed that we actually want the arrow to be centered with respect to the button if possible, only falling back to pointing to the center of the overlay if need be. This was a on my part, sorry for all the back and forth. I've opened a PR based off your work [here](https://github.com/adobe/react-spectrum/pull/4220) so we can hopefully fast track these changes. Thanks for all your work on this once again!

LFDanLu:
We've merged https://github.com/adobe/react-spectrum/pull/4220 and will be putting it through a final round of testing next Monday. Thanks once again for all your work on this and for bearing with all of our design iterations!

PR #4102, opened 2023-02-22
Fix TagGroup label alignment

reidbarber:
- Uses span instead of label since we aren't labelling a form element
- Uses `gap` so that the tags line up with the label

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test TagGroup label is span, and left-most tags align vertically with the label above.

### DIFF
--- a/packages/@adobe/spectrum-css-temp/components/tags/index.css
+++ b/packages/@adobe/spectrum-css-temp/components/tags/index.css
@@ -17,7 +17,13 @@ governing permissions and limitations under the License.
 }
 
 .spectrum-Tags {
-  display: inline;
+  display: contents;
+}
+
+.spectrum-Tags-container {
+  display: inline-flex;
+  flex-wrap: wrap;
+  gap: var(--spectrum-tag-margin);
 }
 
 .spectrum-Tags-item {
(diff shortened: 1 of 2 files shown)


snowystinger (dismissed):
looks like it's working, display contents to the rescue

packages/@adobe/spectrum-css-temp/components/tags/index.css
```diff
@@ -17,7 +17,13 @@ governing permissions and limitations under the License.
 }
 
 .spectrum-Tags {
-  display: inline;
+  display: contents;
```
devongovett, line 20: Unfortunately I think this still breaks accessibility in Safari. https://bugs.webkit.org/show_bug.cgi?id=239479 When I tested it announces the table as having 0 columns and 0 rows and you cannot navigate to the cells with the voiceover cursor, whereas before it announced correctly.

devongovett (approved):
lgtm, we should probably run chromatic just to be sure though

reidbarber:
> lgtm, we should probably run chromatic just to be sure though

We haven't added chromatic stories for TagGroup yet fyi

devongovett:
Oh ok

LFDanLu (approved):
LGTM

PR #4124, opened 2023-02-24
Fix toast ESM build and SSR support

devongovett:
This fixes two issues using Toast in Next.js

1. `use-sync-external-store` doesn't have package.json exports, so it doesn't work in ESM. I filed a PR here to fix it https://github.com/facebook/react/pull/26230 but for now we can import explicitly from `use-sync-external-store/shim/index.js`.
2. `useSyncExternalStore` has an "optional" third argument called `getServerSnapshot` which isn't really optional because it errors if it isn't provided on the server. So now I have implemented it. Also added SSR tests to catch this.

### DIFF
--- a/examples/rsp-next-ts/package.json
+++ b/examples/rsp-next-ts/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@adobe/react-spectrum": "^3.22.0",
     "@react-spectrum/color": "^3.0.0-beta.16",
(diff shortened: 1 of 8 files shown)


LFDanLu (approved):
LGTM, verified that the example app renders the toast properly versus the previous behavior where it fails during build with the `Did you mean to import use-sync-external-store/shim/index.js?` error

reidbarber (approved):
LGTM

PR #4173, opened 2023-03-08
Supporting sections in ListView

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/@react-stately/table/src/TableCollection.ts
```diff
@@ -217,6 +217,12 @@ export class TableCollection<T> extends GridCollection<T> implements ITableColle
             }
           }
           break;
+        case 'section':
```
LFDanLu, line 220: Note: The way we calculate index + how I conflated the `section` as a `row` bothered me a bit so I've done some (extremely rough) work [here](https://github.com/adobe/react-spectrum/commit/d8252866fb56c09126a2600260427475d595b18b) to refactor the Table collection, layout, etc so that the TableSection generates its own row + cell. It does complicate the code a little bit and I can't quite settle on if it is overkill so pausing on the work there

packages/@react-stately/list/src/ListCollection.ts
```diff
@@ -18,17 +18,27 @@ export class ListCollection<T> implements Collection<Node<T>> {
   private iterable: Iterable<Node<T>>;
   private firstKey: Key;
   private lastKey: Key;
+  sections: Node<T>[];
+  rows: Node<T>[];
```
LFDanLu, line 22: TODO: try removing these so migrating to RAC is easier. Have the warning for mixed non-section rows and sections happen in here instead?

LFDanLu:
PR is in a functionally stable state now (behavior/implementation wise). Will either revisit when a RAC collection transition happens to ListView or we can aim to review/move ahead with this as is.

packages/@react-spectrum/list/test/ListView.test.js
```diff
@@ -90,6 +90,8 @@ describe('ListView', function () {
     );
     // Allow for Virtualizer layout to update
     act(() => {jest.runAllTimers();});
+    // Allow section layout info to be updated as a result of the initial render's useVirtualizerItem call
+    act(() => {jest.runAllTimers();});
```
LFDanLu, line 94: This is actually something that we are missing in our tests in main for ListBox, we just haven't encountered it because we don't have PageUp/Down tests. useVirtualizerItem updates the section header height in a useLayoutEffect to match the test's mocked scrollHeight from the default of 48 which ListLayout sets initially. This `virtualizer.updateItemSize` call triggers a raf [here](https://github.com/adobe/react-spectrum/blob/93c6b3c17e205df552f3295f909581367ca53912/packages/%40react-stately/virtualizer/src/Virtualizer.ts#L463-L466), hence the need for a second runAllTimers

PR #4201, opened 2023-03-14
fix(#4200): NumberField issue with flex layout and forms

### LINKED ISSUE #4200: NumberField issue with flex layout and forms
# 🐛 Bug Report 

See https://reactspectrum.blob.core.windows.net/reactspectrum/4a7576870d4dbe80cdbebba2043f593474ad6cb4/storybook/index.html?path=/story/form--form-with-numberfield-and-locale-ar-ae&providerSwitcher-express=false

## 🤔 Expected Behavior

Screenshot of NumberField decrement button layout bug fix within a flex layout

## 😯 Current Behavior

Screenshot of NumberField decrement button layout bug within a flex layout

## 💁 Possible Solution

At
https://github.com/adobe/react-spectrum/blob/4a7576870d4dbe80cdbebba2043f593474ad6cb4/packages/%40adobe/spectrum-css-temp/components/fieldlabel/index.css#L84-L88

add `align-self: start;`:
```css
    .spectrum-Field-field {
      /* If the user overrides the width of the field, propagate to the inner component */
      width: 100%;
      grid-area: field;
      align-self: start;
    }
```

and at:
https://github.com/adobe/react-spectrum/blob/4a7576870d4dbe80cdbebba2043f593474ad6cb4/packages/%40adobe/spectrum-css-temp/components/fieldlabel/index.css#L119-L122

add `align-self: start;`:
```css
      .spectrum-Field-field {
        /* If the user overrides the width of the field, propagate to the inner component */
        width: 100%;
      }
```

## 🌍 Your Environment

| react-spectrum   | @adobe/react-spectrum@3.25.1
| Browser          | Chrome Version 110.0.5481.177 (Official Build) (arm64)

## 🧢 Your Company/Team

Adobe/Accessibility

majornista:
Closes #4200 

## ✅ Pull Request Checklist:

- [x] Included link to corresponding [React Spectrum GitHub Issue #4200](https://github.com/adobe/react-spectrum/issue/4200).

## 📝 Test Instructions:

Examine https://reactspectrum.blob.core.windows.net/reactspectrum/47b4b51649cf63344d204e8ce229197bdff9d320/storybook/index.html?path=/story/form--form-with-numberfield-and-locale-ar-ae&providerSwitcher-express=false to verify that there is no longer a layout problem with the first NumberField instance.

## 🧢 Your Project:

Adobe/Accessibility


majornista:
GET_BUILD

snowystinger:
Will need to run chromatic on this

reidbarber (approved):
LGTM in Chrome/FF/Safari

majornista:
@snowystinger Can I go ahead and accept the issues I fixed as new baselines?

snowystinger (approved):
looks correct now, and i can still use the prop on the component to alignSelf end or center, whatever I like

devongovett:
Only question is if this might cause an issue when placed in a container that has a different alignment. For example, what if a NumberField is in a flex container with `align-items: center`. I think this would override that and put it at the start which might be unexpected. I guess we only want to avoid stretching, but not override the alignment of the parent?

devongovett:
Also what happens when there's no label so it's not wrapped in a Field? Also what about if there is help text? Right now the Field still stretches but the NumberField inside doesn't anymore. But if the Field stretches does that push the help text to the bottom away from the NumberField?

snowystinger:
> Only question is if this might cause an issue when placed in a container that has a different alignment. For example, what if a NumberField is in a flex container with `align-items: center`. I think this would override that and put it at the start which might be unexpected. I guess we only want to avoid stretching, but not override the alignment of the parent?

acht, yeah, it would override the container trying to align children, that's unfortunate

maybe the original proposal @majornista had of max-content height would be better?

majornista:
How about `block-size: max-content;` ?

snowystinger (approved):
checked with a few more examples i added locally to the 'flexed' NumberField story. Works with/without label, works with/without helptext, parent can set the alignment

devongovett (dismissed):

devongovett:
hmm actually one remaining problem is with help text

image

snowystinger (approved):

devongovett (approved):

PR #4210, opened 2023-03-15
(WIP) Sections in table

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/@react-stately/table/src/TableCollection.ts
```diff
@@ -217,6 +217,10 @@ export class TableCollection<T> extends GridCollection<T> implements ITableColle
             }
           }
           break;
+        case 'section':
```
LFDanLu, line 220: Note: The way we calculate index + how I conflated the `section` as a `row` bothered me a bit so I've done some (extremely rough) work [here](https://github.com/adobe/react-spectrum/commit/d8252866fb56c09126a2600260427475d595b18b) to refactor the Table collection, layout, etc so that the TableSection generates its own row + cell and the index is calculated with respect to the node's position within its parent. It does complicate the code a little bit and I can't quite settle on if it is overkill so pausing on the work there

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -504,6 +518,53 @@ export class TableLayout<T> extends ListLayout<T> {
         }
         break;
       }
+      case 'section': {
```
snowystinger, line 521: this case seems very similar to rowgroup, can we combine some logic?

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -504,6 +518,53 @@ export class TableLayout<T> extends ListLayout<T> {
...
+          // Skip persisted rows that overlap with visible cells.
+          while (persistedRowIndices && persistIndex < persistedRowIndices.length && persistedRowIndices[persistIndex] < i) {
+            persistIndex++;
+          }
+
+          res.push(node.children[i].layoutInfo);
+          this.addVisibleLayoutInfos(res, node.children[i], rect);
+        }
+
+        // Add persisted rows after the visible rows.
```
snowystinger, line 556: I'm not seeing a focused tablerow persisted when I scroll it out of view

packages/@react-stately/table/src/TableBody.ts
```diff
@@ -31,17 +31,36 @@ TableBody.getCollectionNode = function* getCollectionNode<T>(props: TableBodyPro
         }
 
         for (let item of items) {
-          yield {
-            type: 'item',
-            value: item,
-            renderer: children
-          };
+          let element = children(item);
```
snowystinger, line 34: isn't this going to be expensive?

snowystinger (commented):
In general this appears to be in a pretty good place, can we write some tests so we know we don't break things as we continue to review?

packages/@react-aria/grid/src/useGridSectionAnnouncement.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function useGridSectionAnnouncement<T>(state: GridSelectionState<T>) {
+  let {collection, selectionManager} = state;
+  let stringFormatter = useLocalizedStringFormatter(intlMessages);
+  let focusedItem = selectionManager.focusedKey != null && state.selectionManager.isFocused
+    ? state.collection.getItem(state.selectionManager.focusedKey)
+    : undefined;
+
+  // VoiceOver on MacOS doesn't announce TableView/ListView sections when navigating with arrow keys so we do this ourselves
+  // TODO: what other information should be conveyed? Should we announce the number of rows in the section and the current focused row/cell?
```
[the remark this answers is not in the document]
LFDanLu: The amount of information announced by the other screen readers is pretty lacking as well (NVDA being the best at announcing a grouping change), this introduces parity. If we are interested in announcing the number of rows in the section, this could be expanded to cover other screen readers as well, will have to test for any double announcements

packages/@react-aria/grid/src/useGridSectionAnnouncement.ts
```diff
@@ -0,0 +1,62 @@
...
+  let parentNode = collection.getItem(focusedItem?.parentKey ?? null);
+  while (sectionKey == null && parentNode) {
+    if (parentNode.type === 'section') {
+      sectionKey = parentNode.key;
+    }
+    parentNode = collection.getItem(parentNode?.parentKey ?? null);
+  }
+  let lastSection = useRef(sectionKey);
+  useUpdateEffect(() => {
+    if (isMac() && focusedItem != null && sectionKey !== lastSection.current) {
```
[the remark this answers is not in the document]
LFDanLu: good idea, I'll add that check

packages/@react-aria/table/src/useTableSection.ts
```diff
@@ -0,0 +1,65 @@
...
+  rowProps: DOMAttributes,
+  /** Props for the table section's grid cell element. */
+  gridCellProps: DOMAttributes
+}
+
+/**
+ * Provides the behavior and accessibility implementation for a section in a table.
+ * @param props - Props for the section.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
LFDanLu: Ah, seems to be some leftovers from a revert

packages/@react-stately/layout/src/TableLayout.ts
LFDanLu: ya probably, this was a separated in a first pass to make sure the logic checked out

packages/@react-stately/layout/src/TableLayout.ts
LFDanLu: What story? I'm seeing the focused row + section persist when scrolled out of view:
image
Focused row is the `aria-rowindex=3` with the section row header above it. The next row in the dom is `aria-rowindex=6` which is non-contiguous since that was the next row that isn't persisted since it is close enough to being in view. https://reactspectrum.blob.core.windows.net/reactspectrum/0c9a33167b9201eea8842469e0d613fed710bb12/storybook/index.html?path=/story/tableview--dynamic-sections&providerSwitcher-express=false

snowystinger: Ah, that story, but focus row 8 and scroll it off the bottom.

packages/@react-stately/table/src/TableBody.ts
LFDanLu: potentially, I'll need to test with a section with many rows. Open to any suggestions since I don't think I can distinguish between the Rows and Sections without doing something like that

packages/@react-stately/layout/src/TableLayout.ts
LFDanLu: Ah good catch, lemme fix

LFDanLu:
> In general this appears to be in a pretty good place, can we write some tests so we know we don't break things as we continue to review?

Yep, tests + WHCM + DnD are still to come, still very much in a WIP status

packages/@react-stately/grid/src/GridCollection.ts
```diff
@@ -23,92 +23,106 @@ export class GridCollection<T> implements IGridCollection<T> {
...
   rows: GridNode<T>[];
+  sections: GridNode<T>[];
 
   constructor(opts?: GridCollectionOptions<T>) {
     this.keyMap = new Map();
     this.columnCount = opts?.columnCount;
     this.rows = [];
+    this.sections = [];
+
+    let visit = (node: GridNode<T>, i?: number) => {
```
LFDanLu, line 34: Refactored GridCollection a bit via [6f60708](https://github.com/adobe/react-spectrum/pull/4210/commits/6f607084ff8ed1ec638ce72fc8679d81d75e724f), see that commit and [ca029a9](https://github.com/adobe/react-spectrum/pull/4210/commits/ca029a9e517876169b23042faf416492a48fda53) for a bit of background. 6f60708 might be a bit excessive perhaps given how we'll be eventually converting to the new way of doing Collections from RAC, but it felt more correct to do the refactor instead of passing sections with rows to GridCollection. Still checking for any breaks, can always revert that commit if things go wrong since the commit before it resolves the weird `state.collection.getItem` (aka getting node via the collection's keymap)  vs `getChildNodes(node, state.collection)` (aka getting node via a node's `childNodes`) disparity

packages/@react-stately/grid/src/GridCollection.ts
```diff
@@ -23,92 +23,106 @@ export class GridCollection<T> implements IGridCollection<T> {
...
-      if (last) {
-        last.nextKey = null;
-      }
+      this.keyMap.set(node.key, node);
 
-      // Remove deleted nodes and their children from the key map
-      if (prevNode) {
-        for (let child of prevNode.childNodes) {
-          if (!childKeys.has(child.key)) {
-            remove(child);
```
LFDanLu, line 71: Not sure when this old code comes into play, ran all the tests on main and never seems to fire. None of the table stories call it either

LFDanLu:
first chromatic build: https://www.chromatic.com/build?appId=5f0dd5ad2b5fc10022a2e320&number=533

packages/@react-stately/grid/src/GridCollection.ts
```diff
@@ -29,86 +29,94 @@ export class GridCollection<T> implements IGridCollection<T> {
...
 
-    let remove = (node: GridNode<T>) => {
-      this.keyMap.delete(node.key);
-      for (let child of node.childNodes) {
-        if (this.keyMap.get(child.key) === child) {
-          remove(child);
-        }
+      if (node.type === 'item' || node.type === 'headerrow' || node.type === 'header') {
+        this.rows.push(node);
+        rowIndex++;
```
LFDanLu, line 110: Recommenting:
From testing, it feels like we either keep aria-rowindex as an absolute position index (aka using this rowIndex, may not be feasible in RAC land) or use aria-posinset + aria-level + aria-setsize with relative position indicies (aka do away with rowIndex). Making aria-rowindex relative to the immediate parent makes announcements in VO confusing since it will announce values relative to the total number of rows rather than the section.

packages/@react-aria/dnd/src/useDropIndicator.ts
```diff
@@ -23,7 +23,9 @@ import {useLocalizedStringFormatter} from '@react-aria/i18n';
 
 export interface DropIndicatorProps {
   /** The drop target that the drop indicator represents. */
-  target: DropTarget
+  target: DropTarget,
+  /** A delegate object that implements behavior for keyboard focus movement. */
+  keyboardDelegate?: KeyboardDelegate<unknown>
```
LFDanLu, line 28: Update to useDropIndicator api, see discussion [here](https://github.com/adobe/react-spectrum/pull/4008#discussion_r1137904263) for background.

packages/@react-aria/dnd/src/useDropIndicator.ts
```diff
@@ -63,16 +65,30 @@ export function useDropIndicator(props: DropIndicatorProps, state: DroppableColl
   } else {
     let before: Key | null;
     let after: Key | null;
-    if (collection.getFirstKey() === target.key && target.dropPosition === 'before') {
-      before = null;
+
+    let keyGetter = keyboardDelegate ?? collection;
+    // TODO: I've changed it so it will announce "insert after"/"insert before" target if the drop position is between the target and a following/preceeding section
+    // instead of announcing "insert between target and NEXT/PREV_SECTION_ROW". Gather opinions
+    let keyBefore = keyboardDelegate != null ? keyboardDelegate.getKeyAbove(target.key) : collection.getKeyBefore(target.key);
```
LFDanLu, line 72: open to opinions here

packages/@react-aria/grid/src/useGridSectionAnnouncement.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+    parentNode = collection.getItem(parentNode?.parentKey ?? null);
+  }
+  let lastSection = useRef(sectionKey);
+  // VoiceOver on MacOS doesn't announce TableView/ListView sections when navigating with arrow keys so we do this ourselves
+  // TODO: NVDA announces the section title when navigating into it with arrow keys as "SECTION TITLE grouping" by defualt, so removing isMac() doubles up on the section title announcement
+  // a bit. However, this does add an announcement for the number of rows in a section which might be
+  // Mobile screen readers don't cause this announcement to fire until focus happens on a row via double tap which is pretty strange
+  useUpdateEffect(() => {
+    if (isMac() && focusedItem != null && selectionManager.isFocused && sectionKey !== lastSection.current) {
```
LFDanLu, line 51: Open to opinions here if people feel like we wanna make this custom announcement in other non-Mac screenreaders

packages/@react-aria/table/src/useTableSection.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  if (isVirtualized) {
+    rowIndex = (getFirstItem(getChildNodes(node, state.collection)) as GridNode<T>).rowIndex;
+  }
+
+  return {
+    rowGroupProps: {
+      // TODO: Chrome VO states "empty row group", doesn't happen in Safari VO. Doesn't happen in NVDA or TalkBack
+      role: 'rowgroup',
+      'aria-labelledby': headerId
```
LFDanLu, line 52: As stated here, Chrome VO doesn't have the best experience when navigating with arrows keys. Using control + option + arrow keys doesn't run into this problem and may be the preferred way that VO users will navigate through the table in general though. Will require some more exploration with different aria configurations but IMO this is ok for a first iteration (e.g. trying aria-posinset, aria-level, and aria-setsize instead of aria-rowindex resulted in a subpar experience)

packages/@react-spectrum/table/src/index.ts
```diff
@@ -13,8 +13,13 @@
...
 
 export {TableView} from './TableView';
-import {Column} from '@react-stately/table';
+import {Column, TableSection} from '@react-stately/table';
 import {SpectrumColumnProps} from '@react-types/table';
+import {SpectrumTableSectionProps} from './TableView';
+
+// Override TS for TableSection to require title prop.
+const SpectrumTableSection = TableSection as <T>(props: SpectrumTableSectionProps<T>) => JSX.Element;
+export {SpectrumTableSection as TableSection};
```
LFDanLu, line 22: Note we are exporting a TableSection and not reusing the Section collection. This feels like it makes more sense since all of Table's collection components are table specific anyways

packages/@react-stately/layout/src/ListLayout.ts
```diff
@@ -431,41 +431,41 @@ export class ListLayout<T> extends Layout<Node<T>> implements KeyboardDelegate,
     return this.contentSize;
   }
 
-  getKeyAbove(key: Key): Key | null {
+  getKeyAbove(key: Key, itemFilter: (item: Node<T>) => boolean = item => item.type === 'item'): Key | null {
```
LFDanLu, line 434: Note the API change here to the getKey methods for ListLayout. This is so we can include sections as valid return keys from these methods ONLY during drag and drop. I'm a bit torn if modifying ListLayout is the best place to do this, but it does allow ListView's DnD to use the same code this way since they both provide a layout to useDroppableCollection.

Alternatives:
- create a Table DnD specific keyboard delegate 
  - Where would this live? Is it too specific? Opted not to do this for these reasons
- Modifying the Table/GridKeyboardDelegate `getKey` getters instead
  - Would need the delegate to be returned by useTable, changes felt odd. Code for this can be found before this [commit](https://github.com/adobe/react-spectrum/pull/4210/commits/fcf0c6e1dd908647ea83c17701835bacbc7615fb) aka in https://github.com/adobe/react-spectrum/pull/4210/commits/96f4921019619f27c56af427f65ee473670f0902

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -431,79 +491,125 @@ export class TableLayout<T> extends ListLayout<T> {
...
             this.addVisibleLayoutInfos(res, node.children[idx], rect);
           }
-          persistIndex++;
-        }
-
-        for (let i = firstVisibleRow; i <= lastVisibleRow; i++) {
-          // Skip persisted rows that overlap with visible cells.
-          while (persistedRowIndices && persistIndex < persistedRowIndices.length && persistedRowIndices[persistIndex] < i) {
+        } else {
+          // Note:
```
LFDanLu, line 517: A variety of notes here to remind myself of the format of various pieces of data, happy to remove if people feel like it isn't helpful

packages/@react-stately/selection/src/SelectionManager.ts
```diff
@@ -481,7 +481,10 @@ export class SelectionManager implements MultipleSelectionManager {
     }
 
     let item = this.collection.getItem(key);
-    if (!item || (item.type === 'cell' && !this.allowsCellSelection)) {
+    // TODO: discuss how best to indicate that a row/item is not selectable without making it disabled
+    // Some general item prop (isInert, notSelectable)? isSectionHeader might be too specific.
+    // Maybe not an item prop? Maybe I shouldn't be making the section header an actual row/item in the collection
+    if (!item || (item.type === 'cell' && !this.allowsCellSelection) || item.props?.isSectionHeader) {
```
LFDanLu, line 487: Open question, maybe it would be useful to have a generic prop that marks a node as non-selectable? Perhaps a non-factor when moving to RAC collections

packages/@react-types/shared/src/collections.d.ts
```diff
@@ -92,12 +92,13 @@ export interface SortDescriptor {
 
 export type SortDirection = 'ascending' | 'descending';
 
-export interface KeyboardDelegate {
+export interface KeyboardDelegate<T> {
+  // TODO: what do we think about this update to the keyboard delegate api?
```
LFDanLu, line 96: As mentioned previously, changes in this PR include a update to the options provided to KeyboardDelegate's `getKey` getters

packages/@react-stately/table/src/TableBody.ts
LFDanLu: I don't see much of a difference in performance in the table with many sections story, but this will be changing when we move over to RAC collections so IMO this is fine for now

packages/@react-aria/dnd/src/useDropIndicator.ts
LFDanLu: Update so it announces the section name as well

packages/@react-types/shared/src/collections.d.ts
LFDanLu: Try doing `T extends object` instead to make this a non-breaking change for someone who made their own keyboard delegate

packages/@react-spectrum/table/src/index.ts
```diff
@@ -23,7 +28,6 @@ export {SpectrumColumn as Column};
 export {
   TableHeader,
   TableBody,
-  Section,
```
LFDanLu, line 26: Keep this in case someone was using it

packages/@react-stately/table/src/TableBody.ts
LFDanLu: todo: add story with sections w/ 100+rows in each

PR #4212, opened 2023-03-15
Convert React Aria Components examples to TypeScript

devongovett:
This converts the examples in the React Aria Components docs to TypeScript, and fixes some type issues in the implementation.

* forwardRef on components with generics. I added a helper type to the utils that we can cast the `forwardRef` function to which supports generics unlike the default type.
* Added missing exports to react-aria and react-stately monopackages to reduce use of individual types packages.
* Made slider types generic. Unfortunately this breaks inference until we can turn on `strictFunctionTypes`. For now I had to add additional annotations.


snowystinger (approved):
Docs look good for modified examples

LFDanLu (approved):
LGTM

PR #4220, opened 2023-03-17
Fix overlay position clamping and arrow position

LFDanLu:
Relates to https://github.com/adobe/react-spectrum/pull/1297 and https://github.com/adobe/react-spectrum/pull/3756
Most of the work here is courtesy of @filipw01, split out from https://github.com/adobe/react-spectrum/pull/4097 with updates in desired behavior discovered after our latest testing session (3/16/2023).

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Open newly created story in UseOverlayPosition story and play with different positions and offsets. More details about the issue that was fixed here https://github.com/adobe/react-spectrum/pull/1297

Open any TooltipTrigger story and play with offset controls to see how the and clamping arrow behaves. Additionally, use the DialogTrigger: popover story and try a combination of different button widths/height + cross offsets + positions. The arrow should prefer being positioned in the center of the button's edge. If placing the arrow in the center of the button would cause the arrow to no longer be contained in the overlay, the arrow should move the minimum amount necessary to still be within the overlay and the trigger

## 🧢 Your Project:

RSP + https://www.chilipiper.com/


packages/@react-spectrum/tooltip/stories/TooltipTrigger.stories.tsx
```diff
@@ -263,18 +263,6 @@ export const CrossoffsetExamples: TooltipTriggerStory = {
           <ActionButton>Tooltip Trigger -10</ActionButton>
           <Tooltip>Tooltip message.</Tooltip>
         </TooltipTrigger>
-        <TooltipTrigger delay={0} placement="left top" crossOffset={10}>
```
LFDanLu, line 266: Extraneous, mirrors the three tooltip triggers above them exactly except for the contents of the action button

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -331,12 +334,32 @@ export function calculatePositionInternal(
...
+  const arrowMaxPosition = overlaySize[crossSize] - (arrowCrossSize / 2);
+
+  if (preferredArrowPosition < arrowMinPosition || preferredArrowPosition > arrowMaxPosition) {
+    // Prefer the arrow being in the center of the overlay if placing it in the center of the button would mean it doesn't also overlap the overlay
+    preferredArrowPosition = overlaySize[crossSize] / 2;
+  }
+
+  // Min/Max position limits for the arrow with respect to the trigger/overlay anchor element
+  const arrowOverlappingChildMinEdge = childOffset[crossAxis] - overlaySize[crossAxis] + (arrowCrossSize / 2);
+  const arrowOverlappingChildMaxEdge = childOffset[crossAxis] + childOffset[crossSize] - overlaySize[crossAxis] - (arrowCrossSize / 2);
```
LFDanLu, line 358: Open question for the team about what boundaries we want for the arrow tip positioning vs the trigger element's edges. At the moment the `(arrowCrossSize / 2)` makes it so the entire arrow must fit in the trigger, with some additional fudge factor applied in Popover/TooltipTrigger so it doesn't quite hit the edges of the overlay/trigger.

If we are ok with something like the following
image (5)
I can get rid of the `(arrowCrossSize / 2)` here

packages/@react-spectrum/tooltip/src/TooltipTrigger.tsx
```diff
@@ -45,20 +45,20 @@ function TooltipTrigger(props: SpectrumTooltipTriggerProps) {
...
   useLayoutEffect(() => {
-    if (overlayRef.current) {
+    if (overlayRef.current && state.isOpen) {
       let spectrumBorderRadius = window.getComputedStyle(overlayRef.current).borderRadius;
       if (spectrumBorderRadius !== '') {
         setBorderRadius(parseInt(spectrumBorderRadius, 10));
       }
     }
-  }, [overlayRef]);
+  }, [state.isOpen]);
```
snowystinger, line 54: this should still have overlayRef, along with setBorderRadius

packages/@react-spectrum/tooltip/src/TooltipTrigger.tsx
```diff
@@ -45,20 +45,20 @@ function TooltipTrigger(props: SpectrumTooltipTriggerProps) {
...
   let arrowRef = useRef(null);
   let [arrowWidth, setArrowWidth] = useState(0);
   useLayoutEffect(() => {
-    if (arrowRef.current) {
+    if (arrowRef.current && state.isOpen) {
       setArrowWidth(arrowRef.current.getBoundingClientRect().width);
     }
-  }, [arrowRef]);
+  }, [state.isOpen]);
```
snowystinger, line 61: similar comment here

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -340,16 +340,24 @@ export function calculatePositionInternal(
 
   let arrowPosition: Position = {};
 
-  const preferredArrowPosition = overlaySize[crossSize] / 2;
-
-  const arrowOverlappingChildMinEdge = childOffset[crossAxis] - overlaySize[crossAxis] + (arrowCrossSize / 2);
-  const arrowOverlappingChildMaxEdge = childOffset[crossAxis] + childOffset[crossSize] - overlaySize[crossAxis] - (arrowCrossSize / 2);
+  // All coords are transformed so that (0, 0) is at the top left of the overlay
+  // Prefer the arrow being in the center of the trigger/overlay anchor element
+  let preferredArrowPosition = childOffset[crossAxis] + .5 * childOffset[crossSize] - overlaySize[crossAxis];
```
snowystinger, line 345: why `- overlaySize[crossAxis]`? seems like the preferred ArrowPosition has nothing to do with the overlay now

snowystinger: Still need arrowOverlappingChild... variables i think
edit: looks like you moved them down below, but lets keep them here

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -340,16 +340,24 @@ export function calculatePositionInternal(
...
 
+  // Min/Max position limits for the arrow with respect to the overlay
   const arrowMinPosition = arrowCrossSize / 2;
   const arrowMaxPosition = overlaySize[crossSize] - (arrowCrossSize / 2);
 
-  const arrowPositionOverlappingChild = clamp(preferredArrowPosition, arrowOverlappingChildMinEdge, arrowOverlappingChildMaxEdge);
+  if (preferredArrowPosition < arrowMinPosition || preferredArrowPosition > arrowMaxPosition) {
```
snowystinger, line 351: this should get taken care of in the clamp?

packages/@react-aria/overlays/src/calculatePosition.ts
snowystinger: I don't quite follow, maybe we can spend a little time tomorrow looking at this after standup if my suggestions don't address this?

packages/@react-spectrum/tooltip/src/TooltipTrigger.tsx
LFDanLu: I can add overlayRef, but isn't setBorderRadius guaranteed to be stable?

packages/@react-aria/overlays/src/calculatePosition.ts
LFDanLu: So the `- overlaySize[crossAxis]` is used to transform the coordinates in such a way that they are w/ respect to the overlay rather than being with respect to the document (i.e. 0 now refers to the top/left of the overlay instead of the top/left of the document/boundary element). This is done since the arrow is `position: absolute` inside the overlay div itself, hence we need a value that uses the top/left of the overlay as the origin.

packages/@react-aria/overlays/src/calculatePosition.ts
LFDanLu: ah, I actually can just skip the clamps entirely if the center of the button is within the arrowMinPosition/maxPosition. The clamps are more for attempting to position the arrow in the center of the overlay or as close as possible if it exceeds the bounds. I'll refactor a bit

packages/@react-spectrum/tooltip/src/TooltipTrigger.tsx
snowystinger: it is, i forget if the eslint thing can figure that out?

LFDanLu:
For reviewers, feel free to leave any opinions on the difference in arrow positioning experience between [this](https://reactspectrum.blob.core.windows.net/reactspectrum/3e73b3ba5d3446e2e2d7732b634dbc42d2a67c0f/storybook/index.html?path=/story/dialogtrigger--type-popover&args=crossOffset:230;buttonWidth:200&providerSwitcher-express=false) and [this](https://reactspectrum.blob.core.windows.net/reactspectrum/89cc2eb4c276bfb9930e27b96736b20a7de9a4ed/storybook/index.html?path=/story/dialogtrigger--type-popover&args=crossOffset:230;buttonWidth:200&providerSwitcher-express=false). The difference between the two is where the arrow is placed if the center of the button is not a viable position, with the first moving as close to the center of the overlay as possible and the second moving towards as close to the center of the button as possible.

packages/@react-aria/overlays/src/useOverlayPosition.ts
```diff
@@ -48,7 +53,11 @@ export interface AriaPositionProps extends PositionProps {
    * The maxHeight specified for the overlay element.
    * By default, it will take all space up to the current viewport height.
    */
-  maxHeight?: number
+  maxHeight?: number,
+  /**
+   * The minimum distance the arrow's edge should be from the edge of the overlay element.
+   */
+  minOverlayArrowOffset?: number
```
snowystinger, line 60: should have a default 0?

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -64,6 +73,8 @@ function Popover(props: PopoverProps, ref: DOMRef<HTMLDivElement>) {
   );
 }
 
+let ROOT_2 = Math.sqrt(2);
```
snowystinger, line 76: btw, just learned this exists `Math.SQRT2` so we don't need to calculate it at all

snowystinger (commented):
looks good, 2 nits

ktabors:
> second moving towards as close to the center of the button as possible.

I'm in favor of the second one that keeps the arrow as close to the center of the button as possible.

ktabors:
Testing the default popover story with the scrolling turning on and it should flip and doesn't. 
image

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -115,54 +133,67 @@ const PopoverWrapper = forwardRef((props: PopoverWrapperProps, ref: RefObject<HT
...
-
-  let borderDiagonal = borderWidth * ROOT_2;
-  let halfBorderDiagonal = borderDiagonal / 2;
+  }, []);
+  return {size, borderWidth, arrowRef};
+}
 
-  let secondary = 2 * size + 2 * borderDiagonal;
-  let primary = size + borderDiagonal;
+function Arrow(props: ArrowProps) {
```
ktabors, line 186: Why wasn't `props` deconstructed? I see `primary`, `secondary` and `isLandscape` each used a few times.

LFDanLu: just old code from main, will update

LFDanLu:
TODO: rerun chromatic when the [specificity pr](https://github.com/adobe/react-spectrum/pull/4241) goes in

devongovett (approved):

Issue #4163, opened 2023-03-06
Doc Issue: useFocusRing differs between Listbox Option in useComboBox and useListbox docs


ashleyryan:
# 🙋 Documentation Request

I first built a listbox component using useListBox and now I'm building a combobox on top of that following the useComboBox documentation. For the longest time I couldn't figure out why no options were showing as focused using keyboard navigation. As it turns out, the `useListBox` documentation shows the `Option` component using `useFocusRing` to determine `isFocusVisible`, which has different behavior than getting `isFocused` from `useOption` which is used in the useComboBox documentation.

- [ListBox documentation](https://react-spectrum.adobe.com/react-aria/useListBox.html#example)
- [ComboBox documentation - listbox code sample](https://react-spectrum.adobe.com/react-aria/useComboBox.html#listbox)

`isFocusVisible` doesn't appear to work with combobox, I'm assuming because focus remains on the input. Is it safe to not use `useFocusRing()` with a listbox?

This might be a bug, not a documentation issue, but I'd love to see the code examples on the two pages match.

LFDanLu:
You're hunch about combobox and `useFocusRing`'s `isFocusVisible` is correct. A listbox option's visual focus state is a bit more complicated than portrayed in the `useListBox` docs and the `useComboBox` docs for that reason, really it should be a combination of the two implementations if the same ListBox is going to be used for both combobox and non-combobox components. In our Spectrum implementation of [ListBoxOption](https://github.com/adobe/react-spectrum/blob/05be974f59eece54dd2f47f1866e1c30031c0999/packages/%40react-spectrum/listbox/src/ListBoxOption.tsx#L72-L84) we use `useOption`'s `isFocused` and a keyboard modality check to display the focused state for the combobox's "virtual" focus case and otherwise rely on `FocusRing` to apply the visual focus style in the standard ListBox use case.

For the sake of simplicity, we didn't go into that level of implementation detail for the aria docs but we can try to refine the docs to make that clearer. [Aria Components](https://github.com/adobe/react-spectrum/pull/4058) will also help alleviate this issue as well for users who just want to use an stylable aria ListBox without needing to implement these hard to find details themselves

ashleyryan:
Ok, that all makes sense - thanks for the link to the Spectrum implementation. I've been wading into all of the source code for react-stately and react-aria, but not the actual spectrum components. 

[...]

PR #4221, opened 2023-03-17
Fixes for Aria Components

devongovett:
Depends on #4212.

* Fixes animation issue with Select popover
* Switches to using `--trigger-width` as a CSS custom property for both ComboBox and Select popovers for consistency
* Fixes animation of side modal example in Chrome, which previously got stuck open. This was due to an ordering issue where child animations completed before parents.
* Fix all docs examples with select all from displaying `a,l,l`
* Fix question mark in circle character on Windows - switched to i in circle. No font seems to support the question mark.
* Fixed some high contrast styling
* Fixes isFocused in useOption to only be true when the collection is actually focused.
* Added an isFocusVisible return value to useOption which works with virtual focus, and allows us to get rid of another dependency in aria components. Closes https://github.com/adobe/react-spectrum/issues/4163.

### DIFF
--- a/packages/@react-aria/listbox/docs/useListBox.mdx
+++ b/packages/@react-aria/listbox/docs/useListBox.mdx
@@ -251,7 +251,7 @@ function Example() {
         <Item key="egg">Egg Salad</Item>
         <Item key="ham">Ham</Item>
       </ListBox>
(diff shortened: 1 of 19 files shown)


snowystinger (approved):

reidbarber (dismissed):
LGTM

PR #4248, opened 2023-03-23
React Aria Components fixes from testing

devongovett:
Fixes bugs we found during testing. See individual commit messages for details.

Also moved useHasTabbableChild from https://github.com/adobe/react-spectrum/pull/3581 out of Virtualizer and into the hooks so it also applies to aria components.


packages/@react-stately/selection/src/SelectionManager.ts
```diff
@@ -173,7 +173,7 @@ export class SelectionManager implements MultipleSelectionManager {
     let first: Node<unknown> | null = null;
     for (let key of this.state.selectedKeys) {
       let item = this.collection.getItem(key);
-      if (!first || item?.index < first.index) {
+      if (!first || (item && compareNodeOrder(this.collection, item, first) < 0)) {
```
devongovett, line 176: This accounts for the new collection implementation using indexes of items in sections relative to the parent node rather than the whole collection.

snowystinger (commented):
One interesting thing, in chrome, the drag preview for ListBox flies in from the top of the page, way way far away

snowystinger (approved):

packages/@react-stately/dnd/src/useDraggableCollectionState.ts
```diff
@@ -98,11 +98,11 @@ export function useDraggableCollectionState(props: DraggableCollectionStateOptio
     preview,
     getAllowedDropOperations,
     startDrag(key, event) {
-      selectionManager.setFocused(false);
-      setDragging(true);
       let keys = getKeys(key);
       draggingKeys.current = keys;
       draggedKey.current = key;
+      selectionManager.setFocused(false);
+      setDragging(true);
```
LFDanLu, line 105: I'm surprised that this was an issue, what was broken that relied on the dragging/draggedKeys refs being updated before these these lines were set?

reidbarber (approved):
LGTM

LFDanLu (approved):
Looks good to me for the most part, will test the WHCM stuff on Monday. Just two questions/observations, nothing blocking

packages/@react-stately/collections/src/getChildNodes.ts
```diff
@@ -49,3 +50,36 @@ export function getLastItem<T>(iterable: Iterable<T>): T | undefined {
...
+
+  if (firstNonMatchingAncestor !== -1) {
+    // Compare the indices of two children within the common ancestor.
+    a = aAncestors[firstNonMatchingAncestor];
+    b = bAncestors[firstNonMatchingAncestor];
+    return a.index - b.index;
+  }
+
+  // 🤷
+  return -1;
```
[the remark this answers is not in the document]
devongovett: ah yeah true. I guess it would ideally compare the index of the first item in b's ancestors with the index of a itself in that case. But yeah we don't really support that case at the moment (and I'd like to disallow it actually).

packages/@react-stately/dnd/src/useDraggableCollectionState.ts
devongovett: Well really the other one further down was the problem with `isDragging` persisting after drop, because the refs were updated after state. I also changed this one just in case React decided to re-render synchronously for some reason and we accessed the dragging keys before they were set.

packages/@react-aria/tabs/src/useTabPanel.ts
```diff
@@ -29,35 +29,10 @@ export interface TabPanelAria {
...
-        childList: true,
-        attributes: true,
-        attributeFilter: ['tabIndex', 'disabled']
-      });
-
-      return () => {
-        observer.disconnect();
-      };
-    }
-  }, [ref]);
```
malykhinvi, line 60: @devongovett hi! I've noticed that after extraction of this layoutEffect to `useHasTabbableChild` it does not have a dependencies array anymore, which makes it run on every re-render. I am wondering if that was done on purpose or it was overlooked?

devongovett: I think the comment explains it pretty well. We want synchronous updates when React re-renders, but need to detect changes when child components re-render too. That's where the MutationObserver comes in. https://github.com/adobe/react-spectrum/blob/73414999fcc6afab5c4bc8ec4740fe3dc1eb8502/packages/%40react-aria/focus/src/useHasTabbableChild.ts#L56

PR #4253, opened 2023-03-23
Simplify aria table resizing docs example

devongovett:
Based on #3982.

This is a slightly simpler implementation of the column resizing example that doesn't require overriding the CSS table layout to use flex. This way we can simply reuse the TableRowGroup, TableHeaderRow, TableRow, and TableCell components from before instead of reimplementing them with minor styling changes.

It works by adding an additional div wrapper around the table element that scrolls. The table itself is set to `width: fit-content` and `table-layout: fixed` to ensure it always takes the widths defined on the column headers. Then the browser automatically handles setting the widths of the cells in the body based on the header widths, and the entire table grows when the widths get bigger (overflowing in the parent wrapper).

You do lose the sticky headers, but in this example there is no vertical scrolling anyway so I'd rather keep the example simpler to focus on the concepts related to resizing itself. The styled example goes into more detail there as well.

Edit: sticky headers can actually be achieved as well. If you add `position: sticky; top: 0` to the tr element in the header and a max-height on the wrapper div it does stick as expected. But anyway it wouldn't be shown in the example because it doesn't vertically scroll.

### DIFF
--- a/packages/@react-aria/table/docs/useTable.mdx
+++ b/packages/@react-aria/table/docs/useTable.mdx
@@ -840,35 +840,25 @@ omitting support for selection, sorting, and nested columns.
 
 As mentioned previously, we first need to call <TypeLink links={statelyDocs.links} type={statelyDocs.exports.useTableColumnResizeState} /> to initialize the widths for our table's columns.
 We'll pass the state returned by <TypeLink links={statelyDocs.links} type={statelyDocs.exports.useTableColumnResizeState} /> along with any user defined `onResize` handlers
-to our `ResizableTableColumnHeaders` so it can be used by <TypeLink links={docs.links} type={docs.exports.useTableColumnResize} />. The `widths` from this state are provided to each table element so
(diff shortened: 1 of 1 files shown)


packages/@react-aria/table/docs/useTable.mdx
```diff
@@ -883,62 +873,49 @@ function ResizableColumnsTable(props) {
 
   let layoutState = useTableColumnResizeState({
     // Matches the width of the table itself
-    tableWidth: 300,
+    tableWidth: 400,
```
LFDanLu, line 876: Can we keep this as 300 or at least something a little less than 400? Its a bit too wide on my phone
Screenshot_20230323_141212_Chrome

LFDanLu (commented):
Just one small nit, I'll push an update

LFDanLu (approved):

devongovett:
Oh hmm I was wondering why it was so small. Kinda annoying on large screens but oh well

PR #4254, opened 2023-03-23
Refine table column resize hook props and returned values 

LFDanLu:
Closes 
Split out from https://github.com/adobe/react-spectrum/pull/3982

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/@react-stately/table/src/useTableColumnResizeState.ts
```diff
@@ -105,8 +103,8 @@ export function useTableColumnResizeState<T>(props: TableColumnResizeStateProps<
     setResizingColumn(null);
   }, [setResizingColumn]);
 
-  let columnWidths = useMemo(() =>
-      columnLayout.buildColumnWidths(tableWidth, state.collection, colWidths)
+  useMemo(() =>
+    columnLayout.buildColumnWidths(tableWidth, state.collection, colWidths)
```
LFDanLu, line 107: I'm not entirely sure why this is required, but without it getColumnWidth returns 0. It also must be done in the render phase. Still digging

@snowystinger do you recall any context behind this?

LFDanLu: Talked with @snowystinger, looks like TableColumnLayout was structured in such a way so TableLayout could calculate controlled/uncontrolled column widths in its constructor. I'm going to leave this useMemo as is for this PR, looking to see if I can refactor TableColumnLayout in such a way that buildColumnWidths is updated by creating a new layout or by calling a layout method via resize

snowystinger (approved):

devongovett (approved):

PR #4272, opened 2023-03-27
useTagGroup: add description and update 'remove' translation string

### LINKED ISSUE #4278: TagGroup: add aria-describedby with instructions for Removing each row, simplify remove button label
# 🐛 Bug Report 

## 🤔 Expected Behavior

- “Press Delete to Remove” should be a separate instruction for the Tag row. We should add `aria-describedby` the the Tag row to include the IDREF for the hint text.
- The label on the button should just be “Remove” + tag name.

## 😯 Current Behavior

With screen readers, where one can navigate using either virtual cursor/browse mode, table navigation, or forms mode, the label, "Press delete to remove." on the remove button can be confusing and might lead a user to delete the wrong Tag accidentally.

When a user navigates the page with JAWS or NVDA and lands on the Tag row, unless the user has explicitly overwritten the navigation mode, the screen reader switches into forms mode, which is fine. However, in forms mode, the screen reader announces only the accessible name for the Tag, and fails to announce “Press Delete to remove.” So there is no indication that the focused Tag can be removed. 

If a user navigates with table navigation shortcuts, virtual cursor/browse mode, or the VoiceOver cursor, rather than ArrowKeys, it is possible to hear the “Press Delete to remove.” button, but the screen reader cursor might become separated from the keyboard focus. In other words, it’s possible to browse the remove button for one Tag, and here the instruction “Press Delete to remove,” while focus remains on another Tag. If the user follows the instruction and presses Delete, rather than pressing Space or Enter to activate the button, the Tag group collection will remove the Tag that has keyboard focus, the one matching the focusedKey, rather than the Tag containing the button that the user just read.

## 💁 Possible Solution

- We should add `aria-describedby` the the Tag row to include the IDREF for the hint text.
- The label on the button should just be “Remove” + tag name.

## 🔦 Context

Discussion pertaining to https://github.com/adobe/react-spectrum/pull/4272#pullrequestreview-1359948146

## 🌍 Your Environment

| react-spectrum   |
| Browser          |
| Operating System |

## 🧢 Your Company/Team

Adobe/Accessibility

reidbarber:
Closes #4278 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Check screen reader announcements for onRemove TagGroup story.

### DIFF
--- a/packages/@react-aria/tag/src/useTag.ts
+++ b/packages/@react-aria/tag/src/useTag.ts
@@ -11,7 +11,7 @@
  */
 
 import {AriaButtonProps} from '@react-types/button';
-import {chain, filterDOMProps, mergeProps, useId} from '@react-aria/utils';
+import {chain, filterDOMProps, mergeProps, useDescription, useId} from '@react-aria/utils';
 import {DOMAttributes, FocusableElement} from '@react-types/shared';
 // @ts-ignore
 import intlMessages from '../intl/*.json';
@@ -46,7 +46,6 @@ export function useTag<T>(props: TagProps<T>, state: TagGroupState<T>, ref: RefO
     item
   } = props;
   let stringFormatter = useLocalizedStringFormatter(intlMessages);
-  let removeString = stringFormatter.format('remove');
   let labelId = useId();
   let buttonId = useId();
 
@@ -66,11 +65,14 @@ export function useTag<T>(props: TagProps<T>, state: TagGroupState<T>, ref: RefO
     }
   };
 
+  let description = allowsRemoving ? stringFormatter.format('removeDescription') : '';
+  let descProps = useDescription(description);
+
   isFocused = isFocused || state.selectionManager.focusedKey === item.key;
   let domProps = filterDOMProps(props);
   return {
     clearButtonProps: {
-      'aria-label': removeString,
+      'aria-label': stringFormatter.format('removeButtonLabel', {label: item.textValue}),
       'aria-labelledby': `${buttonId} ${labelId}`,
       id: buttonId,
       onPress: () => allowsRemoving && onRemove ? onRemove(item.key) : null,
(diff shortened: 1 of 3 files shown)


snowystinger (dismissed):
Were we going to make this only apply to keyboard modality as well?

packages/@react-aria/tag/src/useTag.ts
```diff
@@ -66,11 +66,18 @@ export function useTag<T>(props: TagProps<T>, state: TagGroupState<T>, ref: RefO
...
+  }
+  let description = allowsRemoving && (modality === 'keyboard' || modality === 'virtual') ? stringFormatter.format('removeDescription') : '';
+  let descProps = useDescription(description);
+
   isFocused = isFocused || state.selectionManager.focusedKey === item.key;
   let domProps = filterDOMProps(props);
   return {
     clearButtonProps: {
-      'aria-label': removeString,
+      'aria-label': stringFormatter.format('removeButtonLabel', {label: item.textValue}),
```
devongovett, line 80: I don't think you should need the label in the `aria-label` because you already have an `aria-labelledby` below that points to the label element in addition to the button. Right now I'm hearing the label announced twice. So just the word "Remove" should be enough here?

packages/@react-aria/tag/src/useTag.ts
```diff
@@ -82,7 +89,8 @@ export function useTag<T>(props: TagProps<T>, state: TagGroupState<T>, ref: RefO
     rowProps: {
       ...rowProps,
       tabIndex: (isFocused || state.selectionManager.focusedKey == null) ? 0 : -1,
-      onKeyDown: allowsRemoving ? onKeyDown : null
+      onKeyDown: allowsRemoving ? onKeyDown : null,
+      'aria-describedby': descProps['aria-describedby']
```
devongovett, line 93: VoiceOver appears not to read this at all...

LFDanLu: Reid and I discussed this briefly yesterday, the row's `aria-label` seems to be preventing the `aria-describedby` from announcing which is bizarre. Given how close we are to release and the need for translations, do we wanna punt on this here until after release? The behavior on VoiceOver prior to change is essentially the same, but the changes make it better for NVDA for sure (need to retest TalkBack with the modality check)

LFDanLu (approved):
Tested in VO, Talkback, and NVDA, LGTM

snowystinger (approved):

PR #4282, opened 2023-03-29
Add React Aria Components + Tailwind CSS examples

devongovett:
Depends on #4275.

This adds some examples of using React Aria Components with Tailwind CSS. Not sure where we want these so they are discoverable long term but for now it's just in the examples folder.

Screen Shot 2023-03-22 at 11 17 17 AM

Screen Shot 2023-03-28 at 10 18 26 AM


reidbarber (approved):
Tested locally, LGTM.

We should consider adding this to the verdaccio build.

I did notice a few issues outside of the scope of this PR though: 
1. When showing the dialog, the ContactListExample section headers show above the blur background.
2. In ImageGridExample, keyboard left and right don’t work for navigation (I did see the TODO though).

LFDanLu (dismissed):
Tested locally, LGTM. So colorful!

snowystinger (approved):
so pretty!
combobox is afflicted with the same no visible focus while keyboard navigating that we have in our components right now

snowystinger (approved):

PR #4290, opened 2023-03-30
Generate table row textValue from row header cells by default

devongovett:
When a `textValue` is not explicitly provided to the `<Row>` component, this generates it based on the row header columns within that row. We already had this for our custom selection announcements, but not in drag and drop, so this moves the `getRowText` function that we had in `useTable` into `TableCollection` in order to allow it to be reused. A new `getTextValue` method is added to the `Collection` interface to allow collections to generate a text value for their nodes dynamically if they need to.

For now there are two implementations: one in the old TableCollection in stately, and the other in react-aria-components. Eventually we'll consolidate this when moving to the new collections in spectrum.


snowystinger (approved):

LFDanLu (approved):
LGTM, verified that the row header cells are being used as the text values by default in the DnD stories and RAC table examples.

PR #4297, opened 2023-03-30
Fix SSR issues with React Aria Components

### LINKED ISSUE #3293: `useIsSSR()` hook always returns `true` and then `false` after hydration on mount
# 🐛 Bug Report 

I may have misunderstood the [useIsSSR()](https://react-spectrum.adobe.com/react-aria/useIsSSR.html) documentation, but it isn't working the way I would expect it to.

## 🤔 Expected Behavior

The `useIsSSR()` hook should always return `false` after the app has hydrated on the client.

## 😯 Current Behavior

Currently, the hook always returns `true` (meaning the app is server side rendering, or hydrating) before flipping to `false` every time it's mounted, even after hydration. I would expect it to always return `false` after hydration instead of `true` and then `false`.

## 💁 Possible Solution

No possible solution, but [this line](https://github.com/adobe/react-spectrum/blob/373244834584954dd407853c12e2bc4d68d96162/packages/%40react-aria/ssr/src/SSRProvider.tsx#L93) seems suspect to me. The `isSSR` state is always initially `true`, even if the app has already hydrated.

## 🔦 Context

My use case involves using `useIsSSR()` to create a hook that returns the device's pixel ratio. On the server and during hydration this hook returns a default value. After hydration it should always return `window.devicePixelRatio`.

```js
import { useIsSSR } from '@react-aria/ssr';

function useDevicePixelRatio(defaultRatio = 1) {
  let isSSROrHydration = useIsSSR();

  return isSSROrHydration
    ? defaultRatio
    : window.devicePixelRatio;
}
```

## 💻 Code Sample

This example is a bit contrived since there isn't any server side rendering happening, but it still illustrates the potential issue.

https://codesandbox.io/s/tender-violet-kunchu?file=/src/App.js

## 🌍 Your Environment

| react-spectrum   | `@react-aria/ssr@3.2.0`
| Browser          | `Chrome`
| Operating System | `MacOS`

## 🧢 Your Company/Team

SmugMug

### LINKED ISSUE #3758: `Select` SSR `tabIndex` mismatch
### Discussed in https://github.com/adobe/react-spectrum/discussions/3755

<sup>Originally posted by **duckies** November 16, 2022</sup>
Hello, I have hit a wall trying to figure out why the `Select` Tailwind example is causing console warnings about `tabindex` under Nextjs's SSR. I have created a Codesandbox following the example as closely as possible.

## Error
```
Warning: Prop `tabIndex` did not match. Server: "-1" Client: "0"
    at input
    at div
    at $bdd25dc72710631f$export$cbd84cdb2e668835 (webpack-internal:///./node_modules/@react-aria/select/dist/module.js:202:18)
    at div
    at Select (webpack-internal:///./components/Select.tsx:27:78)
    at div
    at IndexPage
    at $704cf1d3b684cc5c$export$9f8ac96af4b1b2ae (webpack-internal:///./node_modules/@react-aria/ssr/dist/module.js:23:64)
    at Application (webpack-internal:///./pages/_app.tsx:11:11)
    at PathnameContextProviderAdapter (webpack-internal:///./node_modules/next/dist/shared/lib/router/adapters.js:62:11)
    at ErrorBoundary (webpack-internal:///./node_modules/next/dist/compiled/@next/react-dev-overlay/dist/client.js:2:3325)
    at ReactDevOverlay (webpack-internal:///./node_modules/next/dist/compiled/@next/react-dev-overlay/dist/client.js:2:6707)
    at Container (webpack-internal:///./node_modules/next/dist/client/index.js:60:1)
    at AppContainer (webpack-internal:///./node_modules/next/dist/client/index.js:173:11)
    at Root (webpack-internal:///./node_modules/next/dist/client/index.js:346:11)
```

## [Codesandbox Reproduction](https://codesandbox.io/s/react-aria-next-js-ssr-select-prop-mismatch-dtnyc5?file=/components/Select.tsx)

- I was following the [Tailwind CSS](https://codesandbox.io/s/hardcore-moon-xzc4r?file=/src/Select.tsx) example linked from [useSelect — React Aria](https://react-spectrum.adobe.com/react-aria/useSelect.html#styled-examples).
- The issue is the same regardless of using the nightly builds found in the example (`react-aria@3.20.1-nightly.3567` and `react-stately@3.18.1-nightly.3567`) or latest stable versions (`react-aria@3.21.0` and `react-stately@3.19.0`), so I left the reproduction at the latter.
- The error is not consistent on every reload. Trying to interact with the `Select` as quickly as possible after reloading more reliably causes the warning.
</div>

### LINKED ISSUE #3787: SSRProvider causes an error when conditionally rendering an Overlay (Next.js)
# 🐛 Bug Report 

When wrapping the whole application with `SSRProvider` and then conditionally rendering an `Overlay` or `OverlayContainer` component with the `useOverlayTriggerState` state hook, an Unhandled Runtime Error is thrown. 

```
Unhandled Runtime Error
TypeError: Cannot read properties of null (reading 'contains')
```

It doesn't happen when the `SSRProvider` is removed in `_app.tsx`. 

I've tried all sorts of different things like reimplementing a Modal in different ways with react-aria. I've followed the official documentation etc. It all seems to be tied to the `SSRProvider` component somehow. I don't understand the code well enough to find a solution myself. 

## 🤔 Expected Behavior

I expect the `SSRProvider` not to cause unhandled runtime errors when conditionally rendering react-aria components. 

## 😯 Current Behavior

Conditionally rendering the `Overlay` component from `react-aria` causes an Unhandled Runtime Error.

```
Unhandled Runtime Error
TypeError: Cannot read properties of null (reading 'contains')

Call Stack
eval
node_modules/@react-aria/overlays/dist/module.js (1102:0)
Array.some
<anonymous>
MutationObserver.eval
node_modules/@react-aria/overlays/dist/module.js (1102:0)
```

## 💻 Code Sample

The code sample is a standard Nextjs setup. 
Relevant files are `components/modal.tsx` and `pages/index.tsx`. 
https://codesandbox.io/p/sandbox/react-aria-conditional-overlay-ssrprovider-7rvohm?file=%2Fpages%2Findex.tsx

## 🌍 Your Environment

| react-aria   | 3.21.0
| react-stately   |  3.19.0
| Browser  | Chrome
| Operating System | Macbook Pro, Ventura 13.0.1

devongovett:
This fixes some issues I found while testing React Aria Components with Next.js:

* Adds a Node ESM build to the package like our others
* Skips rendering a portal in collection components on the server and during hydration because portals aren't supported by React in SSR
* Only returns true from `useIsSSR` during initial hydration. Previously, it would return true when any component first mounted within an SSRProvider, even after hydration. This was breaking overlay positioning, and caused the error in combobox with ariaHideOutside where the popoverRef wasn't defined yet due to an SSR check on mount. Fixes #3293. Fixes #3787.
* Makes `useInteractionModality` check `useIsSSR` and return `null` if it is on the server or hydrating. This was causing a hydration warning to sometimes appear in Picker due to un-matched tab index if you moved the mouse before hydration completed. Fixes #3758.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ publish-nightly: build
 build:
 	parcel build packages/@react-{spectrum,aria,stately}/*/ packages/@internationalized/{message,string,date,number}/ packages/react-aria-components --no-optimize
 	yarn lerna run prepublishOnly
-	for pkg in packages/@react-{spectrum,aria,stately}/*/  packages/@internationalized/{message,string,date,number}/ packages/@adobe/react-spectrum/ packages/react-aria/ packages/react-stately/; \
+	for pkg in packages/@react-{spectrum,aria,stately}/*/  packages/@internationalized/{message,string,date,number}/ packages/@adobe/react-spectrum/ packages/react-aria/ packages/react-stately/ packages/react-aria-components/; \
 		do cp $$pkg/dist/module.js $$pkg/dist/import.mjs; \
 	done
 	sed -i.bak s/\.js/\.mjs/ packages/@react-aria/i18n/dist/import.mjs
--- a/packages/@react-aria/interactions/package.json
+++ b/packages/@react-aria/interactions/package.json
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "@react-aria/utils": "^3.15.0",
+    "@react-aria/ssr": "^3.5.0",
     "@react-types/shared": "^3.17.0",
     "@swc/helpers": "^0.4.14"
   },
--- a/packages/@react-aria/interactions/src/useFocusVisible.ts
+++ b/packages/@react-aria/interactions/src/useFocusVisible.ts
@@ -17,6 +17,7 @@
 
 import {isMac, isVirtualClick} from '@react-aria/utils';
 import {useEffect, useState} from 'react';
+import {useIsSSR} from '@react-aria/ssr';
 
 export type Modality = 'keyboard' | 'pointer' | 'virtual';
 type HandlerEvent = PointerEvent | MouseEvent | KeyboardEvent | FocusEvent;
@@ -192,7 +193,7 @@ export function useInteractionModality(): Modality {
     };
   }, []);
 
-  return modality;
+  return useIsSSR() ? null : modality;
 }
 
 /**
--- a/packages/@react-aria/ssr/src/SSRProvider.tsx
+++ b/packages/@react-aria/ssr/src/SSRProvider.tsx
@@ -23,7 +23,8 @@ import React, {ReactNode, useContext, useLayoutEffect, useMemo, useRef, useState
 // consistent ids regardless of the loading order.
 interface SSRContextValue {
   prefix: string,
-  current: number
+  current: number,
+  isSSR: boolean
 }
 
 // Default context value to use in case there is no SSRProvider. This is fine for
@@ -33,7 +34,8 @@ interface SSRContextValue {
 // SSR case multiple copies of React Aria is not supported.
 const defaultContext: SSRContextValue = {
   prefix: String(Math.round(Math.random() * 10000000000)),
-  current: 0
+  current: 0,
+  isSSR: false
 };
 
 const SSRContext = React.createContext<SSRContextValue>(defaultContext);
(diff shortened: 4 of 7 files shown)


snowystinger (approved):

reidbarber (approved):
LGTM. Nice to see those existing issues handled too!

PR #4306, opened 2023-03-31
Add overview page for React Aria Components

devongovett:
This adds an overview page for React Aria Components, which describes what the library is, its status, and general info about how to install and use it. Eventually most of this will move to the general React Aria getting started page and other concepts pages, but while we are in alpha, we'll keep it on a separate page.

Also re-exported `SSRProvider` from `react-aria-components` since that will be needed for use in SSR environments.

### DIFF
--- a/packages/react-aria-components/docs/react-aria-components.mdx
+++ b/packages/react-aria-components/docs/react-aria-components.mdx
@@ -1,0 +1,681 @@
+{/* Copyright 2023 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License. */}
+
+import {Layout} from '@react-spectrum/docs';
+export default Layout;
+
+import ChevronRight from '@spectrum-icons/workflow/ChevronRight';
+import styles from '@react-spectrum/docs/src/docs.css';
+
+import {ExampleCard} from '@react-spectrum/docs/src/ExampleCard';
+import ComboBoxAnatomy from '@react-spectrum/docs/pages/assets/component-illustrations/ComboBox.svg';
+import SelectAnatomy from '@react-spectrum/docs/pages/assets/component-illustrations/Picker.svg';
+import Button from '@react-spectrum/docs/pages/assets/component-illustrations/Button.svg';
+import ToggleButton from '@react-spectrum/docs/pages/assets/component-illustrations/ToggleButton.svg';
+import Checkbox from '@react-spectrum/docs/pages/assets/component-illustrations/Checkbox.svg';
+import CheckboxGroup from '@react-spectrum/docs/pages/assets/component-illustrations/CheckboxGroup.svg';
+import RadioGroup from '@react-spectrum/docs/pages/assets/component-illustrations/RadioGroup.svg';
+import Switch from '@react-spectrum/docs/pages/assets/component-illustrations/Switch.svg';
+import TextField from '@react-spectrum/docs/pages/assets/component-illustrations/TextField.svg';
+import NumberField from '@react-spectrum/docs/pages/assets/component-illustrations/NumberField.svg';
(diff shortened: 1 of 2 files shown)


reidbarber (approved):
LGTM

snowystinger (approved):

packages/react-aria-components/docs/react-aria-components.mdx
```diff
@@ -0,0 +1,681 @@
...
+<ExampleCard
+  url="RangeCalendar.html"
+  title="RangeCalendar"
+  description="A range calendar allows a user to select a contiguous range of dates.">
+  <RangeCalendar />
+</ExampleCard>
+
+</section>
+
+### Overlays
```
LFDanLu, line 547: Missing Modal here

LFDanLu (commented):
Looks good to me, just noticed one thing

LFDanLu:
Eventually would be nice to have an example of dropping down to hooks and mixing and matching it with RAC but can be done in the future when we have a solid example.

packages/react-aria-components/docs/react-aria-components.mdx
devongovett: Yeah same on the react-aria homepage. I think the problem was the picture would look the same as the dialog one maybe...

packages/react-aria-components/docs/react-aria-components.mdx
```diff
@@ -0,0 +1,681 @@
...
+    --text-color-disabled: GrayText;
+  }
+}
+```
+
+</details>
+
+If you're using SSR via a framework like Next.js or Remix, you'll also need to wrap your application in an `<SSRProvider>`. See the [Server Side Rendering](ssr.html) docs for more details.
+
+## Styling
```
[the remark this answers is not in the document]
devongovett: hmm yeah I guess the subheadings ("States", "Render props", and "Animation") aren't really different options that do the same things though, they are about different aspects of styling. 🤷

packages/react-aria-components/docs/react-aria-components.mdx
```diff
@@ -0,0 +1,681 @@
...
+
+React Aria Components is currently in **alpha**. This means APIs will likely change in future updates as we discover the best ways to use it, and there are some known bugs and limitations. That said, it is based on a solid and battle-tested foundation in React Aria, and we would love for you to try it out and give us feedback! This will directly help us shape the APIs and make it the best library it can be. Please report issues and feature requests [on GitHub](https://github.com/adobe/react-spectrum/issues).
+
+## Installation
+
+React Aria Components can be installed using a package manager like [npm](https://docs.npmjs.com/cli/npm) or
+[yarn](https://classic.yarnpkg.com/lang/en/).
+
+```
+yarn add react-aria-components
```
ktabors, line 77: Will this be version 1 alpha? Or version 3? Do we need to warn them that the 0.4.0 version exists and isn't this project?

packages/react-aria-components/docs/react-aria-components.mdx
```diff
@@ -0,0 +1,681 @@
...
+```
+
+All components are available in this one package for ease of dependency management.
+
+## Implementing a component
+
+Once installed, you can import and render the components you need. Each component may include several parts, as documented on the corresponding component page. The API is designed around composition, where each component generally has a 1:1 relationship with a single DOM element. This makes it easy to style every element, and control the layout and DOM order as needed to implement your design.
+
+This example renders a custom [Select](Select.html).
```
ktabors, line 86: Do we ever have to worry about URLs collision between info pages and component pages? Should react-aria components be `react-aria/components/NAME.html`? The hooks will always have `use` in front so they should be safe. For example, we have VisuallyHidden as a page for useVisuallyHidden, so we can't have a component named that (not that I expect us to create this).

packages/react-aria-components/docs/react-aria-components.mdx
```diff
@@ -0,0 +1,681 @@
...
+
+Compared with the React Aria hooks, the components provide a default DOM structure and styling API, and abstract away the glue code necessary to connect the hooks together. The components and hooks also work together, allowing them to be mixed and matched depending on the level of customization required. Eventually, React Aria Components will be recommended as a starting point when building a new component, dropping down to hooks only when additional flexibility is needed. Both hooks and components will continue to be maintained and developed going forward.
```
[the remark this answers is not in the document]
devongovett: Yeah when it is out of alpha and we feel confident recommending it

packages/react-aria-components/docs/react-aria-components.mdx
devongovett: Yeah 1.0.0-alpha.0 should get automatically installed as the latest version.

PR #4326, opened 2023-04-03
Fix combobox section filtering and Firefox gap issue

devongovett:
From testing.


snowystinger (approved):

LFDanLu (approved):
LGTM, verified that ComboBox RAC shows a section header if at least one of its items match the filter and that FF GridList DnD doesnt have anymore shifting

PR #4334, opened 2023-04-05
Improve resize cursors on 1x screens

devongovett:
The points in the SVG paths seem to be on half pixels, so this transforms the shape by half a pixel so it renders on the pixel grid and doesn't look fuzzy on 1x screens. Looks much better on my Windows laptop.


snowystinger (approved):
Still looks good on Mac with Safari/Chrome/FF

LFDanLu (approved):
LGTM, much better on windows for sure

PR #4335, opened 2023-04-05
Build RAC Tailwind example in Verdaccio

devongovett:
This builds the React Aria Components + Tailwind example in CI on main as part of the verdaccio build. This way we can get a link to a build of that app. I'm thinking about potentially linking to such a build from the docs as an example temporarily as well but not sure.

### DIFF
--- a/examples/rac-tailwind/package.json
+++ b/examples/rac-tailwind/package.json
@@ -2,14 +2,9 @@
   "name": "rac-tailwind-example",
   "private": true,
   "scripts": {
-    "start": "parcel src/index.html"
+    "start": "parcel src/index.html",
+    "build": "parcel build src/index.html"
   },
-  "workspaces": [
-    "../../packages/react-aria-components",
-    "../../packages/react-aria",
-    "../../packages/react-stately",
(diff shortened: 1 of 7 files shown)


examples/rac-tailwind/package.json
```diff
@@ -2,14 +2,9 @@
...
-    "start": "parcel src/index.html"
+    "start": "parcel src/index.html",
+    "build": "parcel build src/index.html"
   },
-  "workspaces": [
-    "../../packages/react-aria-components",
-    "../../packages/react-aria",
-    "../../packages/react-stately",
-    "../../packages/*/*"
-  ],
```
devongovett, line 12: These are nice for local development but would mean that in CI we don't use the published version. Not sure how we could have both...

LFDanLu (approved):
LGTM, tested via https://reactspectrum.blob.core.windows.net/reactspectrum/e17fbc7e65d2a740327b251468cbedddfdef5690/verdaccio/rac-tailwind/index.html

snowystinger (approved):
lgtm!

PR #4338, opened 2023-04-05
Add tailwind example link to RAC overview page

devongovett:


packages/react-aria-components/docs/react-aria-components.mdx
```diff
@@ -421,6 +421,10 @@ If you are using Tailwind CSS, we recommend using the [tailwindcss-animate](http
 </Popover>
 ```
 
+## Examples
+
+The documentation for each component includes many examples styled using vanilla CSS and the default class names. We also have an example of many of the components using Tailwind CSS. You can find the [code](https://github.com/adobe/react-spectrum/blob/main/examples/rac-tailwind/src/App.js) in the repo, as well as a [live demo](https://reactspectrum.blob.core.windows.net/reactspectrum/5aa969fbb89086a5e2c7bdc32114df44c48aa01a/verdaccio/rac-tailwind/index.html).
```
devongovett, line 426: Ideally the demo link would update when we do releases but I guess it's only temporary so probably fine for now

reidbarber: Should we bit.ly it?

devongovett: haha, can you edit where a bitly goes later? guess not that different from just editing this file

reidbarber: Yeah, I think so if you log in. I think you could even do like `bit.ly/rac-demo`. Might not be worth it to manage though.

reidbarber (approved):

packages/react-aria-components/docs/react-aria-components.mdx
devongovett: Ah that's a good idea, then I can even change the link if I post it on twitter

reidbarber (approved):

packages/react-aria-components/docs/react-aria-components.mdx
devongovett: ugh seems like you cannot edit unless you pay for bitly

LFDanLu (approved):

PR #4357, opened 2023-04-10
Fix React Aria Components tree shaking

### LINKED ISSUE #4356: Very large size bundle size (80KB+) with only Button element (react-aria-components)
# 📝 Feedback

I'm just tried react-aria-components on a fresh next.js app and by just importing only the button component, bundle size increased by 80KB+. This is not the case with react-aria hooks version.
- 
Screenshot 2023-04-10 at 7 10 03 PM

## 🌍 Your Environment

| [Package name]   | react-aria-components
| Browser          | Chrome 111
| Operating System | Mac OS 12.6.3

devongovett:
Fixes #4356

Normally SWC inserts a PURE annotation comment on React `forwardRef` calls so that they can be tree-shaken when unused. But in the cases where we use a TypeScript cast this doesn't work, so all of those components end up getting included in application bundles even when not used. This PR adds the PURE annotation comments manually in those cases. A follow up could be to make SWC smarter about this.

### DIFF
--- a/packages/react-aria-components/src/Breadcrumbs.tsx
+++ b/packages/react-aria-components/src/Breadcrumbs.tsx
@@ -56,7 +56,7 @@ function Breadcrumbs<T extends object>(props: BreadcrumbsProps<T>, ref: Forwarde
 /**
  * Breadcrumbs display a heirarchy of links to the current page or resource in an application.
  */
-const _Breadcrumbs = (forwardRef as forwardRefType)(Breadcrumbs);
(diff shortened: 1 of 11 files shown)


snowystinger (dismissed):
just to make sure, this comment doesn't get removed by anything? I assume that minifiers are smart enough to not remove it and swc won't remove ones it didn't add?

devongovett:
I ran the build and the comment was preserved.

snowystinger (approved):

reidbarber (approved):

PR #4358, opened 2023-04-10
TagGroup: add default empty state and support for custom empty state

reidbarber:
Adds a default empty state of 'None', and support for a custom empty state via `renderEmptyState`.

If the empty state contains no focusable children, then the group itself should be focusable.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test new stories at TagGroup -> Empty state and Custom empty state


packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -64,6 +67,9 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
   delete props.onAction;
   let {gridProps, labelProps, descriptionProps, errorMessageProps} = useTagGroup({...props, keyboardDelegate}, state, tagsRef);
   let actionsId = useId();
+  if (!renderEmptyState) {
+    renderEmptyState = () => <div>{stringFormatter.format('noTags')}</div>;
```
snowystinger, line 71: can move to line 51 as a default value for the destructuring

packages/@adobe/spectrum-css-temp/components/tags/index.css
```diff
@@ -18,12 +18,20 @@ governing permissions and limitations under the License.
 
 .spectrum-Tags {
   display: inline;
+
+  &.focus-ring {
+    outline: none;
+    box-shadow: 0 0 0 2px var(--spectrum-tag-border-color-key-focus);
+    display: block;
```
snowystinger, line 25: because we need to use a block level display for box-shadow, maybe it'd be better to have spectrum-Tags display as `inline-block`? otherwise, swapping between inline and block can drastically affect layout, better to keep it the same at all times. Another option is to put a display block level element (ie div) in between tag and the tags themselves

also, I can still see an outline if I tab to the group, then click on it. seems to be a result of focus-visible
Screenshot 2023-04-10 at 6 24 24 PM
I do not see this issue with ListView for some reason, so could start by comparing between the two

snowystinger (approved):

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -44,7 +46,8 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
     children,
     actionLabel,
     onAction,
-    labelPosition
+    labelPosition,
+    renderEmptyState = () => <div>{stringFormatter.format('noTags')}</div>
```
LFDanLu, line 50: Should we make this div the same height as a Tag? Right now the height of the container will shrink when the final tag is removed and replaced with this div.

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -173,24 +177,35 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
+                  {...item.props}
+                  key={item.key}
+                  item={item}
+                  state={state}
+                  allowsRemoving={allowsRemoving}
+                  onRemove={onRemove}>
+                  {item.rendered}
+                </Tag>
+              ))}
+              {isEmpty && renderEmptyState()}
```
LFDanLu, line 205: Noticed that focus is being lost to the body when all the tags are removed, is that going be handled in this PR or a separate one?

packages/@react-spectrum/tag/src/TagGroup.tsx
```diff
@@ -173,24 +177,35 @@ function TagGroup<T extends object>(props: SpectrumTagGroupProps<T>, ref: DOMRef
...
+          className={
+            classNames(
+              styles,
+              'spectrum-Tags-container',
+              {
+                'spectrum-Tags-container--empty': isEmpty
+              }
+            )
+          }>
+          <FocusRing focusRingClass={classNames(styles, 'focus-ring')}>
```
ktabors, line 189: Not sure the FocusRing makes sense when ShowAll hides all items. The hiding all items in this story is an existing bug. Would this go away if we fix that bug and always show one tag?

packages/@react-spectrum/tag/src/TagGroup.tsx
reidbarber: It should now go to the container when the last tag is removed.

packages/@react-aria/tag/src/useTag.ts
```diff
@@ -63,9 +63,20 @@ export function useTag<T>(props: AriaTagProps<T>, state: ListState<T>, ref: RefO
...
+
+      // Focus container if last tag was removed.
+      if (state.collection.size === 1 && state.collection.getFirstKey() === item.key) {
+        let lastTag = ref;
+        let container = ref.current.parentElement;
+        setTimeout(() => {
+          if (!lastTag.current) {
+            container.focus();
+          }
+        }, 50);
```
snowystinger, line 75: why 50ms out of curiosity? why not immediately? I assume it's because the tag container doesn't get a tabIndex of -1 until the state change has taken effect? we could instead use a useEffect and a ref, the ref tracks previous count of tags and if the tag group has focus, if the count of tags goes from > 0 to === 0, then we focus the container

does focus get set to the container if we mouse click to remove the last tag already?

reidbarber: The intention was to wait and check that the list tag actually got removed from the DOM. Mostly in order to avoid the case that it wasn't removed immediately (like a confirmation dialog before it actually gets removed). I think your suggestion is much better though, so I just changed it to that.

snowystinger (approved):
LGTM, tested in chrome/safari/ff

packages/@react-aria/tag/src/useTagGroup.ts
```diff
@@ -53,14 +53,20 @@ export function useTagGroup<T>(props: AriaTagGroupProps<T>, state: ListState<T>,
...
   });
   let domProps = filterDOMProps(props);
+
+  // If the last tag is removed, focus the container.
+  let prevCount = useRef(state.collection.size);
+  useEffect(() => {
+    if (prevCount.current > 0 && state.collection.size === 0 && isFocusWithin) {
+      ref.current.focus();
+    }
+  }, [state.collection.size, isFocusWithin, ref]);
```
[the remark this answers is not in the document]
snowystinger: eagle eyes, good catch, yes, this should always be updated at the end of the useEffect body regardless of anything else

LFDanLu (approved):
LGTM

LFDanLu (approved):
LGTM

snowystinger (approved):

Issue #4406, opened 2023-04-19
CMD+A keyboard shortcut triggers error when all rows in table are already selected


tianenpang:
# 🐛 Bug Report 

When all rows of a Table component are already selected and `CMD`+`A` keyboard shortcut is used, it triggers a `setB.has is not a function` error.

## 🤔 Expected Behavior

The `CMD`+`A` keyboard shortcut should toggle the selection state of all rows in the Table component.

## 😯 Current Behavior

It triggers the [`setB.has is not a function`](https://github.com/adobe/react-spectrum/blob/b92f6029e71aa29a9f7b35e403586856a90004ad/packages/%40react-stately/selection/src/useMultipleSelectionState.ts#L25) error.

## 💻 Code Sample

https://codesandbox.io/s/react-aria-table-select-all-via-kbd-tu9jkl

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| react-spectrum   | Latest
| Browser          | Chrome
| Operating System | Windows

reidbarber:
I think we could probably get better behavior for this by updating the following line to `manager.toggleSelectAll()`:
https://github.com/adobe/react-spectrum/blob/b92f6029e71aa29a9f7b35e403586856a90004ad/packages/%40react-aria/selection/src/useSelectableCollection.ts#L221

But also add a `keys === 'all' && selectedKeys === 'all'` condition here to fix the actual issue shown:

https://github.com/adobe/react-spectrum/blob/b92f6029e71aa29a9f7b35e403586856a90004ad/packages/%40react-stately/selection/src/useMultipleSelectionState.ts#L112-L116

PR #4415, opened 2023-04-20
fix: table select-all error when triggered twice with keyboard shortcut

tianenpang:
Closes https://github.com/adobe/react-spectrum/issues/4406

- Fixed `setB.has is not a function` error.

Thanks for @reidbarber's guidance, all credit 🙌

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-aria/selection/src/useSelectableCollection.ts
+++ b/packages/@react-aria/selection/src/useSelectableCollection.ts
@@ -218,7 +218,7 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
       case 'a':
         if (isCtrlKeyPressed(e) && manager.selectionMode === 'multiple' && disallowSelectAll !== true) {
           e.preventDefault();
-          manager.selectAll();
+          manager.toggleSelectAll();
         }
         break;
       case 'Escape':
--- a/packages/@react-aria/selection/src/useSelectableCollection.ts
+++ b/packages/@react-aria/selection/src/useSelectableCollection.ts
@@ -218,7 +218,7 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
       case 'a':
         if (isCtrlKeyPressed(e) && manager.selectionMode === 'multiple' && disallowSelectAll !== true) {
           e.preventDefault();
-          manager.toggleSelectAll();
+          manager.selectAll();
         }
         break;
       case 'Escape':
--- a/packages/@react-stately/selection/src/useMultipleSelectionState.ts
+++ b/packages/@react-stately/selection/src/useMultipleSelectionState.ts
@@ -110,7 +110,8 @@ export function useMultipleSelectionState(props: MultipleSelectionStateProps): M
     },
     selectedKeys,
     setSelectedKeys(keys) {
-      if (allowDuplicateSelectionEvents || !equalSets(keys, selectedKeys)) {
+      let isKeysMatched = keys === 'all' && selectedKeys === 'all';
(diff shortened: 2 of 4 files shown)


packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -218,7 +218,7 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
       case 'a':
         if (isCtrlKeyPressed(e) && manager.selectionMode === 'multiple' && disallowSelectAll !== true) {
           e.preventDefault();
-          manager.selectAll();
+          manager.toggleSelectAll();
```
reidbarber, line 221: After discussing with the team, I think we actually want `manager.selectAll()` instead, since it seems Ctrl+A should be 'select all': https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/grid_role#keyboard_interactions

Sorry for the confusion! Feel free to revert this.

packages/@react-stately/selection/src/useMultipleSelectionState.ts
```diff
@@ -110,7 +110,8 @@ export function useMultipleSelectionState(props: MultipleSelectionStateProps): M
     },
     selectedKeys,
     setSelectedKeys(keys) {
-      if (allowDuplicateSelectionEvents || !equalSets(keys, selectedKeys)) {
+      let isKeysMatched = keys === 'all' && selectedKeys === 'all';
+      if (isKeysMatched && allowDuplicateSelectionEvents || !equalSets(keys, selectedKeys)) {
```
reidbarber, line 114: I think you'll want something like this
```suggestion
      let isBothAll = keys === 'all' && selectedKeys === 'all';
      if (allowDuplicateSelectionEvents || (!isBothAll && !equalSets(keys, selectedKeys)))   
```

so that `equalSets` doesn't get called if both are 'all'.

After taking another look, it might actually be a bit easier to just update `equalSets` to handle 'all' being passed in. It could be renamed to `equalSelections`:

```ts
function equalSelections(setA, setB) {
  if (setA === 'all' && setB === 'all') {
    return true
  } else if (setA === 'all' || setB === 'all') {
    return false.
  }

  if (setA.size !== setB.size) {
    return false;
  }

  for (let item of setA) {
    if (!setB.has(item)) {
      return false;
    }
  }

  return true;
}
```

Either of those should work. The second feels a bit cleaner and that function is only used there, so it isn't a huge change.

reidbarber (commented):
Thanks for the fix!

If you don't mind adding an extra check to a test as well. In `Table.test.js` there is a test called `should support selecting all via ctrl + A`. You could just copy the following lines and paste them right below so we check that the behavior of selecting all twice is correct.

https://github.com/adobe/react-spectrum/blob/0e060e1ea73a38680389e6855720f4c381ef4217/packages/%40react-spectrum/table/test/Table.test.js#L2344-L2349

packages/@react-aria/selection/src/useSelectableCollection.ts
tianenpang: No problem, I just think that `selectAll` makes more sense than `toggleSelectAll`.

packages/@react-stately/selection/src/useMultipleSelectionState.ts
reidbarber: That should work too! Feel free to do that instead.

reidbarber:
GET_BUILD

snowystinger (approved):

reidbarber (approved):
Looks great, thanks!!

PR #4424, opened 2023-04-21
Fix: Table Header does not support className Prop

### LINKED ISSUE #4420: TableHeader(react-aria-components): Type does not support `className`
# 🐛 Bug Report 

When using `TableHeader` component from `react-aria-components` and trying to pass `className` prop ([as in docs](https://react-spectrum.adobe.com/react-aria/Table.html#tableheader-1)), TypeScript says that `TableHeaderProps` does not support `className` property.

## 🤔 Expected Behavior

User should be able to pass `className` without any type warnings

## 😯 Current Behavior

```ts
Type '{ children: Element[]; className: string; }' is not assignable to type 'IntrinsicAttributes & TableHeaderProps<object>'.
  Property 'className' does not exist on type 'IntrinsicAttributes & TableHeaderProps<object>'
```

## 💁 Possible Solution

Extend `TableHeaderProps` with `RenderProps` or `StyleRenderProps` interface.

## 💻 Code Sample

https://codesandbox.io/s/goofy-marco-r8c6dx?file=/src/App.tsx

akash1412:
Closes   #4420 

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/react-aria-components/src/Table.tsx
+++ b/packages/react-aria-components/src/Table.tsx
@@ -3,7 +3,7 @@ import {BaseCollection, CollectionContext, CollectionProps, CollectionRendererCo
 import {buildHeaderRows} from '@react-stately/table';
 import {ButtonContext} from './Button';
 import {CheckboxContext} from './Checkbox';
-import {ContextValue, defaultSlot, Provider, RenderProps, SlotProps, StyleRenderProps, useContextProps, useRenderProps} from './utils';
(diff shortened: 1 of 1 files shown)


packages/react-aria-components/src/Table.tsx
```diff
@@ -368,7 +368,7 @@ export function useTableOptions(): TableOptionsContextValue {
   return useContext(TableOptionsContext)!;
 }
 
-export interface TableHeaderProps<T> {
+export interface TableHeaderProps<T> extends StyleRenderProps<T> {
```
devongovett, line 371: I think this should just extend StyleProps. Otherwise this would have to support a function with T as the argument which doesn't really seem that useful.

PR #4440, opened 2023-04-26
Update datepicker documentation which has missing import

samrobbins85:
## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Copy the code into an IDE and see that there is no longer an error that `createCalendar` is not defined


samrobbins85:
I'm not sure on the best approach for the failing type check here. It's failing because separate code blocks both contain the import, but the other import is later in the list of components, which creates initial confusion due to `createCalendar` not being defined. I could have only the `createCalendar` import in the first usage and just `getWeeksInMonth` in the second usage, but that could also be confusing for the second usage

devongovett:
Not all of our examples include all of the imports because that would be very repetitive. But we might be able to improve this for copy past in our upcoming work. Going to close for now.

Issue #4429, opened 2023-04-24
React Aria Components: Expose more states on wrapper elements


devongovett:
_Originally reported here: https://twitter.com/devcassa/status/1650282210690203650_

# 🙋 Feature Request

In components like `<TextField>`, `<ComboBox>`, and others, it would be useful to expose some states on the wrapper div element in addition to the `<Input>`. For example right now we only expose `data-focus-visible` on the `<input>` element, but this can be useful for styling adjacent elements such as icons and such. You could do this with CSS `:has()` but this is not yet supported across all browsers and adds a bit more complexity.

The most obvious states to expose are data-focused and data-focus-visible, but others such as disabled and validation states might be useful as well. Hover and press are a bit weird because it's unclear if that means hovering the label should also set that state or only if you're over the input. We should do a full audit of the states exposed by each component to ensure they are consistent.

## 🤔 Expected Behavior

Some states could be exposed on the wrapper div so it's easy to style all of the elements inside based on that state.

## 😯 Current Behavior

States are only exposed on Input element, which makes it hard to style adjacent elements.

PR #4451, opened 2023-04-28
rac dropzone

### LINKED ISSUE #993: Consider having support for custom file upload inputs within react aria.
First of all thanks for open sourcing `@react-aria/*` and `@react-stately/*`.

It would be nice to have a hook within react-aria which can be used to create an accessible but customisable file upload.
I'm willing to implement it myself, I believe the first step would be to draft an RFC?

Before I get started with it, are you willing to accept a PR for it? And do you have already some thoughts how the API would look like?

yihuiliao:
DropZone allows for data to be dragged and dropped or pasted into it. In its current implementation, there is a hidden button that is rendered when there is no FileTrigger which receives focus. 

Accessing the file system is optional. If someone wants to access the file system, they can pass FileTrigger as a child which contains an hidden input. The FileTrigger accepts a Button or Link which the user can interact with to open the file system. 

Note: Unlike previous implementations of DropZone, the only way to access the file system with a mouse is by clicking on either a button or link within FileTrigger. There is no use case in which the entire DropZone should be clickable to open the file system. This behavior is similar to the dropzone on Google Sheets when importing images.

Might also close #993 with FileTrigger

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

**DropZone (no FileTrigger):**
Test for dnd behavior and pasting. We should expect to see an onDrop action any time a file is dropped. 
Stories to test:
- DropZone with Draggable Object
- DropZone with Copyable Object
- DropZone with Render Props
- DropZone with Capture (MOBILE ONLY)

**DropZone With FileTrigger:**
Should be able to drag/drop files into the dropzone as well as click on the button/link to open the file system. We should expect to see an onDrop/onChange action any time a file is dropped into the DropZone.
Stories to test:
- DropZone with FileTrigger Link
- DropZone with FileTrigger Button
- DropZone With Draggable And FileTrigger
- DropZone Only Accept PNG With FileTrigger
   - should only be able to drop png files regardless if you are dragging/dropping or accessing the file system

**FileTrigger:**
Should be able to click on the button/link to open the file system. We should expect an onChange action
Stories to test:
- FileTrigger Button
- FileTrigger Link


packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -674,3 +676,37 @@ function CustomTab(props) {
...
+      {...dragProps}
+      role="button"
+      tabIndex={0}
+      className={`draggable ${isDragging ? 'dragging' : ''}`}>
+      Drag me
+    </div>
+  );
+}
+
+export const DropZoneExample = () => (
```
reidbarber, line 701: Can we also get an example with render props? It could look something like this:

```tsx
    <DropZone 
      onDrop={action('On Drop')}
      onDropEnter={action('Drop Enter')}
      onDropExit={action('Drop Exit')}
      style={{border: '2px solid white', margin: '20px', padding: '20px'}}>
      {({isHovered, isFocused, isFocusVisible, isDropTarget}) => (
        <div>
          <div>DropZone</div>
          <div>isHovered: {isHovered ? 'true' : 'false'}</div>
          <div>isFocused: {isFocused ? 'true' : 'false'}</div>
          <div>isFocusVisible: {isFocusVisible ? 'true' : 'false'}</div>
          <div>isDropTarget: {isDropTarget ? 'true' : 'false'}</div>
        </div>
      )}
    </DropZone>
```

packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -674,3 +676,37 @@ function CustomTab(props) {
...
+  <div>
+    <Draggable />
+    <DropZone 
+      onDrop={action('OnDrop')}
+      onDropEnter={action('OnDropEnter')}
+      onDropExit={action('OnDropExit')}
+      style={{border: '2px solid white', margin: '20px', padding: '20px'}}>
```
reidbarber, line 708: We could add some styles to `/example/index.css`, including for the different states.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,51 @@
...
+export interface DropZoneProps extends Omit<DropOptions, 'onDropActivate' | 'onInsert' | 'onRootDrop' | 'onItemDrop' | 'onReorder'>, DOMProps, SlotProps {}
+
+export const DropZoneContext = createContext<ContextValue<DropZoneProps, HTMLDivElement>>(null);
+
+function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
+  [props, ref] = useContextProps(props, ref, DropZoneContext);
+  let {dropProps, isDropTarget} = useDrop({...props, ref});
+  let {focusProps, isFocused, isFocusVisible} = useFocusRing();
+  let {hoverProps, isHovered} = useHover({});
+  let {buttonProps} = useButton({elementType: 'div'}, ref);
```
reidbarber, line 26: More of a question for the team, but maybe we want `usePress` in the case that we want to support interactive children?

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {ContextValue, DOMProps, SlotProps, useContextProps, useRenderProps} from './utils';
+import {DropOptions, mergeProps, useButton, useDrop, useFocusRing, useHover} from 'react-aria';
+import React, {createContext, ForwardedRef, forwardRef} from 'react';
+
+export interface DropZoneProps extends Omit<DropOptions, 'onDropActivate' | 'onInsert' | 'onRootDrop' | 'onItemDrop' | 'onReorder'>, DOMProps, SlotProps {}
```
reidbarber, line 17: I think we'll want `DropZoneRenderProps` too. Check for how Button does this.

reidbarber (commented):
Looks great so far!

packages/react-aria-components/src/DropZone.tsx
yihuiliao: for now, do we just want to include isHovered, isFocused, and isFocusVisible?

reidbarber: Yeah, that looks right. And we can have DropZoneProps extend `RenderProps<DropZoneRenderProps>`.

reidbarber: Also, I think we can omit `'getDropOperationForPoint'` here too.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,51 @@
...
+  let {dropProps, isDropTarget} = useDrop({...props, ref});
+  let {focusProps, isFocused, isFocusVisible} = useFocusRing();
+  let {hoverProps, isHovered} = useHover({});
+  let {buttonProps} = useButton({elementType: 'div'}, ref);
+  let renderProps = useRenderProps({ 
+    ...props,
+    values: {isHovered, isFocused, isFocusVisible, isDropTarget},
+    defaultClassName: 'react-aria-DropZone'
+  });
+
```
reidbarber, line 32: We also want to support pasting onto the DropZone, so I think we'll want to add this:

```tsx
  let {clipboardProps} = useClipboard({
    onPaste: (items) => props.onDrop({
      type: 'drop',
      items,
      x: 0,
      y: 0,
      dropOperation: 'copy'
    })
  });
```

Not 100% certain about x, y, and dropOperation, but I think that is what we want.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,137 @@
...
+  let onInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    if (props.onDrop) {
+      props.onDrop({
+        type: 'drop',
+        dropOperation: 'copy',
+        x: 0,
+        y: 0,
+        items: [...(e.target.files || [])].map((file) =>
```
yihuiliao, line 95: there was a linting error here that said roughly "type Object must have a Symbol.iterator method that returns an iterator". originally it was just `items: [...e.target.files]`and this was the solution i was able to come up with but i assume there is a better solution out there

reidbarber: You also might be able to update the if condition to `if (props.onDrop && e.target.files.length > 0)` instead.

yihuiliao: hmmm it's still giving the same linting error if i update the condition. it also introduces a new linting error on the line with the updated condition saying that the "object is possibly null".

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+export interface DropZoneProps extends Omit<DropOptions, 'getDropOperationForPoint' | 'onDropActivate' | 'onInsert' | 'onRootDrop' | 'onItemDrop' | 'onReorder'>, RenderProps<DropZoneRenderProps>, SlotProps {}
+
+export const DropZoneContext = createContext<ContextValue<DropZoneProps, HTMLDivElement>>(null);
+
+function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
+  [props, ref] = useContextProps(props, ref, DropZoneContext);
+  let {dropProps, isDropTarget} = useDrop({...props, ref});
+  let {hoverProps, isHovered} = useHover({});
+  let {focusProps, isFocused, isFocusVisible} = useFocusRing();
```
snowystinger, line 50: the focus props for this are going to the wrong element I think, they should be on the hidden button

otherwise, if this should show a focus ring around the entire set of controls regardless, then you can specify `{within: true}` on the props for `useFocusRing`

but I think the focus props should be hooked up to the hidden button

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,137 @@
...
+   * @selector [data-drop-target]
+   */
+  isDropTarget: boolean
+}
+
+export interface DropZoneProps extends Omit<DropOptions, 'getDropOperationForPoint' | 'onDropActivate' | 'onInsert' | 'onRootDrop' | 'onItemDrop' | 'onReorder'>, RenderProps<DropZoneRenderProps>, SlotProps {}
+
+export const DropZoneContext = createContext<ContextValue<DropZoneProps, HTMLDivElement>>(null);
+
+function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
```
snowystinger, line 46: Note to add in the future, possibly isDisabled prop

packages/react-aria-components/example/index.css
```diff
@@ -175,3 +175,9 @@ html {
     }
   }
 }
+
+.dropzone {
+  border: 2px solid white;
+  margin: 20px;
+  padding: 20px;
+}
```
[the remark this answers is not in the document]
reidbarber: Style nit, but I would also make the border dashed instead of solid.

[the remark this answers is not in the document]
snowystinger: yes, please make it whatever you like, i was mostly demonstrating the selector to use

packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -674,3 +676,145 @@ function CustomTab(props) {
...
+      <Input />
+      <Button slot="file">upload</Button>
+    </DropZone>
+  </div>
+);
+
+
+export const DropzoneExampleWithDraggableObject = (props) => (
+  <div>
+    <Draggable />
```
snowystinger, line 761: this example is a bit odd to use. I tab to the Draggable, hit enter, but then from what I can tell, focus doesn't move to the dropzone. @reidbarber and I are happy to help you debug this if you need it

packages/react-aria-components/example/index.css
```diff
@@ -175,3 +175,17 @@ html {
...
+.dropzone {
+  border: 2px solid white;
+  margin: 20px;
+  padding: 20px;
+
+  &[data-focus-visible][data-focused] {
+    border-color: blue;
+  }
+
+  &[data-hovered]{
```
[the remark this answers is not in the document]
yihuiliao: could we do `data-hovered={isHovered || isDropTarget || undefined} `?

packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -674,3 +676,145 @@ function CustomTab(props) {
...
+      onDropExit={action('OnDropExit')}>
+      DropZone Area
+    </DropZone>
+  </div>
+);
+
+// copying does not currently work 
+export const DropzoneExampleWithCopyableObject = (props) => (
+  <div>
+    <Copyable />
```
snowystinger, line 776: Not sure why, but when I copy and paste this one, the onDrop `action` doesn't seem to have the text in it. Probably just a fix in the story needed, but we should update that. Again, we can help you with this one if it's not easy

packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -674,3 +676,145 @@ function CustomTab(props) {
...
+          <div>isFocused: {isFocused ? 'true' : 'false'}</div>
+          <div>isFocusVisible: {isFocusVisible ? 'true' : 'false'}</div>
+          <div>isDropTarget: {isDropTarget ? 'true' : 'false'}</div>
+        </div>
+      )}
+    </DropZone>
+  </div>
+);
+
+export const DropzoneExample = (props) => (
```
snowystinger, line 811: unsure what this example is for

packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -674,3 +676,145 @@ function CustomTab(props) {
...
+      onDropEnter={action('OnDropEnter')}
+      onDropExit={action('OnDropExit')}>
+      DropzoneArea
+    </DropZone>
+  </div>
+);
+
+export const DropzoneWithRenderProps = (props) => (
+  <div>
+    <Draggable />
```
snowystinger, line 790: similar issue to the above one, if I keyboard drag this, there is no visual indication via the focus ring that focus is on the dropzone, but we do get 'isDropTarget', so maybe we use that instead of the dragover data-attribute I suggested previous and instead have a data-attribute data-is-drop-target or something

I suggest trying them both locally and form an opinion of which one makes more sense to you, then we can discuss which way to go

packages/react-aria-components/stories/index.stories.tsx
yihuiliao: just to clarify, what you're seeing is something like OnDrop: { } in the actions tab when you complete the paste?

snowystinger: no, i see the `{type: "drop"... etc` but when I drill into items, I only get this
Screenshot 2023-05-09 at 3 25 21 PM

packages/react-aria-components/example/index.css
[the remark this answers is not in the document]
snowystinger: I don't think we want to always show the "this is a drop target that you're dropping on" when the user is just hovering without dragging. I feel like that would be confusing because it'd lead them to think that they were dragging something.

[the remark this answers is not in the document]
reidbarber: Maybe something like:

```tsx
data-drop-target={isDropTarget || undefined}
```

packages/react-aria-components/stories/index.stories.tsx
yihuiliao: im thinking maybe using isDropTarget since we already define isDropTarget in useDrop as such: "Whether the drop target is currently focused or hovered" which i think fits how we would want this to work

packages/react-aria-components/stories/index.stories.tsx
yihuiliao: just making that a note that is related to how storybook actions work so we won't be focusing on this for now

packages/react-aria-components/example/index.css
```diff
@@ -175,3 +175,38 @@ html {
...
+  margin: 20px;
+  padding: 20px;
+  outline: none;
+
+  &[data-focus-visible][data-focused] {
+    border: 2px dashed blue;
+  }
+
+  &[data-hovered]{
+    border-color: 2px dashed green;
```
snowystinger, line 190: this isn't valid `border-color`, it should only be a color
however, I think we should drop the hovered style completely, as there is nothing to interact with, so changing the style on hover is misleading. instead, we're using the drop-target to indicate when an interaction has a meaningful result

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,165 @@
...
+   */
+  isFocusVisible: boolean,
+  /**
+   * Whether the dropzone is the drop target.
+   * @selector [data-drop-target]
+   */
+  isDropTarget: boolean
+}
+// note: possibly add isDisabled prop in the future
+export interface DropZoneProps extends Omit<DropOptions, 'getDropOperationForPoint' | 'onDropActivate' | 'onInsert' | 'onRootDrop' | 'onItemDrop' | 'onReorder'>,
```
snowystinger, line 46: why'd we decide to get rid of `onDropActivate`?

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,165 @@
...
+    })
+  });
+
+  let renderProps = useRenderProps({ 
+    ...props,
+    values: {isHovered, isFocused, isFocusVisible, isDropTarget},
+    defaultClassName: 'react-aria-DropZone'
+  });
+
+  let {pressProps} = usePress({
```
snowystinger, line 100: I think we should get rid of this usePress. I think if people want their DropZone to have a file input, then they should have to provide a button or link to access it. Otherwise, I think the assistive technology will read something that won't be so obvious, may not even make a mention of a file input.

Did you find an example where the whole dropzone was clickable as a file input? I know the google sheets one has a button in it for handling it.

packages/react-aria-components/test/DropZone.test.js
```diff
@@ -0,0 +1,455 @@
...
+    let {getByTestId} = render(
+      <DropZoneContext.Provider value={{slots: {test: {'aria-label': 'test'}}}}>
+        <DropZone data-testid="foo" slot="test" />
+      </DropZoneContext.Provider>
+    );
+
+    let dropzone = getByTestId('foo');
+    expect(dropzone).toHaveAttribute('class', 'react-aria-DropZone');
+
+    console.log(dropzone.outerHTML);
```
snowystinger, line 65:
```suggestion
```

packages/react-aria-components/test/DropZone.test.js
```diff
@@ -0,0 +1,455 @@
...
+        <Button slot="file">Upload</Button>
+        <Input data-testid="boo" />
+      </DropZone>);
+
+    let dropzone = getByTestId('foo');
+    let button = getByRole('button');
+    expect(button).toHaveAttribute('class', 'react-aria-Button');
+
+    expect(dropzone).not.toHaveAttribute('data-focus-visible');
+    expect(dropzone).not.toHaveClass('focus');
```
snowystinger, line 158: is this a class we actually have? i thought we only applied a data attribute? we only seem to be checking for the negative of it, I don't see anywhere this exists

packages/react-aria-components/test/DropZone.test.js
```diff
@@ -0,0 +1,455 @@
...
+    let {getByRole} = render(
+      <DropZone>
+        <Input />
+        <Text slot="heading">
+          Test
+        </Text>
+      </DropZone>
+    );
+    let button = getByRole('button');
+    console.log(button.outerHTML);
```
snowystinger, line 434:
```suggestion
```

snowystinger (commented):
would be good to include a couple tests where we specify allowed types of items for drops or files for upload

packages/react-aria-components/src/DropZone.tsx
yihuiliao: i don't think there was a particular reason? these props are based off of reid's notes at the beginning but i can add it back in

snowystinger: let's add it back for the moment, thanks!

packages/react-aria-components/src/DropZone.tsx
yihuiliao: i don't have any specific examples of where the entire dropzone is clickable as a file input but i think it's something that people might want/assume of a dropzone. that's why in its current implementation, when there's just a file input (so no visible button or link), VO will read out "press enter to open files" to make dropzone's interactions more obvious

that said, i think that we could follow the pattern you suggested and require that people include either a link or button in order to access the file input

packages/react-aria-components/test/DropZone.test.js
yihuiliao: yeah that's my bad. we can remove those checks

packages/react-aria-components/src/DropZone.tsx
snowystinger: Ah, right, the hidden button. You can leave as is, but let's get more opinions.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,165 @@
...
+        tabIndex={-1}   
+        data-hovered={isHovered || undefined}
+        data-focused={isFocused || undefined} 
+        data-focus-visible={isFocusVisible || undefined}
+        data-drop-target={isDropTarget || undefined} > 
+        {!hasButton && !hasLink && <VisuallyHidden>
+          <button 
+            {...mergeProps(buttonProps, focusProps, clipboardProps, labelProps)}
+            ref={buttonRef}  /> 
+        </VisuallyHidden>}
```
reidbarber, line 154: Formatting nit:

```suggestion
        {!hasButton && !hasLink &&
          <VisuallyHidden>
            <button 
              {...mergeProps(buttonProps, focusProps, clipboardProps, labelProps)}
              ref={buttonRef}  /> 
          </VisuallyHidden>
        }
```

reidbarber (dismissed):
LGTM

packages/react-aria-components/src/DropZone.tsx
LFDanLu: Can/should users be able to opt out of the "click to open file picker" behavior? What if they want to customize the DropZone click experience? Should DropZone offer any callbacks that fire on dropzone press for analytics/etc?

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,160 @@
...
+function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
+  [props, ref] = useContextProps(props, ref, DropZoneContext);
+  let {dropProps, isDropTarget} = useDrop({...props, ref});
+  let {hoverProps, isHovered} = useHover({});
+  let {focusProps, isFocused, isFocusVisible} = useFocusRing({within: true});
+
+  let buttonRef = useRef<HTMLButtonElement>(null);
+  let inputRef = useRef<HTMLInputElement>(null);
+  let uploadLinkRef = useRef<HTMLAnchorElement>(null);
+  let uploadButtonRef = useRef<HTMLButtonElement>(null);
```
LFDanLu, line 60: Are these refs used for anything right now? Can we remove them?

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,160 @@
...
+  let hasButton = useHasChild('button[slot=file]', ref);
+  let hasLink = useHasChild('span[slot=file]', ref);
+
+  let hasInput = useHasChild('input[type=file]', ref);
+  let textId = useId();
+  let isVirtualDrag = isVirtualDragging();
+  let labelText = useMemo(() => {
+    if (!isVirtualDrag && hasInput) {
+      return 'Press enter to select a file';
```
LFDanLu, line 69: This label is pretty desktop specific, should we add a check to see if it is a mobile device and have a screen reader specific label for "double tap to select a file" instead? Additionally, it looks like a mobile (Android) screenreader user can focus the DropZone as a whole and double tap that to trigger the file picker instead, will need to double check on iPad. Maybe don't need the visually hidden button if both can focus the dropzone in that case

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,160 @@
...
+      y: 0,
+      dropOperation: 'copy'
+    })
+  });
+
+  let renderProps = useRenderProps({ 
+    ...props,
+    values: {isHovered, isFocused, isFocusVisible, isDropTarget},
+    defaultClassName: 'react-aria-DropZone'
+  });
```
LFDanLu, line 92: With the update to useRenderProps from https://github.com/adobe/react-spectrum/pull/4520, you'll need to add `filterDOMProps` to propagate `data-*` attributes to the DropZone div. Maybe handled by usePress already?

packages/react-aria-components/package.json
```diff
@@ -23,8 +23,10 @@
   },
   "dependencies": {
     "@internationalized/date": "^3.2.0",
+    "@react-aria/dnd": "^3.2.0",
     "@react-aria/focus": "^3.12.0",
     "@react-aria/utils": "^3.16.0",
+    "@react-spectrum/utils": "^3.9.1",
```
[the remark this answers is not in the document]
yihuiliao: i'll copy over useHasChild over to react-aria/utils and as for isVirtualDragging, maybe get additional opinions for that since im not too sure

packages/react-aria-components/src/DropZone.tsx
yihuiliao: i think they were used for a previous check that we now use useHasChild for but i think we can remove them

packages/react-aria-components/src/DropZone.tsx
yihuiliao: i can confirm that the same behavior occurs on the iPad as well

packages/react-aria-components/src/DropZone.tsx
reidbarber: > Can/should users be able to opt out of the "click to open file picker" behavior?

I think if they don't explicitly include a button or link to handle this, it's an okay assumption, otherwise I don't know how the file picker would be accessible to mouse/touch users. By including an `<Input />`, they're opting in to including a file picker.

> Should DropZone offer any callbacks that fire on dropzone press for analytics/etc?

Users should be able to pass `onChange` to their provided `<Input />` to know when a file is selected.

reidbarber: I guess we could also pass through `onPress` from the DropZone to `usePress`.

LFDanLu: Yeah, I think we should still have some top level DropZone events that get chained with any respective props the user provides to the inner component. It could be the `onPress` that you mentioned or maybe callbacks that trigger before the file picker opens and after an item is selected.

packages/react-aria-components/src/DropZone.tsx
LFDanLu: Gotcha, we can wait for further testing and decide if the visually hidden button should only be rendered for non-mobile experiences then

packages/react-aria-components/src/DropZone.tsx
yihuiliao: just to document, i added all the press events to dropzone and also added isPressed to render props as a result

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,182 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+import {AriaLabelingProps, PressEvents} from '@react-types/shared';
+import {ButtonContext} from './Button';
+import {ContextValue, Provider, RenderProps, SlotProps, useContextProps, useRenderProps} from './utils';
+import {DropOptions, mergeProps, useButton, useClipboard, useDrop, useFocusRing, useHover, useId, usePress, VisuallyHidden} from 'react-aria';
+import {filterDOMProps, useHasChild, useIsMobileDevice} from '@react-aria/utils';
+import {InputContext} from './Input';
+import {isVirtualDragging} from '@react-aria/dnd';
```
yihuiliao, line 19: see daniel's [comment](https://github.com/adobe/react-spectrum/pull/4451#discussion_r1202851237)

i left it as is but would like other opinions on this

LFDanLu: I think I'll revise my opinion here actually. I don't think `isVirtualDragging` should be exported from the `react-aria` monopackage actually, this is probably fine

packages/@react-aria/utils/src/index.ts
```diff
@@ -35,3 +35,5 @@ export {scrollIntoView, scrollIntoViewport} from './scrollIntoView';
 export {clamp, snapValueToStep} from '@react-stately/utils';
 export {isVirtualClick, isVirtualPointerEvent} from './isVirtualEvent';
 export {useEffectEvent} from './useEffectEvent';
+export {useIsMobileDevice} from './useIsMobileDevice';
+export {useHasChild} from './useHasChild';
```
LFDanLu, line 39: Now that you have `useIsMobileDevice` and `useHasChild` in the `aria/utils` package, you can delete the `spectrum/utils` versions and re-export them here https://github.com/adobe/react-spectrum/blob/bb5259d1bdcd0ce5e9d3636af64e56733a6e5a04/packages/%40react-spectrum/utils/src/index.ts#L36

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -41,38 +42,49 @@ export interface DropZoneRenderProps {
...
+  let isMobile = useIsMobileDevice();
   let hasInput = useHasChild('input[type=file]', ref);
   let textId = useId();
   let isVirtualDrag = isVirtualDragging();
   let labelText = useMemo(() => {
-    if (!isVirtualDrag && hasInput) {
+    if (!isVirtualDrag && hasInput && !isMobile) {
       return 'Press enter to select a file';
+    } else if (!isVirtualDrag && hasInput) {
+      return 'Double tap to select a file';
```
LFDanLu: Mega nit: perhaps the following for more clarity on the logic flow?
```suggestion
    if (!isVirtualDrag && hasInput) {
      return isMobile ? 'Double tap to select a file' : 'Press enter to select a file';
```

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -41,38 +42,49 @@ export interface DropZoneRenderProps {
...
   let {focusProps, isFocused, isFocusVisible} = useFocusRing({within: true});
 
   let buttonRef = useRef<HTMLButtonElement>(null);
   let inputRef = useRef<HTMLInputElement>(null);
-  let uploadLinkRef = useRef<HTMLAnchorElement>(null);
-  let uploadButtonRef = useRef<HTMLButtonElement>(null);
   let hasButton = useHasChild('button[slot=file]', ref);
   let hasLink = useHasChild('span[slot=file]', ref);
 
+  let isMobile = useIsMobileDevice();
```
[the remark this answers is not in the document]
snowystinger: that's a good point
thinking about this more though, an iPad could have a keyboard attached to it, so I think we actually want to check current modality

then if it's keyboard modality, we say "press enter to select a file", vs if it's touch, we say "double tap..."
though is double tap for touch screen readers? i think users will know what to do if we just label it as "File system picker" or something along those lines. A VoiceOver user knows that they need to double tap because it'll announce as a button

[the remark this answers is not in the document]
snowystinger: https://github.com/adobe/react-spectrum/blob/5a24e8147db1f596a0b047ffd88570ef935f5c29/packages/%40react-aria/interactions/src/useFocusVisible.ts#L181

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,182 @@
...
+        onPress(e);
+      }
+      if (inputRef.current && !hasButton  && !hasLink) {
+        inputRef.current.click();
+      }
+    },
+    onPressChange,
+    onPressEnd,
+    onPressStart,
+    onPressUp
```
snowystinger, line 121:
```suggestion
    {...mergeProps({onPress: (e) => {
      if (onPress) {
        onPress(e);
      }
      if (inputRef.current && !hasButton  && !hasLink) {
        inputRef.current.click();
      }
    }},
    props
    )}
```

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,182 @@
...
+  return (
+    <Provider
+      values={[
+        [InputContext, {ref: inputRef, type: 'file', style: {display: 'none'}, onChange: onInputChange}],
+        [LinkContext, {slot: 'file', onPress: () => inputRef.current?.click()}], 
+        [ButtonContext, {slot: 'file', onPress: () => inputRef.current?.click()}],
+        [TextContext, {id: textId, slot: 'heading'}]
+      ]}>
+      <div
+        {...mergeProps(dropProps, hoverProps, pressProps, DOMProps, buttonProps)}
```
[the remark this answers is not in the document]
snowystinger: I don't think so, maybe focused-within? but exposing focus here for the visible button/link inside is focused seems confusing since they are on an independently focusable element

packages/react-aria-components/src/DropZone.tsx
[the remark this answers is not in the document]
LFDanLu: ah, right iPad + keyboard, good catch. I think labeling the button with "Open File system" can be sufficient, we can test it out. If it isn't, I'd recommend doing something like https://github.com/adobe/react-spectrum/blob/5a24e8147db1f596a0b047ffd88570ef935f5c29/packages/%40react-aria/table/src/useTableColumnResize.ts#L173-L176 and then checking if modality is "touch" for the mobile screen reader specific announcement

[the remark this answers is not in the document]
yihuiliao: so to clarify, we don't really need to check how the user is interacting with the dropzone. instead, we can just change the message to be more general and we assume that users will know how to interact with it.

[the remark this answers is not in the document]
LFDanLu: yep

packages/react-aria-components/src/DropZone.tsx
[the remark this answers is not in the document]
LFDanLu: This is in a similar vein to https://github.com/adobe/react-spectrum/issues/4429, I think it would be valuable to expose these focus attributes on the wrapper div so the user has the option of easily applying styles to the wrapper or other elements they may render inside the DropZone when something inside is focused.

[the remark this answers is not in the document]
snowystinger: I think that's fine when it's a group control, but this isn't a group control in the same way.
Those buttons/links may do something else entirely such as take you to another page.
The hidden button can already receive focus and is the focus state we already expose on the wrapper.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,178 @@
...
+  return (
+    <Provider
+      values={[
+        [InputContext, {ref: inputRef, type: 'file', style: {display: 'none'}, onChange: onInputChange}],
+        [LinkContext, {slot: 'file', onPress: () => inputRef.current?.click()}], 
+        [ButtonContext, {slot: 'file', onPress: () => inputRef.current?.click()}],
+        [TextContext, {id: textId, slot: 'heading'}]
+      ]}>
+      <div
+        {...mergeProps(dropProps, hoverProps, pressProps, DOMProps, buttonProps, focusProps)}
```
yihuiliao, line 152: added focusProps to wrapper div based on this [comment](https://github.com/adobe/react-spectrum/pull/4451#discussion_r1206023519) but we can test it out and decide whether we want to continue to include these props

packages/@react-spectrum/utils/src/index.ts
```diff
@@ -29,8 +29,8 @@ export {
   getResponsiveProp
 } from './styleProps';
 export {useSlotProps, cssModuleToSlots, SlotProvider, ClearSlots} from './Slots';
-export {useHasChild} from './useHasChild';
-export {useIsMobileDevice} from './useIsMobileDevice';
+export {useHasChild} from '@react-aria/utils';
+export {useIsMobileDevice} from '@react-aria/utils';
```
snowystinger, line 33: I was thinking about it more, we don't need to move either of these

useIsMobileDevice is no longer used

and in place of useHasChild, we can actually find out if the child exists by providing a slotCallbackSymbol, if it's called, then we know we have that child

can see https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/ComboBox.tsx#L128
we'll need to be careful not to put the props into state, because that may cause extra re-renders, just if that callback is invoked, we should set hasInput to to true

devongovett: There's also `useSlot` in react-aria-components utils.tsx, which returns a callback ref to register when a slot is present.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,178 @@
...
+  let {dropProps, isDropTarget} = useDrop({...props, ref});
+  let {hoverProps, isHovered} = useHover({});
+  let {focusProps, isFocused, isFocusVisible} = useFocusRing({within: true});
+
+  let buttonRef = useRef<HTMLButtonElement>(null);
+  let inputRef = useRef<HTMLInputElement>(null);
+  let hasButton = useHasChild('button[slot=file]', ref);
+  let hasLink = useHasChild('span[slot=file]', ref);
+
+  let hasInput = useHasChild('input[type=file]', ref);
```
devongovett, line 69: Related to my above comment, I would do these as something like this:

```js
let [buttonRef, hasButton] = useSlot();
```

Then you can pass `buttonRef` to the ButtonContext via the `ref` property. You shouldn't need `slot` then, and it would allow users to provide their own `slot` prop to e.g. `<Input>`.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,178 @@
...
+        [LinkContext, {slot: 'file', onPress: () => inputRef.current?.click()}], 
+        [ButtonContext, {slot: 'file', onPress: () => inputRef.current?.click()}],
+        [TextContext, {id: textId, slot: 'heading'}]
+      ]}>
+      <div
+        {...mergeProps(dropProps, hoverProps, pressProps, DOMProps, buttonProps, focusProps)}
+        {...renderProps}
+        ref={ref}
+        slot={props.slot}
+        tabIndex={-1}
```
devongovett, line 156: Why does this element need to be focusable? Shouldn't the button be the focusable element in a drop zone?

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,178 @@
...
+        data-hovered={isHovered || undefined}
+        data-focused={isFocused || undefined}
+        data-focus-visible={isFocusVisible || undefined}
+        data-drop-target={isDropTarget || undefined} 
+        data-pressed={isPressed || undefined} >
+        {!hasButton && !hasLink &&
```
devongovett, line 162: Why is the button only rendered if there are no other buttons or links in the children? Wouldn't that prevent most of the behavior from working if the below props were not used?

packages/react-aria-components/src/DropZone.tsx
devongovett: looks like it's getting added in two places now? it's still on the button below as well. same with a bunch of these. I think we should decide on one focusable element (I would guess the visually hidden button). We can include the data attributes on this div but still handle all the events at the button level.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,178 @@
...
+    }
+  };
+  
+  return (
+    <Provider
+      values={[
+        [InputContext, {ref: inputRef, type: 'file', style: {display: 'none'}, onChange: onInputChange}],
+        [LinkContext, {slot: 'file', onPress: () => inputRef.current?.click()}], 
+        [ButtonContext, {slot: 'file', onPress: () => inputRef.current?.click()}],
+        [TextContext, {id: textId, slot: 'heading'}]
```
devongovett, line 149: I'm wondering if these slots are a bit too opinionated for RAC. A file upload button seems somewhat unrelated from dnd itself. Hiding it by default, and connecting it to Buttons and Links is also kinda specific. Maybe instead we should split this into two components: DropZone, and FileTrigger or something like that (name tbd). Then you could nest those, or use them independently.

```jsx
<FileTrigger>
  <Button /> or <Link />
</FileTrigger>
```

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  let hasInput = useHasChild('input[type=file]', ref);
+  let textId = useId();
+  let isVirtualDrag = isVirtualDragging();
+  let labelText = useMemo(() => {
+    if (!isVirtualDrag && hasInput) {
+      return 'Open file system';
+    }
+  }, [isVirtualDrag, hasInput]);
+  let labelProps = useLabels({'aria-label': labelText, 'aria-labelledby': textId});
```
ktabors, line 77: I see where we are defining these and I see where we are testing these, but I'm not seeing this in the Storybook examples.

packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -750,3 +752,173 @@ function CustomTab(props) {
...
+        className={styles.copyable}>
+        Copy me 
+      </div>
+    </FocusRing>
+  );
+}
+
+export const DropzoneExampleWithInput = (props) => (
+  <div>
+    <DropZone
```
ktabors, line 802: Should DropZone's have aria-labels?

packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -750,3 +752,173 @@ function CustomTab(props) {
...
+        DropZone Area
+      </Text>
+    </DropZone>
+  </div>
+);
+
+export const DropzoneExampleWithCopyableObject = (props) => (
+  <div>
+    <Copyable />
+    <DropZone
```
ktabors, line 867: Is the VO for this complete? When I test Chrome on OSX it seems like it's not announcing some drop events.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,178 @@
...
+export interface DropZoneProps extends Omit<DropOptions, 'getDropOperationForPoint' | 'onInsert' | 'onRootDrop' | 'onItemDrop' | 'onReorder'>, RenderProps<DropZoneRenderProps>, SlotProps, AriaLabelingProps, PressEvents {}
+
+export const DropZoneContext = createContext<ContextValue<DropZoneProps, HTMLDivElement>>(null);
+
+function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
+  [props, ref] = useContextProps(props, ref, DropZoneContext);
+  let {onPress, onPressStart, onPressEnd, onPressChange, onPressUp} = props;
+  let {dropProps, isDropTarget} = useDrop({...props, ref});
+  let {hoverProps, isHovered} = useHover({});
+  let {focusProps, isFocused, isFocusVisible} = useFocusRing({within: true});
```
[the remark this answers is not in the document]
yihuiliao: it's something worth considering but now that i'm working on splitting out the responsibilities of DropZone into DropZone + FileTrigger per Devon's suggestion, i think we'll wait and see

packages/react-aria-components/stories/index.stories.tsx
yihuiliao: the VO should be very similar to dnd so do you know what drop events that it wasn't annoucning?

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -0,0 +1,57 @@
...
+  // onChange event for the input 
+  handleChange?: React.ChangeEventHandler<HTMLInputElement>
+}
+
+export const FileTriggerContext = createContext<ContextValue<FileTriggerProps, HTMLDivElement>>(null);
+
+function FileTrigger(props: FileTriggerProps, ref: ForwardedRef<HTMLDivElement>) {
+  [props, ref] = useContextProps(props, ref, FileTriggerContext);
+  let {handleChange, accept, className, ...otherProps} = props;
+  let inputRef = useRef<HTMLInputElement>(null);
```
yihuiliao, line 33: do we want to add render props to FileTrigger?

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {Input} from './Input';
+import {LinkContext} from './Link';
+import React, {createContext, ForwardedRef, forwardRef, useRef} from 'react';
+
+export interface FileTriggerProps extends SlotProps, DOMProps, AriaLabelingProps {
+  // want to be able to pass down accept types to the input 
+  accept?: string,
+  // onChange event for the input 
+  handleChange?: React.ChangeEventHandler<HTMLInputElement>
+}
```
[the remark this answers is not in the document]
yihuiliao: yeah i think `multiple` is definitely something we'll want to include. i'm a little bit more on the fence with `capture` tho because on mobile, users will already be given the option to take a photo/video or upload a photo/video which seems to take care of `capture` in a sense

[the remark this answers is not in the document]
reidbarber: I think it's something we should support, so we match the HTML spec. Some applications might want to be opinionated about this, and default to one or the other, so we shouldn't prevent them from that.

packages/react-aria-components/src/DropZone.tsx
yihuiliao: some of this might be outdated now that we've decided to implement dropzone alongside file trigger

packages/react-aria-components/src/FileTrigger.tsx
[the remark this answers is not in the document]
yihuiliao: okay that makes sense. i'll go ahead and add `capture` as a prop

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,132 @@
...
+        {...mergeProps(dropProps, hoverProps, DOMProps, clipboardProps)} 
+        {...renderProps}
+        ref={ref}
+        slot={props.slot}
+        tabIndex={-1}
+        data-hovered={isHovered || undefined}
+        data-focused={isFocused || undefined}
+        data-focus-visible={isFocusVisible || undefined}
+        data-drop-target={isDropTarget || undefined} >
+        {!hasFileTrigger &&
```
devongovett, line 117: I think the button element needs to be non-conditional, and the `dropProps` should go on it rather than the outer div. Right now, this div does not have a role, so when dropping with VoiceOver there the announcement doesn't say it's a button. Also clipboardProps is duplicated right now and should probably only go on the button.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,132 @@
...
+  let renderProps = useRenderProps({ 
+    ...props,
+    values: {isHovered, isFocused, isFocusVisible, isDropTarget},
+    defaultClassName: 'react-aria-DropZone'
+  });
+  let DOMProps = filterDOMProps(props);
+  delete DOMProps.id;
+
+  // passed to the FileTrigger component
+  let onInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
devongovett, line 78: Do we need these to be tied together? Since they control both elements, it seems like a user could do something like this:

```jsx
<DropZone onDrop={onDrop}>
  <FileTrigger onChange={onChange}>
     ...
  </FileTrigger>
</DropZone>
```

Maybe slightly more work but maybe not a big deal?

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -0,0 +1,65 @@
...
+  [props, ref] = useContextProps(props, ref, FileTriggerContext);
+  let {onChange, accept, className, multiple, ...otherProps} = props;
+  let inputRef = useRef<HTMLInputElement>(null);
+
+  return (
+    <Provider
+      values={[
+        [LinkContext, {slot: 'file', onPress: () => inputRef.current?.click()}],  
+        [ButtonContext, {slot: 'file', onPress: () => inputRef.current?.click()}]
+      ]}>
```
devongovett, line 48: We could probably use `PressResponder` here rather than specific slots, that way any pressable element can trigger the input. I also don't think we need the `slot: 'file'` here?

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * Specifies what file types are allowed.
+   */
+  accept?: string,
+  /**
+   * Whether multiple files can be selected.
+   */
+  multiple?: boolean,
+  /**
+   * Handler that is called on submission.
```
devongovett, line 31: To clarify that this is when a user selects a file, not when the whole form is submitted:

```suggestion
   * Handler that is called when a user selects a file.
```

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -75,52 +73,19 @@ function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
...
+        [FileTriggerContext, {ref: fileTriggerRef}],
         [TextContext, {id: textId, slot: 'heading'}]
       ]}>
       <div
         {...mergeProps(dropProps, hoverProps, DOMProps)} 
         {...renderProps}
         ref={ref}
         slot={props.slot} 
-        // tabIndex={-1}
+        // onClick={() => buttonRef.current?.focus()}
```
yihuiliao, line 88: debating whether dropzone should be able to be clicked on in order to paste data. seems like there are may be accessibility issues which i will link [here](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/f0d2ddb65f21278ad29be43fb167a1092287b4b1/docs/rules/no-static-element-interactions.md) and [here](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/f0d2ddb65f21278ad29be43fb167a1092287b4b1/docs/rules/click-events-have-key-events.md). tho dropzone might be a special case because you can still paste using the keyboard.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,108 @@
...
+        {...mergeProps(dropProps, hoverProps, DOMProps)} 
+        {...renderProps}
+        ref={ref}
+        slot={props.slot} 
+        // onClick={() => buttonRef.current?.focus()}
+        data-hovered={isHovered || undefined}
+        data-focused={isFocused || undefined}
+        data-focus-visible={isFocusVisible || undefined}
+        data-drop-target={isDropTarget || undefined} >
+        <VisuallyHidden>
```
yihuiliao, line 93: since this is no longer conditional, if the user does include a FileTrigger, they first will first have to tab to the DropZone before they can tab to the FileTrigger. the reason for having it be conditional before was so that if a user did include a FileTrigger, they could just directly tab to the FileTrigger. not sure which behavior is more appropriate in this case.

packages/@react-aria/dnd/src/useDrop.ts
```diff
@@ -300,9 +312,10 @@ export function useDrop(options: DropOptions): DropResult {
     return allowedOperations[0];
   });
 
-  let {ref} = options;
+  let {ref, buttonRef} = options;
   useLayoutEffect(() => DragManager.registerDropTarget({
```
yihuiliao, line 316: buttonRef was added so that both the button/outer div could act as the drop target (instead of just one or the other) since in the case of dropzone, the user will want to be able to drop over any area of the dropzone rather than just the hidden button which is impossible to access when dropping with the mouse. focus still will be on the button during a drop which leads to the updates in the drag manager where focusElement is the hidden button. overall, there were some changes to the api but none that i believe should affect the other dnd components.

devongovett: do we need a ref to the outer div at all if the button is what gets focus during keyboard drop? does it work to just set `ref` to the button element rather than the div, and remove the extra `buttonRef`/`focusElement`?

yihuiliao: so the issue really isn't so much the keyboard drop as it is with dropping with a mouse. originally i had just set `ref` to the button which worked perfectly fine for the keyboard drop but not with the mouse. in this case, when you tried dropping with a mouse, an error would be thrown because the `e.target` would be the outer div, which was not considered valid drop target since only the button was created with useDrop. however, it would be impossible to target the button if you were only dragging with your mouse. that's why i ended up setting a ref to the outer div as well

reidbarber (dismissed):

packages/react-aria-components/src/DropZone.tsx
LFDanLu: Just tested it, noticed that the visually hidden button is getting focus first hence the need for a second tab. Can we make the visually hidden button have tabIndex=-1 or something to prevent this from happening? Or does the visually hidden button need to be tabbable at all times?

packages/react-aria-components/stories/index.stories.tsx
```diff
@@ -746,3 +748,186 @@ function CustomTab(props) {
...
+    onChange={action('OnChange')} >
+    <Button>Upload</Button>
+  </FileTrigger>
+);
+
+export const FileTriggerLink = (props) => (
+  <FileTrigger 
+    {...props} 
+    onChange={action('OnChange')}
+    multiple >
```
LFDanLu, line 930: Doesn't seem like this story accepts multiple file, looks like the `multiple` isn't being propagated to the `input`

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export interface FileTriggerProps extends SlotProps, DOMProps, AriaLabelingProps {
+  /**
+   * Specifies what file types are allowed.
+   */
+  accept?: string,
+  /**
+   * Whether multiple files can be selected.
+   */
+  allowsMultiple?: boolean,
```
LFDanLu, line 28: ah I see, the previous behavior I noticed is because the prop is `allowsMultiple` instead of `multiple`. Do we wanna just make this `multiple` to bring it in line with `capture` and `accept`? Would be nice to link to the mdn docs as well just like we do for other input props in `TextInputDOMProps`

LFDanLu (commented):
couple other things I noticed when testing

packages/react-aria-components/stories/index.stories.tsx
yihuiliao: oops i forgot to update the story...that's my bad

packages/react-aria-components/src/FileTrigger.tsx
reidbarber: I'm ok with that, for the sake of matching the native API. Maybe `allowsMultiple` could be the RSP prop if we need it to be though.

yihuiliao: i think we discussed in the last otter time that we wanted to have the `multiple` prop have the same naming convention as other boolean props in rsp/rac, hence the allowsMultiple

reidbarber (approved):

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -33,14 +33,14 @@ export interface FileTriggerProps extends SlotProps, DOMProps, AriaLabelingProps
   /**
    * Handler when a user selects a file.
    */
-  onChange?: React.ChangeEventHandler<HTMLInputElement>
+  onChange?: (value: FileList) => void
```
reidbarber, line 36: nit, I think we probably want to call it `files` here
```suggestion
  onChange?: (files: FileList) => void
```

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -52,7 +52,14 @@ function FileTrigger(props: FileTriggerProps, ref: ForwardedRef<HTMLDivElement>)
       <PressResponder onPress={() => inputRef.current?.click()}>
         {children}
       </PressResponder>
-      <Input type="file" style={{display: 'none'}} ref={inputRef} accept={acceptedFileTypes?.toString()} onChange={onChange} capture={defaultCamera} multiple={allowsMultiple} /> 
+      <Input 
+        type="file" 
+        style={{display: 'none'}} 
+        ref={inputRef} 
+        accept={acceptedFileTypes?.toString()} 
+        onChange={(e: ChangeEvent<HTMLInputElement>) => e.target.files && onChange(e.target.files)}
```
reidbarber, line 60: I think this can be shortened to this:

```suggestion
        onChange={(e) => onChange?.(e.target.files)} 
```

and the default `onChange` of `= () => {}` can be removed if so.

yihuiliao:
latest push should have fixed up two things discovered during testing session: (1) labeled the visually hidden button so that it announces DropZone if there is no text (2) onChange should give back the relevant data and not the entire event object

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -52,7 +52,14 @@ function FileTrigger(props: FileTriggerProps, ref: ForwardedRef<HTMLDivElement>)
       <PressResponder onPress={() => inputRef.current?.click()}>
         {children}
       </PressResponder>
-      <Input type="file" style={{display: 'none'}} ref={inputRef} accept={acceptedFileTypes?.toString()} onChange={onChange} capture={defaultCamera} multiple={allowsMultiple} /> 
+      <Input 
+        type="file" 
+        style={{display: 'none'}}
```
reidbarber, line 57: One last nit, for consistency, we can wrap this `Input` in a `VisuallyHidden` instead of applying this style

reidbarber (changes requested):

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -0,0 +1,73 @@
...
+   */
+  allowsMultiple?: boolean,
+  /**
+   * Specifies the use of a media capture mechanism to capture the media on the spot.
+   */
+  defaultCamera?: 'user' | 'environment',
+  /**
+   * Handler when a user selects a file.
+   */
+  onChange?: (files: FileList | null) => void
```
yihuiliao, line 37: opinions on this type? got some lint errors saying `Argument of type 'FileList | null' is not assignable to parameter of type 'FileLists` so this was one solution but there are a couple of alternative solutions that we could try i think

reidbarber (approved):

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -80,19 +80,20 @@ function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
...
+        slot={props.slot}
+        onClick={() => buttonRef.current?.focus()}
         data-hovered={isHovered || undefined}
         data-focused={isFocused || undefined}
         data-focus-visible={isFocusVisible || undefined}
         data-drop-target={isDropTarget || undefined} >
         <VisuallyHidden>
           <button
             {...mergeProps(dropButtonProps, focusProps, clipboardProps, labelProps)}
+            aria-label="DropZone" // will need to update with string formatter
```
LFDanLu, line 96: note, will have to get translations from the globalization team. Also for the team, is this the label we want? Do we want to also allow the user to customize the label here?

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -80,19 +80,20 @@ function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
         [FileTriggerContext, {ref: fileTriggerRef}],
         [TextContext, {id: textId, slot: 'heading'}]
       ]}>
+      {/* eslint-disable-next-line */}
```
LFDanLu, line 83: what was this for?

LFDanLu (approved):

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,109 @@
...
+        slot={props.slot}
+        onClick={() => buttonRef.current?.focus()}
+        data-hovered={isHovered || undefined}
+        data-focused={isFocused || undefined}
+        data-focus-visible={isFocusVisible || undefined}
+        data-drop-target={isDropTarget || undefined} >
+        <VisuallyHidden>
+          <button
+            {...mergeProps(dropButtonProps, focusProps, clipboardProps, labelProps)}
+            aria-label="DropZone" // will need to update with string formatter
```
snowystinger, line 96: need a localized string for this

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,109 @@
...
+    values: {isHovered, isFocused, isFocusVisible, isDropTarget},
+    defaultClassName: 'react-aria-DropZone'
+  });
+  let DOMProps = filterDOMProps(props);
+  delete DOMProps.id;
+  
+  return (
+    <Provider
+      values={[
+        [FileTriggerContext, {ref: fileTriggerRef}],
```
devongovett, line 80: Probably don't need this anymore? I don't see anywhere else that `fileTriggerRef` is used now.

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -0,0 +1,109 @@
...
+   * @selector [data-focus-visible]
+   */
+  isFocusVisible: boolean,
+  /**
+   * Whether the dropzone is the drop target.
+   * @selector [data-drop-target]
+   */
+  isDropTarget: boolean
+}
+// note: possibly add isDisabled prop in the future
```
snowystinger, line 43: delete comment

packages/react-aria-components/src/DropZone.tsx
devongovett: should we do an `aria-labelledby` pointing back to the div so it reads the contents of the drop zone?

devongovett (approved):

packages/react-aria-components/src/DropZone.tsx
yihuiliao: yeah so this was because i added an onClick to dropzone so that when you clicked on it, it would focus on the button allowing you to paste since button has the ClipboardProps. the linting errors had to do with accessibility, wanting to add some sort of key press handler to the dropzone (you can read the errors [here](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/f0d2ddb65f21278ad29be43fb167a1092287b4b1/docs/rules/no-static-element-interactions.md) and [here](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/f0d2ddb65f21278ad29be43fb167a1092287b4b1/docs/rules/click-events-have-key-events.md). but since the dropzone has a hidden button, there is keyboard functionality to paste so i felt like it was a special case in which these could be ignored. if not, i think it's fine we remove onClick to dropzone and have it so you must keyboard paste.

PR #4483, opened 2023-05-05
TableView DnD: improve css specificity

reidbarber:
Fixes cases where drop target styles would potentially get overridden by default styles.

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/table/src/TableView.tsx
+++ b/packages/@react-spectrum/table/src/TableView.tsx
@@ -726,6 +726,7 @@ function TableVirtualizer({layout, collection, focusedKey, renderView, renderWra
                 },
                 classNames(
                   stylesOverrides,
+                  'react-spectrum-Table-body',
                   {
                     'react-spectrum-Table-body--dropTarget': !!isRootDropTarget
                   }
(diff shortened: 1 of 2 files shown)


LFDanLu:
Added another change for folder drop target specificity, using https://codesandbox.io/s/confident-rain-1tb9c9?file=/src/App.js to check for any other issues, but I think your other changes should fix the ones I've noticed so far. Running chromatic

LFDanLu (approved):

ktabors (commented):
Based on the example or two rows in a table with height greater than that, shouldn't we add a DnD example like that? I'm concerned we've missed other things in testing by not having DnD examples like that.

LFDanLu:
@ktabors its wasn't so much a case of having a table with only two rows but rather the order in which the css was loaded in the sandbox that caused the broken drop style issue (aka we were getting lucky locally that the drop target selector loaded in first since its specificity tied with the Table body styles). Unfortunately we can't really test for that in our storybook, at least not that I know of (would be something we can research though) .

snowystinger (approved):

PR #4497, opened 2023-05-09
Implement InlineAlert component #1343

arumsey:
Closes https://github.com/adobe/react-spectrum/issues/1343

- add `inlinealert` CSS
- ~~add `@react-aria/alert`~~
- add `@react-spectrum/inlinealert` component
- add `@react-types/inlinealert`

image

**Usage**

```
  <InLineAlert variant="notice">
    <Header>A title</Header>
    <Content>Additional body content</Content>
  </InLineAlert>
```

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Run `yarn test` locally or checkout http://localhost:9003/?path=/story/inlinealert--default.

## 🧢 Your Project:

RSP


packages/@react-spectrum/inlinealert/stories/InLineAlert.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <Content>{args.content}</Content>
+  </InLineAlert>
+);
+
+export const Default = Template().bind({});
+export const LongContent = Template().bind({});
+LongContent.args = {
+  title: 'Unable to process payment',
+  content: 'There was an error processing your payment. Please check your credit card information is correct, then try again.'
+};
```
snowystinger, line 61:
```suggestion
export const Default = {
  render: (args) => (
      <InLineAlert {...args}>
        <Header>{args.title}</Header>
        <Content>{args.content}</Content>
      </InLineAlert>
  )
};

export const LongContent = {
  ...Default,
  args: {
    title: 'Unable to process payment',
    content: 'There was an error processing your payment. Please check your credit card information is correct, then try again.'
  }
};
```

packages/@react-spectrum/inlinealert/stories/InLineAlert.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+type StoryArgs = SpectrumInLineAlertProps & {title: string, content: string};
+
+const meta: Meta<StoryArgs> = {
+  title: 'InLineAlert',
+  component: InLineAlert,
+  args: {
+    title: 'Title',
+    content: 'Content',
+    variant: 'info'
```
snowystinger, line 30: this isn't required, so better to let the argTypes handle it and delete this line

packages/@react-spectrum/inlinealert/test/InLineAlert.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+import {InlineAlert} from '../';
+import {render} from '@react-spectrum/test-utils';
+import React from 'react';
+
+describe('InlineAlert', function () {
+  it.each`
+    Name | Component      | props
+    ${'InlineAlert'} | ${InlineAlert} | ${{}}
+  `('$Name handles defaults', function ({Component, props}) {
```
snowystinger, line 21: simplifying, we used to use this syntax everywhere for v2 or if there are multiple cases that can be handled by one test, but since we only have one example here, we can do it the less verbose way
```suggestion
  it('InLine Alert handles defaults', function () {
```

packages/@react-types/inlinealert/src/index.d.ts
```diff
@@ -0,0 +1,20 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {DOMProps, StyleProps} from '@react-types/shared';
+import {ReactNode} from 'react';
+
+
+export interface SpectrumInLineAlertProps extends DOMProps, StyleProps {
+  variant?: 'neutral' | 'info' | 'positive' | 'notice' | 'negative',
```
snowystinger, line 18: need to add comments to these for the props table in the docs
```suggestion
   /**
   * The [visual style](https://spectrum.adobe.com/page/in-line-alert/#Options) of the In Line Alert.
   * @default 'neutral'
   */
  variant?: 'neutral' | 'info' | 'positive' | 'notice' | 'negative',
```

snowystinger (commented):
High level comment, I think much like Avatar and StatusLight, the aria for this is so small, that we can just do the React Spectrum level component for now. That'd simplify the PR quite a bit too if we remove the react-aria package. We can always add the hook later if we find that we need to handle more than just applying the role.

snowystinger:
GET_BUILD

reidbarber (changes requested):
Initial review, just a few comments. It looks great so far!

packages/@adobe/spectrum-css-temp/components/inlinealert/index.css
```diff
@@ -0,0 +1,117 @@
...
+}
+
+.spectrum-InLineAlert .spectrum-InLineAlert-grid {
+  display: grid;
+  grid-template-columns: 1fr auto;
+  grid-template-rows: auto auto auto;
+  column-gap: var(--spectrum-inlinealert-spacing-header-to-icon);
+  grid-template-areas:
+    "header      icon       "
+    "content     content    ";
```
snowystinger, line 73: this combination means that the header will always stop "column-gap" away from the containing elements 'inline end'

we'll need to make use of `useHasChild` to determine if we have an icon, then apply a class either to the grid so we can render 
```
grid-template-areas: "header" "content";
```
or so we can change the grid area for `spectrum-InLineAlert-header` to be
```
grid-area: header / icon;
```

packages/@adobe/spectrum-css-temp/components/inlinealert/index.css
```diff
@@ -0,0 +1,117 @@
...
+  .spectrum-InLineAlert {
+    --highcontrast-inlinealert-background-color: Background;
+    --highcontrast-inlinealert-header-color: CanvasText;
+    --highcontrast-inlinealert-content-color: CanvasText;
+    --highcontrast-inlinealert-border-and-icon-color: ButtonBorder;
+  }
+}
+
+.spectrum-InLineAlert-icon {
+  inline-size: var(--spectrum-inlinealert-icon-size);
```
snowystinger, line 87: add grid-area property here instead of inline, same goes for the other classes
```
grid-area: icon;
```

packages/@react-spectrum/inlinealert/src/InLineAlert.tsx
```diff
@@ -0,0 +1,83 @@
...
+    children,
+    variant = 'neutral',
+    ...otherProps
+  } = props;
+
+  let {styleProps} = useStyleProps(otherProps);
+  let domRef = useDOMRef(ref);
+
+  let slots = {
+    icon: {UNSAFE_className: styles['spectrum-InLineAlert-icon'], gridArea: 'icon'},
```
snowystinger, line 47: remove the gridArea from inline on each of these. it's not wrong to do it this way, but we just typically don't do it on the slot and instead handle it through the CSS which is where the grid definition is

packages/@adobe/spectrum-css-temp/components/inlinealert/index.css
snowystinger: Screenshot 2023-05-09 at 2 14 28 PM

packages/@react-spectrum/inlinealert/src/InLineAlert.tsx
```diff
@@ -0,0 +1,83 @@
...
+import {useProviderProps} from '@react-spectrum/provider';
+
+let ICONS = {
+  info: InfoMedium,
+  positive: SuccessMedium,
+  notice: AlertMedium,
+  negative: AlertMedium
+};
+
+function InLineAlert(props: SpectrumInLineAlertProps, ref: DOMRef<HTMLDivElement>) {
```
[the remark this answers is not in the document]
arumsey: I orignally had `InlineAlert` but then noticed `spectrum-css` uses `InLineAlert` since Spectrum calls this component `In-Line Alert` and not `Inline Alert`. Anyway, I'm fine with either but went with consistency.

https://github.com/adobe/spectrum-css/blob/main/components/inlinealert/index.css#L13

packages/@adobe/spectrum-css-temp/components/inlinealert/index.css
```diff
@@ -0,0 +1,117 @@
...
+  font-size: var(--spectrum-inlinealert-heading-font-size);
+  line-height: var(--spectrum-inlinealert-heading-line-height);
+  text-transform: none;
+
+  min-block-size: var(--spectrum-inlinealert-header-min-block-size);
+}
+
+.spectrum-InLineAlert-content {
+  display: block;
+  margin-block-start: var(--spectrum-inlinealert-spacing-header-content-button);
```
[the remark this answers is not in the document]
arumsey: That is the var name `spectrum-css` used since it still supports buttons in the footer. Since RSP implmentation does not allow buttons I can re-name this var to be a better match.

arumsey:
GET_BUILD

packages/@adobe/spectrum-css-temp/components/inlinealert/index.css
snowystinger: that is an interesting idea, neutral is the only one without an icon, we don't let anyone else determine that. I think this would be fine

packages/@adobe/spectrum-css-temp/components/inlinealert/index.css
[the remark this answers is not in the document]
snowystinger: if that's the only thing in spectrum-css that gives the correct margin there, maybe we keep it then... just to make future reconciling marginally easier :-/

packages/@react-spectrum/inlinealert/src/InLineAlert.tsx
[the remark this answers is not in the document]
reidbarber: After talking with the team, it looks like `InlineAlert` is preferable for now. Thanks!

reidbarber (commented):
Looks good! Only some small formatting things.

packages/@react-spectrum/inlinealert/src/InLineAlert.tsx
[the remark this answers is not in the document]
arumsey: Does this include renaming ALL the spectrum-css class names or just the React names?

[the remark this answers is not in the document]
snowystinger: can leave the classnames alone, they should just match whatever spectrum-css has

packages/@react-spectrum/inlinealert/chromatic/InlineAlert.chromatic.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {Content, Header} from '@react-spectrum/view';
+import {InLineAlert} from '../';
```
reidbarber, line 14: Looks like this needs to be updated for the lint to pass

```suggestion
import {InlineAlert} from '../';
```

reidbarber:
GET_BUILD

snowystinger:
GET_BUILD

reidbarber (approved):
LGTM, thanks!

snowystinger:
DNM until after release

packages/@react-spectrum/inlinealert/intl/en-US.json
```diff
@@ -0,0 +1,7 @@
+{
+  "neutral": "Neutral",
+  "info": "Informative",
+  "positive": "Positive",
+  "notice": "Notice",
+  "negative": "Negative"
+}
```
devongovett, line 7: I think we should be able to copy most of the translations from `@react-spectrum/alert` too

devongovett:
I guess we should probably delete the old `@react-spectrum/alert` package and associated css?

packages/@react-spectrum/inlinealert/intl/en-US.json
reidbarber: I can handle both of these in a follow-up

PR #4518, opened 2023-05-12
Fix RAC collections in React canary versions

### LINKED ISSUE #4390: Table isRowHeader error (Next.js 13.3 + react-aria-components)
# 🐛 Bug Report 

Rendering a `Table` (w/ `TableHeader`, `Column` children etc.) throws the following error:

`Error: A table must have at least one Column with the isRowHeader prop set to true`

The error is thrown even if the prop is set on a Column.

- Strict mode is disabled from `next.config.js`
- `SSRProvider` is used

I can only reproduce the issue in Next.js 13.3.0. Works fine in 13.2.4.

## 💻 Code Sample

Next.js 13.3.0 - broken:
https://codesandbox.io/p/sandbox/react-aria-next-js-13-3-0-broken-table-071o7w?file=%2Fapp%2Fpage.tsx

Next.js 13.2.4 (otherwise identical) - works:
https://codesandbox.io/p/sandbox/react-aria-next-js-13-2-4-working-table-t8keyk?file=%2Fapp%2Fpage.tsx

devongovett:
Fixes #4390

This runs our tests against the new [React Canary](https://react.dev/blog/2023/05/03/react-canaries) versions in CI. Next.js is using the canary version in their latest version, and this is now an officially supported setup by the React team.

As reported in #4390, the React Aria Components collection implementation was broken in the canary versions of React because our hack of using the `multiple` prop to pass through an object to the underlying fake DOM node no longer works. This is fixed by using a callback ref to imperatively set the props ourselves. This seems cleaner anyway and also fixes the issue in Next.js.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,6 +108,26 @@ jobs:
             - ~/react-spectrum
           key: react-spectrum17-{{ .Environment.CACHE_VERSION }}-{{ .Environment.CIRCLE_SHA1 }}
 
+  install-canary:
+    executor: rsp-large
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - rsp-yarn-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ checksum "yarn.lock" }}
+            - rsp-yarn-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-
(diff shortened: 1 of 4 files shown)


reidbarber (approved):
LGTM, nice to have this tested and good fix for RAC

snowystinger (approved):
Nice, managed it without the useLayoutEffect

Issue #3970, opened 2023-01-27
[useOverlay] Click Outside in Shadow-DOM context


tnovas:
The callback `shouldCloseOnInteractOutside` in [useOverlay](https://github.com/adobe/react-spectrum/blob/f32367ad4a4639ccb8738b9daefe42d0d870affd/packages/%40react-aria/overlays/src/useOverlay.ts#L95) sends `event.target`, in Shadow-DOM context the `event.target` is Shadow-DOM and not the real point click, so, in this context the result is always click outside.
Could be changed to sends the `event` and not `event.target`?
This way we have more information like `event.composedPath()` and we check the real point click.

I look forward to your feedback.
Regards!

LFDanLu:
Closing as dupe

theomessin:
@LFDanLu instead of closing every Shadow DOM related ticket in favour of #1472 I think it might be best work incrementally and tackle issues one-by-one. #1472 has been around for 2 years and no progress has been made. #4526 addresses this particular issue.

PR #4526, opened 2023-05-15
Add Shadow DOM support for useInteractOutside

theomessin:
Closes https://github.com/adobe/react-spectrum/issues/3970
Partial work on https://github.com/adobe/react-spectrum/issues/1472

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Run the tests. The previous implementation fails, the suggested one passes.

## 🧢 Your Project:

https://triptease.com/
https://github.com/triptease/


packages/@react-aria/interactions/src/useInteractOutside.ts
```diff
@@ -123,5 +123,15 @@ function isValidEvent(event, ref) {
...
+  }
+
+  const refWithChildren = [ref.current, ...Array.from(ref.current.childNodes)];
+  return intersection(refWithChildren, event.composedPath()).length === 0;
+}
+
+// TODO: replace this with lodash.intersection
+function intersection(a: any[], b:any[]) {
+  return a.filter(value => b.includes(value));
 }
```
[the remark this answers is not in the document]
snowystinger: lodash intersection requires a ton of code compared to what you've written here. I think this is probably fine and could be shortened to

```
function hasIntersection(a, b) {
  a.some(value => b.includes(value));
}
```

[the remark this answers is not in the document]
theomessin: Yeah that was silly 😆 I've gone with your implementation

packages/@react-aria/interactions/src/useInteractOutside.ts
```diff
@@ -123,5 +123,14 @@ function isValidEvent(event, ref) {
     }
   }
 
-  return ref.current && !ref.current.contains(event.target);
+  if (!ref.current) {
+    return false;
+  }
+
+  const refWithChildren = [ref.current, ...Array.from(ref.current.childNodes)];
```
snowystinger, line 130: apologies for my lack of knowledge here, I'm sure I'm just overlooking something, but why do we need the childNodes?

packages/@react-aria/interactions/test/useInteractOutside.test.js
```diff
@@ -205,4 +207,55 @@ describe('useInteractOutside', function () {
...
+      const shadowDom = document.getElementById('shadowHost').shadowRoot;
+      const inside = shadowDom.querySelector('#inside');
+      fireEvent.mouseDown(inside);
+      fireEvent.mouseUp(inside);
+      expect(onInteractOutside).not.toHaveBeenCalled();
+
+      const outside = shadowDom.querySelector('#outside');
+      fireEvent.mouseDown(outside);
+      fireEvent.mouseUp(outside);
+      expect(onInteractOutside).toHaveBeenCalledTimes(1);
```
snowystinger, line 258: So this is an interesting test, I didn't realize initially that there were two react trees, one inside the shadow dom which we're interested in and the other is outside. In addition, inside/outside on the target divs confused me with the names, because i originally thought it was trying to reference inside and outside the shadow dom. I think as a result, to make this more manageable to those with less shadow dom experience, we should rename those to something a little more real life, maybe `#popover` and `#outside-popover`? I'm open to suggestions.

I think it might be good to add a button outside as a sibling to `Example` and make sure that calls the useInteractOutside as well. What do you think?

packages/@react-aria/interactions/src/useInteractOutside.ts
theomessin: This is actually a good question. I was just using the old `ref.current.contains` (which checks childNodes as well). However, since we're now using `target.composedPath`, if a child of the `ref.current` was clicked, that _should_ be present in the composedPath. I think we can actually just use `ref.current`

packages/@react-aria/interactions/test/useInteractOutside.test.js
theomessin: @snowystinger thanks for your suggestions. I've improved the implementation by just using `ref.current` and rewrote the tests to aid readability. I've also added one to check that clicking a button outside the shadow root triggers.

snowystinger (approved):
thanks! looks good to me

snowystinger:
Closing for the moment: https://github.com/adobe/react-spectrum/issues/1472#issuecomment-1562081880

PR #4538, opened 2023-05-16
Add support for refs to collection components

devongovett:


snowystinger (approved):
Looks good, tests could be further improved by matching the ref to an element retrieved from `getByRole` instead of just checking that they are HTML elements. These will be API now, so we can't easily go changing what element they actually point to.

ktabors (dismissed):

snowystinger (approved):

PR #4610, opened 2023-05-30
InlineAlert: Fix background color and clean up stories

reidbarber:
- Fixes background to match Spectrum designs
- Removes extra stories
- Adds dynamic story, and adds to Form story
- Updates some 'In-line' in descriptions to 'Inline' to match docs

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/form/stories/Form.stories.tsx
+++ b/packages/@react-spectrum/form/stories/Form.stories.tsx
@@ -16,12 +16,13 @@ import {ButtonGroup} from '@react-spectrum/buttongroup';
 import {chain} from '@react-aria/utils';
 import {Checkbox, CheckboxGroup} from '@react-spectrum/checkbox';
 import {ComboBox} from '@react-spectrum/combobox';
-import {Content} from '@react-spectrum/view';
+import {Content, Header} from '@react-spectrum/view';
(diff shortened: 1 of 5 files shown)


snowystinger (approved):

packages/@react-spectrum/form/stories/Form.stories.tsx
```diff
@@ -689,6 +690,12 @@ function FormWithSubmit() {
 
   return (
     <Form onSubmit={handleSubmit} isReadOnly={formStatus === 'valid'}>
+      {(formStatus === 'invalid' || formStatus === 'valid') &&
+        <InlineAlert variant={formStatus === 'invalid' ? 'negative' : 'positive'}>
+          <Header>{formStatus === 'invalid' ? 'Error' : 'Success'}</Header>
```
ktabors, line 695: Only "concern" is that when the Header text is the same as the variant string it says the same word twice. That is an issue for the developing team... wondering if we'll get complaints.

ktabors (approved):

packages/@react-spectrum/form/stories/Form.stories.tsx
reidbarber: Yeah, I think apps should hopefully use more specific titles.

PR #4615, opened 2023-05-31
Trigger production docs deploy manually

devongovett:
This changes the production docs deploy to run when we publish a release (from Github releases), or by triggering a Github Action manually, which kicks off a circle ci build. This way we can merge documentation changes without them going to prod immediately.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,17 @@
 version: 2.1
+parameters:
+  GHA_Event:
+    type: string
+    default: ""
+  GHA_Actor:
+    type: string
+    default: ""
+  GHA_Action:
+    type: string
+    default: ""
+  GHA_Meta:
+    type: string
+    default: ""
 
 orbs:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,4 +1,17 @@
 version: 2.1
+parameters:
+  GHA_Event:
+    type: string
+    default: ""
+  GHA_Actor:
```
LFDanLu, line 6: Some of these aren't used, what are they for exactly?

snowystinger (approved):

LFDanLu (approved):

.circleci/config.yml
devongovett: They are set by the circle ci plugin. The [article](https://circleci.com/blog/trigger-circleci-pipeline-github-action/) implies that they are required, but not entirely sure. Just added the ones they listed to be safe.

PR #4616, opened 2023-05-31
RAC: Expose additional states

### LINKED ISSUE #4429: React Aria Components: Expose more states on wrapper elements

reidbarber:
Closes #4429

We should clarify what to expose as render props in React Aria Components. 

Current plan:

**What to expose:**
- State object from state hook
  - For Combobox, Select, and Tabs, we expose a subset of the object to avoid an infinite re-render issue.
- Read-only values from behavior hook
  - Example: isPressed
- Focus/hover states
  - Examples: isFocused, isHovered, isFocusVisible
  - For wrapper components, use isFocusWithin instead of isFocused
  - Non-interactive components (i.e. wrappers), won't support hover states 
- DnD states (if supported by component)
  - Examples: isDropTarget, isDragging, allowsDragging
- Entrance/exit animation states (if supported by component)
  - Examples: isEntering, isExiting

## ✅ Pull Request Checklist:


packages/react-aria-components/src/Calendar.tsx
```diff
@@ -21,15 +21,19 @@ import {HeadingContext} from './Heading';
 import React from 'react';
 import {TextContext} from './Text';
 
-export interface CalendarProps<T extends DateValue> extends Omit<BaseCalendarProps<T>, 'errorMessage'>, RenderProps<CalendarState>, SlotProps {
+export interface CalendarRenderProps extends Omit<CalendarState, 'setValue' | 'setFocusedDate' | 'focusNextDay' | 'focusPreviousDay' | 'focusNextRow' | 'focusPreviousRow' | 'focusNextPage' | 'focusPreviousPage' | 'focusSectionStart' | 'focusSectionEnd' | 'focusNextSection' | 'focusPreviousSection' | 'selectFocusedDate' | 'selectDate' | 'setFocused'> {}
```
snowystinger, line 24: instead of all these omits, which if we ever add to the calendar state something we don't want RAC renderProps to pick up, we'd need to add here, we could instead create a more base type that CalendarState extends

naming is the hard part, haha...

snowystinger: I don't follow the "but ..." it sounds like the same thing we just agreed on?

reidbarber: I wasn't following the part about creating a base type that CalendarState extends

packages/react-aria-components/src/Breadcrumbs.tsx
```diff
@@ -9,16 +9,39 @@
...
+   */
+  isFocusWithin: boolean,
+  /**
+   * Whether an element within the breadcrumbs is keyboard focused.
+   * @selector [data-focus-visible]
+   */
+  isFocusVisible: boolean
+}
+
+export interface BreadcrumbsProps<T> extends Omit<CollectionProps<T>, 'disabledKeys' | 'children'>, Omit<AriaBreadcrumbsProps, 'children'>, SlotProps, RenderProps<BreadcrumbsRenderProps> {
```
LFDanLu, line 44: Should be the following right? 
```suggestion
export interface BreadcrumbsProps<T> extends Omit<CollectionProps<T>, 'disabledKeys'>, Omit<AriaBreadcrumbsProps, 'children'>, SlotProps, RenderProps<BreadcrumbsRenderProps> {
```

packages/react-aria-components/src/Slider.tsx
```diff
@@ -63,7 +80,13 @@ function Slider<T extends number | number[]>(props: SliderProps<T>, ref: Forward
 
   let renderProps = useRenderProps({
     ...props,
-    values: state,
+    values: {
+      orientation: state.orientation,
+      isDisabled: state.isDisabled,
+      isFocused,
+      isFocusVisible,
+      state
```
LFDanLu, line 88: should this state (and the states in other components) be spread instead?

LFDanLu (changes requested):

packages/react-aria-components/src/Slider.tsx
reidbarber: I think we decided to include the state object during a discussion in slack. I think this also gives us a pretty clear delimiter for what we should include as a data attribute.

LFDanLu: is it weird though to have `isDisabled: state.isDisabled` but then have the same information available in the state object separately? Might be confusing to the user as to if there is a difference between `{({state, isDisabled}) => use state.isDisabled or isDisabled`.

LFDanLu: I'm fine with this if we are treating the `state` object as a special additional info source, we should make that clear in the docs if so. Also looks like the Slider docs examples will need to be updated to extract state from the set of renderProps

packages/react-aria-components/src/Breadcrumbs.tsx
reidbarber: I think we want to omit `children` since it isn't compatible with `children` in `RenderProps`, and that is one preferred.

packages/react-aria-components/src/Select.tsx
```diff
@@ -28,6 +28,11 @@ export interface SelectRenderProps {
    * @selector [data-focused]
    */
   isFocused: boolean,
+  /**
+   * Whether the select is keyboard focused.
+   * @selector [data-focus-visible]
+   */
+  isFocusVisible: boolean,
```
devongovett, line 35: disabled?

packages/react-aria-components/src/Slider.tsx
```diff
@@ -116,15 +146,21 @@ function SliderOutput({children, style, className, ...otherProps}: SliderOutputP
...
   let domRef = mergeRefs(ref, trackRef);
   let renderProps = useRenderProps({
     ...props,
     defaultClassName: 'react-aria-SliderTrack',
-    values: state
+    values: {
+      orientation: state.orientation,
+      isDisabled: state.isDisabled,
+      isFocused,
+      isFocusVisible,
```
devongovett, line 161: hover might be useful on the track?

packages/react-aria-components/src/Tabs.tsx
```diff
@@ -35,7 +35,11 @@ export interface TabListRenderProps {
    * The orientation of the tab list.
    * @selector [aria-orientation="horizontal | vertical"]
    */
-  orientation: Orientation
+  orientation: Orientation,
+  /**
+   * State of the tab list.
+   */
+  state: TabListState<unknown>
```
devongovett, line 42: only in the tab list not the top-level Tabs? also wonder which interactive states we want (e.g. focus within?).

packages/react-aria-components/src/TextField.tsx
```diff
@@ -10,15 +10,38 @@
...
+   * Whether the text field is currently hovered with a mouse.
+   * @selector [data-hovered]
+   */
+  isHovered: boolean,
+  /**
+   * Whether the text field is disabled.
+   * @selector [data-disabled]
+   */
+  isDisabled: boolean
+}
```
devongovett, line 42: validation state?

packages/react-aria-components/src/ToggleButton.tsx
```diff
@@ -48,7 +52,10 @@ function ToggleButton(props: ToggleButtonProps, ref: ForwardedRef<HTMLButtonElem
       slot={props.slot}
       data-pressed={isPressed || undefined}
       data-hovered={isHovered || undefined}
-      data-focus-visible={isFocusVisible || undefined} />
+      data-focused={isFocused || undefined}
+      data-focus-visible={isFocusVisible || undefined}
+      data-selected={state.isSelected || undefined}
+      data-disabled={props.isDisabled || undefined} />
```
devongovett, line 58: I think button documents these as `:focus` and `:disabled` rather than data attrs, and `isSelected` is documented using `aria-pressed` above.

devongovett (commented):
I think we need to have some more consistent rules about which states are included in which components. Maybe a team discussion.

packages/react-aria-components/src/Calendar.tsx
```diff
@@ -12,23 +12,62 @@
...
 import React, {createContext, ForwardedRef, forwardRef, ReactElement, useContext} from 'react';
 import {TextContext} from './Text';
 
-export interface CalendarProps<T extends DateValue> extends Omit<BaseCalendarProps<T>, 'errorMessage'>, RenderProps<CalendarState>, SlotProps {
+export interface CalendarRenderProps {
+  /**
+   * Whether the calendar is currently hovered with a mouse.
+   * @selector [data-hovered]
+   */
+  isHovered: boolean,
```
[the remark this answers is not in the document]
reidbarber: I think so. I think people still may want to style it based on hover, even though it isn't interactive. Probably not too common, but better to support it than not.

packages/react-aria-components/src/NumberField.tsx
```diff
@@ -10,18 +10,45 @@
...
 import React, {createContext, ForwardedRef, forwardRef, useRef} from 'react';
 import {TextContext} from './Text';
 
-export interface NumberFieldProps extends Omit<AriaNumberFieldProps, 'label' | 'placeholder' | 'description' | 'errorMessage'>, RenderProps<NumberFieldState>, SlotProps {}
+export interface NumberFieldRenderProps {
+  /**
+   * Whether the number field is focused, either via a mouse or keyboard.
+   * @selector [data-focused]
+   */
+  isFocused: boolean,
```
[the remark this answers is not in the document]
reidbarber: Why within? Isn't the input essentially the only focusable element within?

packages/react-aria-components/src/RadioGroup.tsx
```diff
@@ -46,55 +46,59 @@ export interface RadioGroupRenderProps {
    * The validation state of the radio group.
    * @selector [aria-invalid]
    */
-  validationState: ValidationState | null
+  validationState: ValidationState | null,
+  /**
+   * State of the radio group.
+   */
+  state: RadioGroupState
```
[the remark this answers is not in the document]
reidbarber: I think so.

packages/react-aria-components/src/TextField.tsx
reidbarber: The hook uses `aria-invalid`, so I suppose we could derive it from that, but it doesn't return a validationState explicitly.

packages/react-aria-components/src/DateField.tsx
```diff
@@ -64,7 +104,7 @@ function DateField<T extends DateValue>(props: DateFieldProps<T>, ref: Forwarded
           }
         }]
       ]}>
-      <div {...DOMProps} {...renderProps} ref={ref} slot={props.slot} />
+      <div {...mergeProps(hoverProps, focusProps)} {...DOMProps} {...renderProps} ref={ref} slot={props.slot} data-validation-date={state.validationState || undefined} />
```
ktabors, line 107: Why isn't this multiple lines for readability?

packages/react-aria-components/src/Select.tsx
```diff
@@ -60,8 +70,10 @@ function Select<T extends object>(props: SelectProps<T>, ref: ForwardedRef<HTMLD
     children: undefined
   });
 
+  let {isFocusVisible, focusProps} = useFocusRing({within: true});
+
   // Only expose a subset of state to renderProps function to avoid infinite render loop
-  let renderPropsState = useMemo(() => ({isOpen: state.isOpen, isFocused: state.isFocused}), [state.isOpen, state.isFocused]);
+  let renderPropsState = useMemo(() => ({isOpen: state.isOpen, isFocused: state.isFocused, isFocusVisible, isDisabled: props.isDisabled || false}), [isFocusVisible, props.isDisabled, state.isFocused, state.isOpen]);
```
ktabors, line 76: also hard to read

ktabors (approved):

packages/react-aria-components/src/Calendar.tsx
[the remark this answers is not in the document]
snowystinger: I think the cells being hover-able is enough.
All the interactive elements in a calendar would have their own isHovered, so I don't think we need it at the top.
I'd like a use case for supporting it at the top

similar comment for focuswithin, focusvisible

not sure about disabled, might be useful for disabling any extra interactive components people add? but also, it always comes from outside, it's not some internal state we compute, so i'd argue against including it here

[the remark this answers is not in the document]
snowystinger: same feeling for the other places devon commented
CheckboxGroup, not sure when you'd need to know at the group level that focus is in the group, it's already on one of the interactive elements

packages/react-aria-components/src/ComboBox.tsx
```diff
@@ -28,6 +28,16 @@ export interface ComboBoxRenderProps {
...
    */
   isFocused: boolean,
+  /**
+   * Whether the combobox is keyboard focused.
+   * @selector [data-focus-visible]
+   */
+  isFocusVisible: boolean,
+  /**
+   * Whether the combobox is currently hovered with a mouse.
+   * @selector [data-hovered]
```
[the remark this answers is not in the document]
snowystinger: I think it's fine to have hovered show up for the label, click the label sends focus to the input, so it's interactive in a way

But if we're going to have isHovered here, then it should be on the Select and Slider as well

packages/react-aria-components/src/ComboBox.tsx
```diff
@@ -136,9 +154,13 @@ function ComboBox<T extends object>(props: ComboBoxProps<T>, ref: ForwardedRef<H
       <div
         {...DOMProps}
         {...renderProps}
+        {...focusProps}
+        {...hoverProps}
```
snowystinger, line 158: don't forget to merge these

packages/react-aria-components/src/NumberField.tsx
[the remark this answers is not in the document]
snowystinger: I think exposing it as isFocused is fine

packages/react-aria-components/src/SearchField.tsx
```diff
@@ -54,10 +92,16 @@ function SearchField(props: SearchFieldProps, ref: ForwardedRef<HTMLDivElement>)
   return (
     <div
       {...DOMProps}
+      {...focusProps}
+      {...hoverProps}
```
snowystinger, line 96: merge

packages/react-aria-components/src/Tabs.tsx
snowystinger: right anything were the collection is being sent from a child up to the parent, it'll be an issue
can we just remove the problematic data from state such that we can expose the rest?

packages/react-aria-components/src/TextField.tsx
```diff
@@ -31,13 +54,26 @@ function TextField(props: TextFieldProps, ref: ForwardedRef<HTMLDivElement>) {
...
+    ...props,
+    values: {isHovered, isFocused, isFocusVisible, isDisabled: props.isDisabled || false},
+    defaultClassName: 'react-aria-TextField'
+  });
+
   return (
     <div
       {...filterDOMProps(props)}
+      {...focusProps}
+      {...hoverProps}
```
snowystinger, line 69: merge

packages/react-aria-components/src/DateField.tsx
```diff
@@ -12,14 +12,45 @@
...
 import React, {cloneElement, createContext, ForwardedRef, forwardRef, HTMLAttributes, ReactElement, useContext, useRef} from 'react';
 import {TextContext} from './Text';
 
-export interface DateFieldProps<T extends DateValue> extends Omit<AriaDateFieldProps<T>, 'label' | 'description' | 'errorMessage'>, RenderProps<DateFieldState>, SlotProps {}
-export interface TimeFieldProps<T extends TimeValue> extends Omit<AriaTimeFieldProps<T>, 'label' | 'description' | 'errorMessage'>, RenderProps<DateFieldState>, SlotProps {}
+export interface DateFieldRenderProps {
+  /**
+   * State of the date field.
+   */
+  state: DateFieldState,
```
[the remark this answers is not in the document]
reidbarber: Good catch

ktabors (approved):

packages/react-aria-components/src/Calendar.tsx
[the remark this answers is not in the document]
reidbarber: We have decided to remove isHover state for non-interactive elements.

packages/react-aria-components/src/DateField.tsx
```diff
@@ -109,7 +155,7 @@ function TimeField<T extends TimeValue>(props: TimeFieldProps<T>, ref: Forwarded
           }
         }]
       ]}>
-      <div {...DOMProps} {...renderProps} ref={ref} slot={props.slot} />
+      <div {...focusProps} {...DOMProps} {...renderProps} ref={ref} slot={props.slot} data-validation-date={state.validationState || undefined} />
```
ktabors, line 158: Why isn't this multiple lines?

ktabors (dismissed):

packages/react-aria-components/src/Calendar.tsx
```diff
@@ -49,16 +83,31 @@ function Calendar<T extends DateValue>(props: CalendarProps<T>, ref: ForwardedRe
...
 
   return (
-    <div {...renderProps} {...calendarProps} ref={ref} slot={props.slot}>
+    <div
+      {...focusProps}
+      {...renderProps}
+      {...calendarProps}
+      ref={ref}
+      slot={props.slot}
+      data-focus-visible={isFocused || undefined}
```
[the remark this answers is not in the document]
reidbarber: Looks like there were two in that file. Did a search and didn't see any others.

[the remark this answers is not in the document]
snowystinger: looks like that was all of them 👍🏻

packages/react-aria-components/src/Tabs.tsx
```diff
@@ -100,7 +105,8 @@ export interface TabPanelRenderProps {
    * `shouldForceMount` prop is true, and the corresponding tab is not selected.
    * @selector [inert]
    */
-  isInert: boolean
+  isInert: boolean,
+  state: TabListState<unknown>
```
snowystinger, line 109: can we do something like
```
  state: Omit<TabListState<unknown>, 'collection'>
```

snowystinger: or, since that's in reverse of how we construct the new state object, we could also do something like
```
interface SubsetState<T = unknown> {
  disabledKeys: TabListState<T>.disabledKeys
  ... repeat, same way we construct the actual object
}
```

reidbarber: The state used for TabPanel and TabList is the entire state, so I don't think we need to change the type here.

snowystinger: Maybe it just needs a comment? I didn't realize this wasn't a subset
Screenshot 2023-06-28 at 2 04 32 PM

snowystinger (approved):
Otherwise looks good

packages/react-aria-components/src/Tabs.tsx
reidbarber: Thanks, looks like I was missing comments for those. Just added.

ktabors (approved):

snowystinger (approved):

PR #4635, opened 2023-06-05
Upgrade to TypeScript 5

devongovett:
Upgrades to Typescript 5.0.4. 5.1 appears to have one issue breaking the lodash types, which are for some reason loaded (storybook I think). https://github.com/microsoft/TypeScript/issues/53514 https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/65674

### DIFF
--- a/examples/rsp-next-ts/pages/_app.tsx
+++ b/examples/rsp-next-ts/pages/_app.tsx
@@ -44,7 +44,7 @@ function MyApp({ Component, pageProps }: AppProps) {
             </ActionButton>
           </Flex>
           <View>
-            {/* @ts-ignore */}
+            {/* @ts-ignore - https://github.com/vercel/next.js/issues/49610 */}
             <Component {...pageProps} />
           </View>
         </Grid>
--- a/examples/rsp-cra-18/package.json
+++ b/examples/rsp-cra-18/package.json
@@ -17,7 +17,7 @@
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
     "react-scripts": "5.0.1",
-    "typescript": "^4.7.3",
+    "typescript": "5.0.4",
     "web-vitals": "^2.1.4"
   },
   "scripts": {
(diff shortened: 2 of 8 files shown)


examples/rsp-next-ts/pages/_app.tsx
```diff
@@ -44,7 +44,7 @@ function MyApp({ Component, pageProps }: AppProps) {
             </ActionButton>
           </Flex>
           <View>
-            {/* @ts-ignore */}
+            {/* @ts-ignore - https://github.com/vercel/next.js/issues/49610 */}
```
LFDanLu, line 47: Just a heads up but apparently bumping to [TS5.1.3](https://github.com/vercel/next.js/issues/49610#issuecomment-1576936320) fixed this issue. Fine with keeping this ignore for now anyways since it isn't a RSP issue

LFDanLu (approved):

snowystinger (commented):
I think we can remove our override for ResizeObserver as well?
https://github.com/adobe/react-spectrum/blob/main/packages/%40react-spectrum/utils/src/ResizeObserver.d.ts

examples/rsp-next-ts/pages/_app.tsx
devongovett: Yeah but then we have the lodash issue I mentioned in the description 🙃

snowystinger (dismissed):

examples/rsp-next-ts/pages/_app.tsx
LFDanLu: Ahhh, so we want the project's TS version to exactly match the test app's TS version?

devongovett: Ah I can try upgrading just the example to the latest yeah

LFDanLu (approved):

snowystinger (approved):

PR #4639, opened 2023-06-06
fix(#4637): Date picker unusable with screen reader (NVDA)

### LINKED ISSUE #4637: Date picker unusable with screen reader (NVDA)
# 🐛 Bug Report 

From the initial focus of the date picker field there is a large amount of repeated reading from the screen reader that its easy to get lost in the 'noise'. When i go to focus and open the calendar, the reader then proceeds to read out every single day in the calendar. 

## 🤔 Expected Behavior

When navigating the date picker with a screen reader and keyboard, the information conveyed shouldn't get lost in the constant repeating of previous labels or headings and when the calendar is in focus, the reader doesn't begin to read out each and every from the initial focused day.

## 🔦 Context

Ive been using the screen reader software NVDA while ive been developing my design system with react aria to ensure maximum WAI ARIA compliancy. However in this instance, navigating the date picker is problematic.

## 💻 Code Sample

The codesandbox link provided by the docs also has this issue. Download NVDA, then begin to tab through the datepicker and open the calendar. [docs link here](https://codesandbox.io/s/white-breeze-rer8g2?file=/src/Calendar.js)

## 🌍 Your Environment

| react-aria   | ^3.25.0
| @react-aria/calendar | ^3.3.0
| @react-aria/datepicker | ^3.4.1
| @internationalized/date | ^3.2.0
| Browser          | Chrome 114.0.5735.90

majornista:
The issue is that NVDA shifts out forms mode when focus lands on the button within a cell within the calendar table, and as such, after announcing the focused element, starts to read the content of the dialog from that point forward. I think the easiest solution for this would be to use `role="application"` instead of `group` for the container `calendarProps` returned by the `useCalendarBase` hook here:  https://github.com/adobe/react-spectrum/blob/7f63e933e61f20891b4cf3f447ab817f918cb263/packages/%40react-aria/calendar/src/useCalendarBase.ts#L94-L97.

Closes #4637 

## ✅ Pull Request Checklist:

- [x] Included link to corresponding [React Spectrum GitHub Issue 4637](https://github.com/adobe/react-spectrum/issue/4637).

## 📝 Test Instructions:
1. Open https://reactspectrum.blob.core.windows.net/reactspectrum/63ddc5ee08c4e077827ae37bb2323c3c8651850b/docs/react-spectrum/DatePicker.html#example on a Windows machine using NVDA.
2. Navigate to and activate the Calendar dropdown menu button for the DatePicker example labeled "Event date".
3. The Calendar popover dialog will open with focus on the current date. NVDA will be in forms mode and additional dates in the Calendar table after the current should not announce.
4. You should be able to navigate between dates in the Calendar grid using ArrowUp/ArrowDown/ArrowLeft/ArrowRight/Home/End/PageUp/PageDown, and NVDA will stay in forms mode.
5. Closing the Calendar using the Esc key to cancel or Enter to select a date, should restore focus to the Calendar button, and NVDA should exit forms mode.

## 🧢 Your Project:

Adobe/Accessibility


devongovett:
Not sure if this is a good idea to do _only_ for Calendar. The same "problem" exists in all grids with NVDA, for example in ListView and TableView. And I remember having a discussion about using role=application for drag and drop to solve the same problem as well, and deciding not to do it. Do you remember the reasoning there? Why should we treat calendar differently? An NVDA user must be used to needing to manually switch into forms mode when navigating to a button as this would be a very common experience. Just wondering if it is something we should really work around or not.

majornista:
> Not sure if this is a good idea to do _only_ for Calendar. The same "problem" exists in all grids with NVDA, for example in ListView and TableView. And I remember having a discussion about using role=application for drag and drop to solve the same problem as well, and deciding not to do it. Do you remember the reasoning there? Why should we treat calendar differently? An NVDA user must be used to needing to manually switch into forms mode when navigating to a button as this would be a very common experience. Just wondering if it is something we should really work around or not.

Calendar is a controlled environment, where each cell contains the same content, a button for a date, and where we want the  user to navigate using the arrow keys. A ListView or TableView with a button in a cell, might have other content in the cell that a user would need to access in browse mode or using table navigation. With regard to drag and drop, we considered trying to force the user into forms mode when a dragging operation was initiated, but I think we found it didn't improve behavior enough to justify the change.

devongovett:
Ok that makes sense. I imagine this issue came up when testing calendar before but I don't remember if we discussed it. Let's test it on various devices/AT and see how it works.

devongovett:
@majornista we tested this and it seemed to mostly work ok. The only thing I found was that in Firefox + JAWS, the entire contents of the popover (all dates in the grid) are read out when it is opened. It worked in Firefox + NVDA, and Chrome + JAWS/NVDA though, so better than before, but not sure if there is anything else we should do to get JAWS + Firefox working as well.

devongovett:
I'll test again. Maybe my jaws is out of date.

> I think we should only include this visually hidden heading on mobile:

Hmm maybe but let's leave that as a separate issue for now. Really want to minimize the amount of these specific browser checks.

majornista:
> Really want to minimize the amount of these specific browser checks.
Except, the VisuallyHidden Month/Year heading was only added for mobile.

devongovett:
Hmm, not only mobile. It's better for any touch screen reader user, even on windows. That's why these kind of checks are hard.

I re-tested in JAWS 2023.2303.144 (latest version downloaded today) and I don't see it reading all of the dates in the calendar anymore. I also didn't see your issue with repeating the month 4 times.

devongovett (approved):

snowystinger (approved):

PR #4641, opened 2023-06-06
Expandable rows TableView

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -10,221 +10,21 @@
  * governing permissions and limitations under the License.
  */
 
-import {AriaLabelingProps, DOMProps, DOMRef, DropTarget, FocusableElement, FocusableRef, SpectrumSelectionProps, StyleProps} from '@react-types/shared';
```
LFDanLu, line 13: See https://github.com/adobe/react-spectrum/pull/4641/commits/dfcdea154620b418190b4d44193aba7a3ac89e03 for a better diff of TableView changes. Hard to see here due to the renaming of the files

packages/@react-spectrum/table/test/TreeGridTable.test.tsx
```diff
@@ -0,0 +1,1518 @@
...
+  );
+  act(() => {jest.runAllTimers();});
+  return newTree;
+};
+
+enableTableNestedRows();
+
+describe.only('TableView tests with expandable rows flag on', tableTests);
+
+describe.only('TableView with expandable rows', function () {
```
LFDanLu, line 80: these `.only` make it so the table tests don't run an extra time due to the way jest imports work.

packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -0,0 +1,314 @@
...
+    showDragButtons: showDragButtons,
+    selectionMode,
+    columns: []
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }), [children, showSelectionCheckboxes, selectionMode, showDragButtons]);
+
+  let builder = useMemo(() => new CollectionBuilder<T>(), []);
+  // TODO: figure out this ts issue
+  // @ts-ignore
+  let nodes = useMemo(() => builder.build({children}, context), [builder, children, context]);
```
LFDanLu, line 69: TS issue here since children of CollectionBuilder isn't strictly compatible with a TableBody child. I could expand the definition of CollectionElement to `export type CollectionElement<T> = SectionElement<T> | ItemElement<T> | ReactElement<TableBodyProps<T>>;` but that brings in table specific details to the general collection types... Frankly not sure how useTableState doesn't run into this issue

packages/@react-aria/table/src/useTable.ts
```diff
@@ -70,6 +70,10 @@ export function useTable<T>(props: AriaTableProps<T>, state: TableState<T>, ref:
     gridProps['aria-rowcount'] = state.collection.size + state.collection.headerRows.length;
   }
 
+  if ('expandedKeys' in state) {
+    gridProps.role = 'treegrid';
+  }
```
LFDanLu, line 75: Opted to put tree grid specific aria attributes in table hooks level, not grid hook level. If it turns out that we will have other components that only call the grid hooks and need tree grid aria attributes, then we can refactor.

packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -0,0 +1,305 @@
...
+  expandedKeys: 'all' | Set<Key>,
+  /** Toggles the expanded state for a row by its key. */
+  toggleKey(key: Key): void,
+  /** The key map containing nodes representing the collection's tree grid structure. */
+  keyMap: Map<Key, GridNode<T>>,
+  /** The number of columns provided by the user. */
+  userColumnCount: number
+}
+
+export interface TreeGridStateProps<T> extends Expandable, Omit<TableStateProps<T>, 'collection'> {}
```
LFDanLu, line 32: For now I am omitting collection from the props since this is a bit of a special case where we still are using TableCollection but need to generate the flattened nodes for the tree grid and a tree grid key map by running the original TableView's nodes structure through CollectionBuilder ourselves. 

If a user were to pass in their own collection, it would need to able of constructing the tree grid key map and the flattened Table row all on its own which we don't really have a good definition for. Perhaps it would be good to still have a separate TreeGridCollection instead like it was before https://github.com/adobe/react-spectrum/pull/4730, maybe can do this when implementing RAC expandable rows. Open to opinions

packages/@react-types/table/src/index.d.ts
```diff
@@ -122,20 +122,20 @@ export interface SpectrumColumnProps<T> extends ColumnProps<T> {
...
   loadingState?: LoadingState
 }
 
-export interface RowProps {
+export interface RowProps<T> {
   // treeble case? Unsupported props for now
-  // /** A list of child item objects used when dynamically rendering row children. */
-  // childItems?: Iterable<T>,
+  /** A list of child item objects used when dynamically rendering row children. */
+  childItems?: Iterable<T>,
```
LFDanLu, line 138: Note that this API addition will be shown when we release but it won't actually do anything without the feature flag. Perhaps I could mark it as alpha?

packages/@react-types/shared/src/collections.d.ts
```diff
@@ -69,9 +69,9 @@ export interface CollectionStateBase<T, C extends Collection<Node<T>> = Collecti
 
 export interface Expandable {
   /** The currently expanded keys in the collection (controlled). */
-  expandedKeys?: Iterable<Key>,
+  expandedKeys?: 'all' | Iterable<Key>,
   /** The initial expanded keys in the collection (uncontrolled). */
-  defaultExpandedKeys?: Iterable<Key>,
+  defaultExpandedKeys?: 'all' | Iterable<Key>,
   /** Handler that is called when items are expanded or collapsed. */
   onExpandedChange?: (keys: Set<Key>) => any
```
LFDanLu, line 76: Should we mark these as alpha as well? Not really being used except in other alpha packages

packages/@react-types/table/src/index.d.ts
yihuiliao: i think that would be a good idea? or at least something to indicate that it won't work without the feature flag

packages/@react-types/shared/src/collections.d.ts
yihuiliao: if they aren't being used except for in other alpha packages i'd be inclined to mark them as alpha as well

packages/@react-spectrum/table/test/TreeGridTable.test.tsx
```diff
@@ -0,0 +1,1518 @@
...
+          checkSelectAll(treegrid);
+        });
+
+        it('should select a row by pressing the Enter key on a chevron cell', function () {
+          let treegrid = render(<ManyRowsExpandableTable onSelectionChange={onSelectionChange} selectionMode="multiple" selectionStyle="checkbox" disabledKeys={null} onAction={null} />);
+          let rowgroups = treegrid.getAllByRole('rowgroup');
+          let rows = within(rowgroups[1]).getAllByRole('row');
+          let cell = getCell(treegrid, 'Row 1, Lvl 1, Foo');
+
+
```
ktabors, line 931: I've noticed a couple double spaces like this in the tests. The tests that are very comprehensive. :)

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -550,10 +556,8 @@ export class TableLayout<T> extends ListLayout<T> {
           this.persistedIndices.set(layoutInfo.parentKey, indices);
         }
 
-        let index = collectionNode.index;
-        if (layoutInfo.parentKey === 'body') {
-          index -= this.collection.headerRows.length;
-        }
+        // TODO: perhaps fallback to collectionNode.index
```
ktabors, line 559: why do we still have a todo?

ktabors (commented):
Noticed when I changed the locale, there is a css/rendering flicker where the ExpandableRowChevron changes from closed to expanded after the table flips. I know that isn't a real use case but was thinking there are other use cases that would replicate this rendering issue.

Just to confirm, no changes were made when TableView became TableViewBase. They looked the same in my editor.

I need to confirm if this is preexisting. There seems to be a slow re-rendering with the resizable columns example, with wrap turned on. Resize the first column to the min width and then undo it to a normal width where all the columns don't resize at the same time. Might be an existing bug, we don't have an existing story to confirm this. 
image

packages/@react-spectrum/table/src/TableViewWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * The drag and drop hooks returned by `useDragAndDrop` used to enable drag and drop behavior for the TableView.
+   * @version alpha
+   */
+  dragAndDropHooks?: DragAndDropHooks['dragAndDropHooks'],
+  /**
+   * Whether the TableView should support expandable rows. Requires the feature flag to be enabled first, see https://react-spectrum.adobe.com/react-spectrum/TableView.html#expandable-rows.
+   * @version alpha
+   */
+  hasExpandableRows?: boolean
```
LFDanLu, line 62: prefix this UNSTABLE_

packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -0,0 +1,305 @@
...
+
+/**
+ * Provides state management for a tree grid component. Handles building a collection
+ * of columns and rows from props. In addition, it tracks and manages expanded rows, row selection, and sort order changes.
+ */
+export function useTreeGridState<T extends object>(props: TreeGridStateProps<T>): TreeGridState<T> {
```
LFDanLu, line 38: prefix this with UNSTABLE and add an error throw if feature flag isn't turned on

packages/@react-stately/layout/src/TableLayout.ts
LFDanLu: Ah good catch, thought I got all of these. I'll get rid of the todo

LFDanLu:
@ktabors 

> Noticed when I changed the locale, there is a css/rendering flicker where the ExpandableRowChevron changes from closed to expanded after the table flips.

I can't seem to reproduce this, was this in particular story/browser?

> Just to confirm, no changes were made when TableView became TableViewBase. They looked the same in my editor.

There were some changes, but def hard to see with the new diff. https://github.com/adobe/react-spectrum/commit/dfcdea154620b418190b4d44193aba7a3ac89e03#diff-613fe2bf090dadb8c1be12e5cbd07f8257055654cf30e819e1f3d0362dfdc930 is helpful here, shows the TableView changes before the naming changes.

> I need to confirm if this is preexisting. There seems to be a slow re-rendering with the resizable columns example...

I'll see about reproducing this on main

ktabors (approved):

packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -0,0 +1,310 @@
...
+
+export interface TreeGridState<T> extends TableState<T> {
+  /** A set of keys for items that are expanded. */
+  expandedKeys: 'all' | Set<Key>,
+  /** Toggles the expanded state for a row by its key. */
+  toggleKey(key: Key): void,
+  /** The key map containing nodes representing the collection's tree grid structure. */
+  keyMap: Map<Key, GridNode<T>>,
+  /** The number of columns provided by the user. */
+  userColumnCount: number
```
devongovett, line 30: Perhaps we can remove this if we use my suggestion in a previous comment to compute some of this in Row itself?

packages/@react-stately/table/src/useTreeGridState.ts
devongovett: Why can't you use `useCollection` here rather than CollectionBuilder directly? It looks like that accepts a more general type. I guess the factory function could call generateTreeGridCollection and then construct a TableCollection?

packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -0,0 +1,310 @@
...
+        }
+      }
+      let clone = {...node, childNodes: childNodes, parentKey: body.key, level: 1, index: globalRowCount++};
+      flattenedRows.push(clone);
+    }
+
+    let newProps = {};
+    // Associate the parent column node to the cell/column cell.
+    if (node.type === 'cell' || node.type === 'column') {
+      newProps['column'] = columns[i];
```
devongovett, line 241: I wonder if we can avoid assigning column to the treegrid nodes and instead rely on the flattened nodes to have this property? It should already be assigned by TableCollection there. That would allow you to remove the duplicate logic above that finds the columns, adds the checkbox column, etc. Should be easy enough to find the corresponding flattened node by key whenever you need the `column`?

devongovett:
btw, great work on this. It's looking really awesome overall. Only really had minor nits. 🥳

packages/@react-aria/table/src/useTableRow.ts
```diff
@@ -13,21 +13,67 @@
...
+    rowProps['aria-rowindex'] = node.index + 1 + state.collection.headerRows.length; // aria-rowindex is 1 based
+  } else {
+    delete rowProps['aria-rowindex'];
+  }
+
+  let treeGridRowProps: HTMLAttributes<HTMLElement> = {};
+  if ('expandedKeys' in state) {
+    let treeNode = state.keyMap.get(node.key);
+    if (treeNode != null) {
+      let hasChildRows = treeNode.props?.childItems || treeNode.props?.children?.length > state.userColumnCount;
```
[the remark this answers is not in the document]
LFDanLu: I can give this a shot. Was a bit wary of doing this since CollectionBuild has some logic for hasChildNodes and not sure if RAC collections could/would make this distinction either compatibility wise

packages/@react-spectrum/table/src/TableViewBase.tsx
```diff
@@ -0,0 +1,1482 @@
...
+        }>
+          {cell.rendered}
+        </span>
+      </div>
+    </FocusRing>
+  );
+}
+
+function ExpandableRowChevron({cell}) {
+   // TODO: move some/all of the chevron button setup into a separate hook?
```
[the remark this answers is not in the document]
LFDanLu: sounds good, will handle this when we work on an aria/RAC implementation of expandable rows

packages/@react-spectrum/table/src/TreeGridTableView.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export interface TreeGridTableProps<T> extends Omit<SpectrumTableProps<T>, 'UNSTABLE_hasExpandableRows'> {}
+
+function TreeGridTableView<T extends object>(props: TreeGridTableProps<T>, ref: DOMRef<HTMLDivElement>) {
+  let {
+    selectionStyle,
+    dragAndDropHooks
+  } = props;
+  let [showSelectionCheckboxes, setShowSelectionCheckboxes] = useState(selectionStyle !== 'highlight');
+  let isTableDraggable = !!dragAndDropHooks?.useDraggableCollectionState;
```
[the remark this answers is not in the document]
LFDanLu: DnD hasn't been tested at all yet, definitely feels like its gonna bit a bit complicated. I'll add a warning/error

devongovett:
Two additional thoughts:

1. I think we should temporarily prefix `expandedKeys`, `defaultExpandedKeys`, `onExpandedChange`, and `childItems` (on `Row`) with `UNSTABLE_` as well. In addition we should add something to the prop description that explains that they only work if `hasExpandableRows` is true. I wouldn't change the shared `Expandable` interface for now, so maybe that means copying those type definitions into the table types temporarily.
2. Maybe `hasExpandableRows` should be called `allowsExpandableRows`? It's subtle but to me, "has" implies that there are currently expandable rows in the content, but that might not always be the case if the content is user generated (e.g. if there are no sub folders). "Allows" makes it clearer that it _can_ contain expandable rows but might not always. 🤷

packages/@react-spectrum/table/test/TreeGridTable.test.tsx
```diff
@@ -0,0 +1,1516 @@
...
+      {children}
+    </Provider>
+  );
+  act(() => {jest.runAllTimers();});
+  return newTree;
+};
+
+enableTableNestedRows();
+
+describe.only('TableView tests with expandable rows flag on', tableTests);
```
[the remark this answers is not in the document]
LFDanLu: So we don't actually want `UNSTABLE_hasExpandableRows` turned on for the table tests here since it would then render a TreeGridTableView instead of a TableView, resulting in test suite failures since the table tests look for `aria-rowindex` instead of `aria-posinset` etc. This `.only` essentially just want to turn the flag on but omit the `UNSTABLE_hasExpandableRows` prop and make sure that a typical TableView still renders.

I guess I could add checks in the base tableview tests to check that the treegrid role doesn't exist

packages/@react-spectrum/table/test/TreeGridTable.test.tsx
LFDanLu: > Another way would be to move the tests out into a separate util maybe

Could you elaborate on this a bit? Definitely interested in any way to get rid of the `.only`'s but not sure I understand what that util would be

packages/@react-stately/table/src/useTreeGridState.ts
LFDanLu: The only issue would be extracting the `keyMap` and `columnCount` from the `generateTreeGridCollection` call since it would be in the factory function. I guess I could set up a ref or something to store those

packages/@react-stately/table/src/useTreeGridState.ts
LFDanLu: Hmm, good point, at a glance I think I can do that. I think I was more concerned with just having the tree grid key map data having as much parity as possible to the TableCollection nodes + having fully fleshed out data for the cell nodes in the tree grid key map (aka calculating indexOfType/nextKey/etc) which all required adding to the `columns` array here with the checkbox column

packages/@react-spectrum/table/test/TreeGridTable.test.tsx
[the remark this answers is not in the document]
devongovett: Oh maybe I misunderstood the intention. I assumed this was meant to test that the existing table stuff still passed when hasExpandableRows was true, but I guess you meant to test only that the feature flag didn't break anything? Once we delete the feature flag, this would just be removed entirely?

packages/@react-stately/table/src/useTreeGridState.ts
devongovett: Oh hmm. It's fine if it doesn't work, a ref doesn't sound great either. Maybe the types can help though.

packages/@react-spectrum/table/test/TreeGridTable.test.tsx
devongovett: I meant pulling out the `tableTests` function into a separate file, and importing it from both `Table.test.js` and `TreeGridTable.test.tsx` so that it doesn't have a `describe` in it. Then you shouldn't need the only?

packages/@react-spectrum/table/test/TreeGridTable.test.tsx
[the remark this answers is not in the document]
LFDanLu: correct, we can get rid of it when we delete the feature flag. As for testing the existing table behavior when `hasExpandableRows` is true, I've written tests for selection/keyboard nav/etc but ideally we'd pull some of the existing test cases from TableView and run them for a expandable row table, just need to figure out a good way to share those and refactor them to work for both base Table + expandable row

packages/@react-stately/table/src/useTreeGridState.ts
LFDanLu: tried refactoring this a bit, ran into an infinite rerender when including the generateTreeGridCollection call into the factory function useCallback along with a state setter to store the treeGridCollection:
```
  let [treeGridCollection, setTreeGridCollection] = useState<TreeGridCollection<T>>({keyMap: new Map(), tableNodes: [], flattenedRows: [], userColumnCount: 0});
  let factory = useCallback((nodes) => {
    let treeGridCollection = generateTreeGridCollection<T>(nodes, {showSelectionCheckboxes, showDragButtons, expandedKeys});
    setTreeGridCollection(treeGridCollection);
    return new TableCollection(treeGridCollection.tableNodes, null, context);
  }, [context, showSelectionCheckboxes, showDragButtons, expandedKeys]);

  let collection = useCollection<T, ITableCollection<T>>({children}, factory, context);
```
gonna punt for a bit and handle the other todos before trying to figure out why

packages/@react-aria/table/src/useTableRow.ts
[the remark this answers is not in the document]
LFDanLu: Tried it out, setting `hasChildNodes` in Row to false breaks rendering of the Row entirely since CollectionBuilder [early returns](https://github.com/adobe/react-spectrum/blob/b4e1f1cb7d37ec85be729e2cd68d20df6f395c03/packages/%40react-stately/collections/src/CollectionBuilder.ts#L182-L184) in the childNodes generator if that property is false

packages/@react-stately/table/src/useTreeGridState.ts
LFDanLu: Since https://github.com/adobe/react-spectrum/pull/4641#discussion_r1268670988 didn't pan out, I'm inclined to leave this as is. Another alternative is if I calculate the number of leaf columns there are excluding the checkbox/drag handle columns in places like useTableRow/BaseTableView instead of having `userColumnCount` as a piece of info calculated in TreeGridState but that feels more cumbersome

packages/@react-stately/table/src/useTreeGridState.ts
LFDanLu: Pushed an update removing the column node assignment to the treegrid nodes. Preserved the `columns` array just so it is easier to determine when a treegrid cell is the last cell in the row and add `nextKey: null`.

packages/@react-spectrum/table/test/TreeGridTable.test.tsx
LFDanLu: Ah, good idea. Tried it out and it works, but opted to move the 'TableView tests with expandable rows flag on' test suite into Table.test.js instead since we'll be getting rid of it later anyways (not sure why I didn't just do that to start with). I've stashed the "`tableTests` function into separate file" work in case if we still want that, figured that approach is more appropriate if the Table tests were actually generalized enough to test basic table interactions with the TreeGridView render instead.

I may look into refactoring/generalizing those Table tests and moving it into a util function in the future to make it easier to get full coverage when we make the RAC implementation as well.

packages/@react-stately/table/src/useTreeGridState.ts
LFDanLu: pushed a type cast that gets rid of the `ts-ignore`, feels a bit scuffed

packages/@react-stately/table/src/useTreeGridState.ts
devongovett: Oh you probably didn't need an array for that, you could have a `columnCount` that you add the number of extra columns to?

devongovett (approved):
approving for testing

packages/@react-types/shared/src/collections.d.ts
LFDanLu: Addressed by duplicating the Expandable props in SpectrumTableProps

ktabors (approved):

dannify (approved):

PR #4646, opened 2023-06-07
Fix scroll position jump on mobile for column resize examples

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test TableView resizable column docs on a phone, make sure the scroll position doesn't jump when clicking on the resizable column header

## 🧢 Your Project:

RSP


ktabors (approved):
This fixes it, but a little concerned that a Table outside of the screen width causes scrolling when a menu opens. I feel like there is a bigger bug. Did you try `max-width` instead of just removing? And why do tables with a `width="size-5000"` work when there are two wrapped in a Flex? The resize to screen width in this case.

yihuiliao (commented):
it fixes the problem that was occurring on mobile but because of the removal of width="size-5000", the size of some of the tables look different on the browser. the tables themselves look fine still, just not what they looked like before.

LFDanLu:
> This fixes it, but a little concerned that a Table outside of the screen width causes scrolling when a menu opens. I feel like there is a bigger bug

The issue seems to happen if the `window.innerWidth` differs from the `document.documentElement.clientWidth` -> causing a non-zero padding-right to be applied on the documentElement -> causing the docs page contents to shift in response to the padding, so there are definitely scenarios in the wild that could break this.

> Did you try max-width instead of just removing? And why do tables with a width="size-5000" work when there are two wrapped in a Flex? The resize to screen width in this case.

I got rid of the width so that the table would adapt itself to the screen width rather than limiting it to the maximum value that would still work on mobile even when there is more than enough room to render the whole table. The other examples that have "size-5000" are all within a Flex and have `flex-shrink: 1`  by default as a flex item so they will shrink to fit within the container.

yihuiliao (approved):

yihuiliao (approved):

PR #4648, opened 2023-06-09
Persisting valid DatePicker TimeField value

### LINKED ISSUE #4033: DatePicker is setting time to default after selecting time for the first time
# 🐛 Bug Report 

DatePicker is setting time to default after selecting time for the first time

## 🤔 Expected Behavior

Should set it to the time we selected using the DatePicker

## 😯 Current Behavior

After inserting the time and leaving the DatePicker, the time is set to the default one.

## 🌍 Your Environment

| react-spectrum   | "^3.19.0"
| Browser          | Chrome
| Operating System | Mac

## 🧢 Your Company/Team

 Adobe/Campaign

ktabors:
Closes #4033

The TimeField blur event doesn't happen when a TimeField is selected in a Popover (DateRangePicker) and the Popover is closed. To address that unmount related behavior the useDateField() hook was updated to be aware of valid times when segment updates occur.

- This introduces behavior changes to DatePicker time granularity. Setting a valid time, before setting AM/PM, will set the time and trigger an onChange. 
- Previously a blur on the TimeField would have been needed to persist the time without setting AM/PM.
- This change is general to TimeField. 
- ~~If you test using the DatePicker "all the events" story, you'll see that the the change triggers two onChange events.~~

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Goto the default DatePicker story
Using story Controls, set granularity to hour, minute, or second.
Click on the button to open DatePicker Popover.
Select a date.
Scroll to the time field.
Enter a valid time up to the AM/PM.
Notice that your cursor on the AM/PM and the selected data and input time have been persisted to the DateRangePicker input.
If you tab away or close the Popover (by pressing escape or clicking outside the Popover) the value will persist.

Also test removing am/pm from the input as well as the TimeField in the Dialog, there is some code to prevent weird behavior around that.

It's worth regression testing TimeField and its events behavior.

## 🧢 Your Project:
RSP


packages/@react-spectrum/datepicker/test/DateRangePicker.test.js
```diff
@@ -1020,7 +1022,7 @@ describe('DateRangePicker', function () {
       expect(endField).not.toHaveAttribute('aria-describedby');
 
       let description = group.getAttribute('aria-describedby').split(' ').map(d => document.getElementById(d).textContent).join(' ');
-      expect(description).toBe('Selected Range: February 3, 2020 at 8:00 AM to February 10, 2020 at 10:00 AM');
+      expect(description).toBe('Selected Range: February 3, 2020, 8:00 AM to February 10, 2020, 10:00 AM');
```
ktabors, line 1025: When I recreate this test in Stroybook it behaves like it used to with the "at" instead of a comma.

packages/@react-spectrum/datepicker/src/TimeField.tsx
```diff
@@ -44,6 +44,17 @@ function TimeField<T extends TimeValue>(props: SpectrumTimeFieldProps<T>, ref: F
...
+  /**
+   * When a TimeField is in a Picker, the Picker can close without triggering
+   * a TimeField onBlur which is sometimes used to persist valid dates. This
+   * persists the value the second it is valid.
+   */
+  useEffect(() => {
+    return () => {
+      state.confirmPlaceholder();
+    };
+  }, [state]);
```
devongovett, line 56: I guess this would make it run every time `state` changes, but I we only really want it to run if the component is actually unmounting. Another problem is that in React strict mode, this cleanup will be run on the initial mount as well since effects run twice. That makes it harder to detect a "real" unmount. One way I found is something like this:

```js
let ref = useRef();
useEffect(() => {
  return () => {
    if (ref.current == null) {
      // Actual unmount rather than state change or extra strict mode effect.
    }
  };
}, [state, ref]);
```

If the DOM ref has been detached already we know it's a real unmount rather than a state change. Sandbox to demonstrate: https://codesandbox.io/s/determined-panka-chl6zn?file=/src/App.js:566-713

You could also move this into `useDateField` (which is used by `useTimeField`) so it also works in our hooks examples.

ktabors: I kept this in RSP because I was concerned about it being a breaking change for the hooks.
Thanks for the code suggestion!

packages/@react-spectrum/datepicker/stories/DatePicker.stories.tsx
```diff
@@ -360,7 +360,7 @@ function Example(props) {
 }
 
 function ControlledExample(props) {
-  let [value, setValue] = React.useState(null);
+  let [value, setValue] = React.useState(props.value);
```
ktabors, line 363: Noticed that the story sets a prop, but it wasn't set to the story state.

packages/@react-stately/datepicker/src/useDateFieldState.ts
```diff
@@ -234,15 +236,19 @@ export function useDateFieldState<T extends DateValue = DateValue>(props: DateFi
     if (props.isDisabled || props.isReadOnly) {
       return;
     }
+    let validKeys = Object.keys(validSegments);
+    let allKeys = Object.keys(allSegments);
 
-    if (Object.keys(validSegments).length >= Object.keys(allSegments).length) {
+    // if all the segments are completed or a timefield with everything but am/pm set the time, also ignore when am/pm cleared
+    if (Object.keys(validSegments).length >= Object.keys(allSegments).length || (validKeys.length === allKeys.length - 1 && allSegments.dayPeriod && !validSegments.dayPeriod && clearedSegment !== 'dayPeriod')) {
```
ktabors, line 243: Unmounting with a ref didn't work because it was updating on segment changes and when the dialog dismisses the onChange happens before the component unmounts.

packages/@react-spectrum/datepicker/src/TimeField.tsx
ktabors: I left a comment about how this didn't work as expected so a tried something else. It doesn't work when you use the spin button arrow up or down once (fix is in progress).

ktabors: I tried this using useLayoutEffect in useDateField.js and it still threw two onChange events like the code being commented on. As mentioned in a previous comment, when this is done with useEffect we get the original behavior. It also has the bug of not being able to delete the am/pm segment with backspace mentioned in the refactor related comments.

Interesting note, if you set a time like "9:09" the first onChange has "9:00" while the second has "9:09". If you change the granularity to seconds and set the time to "9:09:09", the first onChange is "9:09:00" and the second is "9:09:09".

yihuiliao (dismissed):
seems to work well in storybook. i think the changes make sense based on what you've told me but the rest of the team may have other opinions

snowystinger (commented):
I followed the instructions. From a blank field, selecting and editing values in the popover did not propagate to the blank field until I hit esc.

I noticed a strange new behavior, when I delete the am/pm in the datefield (not in the popover, as that doesn't have Yihui's fix yet), and then left arrow to the closest segment, then arrow up/down to spin the time, then AM will automatically be selected again.

ktabors:
> I noticed a strange new behavior, when I delete the am/pm in the datefield (not in the popover, as that doesn't have Yihui's fix yet), and then left arrow to the closest segment, then arrow up/down to spin the time, then AM will automatically be selected again.

Correct, this is a behavior change that extends beyond TimeField in a DatePicker to all instances of TimeField and DateField. The change is that when we have a valid date and valid time (am/pm not set), we set the time and assume AM. The change is we do this on the last segment to create a valid date/time change rather than blur.

ktabors:
We've spent additional time looking at the sequence of events on popover close, why there are two events, using useEffect/useLayoutEffect, etc. Currently, we don't see a way forward with those approaches and we'd like to evaluate the desirability and feasibility of the behavior change.

yihuiliao:
> From a blank field, selecting and editing values in the popover did not propagate to the blank field until I hit esc.

This behavior was noted in the this [ticket](https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=30602564). This is just how the current behavior is and not necessarily a result of this PR and will require additional work

packages/@react-stately/datepicker/src/useDateFieldState.ts
```diff
@@ -202,6 +202,8 @@ export function useDateFieldState<T extends DateValue = DateValue>(props: DateFi
     () => props.value || props.defaultValue ? {...allSegments} : {}
   );
 
+  let [clearedSegment, setClearedSegment] = useState<string>();
```
devongovett, line 205: Why is this extra state necessary? Doesn't `validSegments` already keep track when dayPeriod is cleared?

ktabors: On line [354](https://github.com/adobe/react-spectrum/pull/4648/files/18bc4c1e0e322eb03b9706a25002359402ecf909#diff-838657fffbbf57e9688a449c4899b3a29ace82a4503c9827f932f24214faa1fdR354) we remove a segment from `validSegments`. However, `setValue()` needs to know what segment was deleted for line [243](https://github.com/adobe/react-spectrum/pull/4648/files/18bc4c1e0e322eb03b9706a25002359402ecf909#diff-838657fffbbf57e9688a449c4899b3a29ace82a4503c9827f932f24214faa1fdR243) to work. When testing all the cases this came up as an edge case. Because updates are event based `setValue` has no knowledge of what triggered the update and needs this extra piece of info.

Technically this doesn't need to be state because `setValue()` is called before the state updates via `set...` and we set its value as well as calling its setter like we do for `validSegments`.

devongovett: This could probably be a ref because it doesn't need to trigger re-rendering when set.

devongovett (dismissed):

devongovett:
hmm looks like there is a test failing

reidbarber (approved):
LGTM

PR #4655, opened 2023-06-12
Fix width of number field within a form

devongovett:
Specificity of form selector that sets `width: 100%` should override the one for NumberField `isMobile`. Switched to setting a variable rather than the `inline-size` property directly so the specificity is always the same regardless of if it is mobile or desktop.

### DIFF
--- a/packages/@adobe/spectrum-css-temp/components/stepper/index.css
+++ b/packages/@adobe/spectrum-css-temp/components/stepper/index.css
@@ -34,7 +34,8 @@ governing permissions and limitations under the License.
 
   block-size: max-content;
 
-  inline-size: var(--spectrum-stepper-default-width-desktop);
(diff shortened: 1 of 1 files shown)


ktabors (approved):

ktabors:
After approving I took a second look and found the following which is preexisting. We've probably seen this before and documented it. It's the NumberField "min width" story.
image
image

LFDanLu:
Noted the other chromatic failures, seems like we could fix them by increasing the specificity of the Form specific overrides (either by doing something like `.spectrum-Form.spectrum-Form` or changing the `&:where(.spectrum-Form--positionSide) {` to `&.spectrum-Form--positionSide) {`). Noted however that the usage of `:where` seems to be intentional changes from https://github.com/adobe/react-spectrum/commit/595aa1844303390268709226ecc58429fc73a6b3, not entirely sure what it was fixing though

PR #4656, opened 2023-06-12
Add support for standalone RAC popovers

devongovett:
This allows the RAC `Popover` component to be used standalone, outside a `DialogTrigger`, the same way as we already support for `Modal`. This means you can pass the `isOpen` and `onOpenChange` props directly to the popover, as well as the `triggerRef` prop to indicate which element to position it relative to. This enables more complex use cases, such as programmatically showing popovers (e.g. coach marks). The `isOpen` prop is also useful for integration with animation libraries like Framer Motion. It's also nice to have consistency between Modal and Popover here.


snowystinger (approved):

reidbarber (approved):
LGTM

PR #4663, opened 2023-06-14
TableView expandable rows: rendering + interactions

reidbarber:
## ✅ Pull Request Checklist:


packages/@react-spectrum/table/package.json
```diff
@@ -62,7 +62,9 @@
     "@react-types/grid": "^3.1.8",
     "@react-types/shared": "^3.18.1",
     "@react-types/table": "^3.6.1",
+    "@spectrum-icons/express": "3.0.0-alpha.6",
     "@spectrum-icons/ui": "^3.5.2",
+    "@spectrum-icons/workflow": "^4.2.1",
```
LFDanLu, line 67: where did these two imports come from?

packages/@react-aria/table/src/useTableRow.ts
```diff
@@ -34,9 +47,32 @@ export function useTableRow<T>(props: GridRowProps<T>, state: TableState<T> | Tr
...
+        if ((e.key === EXPANSION_TRIGGERS['expand'][direction]) && state.selectionManager.focusedKey === node.key && node.props.childItems && state.expandedKeys !== 'all' && !state.expandedKeys.has(node.key)) {
+          state.toggleKey(node.key);
+          e.stopPropagation();
+          // After some time, if the row was not expanded (via controlled), focus the first cell
+          setTimeout(() => {
+            if (state.expandedKeys !== 'all' && !state.expandedKeys.has(node.key)) {
+              let cells = [...getChildNodes(node, state.collection)].filter(child => child.type === 'cell');
+              state.selectionManager.setFocusedKey(direction === 'ltr' ? cells[0].key : cells[cells.length - 1].key);
+            }
+          }, 10);
```
LFDanLu, line 63: this timeout feels a bit specific for a case that probably won't happen(?) since users would modify their controlled expandedKeys on toggle, I think I can go ahead and change the tests to use defaultExpandedKeys instead of controlled expandedKeys. We can ask for opinions from the team.

packages/@react-spectrum/table/src/TreeGridTableView.tsx
```diff
@@ -1387,16 +1390,21 @@ function TableCheckboxCell({cell}) {
...
+  let {state, toggleKey, direction} = useTableContext();
   let ref = useRef();
   let columnProps = cell.column.props as SpectrumColumnProps<unknown>;
   let isDisabled = state.disabledKeys.has(cell.parentKey);
   let {gridCellProps} = useTableCell({
     node: cell,
     isVirtualized: true
   }, state, ref);
-
+  let showExpandCollapseButton = cell.index === 0 && (state.collection.getItem(cell.parentKey)?.props.childItems?.length > 0 || state.collection.getItem(cell.parentKey)?.props.children.length > state.collection.columnCount);
```
LFDanLu, line 1402: The check on `cell.index === 0` implies that the chevron can only be on the first cell right? I believe there was some discussion about making the chevron attached to the cell which is the first rowColumn header instead, forgot to share that with you, sorry bout that

packages/@react-spectrum/table/src/TreeGridTableView.tsx
```diff
@@ -1387,16 +1390,21 @@ function TableCheckboxCell({cell}) {
...
+  let isExpanded = showExpandCollapseButton && (state.expandedKeys === 'all' || state.expandedKeys.has(cell.parentKey));
+  let levelOffset = (cell.level - (showExpandCollapseButton ? 1 : 2)) * 16 + 4;
```
LFDanLu, line 1404: I'll test it out, but are the numbers included in here and other places in the PR accounting for scale changes as well?

packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -70,21 +70,21 @@ export function useTreeGridState<T extends object>(props: TreeGridStateProps<T>)
...
 // TODO: copied from useTreeState, perhaps make this accept multiple keys?
-function toggleKey<T>(set: 'all' | Set<Key>, key: Key, collection: ITableCollection<T>): Set<Key> {
-  let res;
-  if (set === 'all') {
-    res = new Set(collection.rows.filter(row => row.props.hasChildItems || row.props.children.length > collection.columnCount));
-    res.delete(key);
+function toggleKey<T>(currentExpandedKeys: 'all' | Set<Key>, key: Key, collection: ITableCollection<T>): Set<Key> {
+  let updatedExpandedKeys: Set<Key>;
+  if (currentExpandedKeys === 'all') {
+    updatedExpandedKeys = new Set(collection.rows.filter(row => row.props.childItems?.length > 0 || row.props.children.length > collection.columnCount).map(row => row.key));
```
LFDanLu, line 76: if a row has childItems that aren't loaded in yet and thus is an array of length 0 I think we will want to include it in the set of expanded keys maybe

LFDanLu (commented):
some things I noticed when scanning through the code, will add more comments after testing if I come across anything. Thanks for the speedy work!

packages/@react-aria/table/src/useTableRow.ts
LFDanLu: in testing, the behavior diverges from aria grid examples since the row expands and then focus is moved to the next cell. Felt kinda weird to me IMO, open to opinions

packages/@react-spectrum/table/src/TreeGridTableView.tsx
LFDanLu: addressed

packages/@react-spectrum/table/src/TreeGridTableView.tsx
reidbarber: I can fix this, and also add a comment explaining the logic

packages/@react-spectrum/table/src/TreeGridTableView.tsx
```diff
@@ -1387,15 +1390,23 @@ function TableCheckboxCell({cell}) {
...
   let isDisabled = state.disabledKeys.has(cell.parentKey);
   let {gridCellProps} = useTableCell({
     node: cell,
     isVirtualized: true
   }, state, ref);
+  let isFirstRowHeader = state.collection.rowHeaderColumnKeys.keys().next().value === cell.column.key;
+  let showExpandCollapseButton = isFirstRowHeader && (state.collection.getItem(cell.parentKey)?.props.childItems?.length > 0 || state.collection.getItem(cell.parentKey)?.props.children.length > state.collection.columnCount);
+  let isExpanded = showExpandCollapseButton && (state.expandedKeys === 'all' || state.expandedKeys.has(cell.parentKey));
+  // TODO: would be good to adjust these numbers by scale if it matters
+  let levelOffset = (cell.level - (showExpandCollapseButton ? 1 : 2)) * 16 + 4 + (isFirstRowHeader && !showExpandCollapseButton ? 16 : 0);
```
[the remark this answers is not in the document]
reidbarber: Yep, fixing this now

PR #4730, opened 2023-06-28
(WIP) Simplify TreeGridCollection by building the keymap inside state instead 

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -10,73 +10,91 @@
...
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  }), [props.children, showSelectionCheckboxes, selectionMode, showDragButtons]);
+  }), [children, showSelectionCheckboxes, selectionMode, showDragButtons]);
 
-  let collection = useCollection<T, ITreeGridCollection<T>>(
-    props,
-    useCallback((nodes) => new TreeGridCollection(nodes, {...context, expandedKeys}), [context, expandedKeys]),
-    context
-  );
+  let builder = useMemo(() => new CollectionBuilder<T>(), []);
```
LFDanLu, line 66: Since we don't have a TreeGridCollection anymore, we need to call CollectionBuilder to generate the table's nodes so we can generate the treegrid keymap. Note that this means I have to export CollectionBuilder from `react-stately/collections` which wasn't being exported before

packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -10,73 +10,91 @@
...
+  // TODO: figure out this ts issue
+  // @ts-ignore
+  let nodes = useMemo(() => builder.build({children}, context), [builder, children, context]);
```
LFDanLu, line 69: Not sure why TS complains about this and not in useTableState...

packages/@react-stately/table/src/useTreeGridState.ts
```diff
@@ -10,73 +10,91 @@
...
+  userColumnCount: number
 }
 
-export interface TreeGridStateProps<T> extends Expandable, TableStateProps<T> {
-  collection?: ITreeGridCollection<T>
-}
+// TODO: For now I am omitting collection since this is a bit of a special case where are still using TableCollection but include
+// extra information for the tree grid in the state. That means the user can't really provide their own collection. Perhaps it would
+// be good to still have a TreeGridCollection implementation in that case?
+export interface TreeGridStateProps<T> extends Expandable, Omit<TableStateProps<T>, 'collection'> {}
```
LFDanLu, line 35: Since we are generating a key map to represent the tree grid structure and then flattening the rows for TableCollection instead of making a TreeGridCollection, it doesn't really make sense IMO to accept a user generated collection here (or at the very least it doesn't feel very straight forward to accept one). If a user were to provide a collection that we would use over generating a separate TableCollection, it would have to contain the same information as the TableCollection with flattened rows for layout/keyboardDelegate purpose and provide the extra tree grid key map info. Perhaps an argument for keeping TreeGridCollection and the extra changes to layout/keyboard delegates to accommodate the key getters that exist in the original PR

LFDanLu:
Updated suite of tests pass, gonna merge in. Can revisit the changes in this PR later if need be

PR #4755, opened 2023-07-10
RAC: Additional render props and docs improvements

reidbarber:
Follow-up to https://github.com/adobe/react-spectrum/pull/4616

Includes the following changes:
- New render prop types are exported from from index.ts
- Additional render props tables are added to the docs, including:
  - TableBody, TagList, Calendar, DateField, DatePicker, DateRangePicker, RangeCalendar, TimeField , NumberField, TextField, ComboBox Input.
- Focus state removed from various field/wrapper components, including:
  - Calendar, DateField, SearchField, TextField, NumberField, ComboBox, CheckboxGroup, RadioGroup, Slider
- Add validationState to a few components, including:
  - NumberField, SearchField, RadioGroup, Select, ComboBox
- Add isRequired to Select and ComboBox
- Add isDisabled to ComboBox
- Meter and ProgressBar docs props table updated to show valueText and percentage
- Hover added to Column and Cell
- Remove isOpen from Tooltip (because it is always true if it's rendered)
- Remove value from SearchField
- Other minor docs fixes and improvements

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/react-aria-components/docs/Checkbox.mdx
+++ b/packages/react-aria-components/docs/Checkbox.mdx
@@ -250,7 +250,7 @@ Render props may also be used as children to alter what elements are rendered ba
 
 The states, selectors, and render props for `Checkbox` are documented below.
 
-<StateTable properties={docs.exports.CheckboxRenderProps.properties} />
+<StateTable properties={docs.exports.CheckboxRenderProps.properties} showOptional />
 
 ## Reusable wrappers
 
--- a/packages/react-aria-components/docs/Meter.mdx
+++ b/packages/react-aria-components/docs/Meter.mdx
@@ -181,7 +181,7 @@ The selectors and render props for each component used in a `Meter` are document
 
 A `Meter` can be targeted with the `.react-aria-Meter` CSS selector, or by overriding with a custom `className`. It supports the following states and render props:
 
-<StateTable properties={docs.exports.MeterRenderProps.properties} />
+<StateTable properties={docs.exports.MeterRenderProps.properties} showOptional />
 
 ### Label
 
--- a/packages/react-aria-components/docs/NumberField.mdx
+++ b/packages/react-aria-components/docs/NumberField.mdx
@@ -285,7 +285,9 @@ The states, selectors, and render props for each component used in a `NumberFiel
 
 ### NumberField
 
-A `NumberField` can be targeted with the `.react-aria-NumberField` CSS selector, or by overriding with a custom `className`. It provides a <TypeLink links={statelyDocs.links} type={statelyDocs.exports.NumberFieldState} /> object to its render props, which can be used to customize the `className`, `style`, or `children`.
+A `NumberField` can be targeted with the `.react-aria-NumberField` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
+
+<StateTable properties={docs.exports.NumberFieldRenderProps.properties} />
 
 ### Label
 
@@ -293,13 +295,13 @@ A `Label` can be targeted with the `.react-aria-Label` CSS selector, or by overr
 
 ### Group
 
-A `Group` can be targeted with the `.react-aria-Group` selector, or by overriding with a custom `className`. It supports the following states:
+A `Group` can be targeted with the `.react-aria-Group` selector, or by overriding with a custom `className`. It provides the following render props and states:
 
 <StateTable properties={docs.exports.GroupRenderProps.properties} />
 
 ### Input
 
-An `Input` can be targeted with the `.react-aria-Input` CSS selector, or by overriding with a custom `className`. It supports the following states:
+An `Input` can be targeted with the `.react-aria-Input` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
 
 <StateTable properties={docs.exports.InputRenderProps.properties} />
 
--- a/packages/react-aria-components/docs/NumberField.mdx
+++ b/packages/react-aria-components/docs/NumberField.mdx
@@ -285,7 +285,7 @@ The states, selectors, and render props for each component used in a `NumberFiel
 
 ### NumberField
 
-A `NumberField` can be targeted with the `.react-aria-NumberField` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
+A `NumberField` can be targeted with the `.react-aria-NumberField` CSS selector, or by overriding with a custom `className`. It supports the following states:
 
 <StateTable properties={docs.exports.NumberFieldRenderProps.properties} showOptional />
 
@@ -295,13 +295,13 @@ A `Label` can be targeted with the `.react-aria-Label` CSS selector, or by overr
 
 ### Group
 
-A `Group` can be targeted with the `.react-aria-Group` selector, or by overriding with a custom `className`. It provides the following render props and states:
+A `Group` can be targeted with the `.react-aria-Group` selector, or by overriding with a custom `className`. It supports the following states:
 
 <StateTable properties={docs.exports.GroupRenderProps.properties} />
 
 ### Input
 
-An `Input` can be targeted with the `.react-aria-Input` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
+An `Input` can be targeted with the `.react-aria-Input` CSS selector, or by overriding with a custom `className`. It supports the following states:
 
 <StateTable properties={docs.exports.InputRenderProps.properties} />
 
--- a/packages/react-aria-components/docs/NumberField.mdx
+++ b/packages/react-aria-components/docs/NumberField.mdx
@@ -287,7 +287,7 @@ The states, selectors, and render props for each component used in a `NumberFiel
 
 A `NumberField` can be targeted with the `.react-aria-NumberField` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
 
-<StateTable properties={docs.exports.NumberFieldRenderProps.properties} />
+<StateTable properties={docs.exports.NumberFieldRenderProps.properties} showOptional />
 
 ### Label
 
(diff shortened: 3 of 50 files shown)


packages/react-aria-components/docs/NumberField.mdx
```diff
@@ -285,21 +285,23 @@ The states, selectors, and render props for each component used in a `NumberFiel
 
 ### NumberField
 
-A `NumberField` can be targeted with the `.react-aria-NumberField` CSS selector, or by overriding with a custom `className`. It provides a <TypeLink links={statelyDocs.links} type={statelyDocs.exports.NumberFieldState} /> object to its render props, which can be used to customize the `className`, `style`, or `children`.
+A `NumberField` can be targeted with the `.react-aria-NumberField` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
+
+<StateTable properties={docs.exports.NumberFieldRenderProps.properties} />
```
LFDanLu, line 290: Does this need showOptional?

packages/react-aria-components/docs/Slider.mdx
```diff
@@ -324,11 +324,11 @@ A `Label` can be targeted with the `.react-aria-Label` CSS selector, or by overr
 
 ### SliderOutput
 
-The `SliderOutput` component can be targeted with the `.react-aria-SliderOutput` CSS selector, or by overriding with a custom `className`. It provides a <TypeLink links={statelyDocs.links} type={statelyDocs.exports.SliderState} /> object to its render props, which can be used to customize the `className`, `style`, or `children`.
+The `SliderOutput` component can be targeted with the `.react-aria-SliderOutput` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
 
 ### SliderTrack
 
-The `SliderTrack` component can be targeted with the `.react-aria-SliderTrack` CSS selector, or by overriding with a custom `className`. It provides a <TypeLink links={statelyDocs.links} type={statelyDocs.exports.SliderState} /> object to its render props, which can be used to customize the `className`, `style`, or `children`.
+The `SliderTrack` component can be targeted with the `.react-aria-SliderTrack` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
```
LFDanLu, line 331: These need state tables

packages/react-aria-components/docs/Table.mdx
```diff
@@ -499,13 +499,15 @@ A `Column` can be targeted with the `.react-aria-Column` CSS selector, or by ove
 
 ### TableBody
 
-A `TableBody` can be targeted with the `.react-aria-TableBody` CSS selector, or by overriding with a custom `className`.
+A `TableBody` can be targeted with the `.react-aria-TableBody` CSS selector, or by overriding with a custom `className`. It supports the following states and render props:
```
LFDanLu, line 502: Mega nit: I see changes from `support` to `provide` and vice versa. Quite frankly I think having both is fine, but it seems to be the bulk of the changes so just wanna make sure we weren't trying to only use `support` in these blurbs. Or maybe there is some other reasoning to using each in various situations

packages/react-aria-components/docs/TextField.mdx
```diff
@@ -217,15 +217,17 @@ The states, selectors, and render props for each component used in a `TextField`
 
 ### TextField
 
-A `TextField` can be targeted with the `.react-aria-TextField` CSS selector, or by overriding with a custom `className`.
+A `TextField` can be targeted with the `.react-aria-TextField` CSS selector, or by overriding with a custom `className`. It provides the following render props and states:
+
+<StateTable properties={docs.exports.TextFieldRenderProps.properties} />
```
LFDanLu, line 222: Needs showOptional?

packages/react-aria-components/src/ComboBox.tsx
```diff
@@ -24,20 +24,25 @@ import {useCollection} from './Collection';
 
 export interface ComboBoxRenderProps {
   /**
-   * Whether the combobox is focused, either via a mouse or keyboard.
-   * @selector [data-focused]
```
LFDanLu, line 28: Ugh I'm having a hard time tracking down the discussion which decided that the wrapper here shouldn't have `isFocused`. I thought we wanted some top level wrapper render props to resolve https://github.com/adobe/react-spectrum/issues/4429?

LFDanLu (changes requested):

packages/react-aria-components/docs/Checkbox.mdx
```diff
@@ -250,7 +250,7 @@ Render props may also be used as children to alter what elements are rendered ba
 
 The states, selectors, and render props for `Checkbox` are documented below.
 
-<StateTable properties={docs.exports.CheckboxRenderProps.properties} />
+<StateTable properties={docs.exports.CheckboxRenderProps.properties} showOptional />
```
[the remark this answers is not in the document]
reidbarber: Not sure, maybe @devongovett can provide more info when he's back.

packages/react-aria-components/docs/Table.mdx
reidbarber: There wasn't a specific reason for this, we just needed `and render props` added in lots of places. I can review and make sure these are all the same. Not a big deal which we use, I think I prefer `provides` is a little more.

packages/react-aria-components/src/ComboBox.tsx
reidbarber: I'll double check this. I was also unsure about this change, so I put them all in the same commit to be easily reverted if needed.

reidbarber: I think the plan is to have a separate wrapper that can be used instead, so we don't need it on every component.

packages/react-aria-components/docs/Table.mdx
reidbarber: Went ahead and reverted a bunch of these changes for simplicity. I think it was fine as-is

ktabors: TableBody is only showing the state `isEmpty`, should we edit that sentence or are there missing props?

packages/react-aria-components/docs/TimeField.mdx
```diff
@@ -277,7 +277,9 @@ The states, selectors, and render props for each component used in a `TimeField`
 
 ### TimeField
 
-A `TimeField` can be targeted with the `.react-aria-TimeField` CSS selector, or by overriding with a custom `className`. It provides a <TypeLink links={statelyDocs.links} type={statelyDocs.exports.DateFieldState} /> object to its render props, which can be used to customize the `className`, `style`, or `children`.
+A `TimeField` can be targeted with the `.react-aria-TimeField` CSS selector, or by overriding with a custom `className`. It supports the following states and render props:
```
ktabors, line 280:
```suggestion
A `TimeField` can be targeted with the `.react-aria-TimeField` CSS selector, or by overriding with a custom `className`. It supports the following states:
```
TextField called `isDisabled` and `validationState ` states, and this one has those two and `state` so I think we should remove render props here.

packages/react-aria-components/docs/Table.mdx
ktabors, line 502: TableBody is only showing the state isEmpty, should we edit that sentence or are there missing props?

reidbarber: That is the only render prop for TableBody. I think we can keep the text to be consistent with other sections.

packages/react-aria-components/docs/TimeField.mdx
reidbarber: Since those are render props, I'm fine keeping the text. We have it both ways throughout different examples, and both are right. I didn't change them all in order to keep the PR simple.

LFDanLu (dismissed):

packages/react-aria-components/docs/Table.mdx
ktabors: My comment was that we aren't being consistent.

ktabors (approved):
LGTM

I was looking at the "styling" for all the RAC docs and noticed that the Menu example had `{...props}` to both the MenuTrigger and Menu. Not sure if that would be in scope for this PR.

packages/react-aria-components/docs/Checkbox.mdx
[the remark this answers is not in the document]
devongovett: I guess this was for `validationState`? I'm not sure why that was made optional in the first place. Maybe it shouldn't be?

LFDanLu (approved):

packages/react-aria-components/docs/Table.mdx
reidbarber: I think we can refine these in a follow-up if needed. Wanted to keep the wording changes minimal in here to make it easier to review.

packages/react-aria-components/src/RadioGroup.tsx
```diff
@@ -44,19 +44,9 @@ export interface RadioGroupRenderProps {
   isRequired: boolean,
   /**
    * The validation state of the radio group.
-   * @selector [aria-invalid]
+   * @selector [data-validation-state]
```
devongovett, line 47: Seems like this data attribute is missing on the DOM node?

packages/react-aria-components/docs/Meter.mdx
```diff
@@ -181,7 +181,7 @@ The selectors and render props for each component used in a `Meter` are document
 
 A `Meter` can be targeted with the `.react-aria-Meter` CSS selector, or by overriding with a custom `className`. It supports the following states and render props:
 
-<StateTable properties={docs.exports.MeterRenderProps.properties} />
+<StateTable properties={docs.exports.MeterRenderProps.properties} showOptional />
```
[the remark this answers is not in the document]
reidbarber: I think this was due to type error. I think I can probably update valueText from being optional to `string | undefined` instead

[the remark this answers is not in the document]
devongovett: Ah yeah I think that would make sense for ValidationState too.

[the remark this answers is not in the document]
reidbarber: Made a follow-up here: https://github.com/adobe/react-spectrum/pull/4794

PR #4757, opened 2023-07-11
Rsp dropzone

yihuiliao:
RSP DropZone is built on top of the RAC DropZone implementation so in terms of functionality and behavior, it should be pretty much the same besides the styling and extra banner that appears if the DropZone is filled. 

RSP DropZone takes two additional props: `replaceMessage` and `isFilled`. 

`replaceMessage` allows the user to decide what the message is presented in the banner, although it will default to ‘Drop file to replace’ if nothing is passed. 

`isFilled` is used to determine whether the DropZone has been dropped with valid data or not. This is used to determine when the banner should be shown (which is only when the DropZone is in a filled state). 

Previews are controlled by the user who determine what to show. This implementation simply provides the core structure of a DropZone with additional styling.

**Known Issues:** 

For keyboard interactions, if the FileTrigger is opened, the focus will return to the DropZone's visually hidden button and not the FileTrigger button. A ticket has already been made for this issue. 

Various platforms seem to have issues with announcing the filled message. This will be a follow-up. 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Accessibility: 

"Drop File to Replace" and "This is a custom message" should only be announced when the DropZone is in a filled state.

`with-draggable`: for screen reader, should announce "DropZone" and "Drop File to Replace" 
`with-draggable-text-context`: for screen reader, should announce "DropZone, Drag and Drop Here" and  "Drop File to Replace" 
`with-draggable-custom-aria-label`: for screen reader, should announce "custom label" and "Drop File to Replace"
`custom-banner-message`: for screen reader, should announce "DropZone" and "This is a custom message"


packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,85 @@
...
+  let domProps = filterDOMProps(props, {labelable: true});
+  delete domProps.id;
+
+  // Expose imperative interface for ref
+  useImperativeHandle(ref, () => ({
+    ...createFocusableRef(domRef, inputRef),
+    getInputElement() {
+      return inputRef.current;
+    }
+  }));
```
LFDanLu, line 45: tried printing out the refs for `ref.current.UNSAFE_getDOMNode()` and `ref.current.getInputElement()` but got undef for both. I think the DOM node ref isn't setup [here](https://github.com/adobe/react-spectrum/blob/acc48c862d66fe49f3fc3c3cbf923f89206c0c03/packages/react-aria-components/src/DropZone.tsx#L83-L91) (aka needs ref={ref}) not sure about the inputRef though. As it is now, FileTrigger needs to access the Slots context via useSlotProps or you need to setup a Provider similarly to how its done in the RAC DropZone or you can make RAC DropZone setup the imperative handle instead

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
```diff
@@ -0,0 +1,259 @@
...
+            <Button variant={'accent'}>Browse</Button>
+          </FileTrigger>
+        </Content>
+      </IllustratedMessage>
+      {isFilled && <img style={{width: '100px', position: 'absolute'}} src={filledSrc} alt="test" />}
+    </DropZone>
+  );
+}
+
+function Draggable() {
```
LFDanLu, line 222: Might be easier exporting and importing this from the aria DnD stories like its done for Droppable in ListViewDnD stories

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+import {AriaLabelingProps, DOMProps, FocusableRefValue, StyleProps} from '@react-types/shared';
+import {classNames, createFocusableRef, SlotProvider, useStyleProps} from '@react-spectrum/utils';
+import {DropOptions, mergeProps} from 'react-aria';
+import {filterDOMProps} from '@react-aria/utils';
+import {DropZone as RACDropZone} from 'react-aria-components';
+import React, {ReactNode, Ref, useImperativeHandle, useRef} from 'react';
+import styles from '@adobe/spectrum-css-temp/components/dropzone/vars.css';
+export interface SpectrumDropZoneProps extends Omit<DropOptions, 'getDropOperationForPoint' | 'ref' | 'hasDropButton'>, DOMProps, StyleProps, AriaLabelingProps {
```
reidbarber, line 20: I think we may want to extend RAC's `DropZoneProps` (with `'children'` omitted) here instead of `DropOptions`? Or at least see if that simplifies this.

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,85 @@
...
+                  styles,
+                  'spectrum-Dropzone-banner',
+                  styleProps.className
+                )
+              }>
+              {bannerMessage ? bannerMessage : 'Drop file to replace'}
+            </div>}
+          <SlotProvider
+            slots={{
+              illustration: {UNSAFE_className: classNames(styles, 'spectrum-IllustratedMessage')},
```
reidbarber, line 73: This should be a class name specific to dropzone, like `spectrum-Dropzone--illustratedMessage`, then we can style that instead of the `.spectrum-Dropzone[data-focus-visible] svg` selector currently used.

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
```diff
@@ -0,0 +1,259 @@
...
+  )
+};
+
+export const withLink = {
+  render: (args) => (
+    <DropZoneWithLink {...args} />
+  )
+};
+
+export const withButton = {
```
reidbarber, line 51: IMO we can remove the Button examples if there aren't any Spectrum designs for it. The Spectrum use case might be limited to a Link.

packages/@adobe/spectrum-css-temp/components/dropzone/skin.css
```diff
@@ -18,15 +18,33 @@ governing permissions and limitations under the License.
...
+    border-color: var(--spectrum-dropzone-border-color-selected-hover);
+
+    svg {
+      color: var(--spectrum-dropzone-illustratedmessage-illustration-color-key-focus);
+    }
+  }
+
+  &[data-focused] {
+    border-color: var(--spectrum-dropzone-border-color-selected-hover);
+  }
```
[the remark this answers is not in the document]
yihuiliao: i'm not too sure either. it hasn't been discussed with design so definitely worth bringing it up to them. for now, i think i'll have it match the focus-visible style and we can change it later according to what design says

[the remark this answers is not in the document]
reidbarber: I would lean towards not having the `data-focused` style, since it doesn't point the user towards any particular action they could take.

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
yihuiliao: i think this is something that we can maybe bring up with design and see if they see any use cases with a button? since it's unclear, i'll just leave it for now although you're right, currently there is no spectrum designs with a button, but we can easily choose to remove them later if needed

packages/@adobe/spectrum-css-temp/components/dropzone/skin.css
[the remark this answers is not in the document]
yihuiliao: well...technically they could paste but that's really the only action i can think of. i think it would be a bit weird if you were to click on the dropzone to paste something and nothing changed visually but also maybe it's worth reconsidering if that's an action we want to allow?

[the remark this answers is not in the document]
ktabors: Shouldn't this type of selection/focus be done with the FocusRing component?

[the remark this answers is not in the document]
yihuiliao: hmm well this styling really is only applied when you click on the dropzone with a mouse and i believe that the FocusRing component only applies for keyboard interactions so im not sure if that would fix things but let me know if im missing anything

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,93 @@
...
+  getInputElement(): HTMLInputElement | HTMLDivElement | null
+}
+
+function DropZone(props: SpectrumDropZoneProps, ref: Ref<DropZoneRef>) {
+  let {children, isFilled, bannerMessage, ...otherProps} = props;
+  let {styleProps} = useStyleProps(props);
+  let domRef = useRef<HTMLDivElement>(null);
+  let inputRef = useRef<HTMLInputElement>(null);
+
+  // Expose imperative interface for ref
```
yihuiliao, line 35: leaving imperative handle here for now. we can decide to remove it later but just gonna leave these questions here for us to think about for the future:
- Do we think that RAC components should have imperative refs? 
- Does the FileTriggers input store any important information that we’d like the user to be able to access or have any interactions that the user may want to trigger?

packages/@adobe/spectrum-css-temp/components/dnd/index.css
```diff
@@ -0,0 +1,33 @@
+/*
```
yihuiliao, line 1: created a new folder called dnd and put all dnd css in there because turns out the css for the dnd stories were in dropzone and i had completely messed up the css for the dnd stories up. let me know if we don't want to do this but yeah, just restores the dnd stories back to normal

yihuiliao:
also chose to remove focus styling based on other people's opinions so let me know what you think. we can still bring it up to design but wanted to see how other feel

packages/@adobe/spectrum-css-temp/components/dropzone/skin.css
```diff
@@ -18,27 +18,32 @@ governing permissions and limitations under the License.
...
     }
-    &.focus-ring {
-      border-color: var(--spectrum-dropzone-border-color-selected-hover);
-    }
-    &.is-dragged.focus-ring {
-      .spectrum-IllustratedMessage-illustration {
-        color: var(--spectrum-dropzone-illustratedmessage-illustration-color-selected);
-      }
+  }
+
```
yihuiliao, line 43: also chose to remove focus styling (and just keep focus visible) based on other people's opinions so let me know what you think. we can still bring it up to design but wanted to see how other feel

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,93 @@
...
+import {AriaLabelingProps, DOMProps, FocusableRefValue, StyleProps} from '@react-types/shared';
+import {classNames, createFocusableRef, SlotProvider, useStyleProps} from '@react-spectrum/utils';
+import {DropZoneProps, FileTriggerContext, Provider, DropZone as RACDropZone} from 'react-aria-components';
+import {mergeProps} from '@react-aria/utils';
+import React, {ReactNode, Ref, useImperativeHandle, useRef} from 'react';
+import styles from '@adobe/spectrum-css-temp/components/dropzone/vars.css';
+export interface SpectrumDropZoneProps extends DropZoneProps, DOMProps, StyleProps, AriaLabelingProps {
+  children: ReactNode,
+  isFilled?: boolean, 
+  bannerMessage?: string
```
reidbarber, line 22: More of a naming question for the team to discuss, but I think something like `replaceMessage` or `replaceLabel` might be better here.

reidbarber, line 22: These props need JSDoc comments.

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,93 @@
...
+              {children}
+            </SlotProvider>
+          </>
+        )}
+      </RACDropZone>
+    </Provider>
+  );
+}
+
+let _DropZone = React.forwardRef(DropZone);
```
reidbarber, line 92: JSDoc comment needed here

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+        if (items.length > 0) {
+          const urls = await Promise.all(items.map(async (item) => URL.createObjectURL(await item.getFile())));
+          const stringUrls = urls.map((url) => url.toString());
+          setFilledSrc(stringUrls);
+          setIsFilled(true);
+        }
+      }}
+      onDropEnter={action('onDropEnter')}
+      onDropExit={action('onDropExit')} 
+      onPaste={async (e) => {
```
reidbarber, line 81: Maybe a question for the team, but should we just have pasting trigger a drop automatically? Feels strange to force users to handle both.

packages/@react-spectrum/dropzone/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/// <reference types="css-module-types" />
+
+export {DropZone} from './DropZone';
```
reidbarber, line 15: We should also export types here.

dannify:
Can you add an example of a filled state in the stories? Maybe one with an image (you can use images from the CardView story examples). I think you will need to reduce the padding significantly, remove the illustrated message and center the content.

packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -36,13 +36,20 @@ export interface FileTriggerProps extends SlotProps, DOMProps, AriaLabelingProps
   onChange?: (files: FileList | null) => void
 }
 
-export const FileTriggerContext = createContext<ContextValue<FileTriggerProps, HTMLDivElement>>(null);
+interface FileTriggerContextValue extends FileTriggerProps {
+  inputRef?: RefObject<HTMLInputElement>
```
devongovett, line 40: I think this is a good question for RAC in general about how we want to expose refs for sub-elements. What was the use case for adding this now?

LFDanLu: @yihuiliao  If I remember correctly from our convo, this `inputRef` was added to support the RSP DropZone's imperativeRef since we wanted to match with how refs are treated in other RSP components (aka TextField), where the ref is attached to the wrapper but users can still access the inner input via `getInputElement`.

However, I wonder we even need to expose a way to get the input element for the FileTrigger. It makes sense for TextField since people often do stuff with the input element directly (aka read the value), but what about the FileTrigger's input? Will people need to access to it?

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+              let files = Array.from(e);
+              let urls = files.map(file => URL.createObjectURL(file));
+              setFilledSrc(urls);
+              setIsFilled(true);
+            }}>
+            <Link>Select a file</Link> from your computer
+          </FileTrigger>
+        </Content>
+      </IllustratedMessage>
+      {isFilled && <img alt="test" style={{width: '100px', position: 'absolute'}} src={filledSrc[0]} />}
```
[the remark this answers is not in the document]
reidbarber: I was thinking something along the lines of `object-fit: contain | cover`, but I guess apps could do that themselves.

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
reidbarber: Looks like we do. In that case, this story can be updated to just do `action('onPaste')`

packages/react-aria-components/src/FileTrigger.tsx
yihuiliao: Yup, that's correct. Personally, I'm with Daniel in that I'm not really sure we need to expose the input element for FileTrigger but there could be use cases that I might be missing.

packages/@react-spectrum/dropzone/src/DropZone.tsx
yihuiliao: I prefer `replaceMessage` over `replaceLabel` because it might sound like we're adding a label for accessibility which isn't the case. And with `replaceMessage`, I like how it is clear what the prop does but like you said, up for opinions.

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
yihuiliao: oops my bad! i'll change it to `action('onPaste')`

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,81 @@
...
+export interface SpectrumDropZoneProps extends DropZoneProps, DOMProps, StyleProps, AriaLabelingProps {
+  /** The content to display in the button. */
+  children: ReactNode,
+  /** Whether the dropzone has been filled. */
+  isFilled?: boolean, 
+  /** The message to replace the default banner message that is shown when the dropzone is filled. */
+  replaceMessage?: string
+}
+
+function DropZone(props: SpectrumDropZoneProps, ref: DOMRef<HTMLDivElement>) {
```
yihuiliao, line 28: When writing the docs for `DropZone` and `FileTrigger`, some changes were made that affected the imperative interface for refs in this PR. Notably, FileTrigger no longer has a wrapping div and context which was used to help expose the FileTrigger’s input. Additionally, it seems like the overall consensus is that an imperative ref for DropZone isn’t necessary and so with this latest push, I decided to remove it from the code.

reidbarber (commented):
The replace message doesn't meet color contrast requirements in dark theme, so we may need to consider a different background color.

Screenshot 2023-08-22 at 1 24 16 PM

Screenshot 2023-08-22 at 1 24 36 PM

snowystinger (commented):
I'd love for the stories, "custom-banner-message" and "with-button", to have some styling for the dropped asset like we'd expect to see in applications.

Will continue to review tomorrow

reidbarber (approved):
LGTM

packages/@react-spectrum/dropzone/test/DropZone.test.js
```diff
@@ -0,0 +1,213 @@
...
+          <Content>
+            <FileTrigger>
+              <Link>Select a file</Link>from your computer
+            </FileTrigger>
+          </Content>
+        </IllustratedMessage>
+      </DropZone>
+    );
+    let dropzone = getByTestId('foo');
+    expect(dropzone).toHaveAttribute('class', 'spectrum-Dropzone');
```
snowystinger, line 39: so this test makes sense in React Aria Components, but in RSP we don't usually check classnames as they aren't that important in a unit test, our Chromatic tests will show if these break

Instead, I suggest just moving right on to behavior testing

If there is some reason you think that the ref location is important, then I suggest just doing tests like these
```
dropzoneRef ==== tree.container.firstChild // most common, we want to make sure that the ref on the outer most part of our component goes to the outermost wrapping div, so will be resistant to changes in dom structure

dropzoneRef !== FileTriggerRef // similar to the previous check, just a little less information, maybe we don't care if a particular ref is the outer most, so long as it's not the input, so might be more common to check something like IllustratedMessageRef !== FileTriggerRef, however, IllustratedMessage already has tests, so this isn't really needed

dropzoneRef !== getByRole('fileinput') // if there isn't a ref on the input or something like that
```
 (https://testing-library.com/docs/react-testing-library/api/#container-1)

packages/@react-spectrum/dropzone/test/DropZone.test.js
```diff
@@ -0,0 +1,213 @@
...
+  });
+
+  it('should render an illustrated message as a child', () => {
```
snowystinger, line 42: this test should go away and instead be in chromatic

packages/@react-spectrum/dropzone/test/DropZone.test.js
```diff
@@ -0,0 +1,213 @@
...
+      </DropZone>
+    );
+
+    let heading = getByTestId('foo');
+    let content = getByTestId('bar');
+    expect(heading).toHaveAttribute('class', 'spectrum-IllustratedMessage-heading');
+    expect(content).toHaveAttribute('class', 'spectrum-IllustratedMessage-description');
+  });
+
+  it('should attach a ref to the dropzone if provided as a prop', () => {
```
snowystinger, line 62: this should be covered in the suggestion for the first test in this file if it's important

could be argued it should be placed in the RAC test file instead if it's important, in which case we wouldn't need it here

packages/@react-spectrum/dropzone/test/DropZone.test.js
```diff
@@ -0,0 +1,213 @@
...
+        type: 'dropenter',
+        x: 1,
+        y: 1
+      });
+
+      expect(dataTransfer.dropEffect).toBe('move');
+      expect(dropzone).toHaveAttribute('data-drop-target', 'true');
+
+      let banner = tree.getByText('Drop file to replace');
+      expect(banner).toHaveClass('spectrum-Dropzone--banner');
```
snowystinger, line 168: no need to test the classname

packages/@react-spectrum/dropzone/test/DropZone.test.js
```diff
@@ -0,0 +1,213 @@
...
+                  <Link>Select a file</Link>from your computer
+                </FileTrigger>
+              </Content>
+            </IllustratedMessage>
+          </DropZone>
+        </>
+      );
+      let dropzone = tree.getByTestId('foo');
+      let draggable = tree.getByText('Drag me');
+      expect(dropzone).toHaveClass('spectrum-Dropzone');
```
snowystinger, line 126: drop classname expect

snowystinger (changes requested):
Should VO be announcing the replaceMessage?

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
```diff
@@ -0,0 +1,251 @@
...
+  let [isFilled, setIsFilled] = useState(true);
+  let [filledSrc, setFilledSrc] = useState('https://i.imgur.com/DhygPot.jpg');
+
+  return (
+    <>
+      <DraggableImage />
+      <DropZone
+        {...props}
+        isFilled={isFilled}
+        UNSAFE_className={classNames(styles, 'is-filled')}
```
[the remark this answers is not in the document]
yihuiliao: this one should always be filled since the story really is just about showing some styling in a filled state

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,80 @@
...
+      {...styleProps as Omit<React.HTMLAttributes<HTMLElement>, 'onDrop'>}
+      aria-labelledby={isFilled ? messageId : null}
+      className={
+      classNames(
+        styles,
+        'spectrum-Dropzone',
+        styleProps.className
+      )} 
+      ref={domRef}>
+      {({isDropTarget}) => (
```
snowystinger, line 47: if we do banner display: none in CSS and then only set it to display block when it's drop target, then we don't need this render prop
we could also add a top level state `.spectrum-Dropzone--filled` so we could get rid of the conditional rendering below completely

packages/react-aria-components/src/DropZone.tsx
```diff
@@ -53,7 +53,9 @@ function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
 
   let textId = useSlotId();
   let ariaLabel = props['aria-label'] || 'DropZone';
-  let labelProps = useLabels({'aria-label': ariaLabel, 'aria-labelledby': textId});
+  let messageId = (isDropTarget && props['aria-labelledby']) ? props['aria-labelledby'] : null;
+  let ariaLabelledby = [textId, messageId].filter(str => str !== null && str !== undefined).join(' ');
```
snowystinger, line 57:
```suggestion
  let ariaLabelledby = [textId, messageId].filter(Boolean).join(' ');
```

snowystinger (approved):

snowystinger:
Still need to answer https://github.com/adobe/react-spectrum/pull/4757#pullrequestreview-1592626684

yihuiliao:
> Still need to answer [#4757 (review)](https://github.com/adobe/react-spectrum/pull/4757#pullrequestreview-1592626684)

so after working on this comment on Friday, i found that VO behaves differently for the different browsers so while it will announce the message in Safari and FireFox, it won't announce in Chrome. not sure if there's a way to work around that but it would be good to do some testing to see which combination of browsers + OS the screenreader will/won't announce this message.

packages/@react-spectrum/dropzone/stories/DropZone.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+            setIsFilled(true);
+          }
+        }}
+        onDropEnter={action('onDropEnter')}
+        onDropExit={action('onDropExit')} 
+        onPaste={action('onPaste')}>
+        <IllustratedMessage>
+          <Upload />
+          <Heading>
+            <Text slot="label">
```
yihuiliao, line 229: The stories have been updated so that for each there is some textual content used to label the drop zone (except for the `is-filled` story). Hopefully that will reduce confusion regarding if the story is making the correct announcements.

reidbarber (approved):

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+import {AriaLabelingProps, DOMProps, DOMRef, StyleProps} from '@react-types/shared';
+import {classNames, SlotProvider, useDOMRef, useStyleProps} from '@react-spectrum/utils';
+import {DropZoneProps, DropZone as RACDropZone} from 'react-aria-components';
+import {mergeProps, useId} from '@react-aria/utils';
+import React, {ReactNode} from 'react';
+import styles from '@adobe/spectrum-css-temp/components/dropzone/vars.css';
+
+export interface SpectrumDropZoneProps extends DropZoneProps, DOMProps, StyleProps, AriaLabelingProps {
```
LFDanLu, line 20: Might be some props we don't need here (DropZone's slotProps and the extra AriaLabelingProps) but we can double check when we see the PropsTable in the docs

packages/@react-spectrum/dropzone/src/DropZone.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <div
+        id={messageId}
+        className={
+          classNames(
+            styles,
+            'spectrum-Dropzone-banner',
+            styleProps.className
+          )
+        }>
+        {replaceMessage ? replaceMessage : 'Drop file to replace'}
```
LFDanLu, line 67: Note: will need translated version of the default string

LFDanLu (approved):
Just some extra things to note for the followup PR, otherwise LGTM

PR #4758, opened 2023-07-11
(DNM) Add TableView expandable rows docs

reidbarber:
## ✅ Pull Request Checklist:

## 📝 Test Instructions:

https://reactspectrum.blob.core.windows.net/reactspectrum/9bd8f9fba7c90ace28faa4fd98d682b8bf3f5467/docs/react-spectrum/TableView.html#expandable-rows


packages/@react-spectrum/table/docs/TableView.mdx
```diff
@@ -1904,3 +1904,102 @@ let rows = [
...
+
+You'll first need to enable a feature flag in your app to use this feature.
+
+```tsx example
+import {enableTableNestedRows} from '@react-stately/flags';
+
+enableTableNestedRows();
+```
+
+The example below enables expandable rows, and uses `defaultSelectedKeys` to expand rows with keys "1" and "2".
```
LFDanLu, line 1921: The spacing here is pretty narrow, will see about making it a bit bigger

packages/@react-spectrum/table/docs/TableView.mdx
```diff
@@ -1904,3 +1904,102 @@ let rows = [
...
+          <Cell>San Jose</Cell>
+          <Cell>-</Cell>
+          <Cell>983.4 thousand</Cell>
+        </Row>
+      </Row>
+    <Row key="4">
+      <Cell>Texas</Cell>
+      <Cell>Austin</Cell>
+      <Cell>29.53 million</Cell>
+    </Row>
```
LFDanLu, line 1949: Mega nit, I think it might be useful having a child row for Texas just to highlight that only a subset of keys have been specified for expansion.

packages/@react-spectrum/table/docs/TableView.mdx
```diff
@@ -1904,3 +1904,102 @@ let rows = [
   </TableBody>
 </TableView>
 ```
+
+## Expandable Rows <VersionBadge version="alpha" style={{marginLeft: 4, verticalAlign: 'bottom'}} />
+
+By default, TableView doesn't allow expandable rows, but this can be enabled using the `hasExpandableRows` prop.
+Use `defaultExpandedKeys` to provide a default set of expanded rows. Note that the value of the expanded keys must match the `key` prop of the Row.
```
LFDanLu, line 1911: Would it be valuable to highlight that "all" will expand all the rows? What about highlighting that a expanded child row will only be visible if its parent is also in the expandedKeys set?

LFDanLu (commented):
Some nits that I had, will go ahead and push a update. Lemme know what you think and if any of the info feels excessive

LFDanLu:
oh also what do you think about moving the entire section up before the props table?

packages/@react-spectrum/table/docs/TableView.mdx
```diff
@@ -1587,6 +1587,117 @@ function CustomDragPreviewExample() {
...
+import {enableTableNestedRows} from '@react-stately/flags';
+
+enableTableNestedRows();
+```
+<br />
+
+Use `defaultExpandedKeys` to provide a default set of expanded rows. Note that the value of the expanded keys must match the `key` prop of the Row and that a
+expanded child row will only be visible if its parent row key is also included within the expanded key set.
+
+A expandable row can be statically defined by including one or more Rows as children. Note that all child Rows must be placed after the last Cell within the Row.
```
yihuiliao, line 1605:
```suggestion
An expandable row can be statically defined by including one or more Rows as children. Note that all child Rows must be placed after the last Cell within the Row.
```

yihuiliao, line 1605: I noticed that sometimes the word "row" is capitalized and sometimes it isn't (particularly in the paragraph above). It would probably be good to keep it consistent one way or another. So like either "child Row" or "child row" rather than switching between the two.

LFDanLu (dismissed):

packages/@react-spectrum/table/docs/TableView.mdx
```diff
@@ -1587,6 +1587,117 @@ function CustomDragPreviewExample() {
...
+You'll first need to enable a feature flag in your app to use this feature.
+
+```tsx example
+import {enableTableNestedRows} from '@react-stately/flags';
+
+enableTableNestedRows();
+```
+<br />
+
+Use `defaultExpandedKeys` to provide a default set of expanded rows. Note that the value of the expanded keys must match the `key` prop of the Row and that a
```
yihuiliao, line 1602:
```suggestion
Use `defaultExpandedKeys` to provide a default set of expanded rows. Note that the value of the expanded keys must match the `key` prop of the Row and that an
```

packages/@react-spectrum/table/docs/TableView.mdx
LFDanLu: So the difference between the two was to try to distinguish between referencing the component Row/Cell and the row/cell generically. I'll add back ticks around the capital Rows to make it more clear

yihuiliao (approved):

LFDanLu:
Reminder not to merge this PR for now. Running idea is that the docs built from this PR will be used as the alpha docs for the team using expandable rows, then we'll revisit when doing the full release

packages/@react-spectrum/table/docs/TableView.mdx
```diff
@@ -1587,6 +1587,117 @@ function CustomDragPreviewExample() {
...
+You'll first need to enable a feature flag in your app to use this feature.
+
+```tsx example
+import {enableTableNestedRows} from '@react-stately/flags';
+
+enableTableNestedRows();
+```
+<br />
+
+Use `defaultExpandedKeys` to provide a default set of expanded rows. Note that the value of the expanded keys must match the `key` prop of the `Row` and that an
```
ktabors, line 1602: I got confused because this paragraph mixes how to enable expanded rows and that there is a feature flag. These feel like two different concepts.

packages/@react-spectrum/table/docs/TableView.mdx
```diff
@@ -1587,6 +1587,117 @@ function CustomDragPreviewExample() {
...
+expanded child row will only be visible if its parent row key is also included within the expanded key set.
+
+An expandable row can be statically defined by including one or more `Rows` as children. Note that all child `Rows` must be placed after the last `Cell` within the parent `Row`.
+
+The example below enables expandable rows, and uses `defaultSelectedKeys` to expand rows with keys "1" and "2".
```
ktabors, line 1607: If this were its own component we'd have a separate example for `defaultSelectedKeys`. This is probably okay.

We aren't covering controlled `selectedKeys`.

ktabors (commented):
Do we need to cover accessibility or keyboard navigation?

packages/@react-spectrum/table/docs/TableView.mdx
reidbarber: This should actually read `defaultExpandedKeys`, good catch

packages/@react-spectrum/table/docs/TableView.mdx
```diff
@@ -1587,6 +1587,117 @@ function CustomDragPreviewExample() {
 <CustomDragPreviewExample />
 ```
 
+## Expandable Rows <VersionBadge version="alpha" style={{marginLeft: 4, verticalAlign: 'bottom'}} />
+
+By default, TableView doesn't allow expandable rows, but this can be enabled by toggling the feature flag via `enableTableNestedRows` in your app and then providing the `UNSTABLE_allowsExpandableRows` prop to
+your TableView.
+
+```tsx example
+import {enableTableNestedRows} from '@react-stately/flags';
```
ross-pfahler, line 1596: can we add where in your app this should go? eg the root vs the file that expandable rows is used? or anywhere? does it matter?

reidbarber: I'll update this section with some clarification. Here is the entire explanation:

> This approach is the easiest because the flags can be accessed from anywhere. However, one downside is that flags are enabled for all instances of a component across the whole app, not for one specific table for example. We could do that with context, but then the flags would need to be wired to more places that aren't allowed to access classes (e.g. classes, non-hook functions, etc.). The flags also need to be set before anything is rendered in the app or things may break unexpectedly, so libraries like quarry should not be enabling flags, they should be enabled in apps. Like any other package, there must be only one copy installed or things will also not work.

From https://github.com/adobe/react-spectrum/pull/4617

reidbarber: So you'll want to do this once, likely at the root of your app since it will affect all instances and should be done before rendering the table.

PR #4759, opened 2023-07-12
#3578 Add the ability to add a data attribute to <HiddenSelect /> to handle false a11y positives

abitbetterthanyesterday:
Closes [#3578 ](https://github.com/adobe/react-spectrum/issues/3578)

This PR simply add the ability to expose a data-attribute on the `<HiddenElement />` of `useSelect`.

This is useful when using a tool such as AXE or pa11y giving you false positive - the data attribute is a good way to force the tool to ignore the false positive.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Use the following code (from the `useSelect` documentation example), ensuring that the `props` contains `{hasA11yIgnoreDataAttr: true}`
2. Ensure that the hidden select element has the data attribute `data-rsp-a11y-ignore`.
3. Ensure unit tests are passing (`yarn jest`)

```tsx
import {HiddenSelect, useSelect} from '@react-aria/select';
import {Item} from '@react-stately/collections';
import {useSelectState} from '@react-stately/select';

// Reuse the ListBox, Popover, and Button from your component library. See below for details.
import {ListBox, Popover, Button} from 'your-component-library';

function Select(props) {
  // Create state based on the incoming props
  let state = useSelectState(props);

  // Get props for child elements from useSelect
  let ref = React.useRef(null);
  let {
    labelProps,
    triggerProps,
    valueProps,
    menuProps
  } = useSelect(props, state, ref);

  return (
    <div style={{display: 'inline-block'}}>
      <div {...labelProps}>{props.label}</div>
      <HiddenSelect
        isDisabled={props.isDisabled}
        state={state}
        triggerRef={ref}
        label={props.label}
        name={props.name} />
      <Button
        {...triggerProps}
        buttonRef={ref}
        style={{height: 30, fontSize: 14}}>
        <span {...valueProps}>
          {state.selectedItem
            ? state.selectedItem.rendered
            : 'Select an option'
          }
        </span>
        <span
          aria-hidden="true"
          style={{paddingLeft: 5}}>
          ▼
        </span>
      </Button>
      {state.isOpen &&
        <Popover state={state} triggerRef={ref} placement="bottom start">
          <ListBox
            {...menuProps}
            state={state} />
        </Popover>
      }
    </div>
  );
}

<Select label="Favorite Color">
  <Item>Red</Item>
  <Item>Orange</Item>
  <Item>Yellow</Item>
  <Item>Green</Item>
  <Item>Blue</Item>
  <Item>Purple</Item>
  <Item>Black</Item>
  <Item>White</Item>
  <Item>Lime</Item>
  <Item>Fushsia</Item>
</Select>
```

## 🧢 Your Project:

N/A


packages/@react-aria/select/src/HiddenSelect.tsx
```diff
@@ -88,6 +92,7 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectProps, state: SelectSt
       size: state.collection.size,
       value: state.selectedKey ?? '',
       onChange: (e: React.ChangeEvent<HTMLSelectElement>) => state.setSelectedKey(e.target.value)
+      ['data-rsp-a11y-ignore']: props.hasA11yIgnoreDataAttr,
```
reidbarber, line 95:
```suggestion
      ['data-rsp-a11y-ignore']: props.hasA11yIgnoreDataAttr
```

packages/@react-aria/select/test/HiddenSelect.test.tsx
```diff
@@ -60,4 +60,12 @@ describe('<HiddenSelect />', () => {
     userEvent.selectOptions(select, '5');
     expect(onSelectionChange).toBeCalledWith('5');
   });
+
+  it('should add a data attribute data-rsp-a11y-ignore when specified in the props', () => {
+    render(
+      <HiddenSelectExample items={makeItems(5)} hiddenProps={{ hasA11yIgnoreDataAttr: true }} />
```
reidbarber, line 66:
```suggestion
      <HiddenSelectExample items={makeItems(5)} hiddenProps={{hasA11yIgnoreDataAttr: true}} />
```

packages/@react-aria/select/test/HiddenSelect.test.tsx
```diff
@@ -60,4 +60,12 @@ describe('<HiddenSelect />', () => {
...
+    );
+
+    expect(screen.getByRole('listbox', { hidden: true }).dataset.rspA11yIgnore).toBeTruthy()
```
reidbarber, line 69:
```suggestion
    expect(screen.getByRole('listbox', {hidden: true}).dataset.rspA11yIgnore).toBeTruthy();
```

packages/@react-aria/select/test/HiddenSelect.test.tsx
```diff
@@ -60,4 +60,12 @@ describe('<HiddenSelect />', () => {
...
+  })
```
reidbarber, line 70:
```suggestion
  });
```

packages/@react-aria/select/src/HiddenSelect.tsx
```diff
@@ -88,6 +92,7 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectProps, state: SelectSt
       size: state.collection.size,
       value: state.selectedKey ?? '',
       onChange: (e: React.ChangeEvent<HTMLSelectElement>) => state.setSelectedKey(e.target.value)
```
reidbarber, line 94: Looks like there's a missing comma:

```suggestion
      onChange: (e: React.ChangeEvent<HTMLSelectElement>) => state.setSelectedKey(e.target.value),
```

packages/@react-aria/select/test/HiddenSelect.test.tsx
reidbarber: and this might be able to be shortened to use `.toHaveAttribute('data-rsp-a11y-ignore')`

reidbarber (changes requested):
Thanks for the contribution!!

There were a few lint issues that you can see when you run `yarn lint`. Fixing those should fix the build.

reidbarber (commented):
Also, I think we can just always add the new data attribute, no prop needed to enable.

abitbetterthanyesterday:
@reidbarber I've made the `data-attribute` always available; there is no harm in doing so.
 I agree with you, there is no need to hide it behind a flag.
 
I've also merged in the latest master, but maybe you prefer it rebased?
Let me know.

Linting issues should be also resolved. (We'll see once the CI has finished!)

packages/@react-aria/select/docs/useSelect.mdx
```diff
@@ -520,3 +520,9 @@ content that is passed into the select.
 In right-to-left languages, the select should be mirrored. The arrow should be on the left,
 and the selected value should be on the right. In addition, the content of list options should
 flip. Ensure that your CSS accounts for this.
+
+### False positive on `<HiddenSelect />` in accessibility tools
+
+When using &lt;<TypeLink links={docs.links} type={docs.exports.HiddenSelect} />&gt; some accessibility tools such as AXE or pa11y might give a false positive.
+The option `hasA11yIgnoreDataAttr` exposes a data attribute `data-rsp-a11y-ignore` on the hidden select element that can be used to ignore the false positives.
```
reidbarber, line 527:
```suggestion
The data attribute `data-rsp-a11y-ignore` is included on the hidden select element and can be used to ignore the false positives in these tools.
```

reidbarber (changes requested):
Looks great! Just a small tweak to the docs to reflect the updates and I'm good to approve.

packages/@react-aria/select/docs/useSelect.mdx
reidbarber: Hope its ok, went ahead and committed this change since we are looking to get this merged 🚀

packages/@react-aria/select/src/HiddenSelect.tsx
```diff
@@ -134,7 +135,8 @@ export function HiddenSelect<T>(props: HiddenSelectProps<T>) {
         autoComplete={selectProps.autoComplete}
         name={name}
         disabled={isDisabled}
-        value={state.selectedKey ?? ''} />
+        value={state.selectedKey ?? ''}
+        data-rsp-a11y-ignore />
```
reidbarber, line 139: Looks like we only need this on the container h/t @LFDanLu 

```suggestion
        value={state.selectedKey ?? ''} />
```

packages/@react-aria/select/src/HiddenSelect.tsx
```diff
@@ -87,7 +87,8 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectProps, state: SelectSt
       name,
       size: state.collection.size,
       value: state.selectedKey ?? '',
-      onChange: (e: React.ChangeEvent<HTMLSelectElement>) => state.setSelectedKey(e.target.value)
+      onChange: (e: React.ChangeEvent<HTMLSelectElement>) => state.setSelectedKey(e.target.value),
+      ['data-rsp-a11y-ignore']: true
```
LFDanLu, line 91: Think we can get rid of this one too. I figure people can blanket filter out the container div for aXe and pa11y via `data-rsp-a11y-ignore` (its the one triggering the "ARIA hidden element must not be focusable or contain focusable elements" false positive) and thus we don't need the attribute on the select props. We could even move this line onto the `containerProps` so people using the hook instead of HiddenSelect benefit too.

Did you find that you needed the data attribute on the select element?

reidbarber (approved):

packages/@react-aria/select/src/HiddenSelect.tsx
reidbarber, line 91: Agreed, we can remove this from the hook.

```suggestion
      onChange: (e: React.ChangeEvent<HTMLSelectElement>) => state.setSelectedKey(e.target.value)
```

reidbarber (dismissed):

packages/@react-aria/select/test/HiddenSelect.test.tsx
```diff
@@ -60,4 +60,12 @@ describe('<HiddenSelect />', () => {
     userEvent.selectOptions(select, '5');
     expect(onSelectionChange).toBeCalledWith('5');
   });
+
+  it('should always add a data attribute data-rsp-a11y-ignore', () => {
+    render(
+      <HiddenSelectExample items={makeItems(5)} />
+    );
+
+    expect(screen.getByRole('listbox', {hidden: true}).dataset.rspA11yIgnore).toBeTruthy();
```
reidbarber, line 69:
```suggestion
    expect(screen.getByRole('listbox', {hidden: true})).toHaveAttribute('data-rsp-a11y-ignore');
```

LFDanLu (dismissed):
Thanks for the contribution! I went in and updated it so the data attribute is just on the container props and verified that we could actually use the attribute to suppress the false positive in our own storybook, hope you don't mind. Apologies about highjacking the PR, we wanted to try and get it merged for our testing session on Monday.

Lemme know if you had any thoughts about my comment [here](https://github.com/adobe/react-spectrum/pull/4759#discussion_r1264222127), happy to update in a followup if it turns out we need the attribute on more that just the container

LFDanLu:
GET_BUILD

abitbetterthanyesterday:
Not a worry at all, I'm down in Australia so I was in bed at that time!

Thank you for the finishing touches 🙌

.storybook/preview.js
```diff
@@ -14,7 +14,16 @@ export const parameters = {
   options: {
     storySort: (a, b) => a[1].kind === b[1].kind ? 0 : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
   },
-  a11y: {},
+  a11y: {
+    config: {
+      rules: [
+        {
+          id: 'aria-hidden-focus',
+          selector: 'body main *:not([data-rsp-a11y-ignore])',
```
majornista, line 22: Will this miss a HiddenSelect rendered within a Dialog that appears in a Portal at the end of the DOM after the `main`?

```suggestion
          selector: 'body *:not([data-rsp-a11y-ignore])',
```

majornista (changes requested):
Suggest change to rule selector for storybook.

.storybook/preview.js
LFDanLu: Ah true, good catch.

packages/@react-aria/select/docs/useSelect.mdx
```diff
@@ -520,3 +520,9 @@ content that is passed into the select.
 In right-to-left languages, the select should be mirrored. The arrow should be on the left,
 and the selected value should be on the right. In addition, the content of list options should
 flip. Ensure that your CSS accounts for this.
+
+### False positive on `<HiddenSelect />` in accessibility tools
+
+When using &lt;<TypeLink links={docs.links} type={docs.exports.HiddenSelect} />&gt; some accessibility tools such as AXE or pa11y might give a false positive.
+The data attribute `data-rsp-a11y-ignore` is included on the hidden select element and can be used to ignore the false positives in these tools.
```
LFDanLu, line 527: For the team, open to opinions on where this information should go. Perhaps we should add a new section in this docs page for accessibility gotchas (or simply accessiblity?)

reidbarber: I'm fine with where it is, but not against a dedicated accessibility section either.

reidbarber (approved):

packages/@react-aria/select/docs/useSelect.mdx
yihuiliao: +1 to Reid's comment. I think this is fine where it is but a new section for accessibility gotchas would be nice

LFDanLu: Alright, I'll approve this PR and make a followup one moving it to a separate accessibility section.

LFDanLu (approved):
LGTM, thanks once again for the contribution!

yihuiliao (dismissed):

packages/@react-aria/select/src/HiddenSelect.tsx
```diff
@@ -71,7 +71,8 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectProps, state: SelectSt
   return {
     containerProps: {
       ...visuallyHiddenProps,
-      'aria-hidden': true
+      'aria-hidden': true,
+      ['data-rsp-a11y-ignore']: true
```
devongovett, line 75: This probably should not have `rsp` in the name because it's not spectrum specific, this will also affect React Aria.

packages/@react-aria/select/docs/useSelect.mdx
devongovett: It definitely shouldn't be under "Internationalization" IMO, seems unrelated to that. Also we will need to cover this on all of the other select pages: RAC Select and RSP Picker. If it is going to be a general way for us to indicate false positives, we may also want to document it on a more general page such as the "Accessibility" one.

LFDanLu: good point about covering this on the other Select pages. By the "Accessibility" page I assume you are referencing [this](https://react-spectrum.adobe.com/react-aria/accessibility.html#accessibility)? If so, I think that is a good idea.

@abitbetterthanyesterday IMO this is creeping in scope a bit, would you like to take this on? I'd be perfectly happy making followup PR for this as well instead.

abitbetterthanyesterday:
Is there anything left I can do to get this in?
Sorry it has grown out of scope a bit so fell off the wagon 🙂

LFDanLu:
I'm hoping to get this reviewed by the team and merged in as is by early next week, then will open a followup PR shortly after addressing the rest of the comments.

reidbarber (approved):
LGTM

packages/@react-aria/select/docs/useSelect.mdx
ktabors, line 527:
```suggestion
The data attribute `data-a11y-ignore` is included on the hidden select element and can be used to ignore the false positives in these tools.
```
It got renamed right?

ktabors:
GET_BUILD

devongovett (approved):
thanks for the contribution!

packages/@react-aria/select/docs/useSelect.mdx
LFDanLu: I'll update in a followup

PR #4777, opened 2023-07-17
RAC date - fixes the events and invalid examples crashing

ktabors:
Found in testing.

If you went to one of the RAC date components "Events" or "Invalid" examples and deleted an hour, minute, date, or month segment to make it invalid, the example would crash.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Goto the RAC DateField, DatePicker, DateRangePicker, and TimeField docs. 
For each of these go to the "Events" and "Invalid" examples. 
Delete a day, month, year, hour, or minute segment to make the date/time invalid.
The example should not crash.

## 🧢 Your Project:
RSP


yihuiliao:
for timefield, it doesn't crash anymore if you try deleting a segment, but when i delete a segment and try entering a new time it'll crash. 

and also, i noticed that it'll crash in RSP DateRangePicker docs so was wondering if you could add the same change you made to RAC DateRangePicker to RSP DateRangePicker

packages/@react-aria/datepicker/docs/useTimeField.mdx
```diff
@@ -264,7 +264,7 @@ function Example() {
   return (
     <>
       <TimeField label="Time" value={date} onChange={setDate} />
-      <p>Selected date and time: {formatter.format(date.toDate())}</p>
+      <p>Selected date and time: {(date?.toDate && formatter.format(date.toDate())) || (date && date.toString()) || '--'}</p>
```
ktabors, line 267: Yihui noticed created a valid time after it's not a valid Date/Time caused an error. The code above fixes that. The issue is that TimeField loses the date portion when the state no longer has a valid DateTime and becomes just a Time. This might be related to https://github.com/adobe/react-spectrum/pull/4718#issuecomment-1615205039. I'm not sure if we want to fix what Yihui found via the code above where we display a valid Date or Time or would we rather prevent TimeField from losing the Date portion when it becomes invalid.

yihuiliao: I think ideally we'd like to prevent TimeField from losing the Date portion when it becomes invalid, but I'm not sure when that fix will happen. 

I will say that this fix does prevent the example from crashing but agree that it is kinda a workaround the actual problem. I guess it's better than it crashing, but yeah it's a little strange since it first starts off as a DateTime but after deleting the value and re-entering a value it just becomes a Time. You can see it in the way the value is formatted too.

packages/@react-aria/datepicker/docs/useDateRangePicker.mdx
```diff
@@ -543,7 +543,7 @@ function Example() {
   return (
     <>
       <DateRangePicker label="Date range" value={range} onChange={setRange} />
-      <p>Selected date: {formatter.formatRange(range.start.toDate(getLocalTimeZone()), range.end.toDate(getLocalTimeZone()))}</p>
+      <p>Selected date: {range ? formatter.formatRange(range.start.toDate(getLocalTimeZone()), range.end.toDate(getLocalTimeZone())) : '--'}</p>
```
yihuiliao, line 546: Do we need the '--' for when the value is null? I think it would be okay if it were just left blank (and this goes for the other ones too). I guess it depends if we want to indicate visually that there is no value besides just leaving it blank but open to opinions

ktabors: We were doing the '--' in RSP, so I replicated that to the other examples. I liked that the '--' is like an empty segment, but non-date component examples are more likely to have nothing.

yihuiliao: Ahh I see...thanks for letting me know. In that case, I guess we should just leave it as '--' like you had

packages/@react-aria/datepicker/docs/useTimeField.mdx
LFDanLu: I think this is fine for now, we can try to address this when we do the refactor mentioned in #4718.

LFDanLu (approved):

yihuiliao (approved):

PR #4778, opened 2023-07-17
Allow custom aria-describedby on checkbox and radio group items

devongovett:
The `aria-describedby` set by the group was overriding the manually provided one from props. Now they get combined. I kinda just chose an order here (props one comes first), but can change it if there is a good reason.


ktabors (approved):
The code looks fine. I couldn't find a docs or storybook example where we do this.

dannify (approved):

PR #4779, opened 2023-07-17
Add TextArea component to RAC

### LINKED ISSUE #4595: TextArea component is missing from react-aria-components
### Discussed in https://github.com/adobe/react-spectrum/discussions/4594

<sup>Originally posted by **eashish93** May 25, 2023</sup>
Playing with react-aria-components for a mvp project and found that `TextArea` component is missing from the doc. In the meantime I'm using my own created with hooks version. It's strange it's not in alpha because it's literally a copy of `Input` element. 
</div>

devongovett:
Closes #4595

This adds a `TextArea` component that can be used in place of `Input` within a `TextField` for multi-line text entry. It's basically a copy of `Input` put renders a `<textarea>` and accepts different prop types matching that HTML element. Right now it's only supported inside `TextField` but we could potentially add support for it within other components that currently use `Input` in the future.


reidbarber (dismissed):
LGTM

ktabors (dismissed):

PR #4785, opened 2023-07-18
Table column resizing support for RAC

devongovett:
This adds support for table column resizing to RAC. In order to keep bundle size down, and preserve the ability to use native HTML table layout when the columns are not resizable, resizing is only enabled when the table is wrapped in a `<ResizableTableContainer>`. This container also acts as a scrollable region and uses a resize observer to measure the size of the table. The `onResize` events are also fired here.

The `Column` component adds `width` props for controlling the width using the same units as supported in spectrum. These props only work when wrapped in a `ResizableTableContainer`. I guess we could make pixels and percentages work with CSS, but I don't know of a way to implement fr units with css so for now the props are only useful when resizing is enabled.

To make a column resizable, render a `<ColumnResizer>` element inside the column. You can also render a MenuTrigger like we have in Spectrum and programmatically trigger resizing for a column via the `startResize` function passed to the render props. This can also be done for sorting with the `sort` function as a convenience.

The API is a bit different because of the wrapper that enables additional functionality, but it is kinda similar to the `useDragAndDrop` API in spirit in that it injects additional hooks to avoid adding them to the bundle when not used. Feedback welcome!

See inline comments for more details about the other changes in the PR.


packages/@react-aria/table/src/useTableColumnResize.ts
```diff
@@ -244,24 +251,15 @@ export function useTableColumnResize<T>(props: AriaTableColumnResizeProps<T>, st
...
-            // useMove calls onMoveStart for every keypress, but we want resize start to only be called when we start resize mode
-            // call instead during focus and blur
-            startResize(item);
-            state.tableState.setKeyboardNavigationDisabled(true);
-          }
-        },
         onBlur: () => {
           endResize(item);
-          state.tableState.setKeyboardNavigationDisabled(false);
         },
```
devongovett, line 263: The changes in this file are basically to not assume that if a `triggerRef` is provided that the resizer isn't visible. In the docs example, we have a menu but also the resizer is always visible. Instead, we store whether the resizer itself was focused when resizing starts. If so, focus stays on the resizer when resizing ends, otherwise it goes to the triggerRef.

packages/@react-spectrum/table/src/TableView.tsx
```diff
@@ -914,39 +912,8 @@ function ResizableTableColumnHeader(props) {
...
-      timeout.current = setTimeout(focusResizer, 0);
-    }
-    prevResizingColumn.current = resizingColumn;
-  }, [resizingColumn, column.key, isMobile, onFocusedResizer, resizingRef, prevResizingColumn, timeout]);
-
-  // eslint-disable-next-line arrow-body-style
-  useEffect(() => {
-    return () => clearTimeout(timeout.current);
-  }, []);
-
```
devongovett, line 949: This is handled in the hook now. The onFocusedResizer moved to the onFocus event of the resizer itself.

packages/@react-stately/table/src/TableColumnLayout.ts
```diff
@@ -93,16 +93,16 @@ export class TableColumnLayout<T> {
       let frKey;
       let frValue;
       minWidths.set(column.key, this.getDefaultMinWidth(collection.columns[i]));
-      if (col !== column.key && !column.column.props.width && !isStatic(uncontrolledWidths.get(column.key))) {
+      if (col !== column.key && !column.props.width && !isStatic(uncontrolledWidths.get(column.key))) {
```
devongovett, line 96: These nodes are already columns, so no need to do `column.column`. RAC doesn't have these references either.

packages/react-aria-components/docs/Table.mdx
```diff
@@ -999,6 +999,471 @@ Use the `renderEmptyState` prop to customize what the `TableBody` will display i
...
+```tsx example
+import {ResizableTableContainer, ColumnResizer} from 'react-aria-components';
+
+<ResizableTableContainer>
+  <Table aria-label="Table with resizable columns">
+    <TableHeader>
+      {/*- begin highlight -*/}
+      <Column id="file" isRowHeader maxWidth={500}>
+        <div className="flex-wrapper">
+          <span tabIndex={-1} className="column-name">File Name</span>
```
devongovett, line 1025: Not sure about this. The tabIndex is so that the name can be focused, but idk if this needs a role or something. It seemed to announce ok in VoiceOver (as a group). Not sure if we should explain this somewhere, or even provide a component for this?

packages/@react-aria/table/src/useTableColumnResize.ts
LFDanLu: I'll have to go back and revisit the original PR where this assumption was first made, forget if there was a particular reason we made that assumption. This does look like it works for all the various menu with/without always rendered resizer though, so thats good

packages/react-aria-components/docs/Table.mdx
LFDanLu, line 1: Some minor things I noticed when testing on Android Chrome:
- as you make the column wider, it affects the top search bar + theme button placement
- clicking the column menu headers to open the dropdown causes the entire page to jump
- you can shrink the columns in the column header menu example to the point where the resizer becomes covered

packages/@react-stately/table/src/TableColumnLayout.ts
LFDanLu: I feel like Rob had a reason for `column.column`, trying to dig up the original reason.

packages/react-aria-components/docs/Table.mdx
LFDanLu: Not sure if it needs a role either, it just announced the span text in TalkBack without a role. I think we should explain it at the very least in passing or as a comment at the very least. Making it a component might be a bit overkill, but I could see an argument for making one if we need it for other situations as well.

LFDanLu (changes requested):
Looks really good, I think the `ResizableTableContainer` wrapper is perfectly fine even though it is a bit different. Only some minor things I noticed when testing the docs + opinions

packages/react-aria-components/docs/Table.mdx
devongovett: Good catch! The first one was weird. Seems you need `position: relative` on the container to prevent the table from causing the page to change width. Once that was fixed it fixed the second one which was due to the horizontal scrolling/zoom. Also fixed the last one by truncating the column name.

packages/react-aria-components/docs/Table.mdx
devongovett: cc @majornista is it valid to have a focusable element without a role?

packages/@react-stately/table/src/TableColumnLayout.ts
LFDanLu: Can't seem to find anything in my digging. This should be fine since tests/functionality still passes, will ask him when he gets back.

LFDanLu (approved):
LGTM

reidbarber (approved):
LGTM, tested behavior on desktop and mobile. API looks good

PR #4796, opened 2023-07-20
Fix "Edit this page" link on production docs build

reidbarber:
Currently on production docs, the "Edit this page" link used is `/temp/{hash}/{correct_link}`, so this should fix it to be the correct link.

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/dev/docs/src/Layout.js
+++ b/packages/dev/docs/src/Layout.js
@@ -123,6 +123,7 @@ function Page({children, currentPage, publicUrl, styles, scripts, pathToPage}) {
   let title = currentPage.title + (!INDEX_RE.test(currentPage.name) || isBlog ? ` – ${pageSection}` : '');
   let hero = (parts.length > 1 ? HERO[parts[0]] : '') || heroImageHome;
   let heroUrl = `https://${TLD}/${currentPage.image || path.basename(hero)}`;
+  let githubLink = pathToPage.startsWith('/tmp/') ? pathToPage.replace('/tmp/', '').substring(pathToPage.indexOf('/')) : pathToPage;
 
   return (
     <html
@@ -198,7 +199,7 @@ function Page({children, currentPage, publicUrl, styles, scripts, pathToPage}) {
         <meta property="og:image" content={heroUrl} />
         <meta property="og:description" content={description} />
         <meta property="og:locale" content="en_US" />
-        <meta data-github-src={pathToPage} />
+        <meta data-github-src={githubLink} />
         <script
           type="application/ld+json"
           dangerouslySetInnerHTML={{__html: JSON.stringify(


packages/dev/docs/src/Layout.js
```diff
@@ -123,6 +123,7 @@ function Page({children, currentPage, publicUrl, styles, scripts, pathToPage}) {
   let title = currentPage.title + (!INDEX_RE.test(currentPage.name) || isBlog ? ` – ${pageSection}` : '');
   let hero = (parts.length > 1 ? HERO[parts[0]] : '') || heroImageHome;
   let heroUrl = `https://${TLD}/${currentPage.image || path.basename(hero)}`;
+  let githubLink = pathToPage.startsWith('/tmp/') ? pathToPage.replace('/tmp/', '').substring(pathToPage.indexOf('/')) : pathToPage;
```
yihuiliao, line 126: correct me if im wrong but i think the value of `pathToPage.indexOf('/')` will be incorrect since it won't be using the returned value of` pathToPage.replace('/tmp/', '')`. if you calculate it this way, you'll still get `/{hash}/{correct_link}`

reidbarber: Ahh good catch, I had this multiline originally, and didn't catch that when condensing it. Thanks!

ktabors (dismissed):

majornista (dismissed):

ktabors (approved):

packages/dev/docs/src/Layout.js
```diff
@@ -123,6 +123,11 @@ function Page({children, currentPage, publicUrl, styles, scripts, pathToPage}) {
   let title = currentPage.title + (!INDEX_RE.test(currentPage.name) || isBlog ? ` – ${pageSection}` : '');
   let hero = (parts.length > 1 ? HERO[parts[0]] : '') || heroImageHome;
   let heroUrl = `https://${TLD}/${currentPage.image || path.basename(hero)}`;
+  let githubLink = pathToPage;
```
yihuiliao, line 126: so the current url that the edit page bring us to is something like `.../tmp/{hash}/docs/@react-spectrum/docs/{the result of the url}` and the resulting string from this logic would be `docs/{the rest of the stuff}`. but i think that would still result in a 404 because the url to actually get to the docs is `packages/{the rest of the stuff}` so i think we'd want to replace `docs` with `packages`. 

you can see what i mean here. so this is the url that the edit button currently bring you to: https://github.com/adobe/react-spectrum/tree/main//tmp/38a4410cad4337bcf8bd79cde4df1582/docs/react-aria-components/docs/Button.mdx

whereas this is the url that we would want instead. 
https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/docs/Button.mdx

and this is the url we would presumably get from this pr:
https://github.com/adobe/react-spectrum/tree/main/docs/react-aria-components/docs/Button.mdx

reidbarber: Good catch, should be fixed now

yihuiliao (approved):
lgtm

ktabors (approved):

PR #4821, opened 2023-07-25
fix[RAC]: Update tab navigation for arrow keys

### LINKED ISSUE #4140: Tabs: screen reader users perceive no distinction between horizontal and vertical orientation
# 🐛 Bug Report 

## 🤔 Expected Behavior

Since we can't rely on screen readers announcing the tablist orientation to determine whether to use ArrowUp/ArrowDown instead of ArrowLeft/ArrowRight. Both ArrowUp and ArrowLeft should work to navigate to the previous tab, and both ArrowDown and ArrowRight should work to navigate to the next tab.

## 😯 Current Behavior

With focus on a Tab, a screen reader user would have to guess the orientation to figure out which arrow keys will work to navigate to the next or previous Tab.

## 🌍 Your Environment

| react-spectrum   | [@adobe/react-spectrum@3.25.0](https://github.com/adobe/react-spectrum/releases/tag/%40adobe%2Freact-spectrum%403.25.0)
| Browser          | Chrome Version 110.0.5481.177 (Official Build) (arm64)

## 🧢 Your Company/Team

Adobe/Accessibility

mariareedstrom:
Resolved Tabs issue with screen reader users not perceiving distinction between horizontal and vertical orientation. 

TabsKeybordDelegate now returns the previous tab for LeftArrow and UpArrow, and next tab for RightArrow and DownArrow regardless of tab orientation being horizontal or vertical. 

Followed notes from @LFDanLu per [#4140, comment](https://github.com/adobe/react-spectrum/issues/4140#issuecomment-1611881150). Updated documentation. Verified functionality in Storybook.
Updated tests to cover. 

Closes #4140 

## ✅ Pull Request Checklist:

### Visual Test
1. Open Storybook
2. Under React Aria Components select Tabs
3. Under Tabs select orientation flip
4. With a ‘ltr’ language selected use left arrow and up arrow to select previous tab
5. With a ‘ltr’ language selected, use right arrow and down arrow to select next tab
6. Change Storybook locale to Arabic 
7. With a ‘rtl’ language selected use right arrow and up arrow to select previous tab
8. With a ‘rtl’ language selected, use left arrow and down arrow to select next tab
9. Change orientation from horizontal to vertical 
10. Change Storybook locale to English 
11. With a ‘ltr’ language selected use left arrow and up arrow to select previous tab
12. With a ‘ltr’ language selected, use right arrow and down arrow to select next tab
13. Change Storybook locale to Arabic 
14. With a ‘rtl’ language selected use left arrow and up arrow to select previous tab
15. With a ‘rtl’ language selected, use right arrow and down arrow to select next tab

### Unit Test

Run updated test in `react-spectrum/tabs/test/Tabs.test.js`

- 'allows user to change tab item select via arrow keys with horizontal tabs'
- 'allows user to change tab item select via arrow keys with vertical tabs'

## 🧢 Your Project:
N/A


mariareedstrom:
@LFDanLu  The orientation variable is no longer used in the code, unsure of plans for future use. Keeping it unused causes Typescript issues, which can be suppressed. What do you think?

packages/@react-aria/tabs/src/TabsKeyboardDelegate.ts
```diff
@@ -21,45 +21,31 @@ export class TabsKeyboardDelegate<T> implements KeyboardDelegate {
 
   constructor(collection: Collection<T>, direction: Direction, orientation: Orientation, disabledKeys: Set<Key> = new Set()) {
     this.collection = collection;
-    this.flipDirection = direction === 'rtl' && orientation === 'horizontal';
```
LFDanLu, line 24: > @LFDanLu The orientation variable is no longer used in the code, unsure of plans for future use. Keeping it unused causes Typescript issues, which can be suppressed. What do you think?

I think we'll want to keep `this.flipDirection = direction === 'rtl' && orientation === 'horizontal';` here to keep ArrowLeft/Right behavior in vertical Tabs in LTR/RTL locales in line with similar components (ActionGroup).

mariareedstrom: Understood, I'll update it to match the behaviors of similar components. However, in RTL left is next for horizontal orientation, but the semantics flip when in vertical orientation, where left then means previous. This seems confusing to the end user. Would you suggest I open an RFC for this?

mariareedstrom: Updated to match similar component behaviors. 🙂

LFDanLu: > Understood, I'll update it to match the behaviors of similar components. However, in RTL left is next for horizontal orientation, but the semantics flip when in vertical orientation, where left then means previous. This seems confusing to the end user. Would you suggest I open an RFC for this?

Good point, this makes sense to me and got me revisiting the logic. Looks like we actually had this working [here](https://github.com/adobe/react-spectrum/pull/223) where ArrowLeft/ArrowUp always moved to the previous item and ArrowRight/ArrowDown always moved to the next item in RTL (can be observed [here](https://reactspectrum.blob.core.windows.net/reactspectrum/046831ed00bf013c6f728bf474b270bf5656b208/storybook/index.html?path=/story/buttongroup--default)). However, looks like the behavior changed in [this PR](https://github.com/adobe/react-spectrum/pull/339) to be what it currently is, presumably as a fix for [something separate](https://github.com/adobe/react-spectrum/pull/339/commits/c2c46911ce92cd478b0945492029d93869c905b8). This feels like the wrong behavior as you pointed out, would you mind opening a separate issue for this? 

For this PR, we'll keep it as is to avoid scope creep and so we can discuss with the rest of the team if the current behavior is incorrect.

LFDanLu:
GET_BUILD

packages/@react-aria/tabs/docs/useTabList.mdx
```diff
@@ -319,7 +319,8 @@ tab selection.
 
 ### Orientation
 
-By default, tabs are horizontally oriented. The `orientation` prop can be set to `vertical` to change this. This affects keyboard navigation. You are responsible for styling your tabs accordingly.
+By default, tabs are horizontally oriented. The `orientation` prop can be set to `vertical` to change this. This does not affect keyboard navigation. Regardless of orientation, both ArrowUp and ArrowLeft navigate to the previous tab while ArrowDown and ArrowRight navigate to the next tab. Note that in rtl ArrowLeft and ArrowRight actions are flipped in `horizontal` orientation. 
+You are responsible for styling your tabs accordingly.
```
LFDanLu, line 323:
```suggestion
By default, tabs are horizontally oriented. The `orientation` prop can be set to `vertical` to change this. This does not affects keyboard navigation. You are responsible for styling your tabs accordingly.
```

Mind omitting the details of the keyboard navigation here and in the other docs pages? IMO we can add those details once we settle on the correct RTL behavior

LFDanLu (changes requested):
I'll file a ticket for the RTL keyboard navigation, just one small suggestion for the docs but otherwise the behavior looks good to me.

LFDanLu:
https://github.com/adobe/react-spectrum/issues/4896 for the overall RTL keyboard nav experience in ActionGroup btw

packages/@react-aria/tabs/docs/useTabList.mdx
mariareedstrom: Updated the docs to match your suggestion.

mariareedstrom:
I'll check out #4896. Thanks for opening that!

LFDanLu (approved):

reidbarber (approved):
LGTM, thanks!

PR #4835, opened 2023-07-27
Testing documentation and util RFC

LFDanLu:
This RFC proposes a three part improvement to the test writing experience for consumers of React Spectrum/Aria Components. It will cover possible additions to our documentation for test writing education and gotchas, discuss a list of low level helper/utility function to expose that can aid in test setup/interactions, and propose a new set of high level utility classes that aim to simplify writing various common component interactions during testing.

[View Rendered RFC](https://github.com/adobe/react-spectrum/blob/test_util_rfc/rfcs/2023-rsp-testing-utils.md)


rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+These are non-trivial issues for maintainers of the library and have resulted in various internal testing utils and patterns, but things like common component interactions are typically duplicated across test files instead of being centralized. Explicit documentation of various testing gotchas doesn't really exist and has become internalized knowledge instead.
+
+We hope that documenting these testing gotchas externally will reduce the test debugging frustration, and that exposing testing utilities for common test setup and component interactions will simplify the overall test writing experience, leading towards easier adoption.
+
+
+## Detailed Design
+
+As mentioned earlier, there are three proposed parts to this initiative:
+
+1. Documentation of testing gotchas/common issues
```
reidbarber, line 33: I think this would be really helpful. Maybe the React Spectrum and React Aria sections of the docs can have their own separate testing pages.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+2. Create/expose low level util functions for basic test setup and interactions
+
+Some mocks we use feel like they could be applied universally, such as the mock for `window.screen`'s width when simulating a mobile device. We could expose this as `simulateMobile`/`simulateDesktop` so that users don't have to know the nitty gritty of what breakpoint we use when determining whether to render the mobile version of a React Spectrum component. Another example of a lower level util would be be for triggering a long press event. Instead of needing to know how long it takes before a pointerDown event is considered a long press or that you need to mock support for `PointerEvent` globally, we could provide our internal `triggerLongPress` test util alongside with `installPointerEvent` for mocking `PointerEvent` support. A drag and drop (via mouse and keyboard) interaction utility is another good candidate, as is a render wrapper that wraps arbitrary children in a Provider.
```
reidbarber, line 47: Another idea is maybe a TestProvider that includes this logic, if we end up having a lot of setup applied in most places.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+In general, there are several internal test utils [here](https://github.com/adobe/react-spectrum/blob/07f673acb2d68144156df3aa0db6e91810b12c67/packages/dev/test-utils/src/events.ts) and in our tests that we should consider exposing that handle issuing/simulating the press/touch events that would happen in the browser. Ideally, using the `user-event` library would eliminate the need for most, if not all, of the interaction related util functions, but has proved insufficient in various cases (like for long press). We'll have to convert our own tests to use the `user-event` v14 and see if any other internal test utils can be replaced entirely.
+
+These util functions would live in one of two packages perhaps, `@react-aria/test-utils` and `@react-spectrum/test-utils`. The differentiating factor would be if the util is specific to our React Spectrum components (`simulateMobile`/`simulateDesktop`) or not (`triggerLongPress`).
```
reidbarber, line 51: +1 to two packages

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+    <Picker label="Test" data-testid="test" onSelectionChange={onSelectionChange} onOpenChange={onOpenChange}>
+      <Item key="one">One</Item>
+      <Item key="two">Two</Item>
+      <Item key="three">Three</Item>
+    </Picker>
+  </Provider>
+);
+
+let picker = new SelectTester({element: screen.getByTestId('test'), timerType: 'real'});
+await picker.selectOption('Three');
```
reidbarber, line 75: I think APIs like this would be really useful, and a huge DX improvement.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Backwards Compatibility Analysis
+
+This is a backwards compatible change since no existing test utils are exposed from our library today.
+
+## Alternatives
+
+Unknown, haven't done research here yet.
+
+## Open Questions
```
reidbarber, line 100: I think we may also want to do some research into what testing libraries people are using most frequently.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+* What should be exposed as a test util vs documented? Since it will take a longer amount of time required to spin up the test util classes, should we expose the interaction util functions we already have even though they will be replaced by the test util classes?
+* Where should these utils live? Exported from the monopackage as well? Or separated into its own test-util package? Re-exported from existing packages (react-aria-components, react-aria, react-spectrum)?
```
reidbarber, line 103: Exporting from the existing monopackages seems fine, unless we're concerned with crowding the namespace.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+* Do we have a target version of `user-event` library? Right now we are on 12, but feels like we should move to 14 for broader pointer/keyboard support
+* Feedback for test util classes: https://github.com/adobe/react-spectrum/pull/4836/files#diff-0dc4808158bbefa842fb382ea67f1b9cfa5f35f50496133afca79d908fc49209. Still a WIP as I poke around to get the actual listbox tied to the picker and add examples of other test util classes
```
reidbarber, line 105: API looks great. I think there may be a question on whether we want these to support environments other that jest as well if possible.

reidbarber (approved):
I think these utilities will be super useful. Hope we can get out some alphas at least soon.

rfcs/2023-rsp-testing-utils.md
LFDanLu: I think that is a good idea, perhaps the TestProvider could even take options/util functions for enabling common pre-test setup and tear down, will play around with that idea

rfcs/2023-rsp-testing-utils.md
LFDanLu: Yep, I hope people will chime in with what library they are using, but will probably stick with jest+react testing library for the first iteration

rfcs/2023-rsp-testing-utils.md
LFDanLu: The namespace is really my main concern, might make IDE autocomplete too busy

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Alternatives
+
+Unknown, haven't done research here yet.
+
+## Open Questions
+
+* What should be exposed as a test util vs documented? Since it will take a longer amount of time required to spin up the test util classes, should we expose the interaction util functions we already have even though they will be replaced by the test util classes?
+* Where should these utils live? Exported from the monopackage as well? Or separated into its own test-util package? Re-exported from existing packages (react-aria-components, react-aria, react-spectrum)?
+* Do we have a target version of `user-event` library? Right now we are on 12, but feels like we should move to 14 for broader pointer/keyboard support
```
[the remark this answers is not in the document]
LFDanLu: oh sweet, thanks for digging up the numbers! I'll have to check how backwards compatible these version are

rfcs/2023-rsp-testing-utils.md
LFDanLu: for the first iteration will probably go with just jest to simplify things but will check what most other people are using

rfcs/2023-rsp-testing-utils.md
devongovett, line 47: These would definitely be useful. Not sure whether people will really be simulating mobile/desktop specifically to test RSP components, since they can usually assume that we have already tested that behavior, but no harm in offering it I guess. Long press / pointer event stuff definitely seems useful though.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+To solve this, we could create a test helper class for each ARIA pattern (i.e. table, list, select, button, etc) that contains methods for common interactions specific to the pattern as described in [this issue comment](https://github.com/adobe/react-spectrum/issues/3703#issuecomment-1302727331). The interaction methods would need to work for `realTimers` and `fakeTimers` (if using Jest) and would handle all the requisite element lookup and event firing required for the high level interaction pattern. Where ever possible, these interaction methods should use `user-event` or our other interaction utils. Additionally, it may be valuable for each of the classes to look up and expose the internal nodes that users often want to assert against (e.g. The picker's trigger button, label, listbox, etc). A implementation of this could look like the following:
```
devongovett, line 61: Supporting both real and fake timers might be a lot of work. We could consider only supporting fake timers, at least to start. Another alternative could potentially be to turn on fake timers just before performing an interaction, and restore it afterward if a larger test is using real timers. That way we could potentially avoid having two paths.

The downside to that might be that it's specific to Jest, and if someone is using a test framework other than Jest it wouldn't work. The more we can reduce the complexity and number of scenarios we need to cover though the better.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+);
+
+let picker = new SelectTester({element: screen.getByTestId('test'), timerType: 'real'});
+await picker.selectOption('Three');
+expect(picker.trigger).toHaveTextContent('Three');
+```
+
+Since these classes are specific to each ARIA pattern, they should work for React Aria Components (RAC) as well as React Spectrum (RSP) components and thus need to handle possible differences in what `element` is provided. For instance, the `data-testid` passed to a RSP Picker is attached to the trigger button element but the `data-testid` for a RAC Select could be the one attached to the wrapper element instead. Alternatively, we could just standardize what `element` is expected to be provided to each of the util classes (aka make the user pass in the Select's trigger button element).
+
+These util classes could live in the `@react-aria/test-util` package and re-exported from `react-aria-components`.
```
devongovett, line 81: Not sure about re-exporting it from the monopackages. I wouldn't want test utils to accidentally get bundled into people's production apps. I think using the individual package for test utils should be fine.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Documentation
+
+Documentation for the first two parts described above should go in the existing "Testing" page that we have [in the docs](https://react-spectrum.adobe.com/react-spectrum/testing.html#testing) already. We'll want to consider having a similar page in the React Aria Component/React Aria docs sections as well or moving that page up in hierarchy so it lives on the [main page](https://react-spectrum.adobe.com/index.html). The test util classes could also be documented in the same page as the rest of the testing info. I think we should also include a testing section in each of the individual component pages that highlights what test util class can be used (aka mapping Picker to SelectTester), documents the class's methods and options, and links back to the overall testing FAQ for common issues/gotchas that are related to said component. The main testing page should also show a high level example of using a test util class in a example test.
```
devongovett, line 85: We could also consider having a testing section in the individual component pages if there are specific things to cover for a given component. For example I imagine testing table might be complex. Once we have the test utils for certain aria patterns it might be nice to show an example of that, or at least link to it from each component as well.

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Drawbacks
+
+My primary concern is whether or not some of the mocks truly work in other test suite configurations. Historically, we've seen other teams struggle with a small subset of the mocks not resolving their test rendering issues (specifically with ListBox and the `clientHeight`). Making sure the interaction utils aren't brittle may be a challenge since the way we test in RSP is with components in isolation versus other teams testing entire app flows that have many elements rendered at once. For instance, we can easily find the ListBox tied to a Select element by querying for `role="listbox"` in our RSP test, but doing so in a page filled with other ListBoxes would involve looking up the `id` from the Select's `aria-controls`.
```
devongovett, line 89: Yeah agreed. I think reducing our reliance on mocks overall where we can (including by adding test-only behaviors) would help a lot.

rfcs/2023-rsp-testing-utils.md
devongovett: Yeah I would not mix test and production code as I mentioned above. I think importing from the separate package in tests is totally fine.

devongovett (commented):
Great writeup! 🥳

rfcs/2023-rsp-testing-utils.md
```diff
@@ -0,0 +1,113 @@
...
+
+- Mock of `HTMLElement.prototype`'s `clientWidth`, `clientHeight`, and `scrollHeight`. This is necessary for many of our virtualized collection components like ListBox and TableView. Without these the component often renders with zero height/width and thus zero items. An example of these mocks can be found [here](https://github.com/adobe/react-spectrum/blob/07f673acb2d68144156df3aa0db6e91810b12c67/packages/%40react-spectrum/listbox/test/ListBox.test.js#L71-L76).
+- Mock of `HTMLElement.prototype`'s `getBoundingClientRect`. Often mocked for drag and drop operations, an example can be found [here](https://github.com/adobe/react-spectrum/blob/07f673acb2d68144156df3aa0db6e91810b12c67/packages/%40react-aria/dnd/test/dnd.test.js#L39-L46).
+
+There are other mocks that teams have historically needed, but the above are two of the most commonly brought up. Additionally, the way we've mocked the above in the React Spectrum tests changes it for ALL HTMLElements which maybe too broad for other people's test suites, hence why I'm leaning towards simply documenting this behavior rather than creating a utility function.
```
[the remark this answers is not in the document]
LFDanLu: I think this is a good idea, would definitely make some (if not all) of these brittle mocks unnecessary. Will look into the feasibility

rfcs/2023-rsp-testing-utils.md
LFDanLu: Right, I was mainly thinking along the lines of people writing tests for their app's mobile experience vs desktop experience and thus needing to make our components render in their proper mobile/desktop configuration. I think Quarry had run into this issue previously

rfcs/2023-rsp-testing-utils.md
LFDanLu: I'll look into this, definitely would be nice to reduce the complexity for a first iteration.

>  Another alternative could potentially be to turn on fake timers just before performing an interaction, and restore it afterward if a larger test is using real timers

An interesting idea, I'll have to do some testing around this. Not sure the ramifications of performing timer type switches since I've primarily had experience with sticking with fakeTimers, will have to play around with that.

> The downside to that might be that it's specific to Jest, and if someone is using a test framework other than Jest it wouldn't work. The more we can reduce the complexity and number of scenarios we need to cover though the better.

Yeah, that is a big concern of mine. Will need to test against some other test frameworks, will be a big research task.

snowystinger (commented):
Some notes

with user-event 14, I was able to get rid of our `typeText` and I should be able to get rid of `triggerPress` as well, I think I have a couple kinks to work out with installPointerEvents in https://github.com/adobe/react-spectrum/pull/4426

I'd also like to discuss with you the notion of `users` which user-event 14 allows us to define. I think this would be a useful thing for us to define such that people wouldn't just be always testing mouse users in their tests, they'd also have easy access to touch/virtual cursor/keyboard. Note, this may involve creating a couple functions/flows for various components.
https://testing-library.com/docs/user-event/setup#starting-a-session-per-setup
Also, this user object will be required in for us to call things like user.click, so whatever API we build, we need to require or at least allow the passing of this object.

For instance, what if our API was
```
let user = SetupAriaPointerUser(/*{userEvent.setup options like delay: null}*/) -> // calls userEvent.setup(), maybe adds things like selectFromMenu to the user
// or we could have packages for different components and you'd call something like
// installMenuTestUtils(user)
user.selectFromMenu({query: 'selectorForTrigger', option: 'optionName'})
```
We'd know based on 'delay' if users have mocked the timer or not

We could also still go the route of the PageObject classes you've proposed, but I suggest we require passing the `user` that user-event 14 creates to them for the reason of mocked or real timers and so we eventually get the fixes that user-event comes out with.

devongovett (approved):

PR #4836, opened 2023-07-27
(WIP) Exploration/prototyping of test utils package

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/@react-spectrum/table/test/TableBlah.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+                {key => <Cell>{item[key]}</Cell>}
+              </Row>)
+            }
+          </TableBody>
+        </TableView>
+      </Provider>
+    );
+  };
+
+  beforeAll(function () {
```
jluyau, line 61: this is probably on your radar, but having these mocks and such callable with 1 line of code would be nice too!

packages/@react-aria/test-utils/src/TableTester.ts
```diff
@@ -0,0 +1,133 @@
...
+    } = opts;
+    this.element = element;
+    this.timerType = timerType;
+    // This can potentially become stale? Double check, if so replace with getters that refetch the rows and columns
+    this.rowgroups = within(this.element).getAllByRole('rowgroup');
+    this.columns = within(this.rowgroups[0]).getAllByRole('columnheader');
+    this.rows = within(this.rowgroups[1]).getAllByRole('row');
+  }
+
+  toggleRowSelection(opts: {index?: number, text?: string, needsLongPress?: boolean}) {
```
jluyau, line 55: one helper function that i think would help (since it's come up a bunch) is a method to select a given cell, though i guess it'd require the column and row indexes or text

jluyau (commented):
i think these utils would definitely make writing tests for rsp components much cleaner. another utility class that would help would be interacting with the current dialog (methods to get elements within the content body as well as pressing the dialog buttons)

packages/@react-aria/test-utils/src/TableTester.ts
LFDanLu: sounds good, we have a util for that that we use internally as well actually haha. I think we can start off with searching for cell text first since that is what a user would look for typically and see about supporting a (x row, y column) alternative maybe.

packages/@react-spectrum/table/test/TableBlah.test.tsx
LFDanLu: Yep, I've been mulling over whether or not these mocks are too broad. Ideally we would limit what it actually gets applied on, similar to https://github.com/adobe/react-spectrum/blob/1c32cd5559048ace7b185f4c594d503aea0e6ae7/packages/%40react-aria/overlays/test/useOverlayPosition.test.tsx#L244-L252, but I'm worried that users may need to have finer control over the mocks.

Perhaps we can provide a basic util that does the simple mock and offer a way to override/customize

LFDanLu: Another interesting idea that @devongovett brought up was making our virtualized components render as non-virtualized components when rendered in a test environment specifically. That means we won't have to mock any of this stuff, thus avoid possible mock collisions and assumptions. RSP would still write tests with the virtualized components still being virtualized just to make sure things work as expected

LFDanLu:
Closing this for now, will incorporate the feedback when work resumes on this portion of the testing util work

PR #4839, opened 2023-07-28
Fix InlineAlert docs package data

dannify:
It was pointing to the wrong package


reidbarber:
Dupe of https://github.com/adobe/react-spectrum/pull/4830 :)

Issue #4761, opened 2023-07-12
TableView with Zoom 150% on Windows crashes with Error: Maximum update depth exceeded


synox:
# 🐛 Bug Report 

Our Adobe product uses `TableView` in a React Application. Our customers reported that the React application crashes when using Google Chrome with 150% zoom, only on Windows. We had to rollback the feature. 

## Reproduce

 - use a Windows 11 (VM). 
 - in System settings, open `Display`, set `Scale` to 150%, 
- Open Google Chrome, 
- open https://reactspectrum.blob.core.windows.net/reactspectrum/b57b3af44c345d456658e2895bb9244c65f0e7d3/storybook/index.html?path=/story/tableview--resizing-uncontrolled-static-widths
- if the error does not happen, resize the window a few times. 

The error happens sometimes on macOS, with Scrollbars always enabled, and 150% zoom in Chrome. 

also happens on 
 - https://reactspectrum.blob.core.windows.net/reactspectrum/b57b3af44c345d456658e2895bb9244c65f0e7d3/storybook/index.html?path=/story/tableview--resizing-uncontrolled-column-divider&providerSwitcher-express=false
 - https://reactspectrum.blob.core.windows.net/reactspectrum/b57b3af44c345d456658e2895bb9244c65f0e7d3/storybook/index.html?path=/story/tableview--resizing-uncontrolled-min-max&providerSwitcher-express=false
 - https://reactspectrum.blob.core.windows.net/reactspectrum/b57b3af44c345d456658e2895bb9244c65f0e7d3/storybook/index.html?path=/story/tableview--resizing-uncontrolled-some-not-allowed&providerSwitcher-express=false
 - https://reactspectrum.blob.core.windows.net/reactspectrum/b57b3af44c345d456658e2895bb9244c65f0e7d3/storybook/index.html?path=/story/tableview--resizing-uncontrolled-some-not-allowed&providerSwitcher-express=false
 - 

## 🤔 Expected Behavior

Zoom level should not influence component behavior. 

## 😯 Current Behavior

Crashes with Error message: 
> Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

<img width="1351" alt="Screenshot 2023-07-12 at 16 21 00" src="https://github.com/adobe/react-spectrum/assets/2250964/6b75b445-a490-454a-b7da-eff5d29bf609">

## 🔦 Context

[...]

PR #4846, opened 2023-07-31
Prevent resize infinite loop in TableView when scrollbars are always on 

LFDanLu:
Closes https://github.com/adobe/react-spectrum/issues/4761

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Go to the "allowsResizing, uncontrolled, column divider" story and zoom in 150%. Try resizing the window in such a way that the table begins to shrink to match the window. The story shouldn't crash. Continue to resize the window wider/smaller.

https://reactspectrum.blob.core.windows.net/reactspectrum/973f40ab090541bcb3e27e2f20c80884964c610f/storybook/index.html?path=/story/tableview--resizing-uncontrolled-column-divider&providerSwitcher-express=false can be used as a comparison and will crash as you resizing in 150% zoom

## 🧢 Your Project:

RSP


packages/@react-aria/virtualizer/src/Virtualizer.tsx
LFDanLu, line 1: This is a pretty rough solution, not super happy with it. Some notes on other approaches and investigations:
- This solution is placed here and not in [stately Virtualizer](https://github.com/adobe/react-spectrum/blob/1725d70270dbb247d6224f3082b2a4c59f93a92b/packages/%40react-stately/virtualizer/src/Virtualizer.ts#L163-L168) because we get all kinds of extra rects being set there, to the point where there are 4+ different rects being cyclically applied when the issue occurs. I figure this is a safe enough place to put the check instead since all our Virtualizer hit this file's `setVisibleRect`

- The core issue in the TableView seems to be related to the [width applied in the TableView's head wrapper](https://github.com/adobe/react-spectrum/blob/1725d70270dbb247d6224f3082b2a4c59f93a92b/packages/%40react-spectrum/table/src/TableViewBase.tsx#L653). If we subtract a 1px from that width, the issue no longer seems to reproduce but then causes the resizer handler of the last column to be hidden by default + a odd scroll shift when keyboard focusing the last column of the Table.

LFDanLu: Removed the old patch that early aborts a potential infinite loop of rect sets in favor of increased pixel accuracy in ScrollView when setting the visibleRect (aka the width applied to the TableView header row). Might be good to still have that early abort in case there are other ways it can happen, but for now getting rid of it

packages/@react-aria/virtualizer/src/ScrollView.tsx
```diff
@@ -167,7 +169,7 @@ function ScrollView(props: ScrollViewProps, ref: RefObject<HTMLDivElement>) {
   }
 
   return (
-    <div {...otherProps} style={style} ref={ref} onScroll={onScroll}>
+    <div {...otherProps} style={style} ref={ref} onScroll={onScroll} data-rsp-testid="scrollview" >
```
LFDanLu, line 172: Test id that won't get overwritten by `data-testid` provided by user to ListView/other components that provide a `data-testid` to the grid. I've opted to not remove it via our babel config so others can mock `computedStyles` in the same way in their tests

packages/@react-aria/virtualizer/src/ScrollView.tsx
```diff
@@ -128,8 +129,9 @@ function ScrollView(props: ScrollViewProps, ref: RefObject<HTMLDivElement>) {
       return;
     }
 
-    let w = dom.clientWidth;
-    let h = dom.clientHeight;
+    let computedStyles = getComputedStyle(dom);
+    let w = parseFloat(computedStyles.width) + parseFloat(computedStyles.paddingLeft) + parseFloat(computedStyles.paddingRight);
+    let h = parseFloat(computedStyles.height) + parseFloat(computedStyles.paddingTop) + parseFloat(computedStyles.paddingBottom);
```
LFDanLu, line 134: As mentioned in https://github.com/adobe/react-spectrum/pull/4846/files#r1279884044, the core issue was that the TableView's column head wrapper would occasionally be wider than the actual TableBody due to `clientWidth`'s pixel rounding behavior. This would result in the window/flex container around the TableView rendering scrollbars if the TableView's width was matching the exact width of said window/flex container -> resulting in the TableView resizing itself to match the new (smaller) available width of the container -> a new smaller visibleRect is calculated and applied to the TableView's header width causing the scrollbars to disappear -> cycle begins again

packages/@react-aria/landmark/test/useLandmark.test.tsx
```diff
@@ -72,17 +72,24 @@ function toggleBrowserWindow() {
...
+    jest.spyOn(window, 'getComputedStyle').mockImplementation((element) => {
+      if (element.attributes.getNamedItem('data-rsp-testid')?.value === 'scrollview') {
+        const sty = realGetComputedStyle(element);
+        sty.width = '1000px';
+        sty.height = '1000px';
+        return sty;
+      } else {
+        return realGetComputedStyle(element);
+      }
+    });
```
LFDanLu, line 87: Bit annoying since this will need to be done by any users would was mocking offsetWidth like we did before. Open to any ideas

packages/@react-aria/virtualizer/src/ScrollView.tsx
```diff
@@ -128,8 +129,9 @@ function ScrollView(props: ScrollViewProps, ref: RefObject<HTMLDivElement>) {
       return;
     }
 
-    let w = dom.clientWidth;
-    let h = dom.clientHeight;
+    let computedStyles = getComputedStyle(dom);
+    let w = parseFloat(computedStyles.width) + parseFloat(computedStyles.paddingLeft) + parseFloat(computedStyles.paddingRight) || 0;
+    let h = parseFloat(computedStyles.height) + parseFloat(computedStyles.paddingTop) + parseFloat(computedStyles.paddingBottom) || 0;
```
LFDanLu, line 134: recommenting:
As mentioned in https://github.com/adobe/react-spectrum/pull/4846/files#r1279884044, the core issue was that the TableView's column head wrapper would occasionally be wider than the actual TableBody due to clientWidth's pixel rounding behavior. This would result in the window/flex container around the TableView rendering scrollbars if the TableView's width was matching the exact width of said window/flex container -> resulting in the TableView resizing itself to match the new (smaller) available width of the container -> a new smaller visibleRect is calculated and applied to the TableView's header width causing the scrollbars to disappear -> cycle begins again

devongovett: did you try `getBoundingClientRect`? I think that has sub-pixel precession, and then you wouldn't need to manually add the padding and mock as many things.

LFDanLu: it does have sub-pixel precision but it also includes the border and the scrollbar width which we don't want unfortunately since the TableHeader wrapper's width should only match the ScrollView's width + padding. Doing so results in extra scrolling when scrollbars are enabled:

image

What makes it even more troublesome is that the TableHeader wrapper doesn't have `box-sizing: border-box` so we also have to account for the extra 1px applied on either end when calculating what width should be applied to it (aka we need it to match the scrollable body's layout, border and all, so that the column headers line up with the row cells)

LFDanLu: Doing something like using `width: 100%` for the TableHeader wrapper width instead of using the visible rect + `box-sizing: border-box` and reverting back to using `dom.clientWidth` in scrollView seems to work decently (makes the TableHeader wrapper really wide when we have a lot of rows but doesn't seem to break anything visually), but then results in stuff like missing top border due to element overlap:
image

fixing the above results in:
image

which is the original issue @snowystinger ran into all that time ago with resizing columns (aka the break in the blue line). I haven't been able to find a way to fix the above, otherwise I would've went with the `width: 100%` route rather than the computed styles

packages/@react-spectrum/table/src/TableViewBase.tsx
```diff
@@ -649,7 +648,6 @@ function TableVirtualizer(props) {
             role="presentation"
             className={classNames(styles, 'spectrum-Table-headWrapper')}
             style={{
-              width: visibleRect.width,
```
LFDanLu, line 652: Looks like this width wasn't actually needed? This simplifies the logic greatly since the table header wrapper lets the browser determine its width via its flex container width's, matching the scroll body behavior.

devongovett: 🤞

LFDanLu: 🤞🤞

packages/@react-aria/virtualizer/src/ScrollView.tsx
snowystinger: There's an old PR or commit which moved the blue line up and out so it was a single element instead of two. Forget where it is, but you might look for that.

LFDanLu: Oh thanks for the heads up! It actually turns out the `width: 100%` route was non-viable for other reasons (made the column header too wide in the many rows case) so I think we are good here for now (fingers crossed).

devongovett (approved):

yihuiliao (approved):

PR #4869, opened 2023-08-05
Do not add random prefix to ids during tests in React 18

### LINKED ISSUE #4851: [React-Aria] Snapshot Test Ids change everytime after 2023-06-27 release

After the latest release, ids in my snapshot tests are changing every run. I see the comments on the release notes and on the MR (#4347) but that seems to imply that the id format will only change once - which is fixable - but for me its changing every time.

I'm using React 18 and it seems like the [`didSSR` check seems to be false](https://github.com/adobe/react-spectrum/blob/22ded332388098ca54227afb7d1c4d0ae270e875/packages/%40react-aria/ssr/src/SSRProvider.tsx#L161) (which makes sense, Jest isn't doing SSR) and so `defaultContext.prefix` is being used in the id.

In React 17, the SSRProvider set `isSSR` to true, before switching to false in useLayoutEffect (which I assume ran after snapshots were generated)

old and new:
```
- aria-controls="react-aria-4"
+ aria-controls="react-aria9516198744-:r9a:"
```

I would expect it to be `react-aria-:r9a:`

### 🤔 Expected Behavior?

Snapshot test ids should be stable between snapshot runs

### 😯 Current Behavior

the ids change everytime in jest, so snapshot tests don't work

### 💁 Possible Solution

Re-purpose the SSRProvider (or something similar?) to be used in tests so that the prefix isn't included?

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

https://codesandbox.io/p/sandbox/naughty-fermi-4zlqv2?welcome=true

The terminal tab at the bottom should be running jest watch mode and will show the snapshot differences between runs

(if it's not showing anything in the terminal you may need to fork and run npm run test)

### Version

react-aria 3.21.1

Other

### If other, please specify.

jest/jsdom

MacOS ventura

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
Fixes #4851

This ensures that we do not add a random prefix to ids generated by `useId` in test environments with React 18. Previously it was possible to use `SSRProvider` to ensure this, but `SSRProvider` is now a noop in React 18, and we determine whether we are SSRing in another way. But that's not set in a test environment. This will check whether the `NODE_ENV` environment variable is set to "test" and won't add a random prefix if so.

### DIFF
--- a/packages/@react-aria/ssr/package.json
+++ b/packages/@react-aria/ssr/package.json
@@ -29,5 +29,16 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "engines": {
+    "node": ">= 12"
+  },
+  "targets": {
+    "main": {
+      "context": "node"
+    },
+    "module": {
+      "context": "node"
+    }
   }
 }
--- a/packages/@react-aria/ssr/src/SSRProvider.tsx
+++ b/packages/@react-aria/ssr/src/SSRProvider.tsx
@@ -152,7 +152,8 @@ function useLegacySSRSafeId(defaultId?: string): string {
   }
 
   let counter = useCounter(!!defaultId);
-  return defaultId || `react-aria${ctx.prefix}-${counter}`;
+  let prefix = ctx === defaultContext && process.env.NODE_ENV === 'test' ? 'react-aria' : `react-aria${ctx.prefix}`;
+  return defaultId || `${prefix}-${counter}`;
 }
 
 function useModernSSRSafeId(defaultId?: string): string {
--- a/packages/@react-aria/ssr/src/SSRProvider.tsx
+++ b/packages/@react-aria/ssr/src/SSRProvider.tsx
@@ -159,7 +159,7 @@ function useModernSSRSafeId(defaultId?: string): string {
   // @ts-ignore
   let id = React.useId();
   let [didSSR] = useState(useIsSSR());
-  let prefix = didSSR ? 'react-aria' : `react-aria${defaultContext.prefix}`;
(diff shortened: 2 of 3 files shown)


packages/@react-aria/ssr/package.json
```diff
@@ -29,5 +29,16 @@
...
+    "node": ">= 12"
+  },
+  "targets": {
+    "main": {
+      "context": "node"
+    },
+    "module": {
+      "context": "node"
+    }
   }
```
devongovett, line 43: This is a hack to prevent parcel from inlining `NODE_ENV` during compile time. node 12 to ensure we don't end up with optional chaining in the built output, matching our other browser targeted code. I did a diff from what's currently published and there appear to be no other differences from my additions here.

LFDanLu (dismissed):

reidbarber (approved):
LGTM

packages/@react-aria/ssr/src/SSRProvider.tsx
```diff
@@ -152,14 +152,15 @@ function useLegacySSRSafeId(defaultId?: string): string {
   }
 
   let counter = useCounter(!!defaultId);
-  return defaultId || `react-aria${ctx.prefix}-${counter}`;
+  let prefix = ctx === defaultContext && process.env.NODE_ENV === 'test' ? 'react-aria' : `react-aria${ctx.prefix}`;
```
tbrannam, line 155: `process.env` is always available in node, but on the browser-side it may not be defined, depending on the build-tools?

PR #4879, opened 2023-08-08
Fix popover behavior in Firefox

reidbarber:
In Firefox, if you open a dialog trigger popover via keyboard, the focus ring on the trigger element will stay shown, even though focus has entered the popover.  Also, when pressing Esc and Tab quickly, that focus ring will stay visible even though focus has moved to the next element (so two focus rings are shown).

image

Seems to have broken in https://github.com/adobe/react-spectrum/pull/4773, due to Firefox's behavior with inert.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

In Firefox:

Go to a DialogTrigger type=popover story (or a ContextualHelp story.) Open the popover via keyboard. There should not be a focus ring around the trigger element. Close the popover via Escape, and focus should return to the trigger element. 

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-aria/dnd/test/dnd.test.js
+++ b/packages/@react-aria/dnd/test/dnd.test.js
@@ -2470,7 +2470,6 @@ describe('useDrag and useDrop', function () {
       await act(async () => {
         await tree.rerender(<Test showInput2 />);
       });
-      act(() => jest.runAllTimers());
       expect(() => tree.getAllByRole('textbox')).toThrow();
 
       fireEvent.click(draggable);
--- a/packages/@react-aria/dnd/test/dnd.test.js
+++ b/packages/@react-aria/dnd/test/dnd.test.js
@@ -2470,6 +2470,7 @@ describe('useDrag and useDrop', function () {
       await act(async () => {
         await tree.rerender(<Test showInput2 />);
       });
+      act(() => jest.runAllTimers());
       expect(() => tree.getAllByRole('textbox')).toThrow();
 
       fireEvent.click(draggable);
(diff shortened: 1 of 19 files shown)


packages/@react-aria/dnd/test/dnd.test.js
```diff
@@ -2470,6 +2470,7 @@ describe('useDrag and useDrop', function () {
       await act(async () => {
         await tree.rerender(<Test showInput2 />);
       });
+      act(() => jest.runAllTimers());
```
devongovett, line 2473: Were these timer changes still needed? Is firing a blur event somehow causing a timer to fire?

reidbarber: And there were some additional tests failing, specifically DialogTrigger `input in nested popover should be interactive with a click` and `will not lose focus to body`

reidbarber: Those weren't actually needed. Looks like nested cases for DialogTrigger are failing

packages/@react-aria/overlays/src/usePopover.ts
```diff
@@ -96,7 +96,7 @@ export function usePopover(props: AriaPopoverProps, state: OverlayTriggerState):
     isDisabled: isNonModal
   });
 
-  useLayoutEffect(() => {
+  useEffect(() => {
```
LFDanLu, line 99: I think this was intentionally changed to be a useLayoutEffect in a prior PR? Digging to see why and if it is safe to change this to a useEffect

reidbarber: From @devongovett https://github.com/adobe/react-spectrum/pull/3677/files#diff-02c0eb0f2cdf8d726564712562f36c693aa5d9feb5860837da0a88f3eac01a51R98-R102, I think he's taking a look too.

PR #4886, opened 2023-08-09
Remove unused code

marob:


snowystinger (commented):
I've checked this in an older version of node as well as in the codebase before the original change and test was introduced.
In no scenario is the test failing, with or without the code that fixed the issue

As a result, I'm going to remove the whole test.

snowystinger (approved):

LFDanLu (approved):

PR #4891, opened 2023-08-09
Migrate all RAC render props to data attributes

yihuiliao:
Right now we have a mix of aria and data attributes, and this can be confusing, making RAC harder to learn. It's also inconsistent between components, where the same state might be an aria attribute in one but a data attribute in another (e.g. aria-selected vs data-selected). Obviously the aria attributes have to stay, but we should add additional data attributes for all states for consistency. This will also ensure that if we ever have to switch aria patterns as an implementation detail, people's styling won't break.

credit/source: devon

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Given that some of the css was updated to use the data-attributes, it might be a good idea to just go through the RAC docs and make sure there haven't been any changes. It's also a good place to see the data-attributes being passed if you use the inspector tool.


packages/react-aria-components/src/Tabs.tsx
```diff
@@ -37,7 +37,7 @@ export interface TabListProps<T> extends StyleRenderProps<TabListRenderProps>, A
 export interface TabListRenderProps {
   /**
    * The orientation of the tab list.
-   * @selector [aria-orientation="horizontal | vertical"]
+   * @selector [aria-orientation="horizontal | vertical"], [data-orientation="horizontal | vertical"]
```
yihuiliao, line 40: im aware that this doesn't wrap in the `PropsTable` in the docs so if anyone has any suggestions that would be much appreciated. in the meantime, i'll continue messing around with it

yihuiliao:
do we also want to update the docs so that they say something about each aria attribute having an equivalent data attribute?

packages/react-aria-components/src/Tabs.tsx
devongovett: I'd say we should just drop the aria versions from the docs, and only document the data attributes.

yihuiliao: https://github.com/adobe/react-spectrum/blob/0e359399ade3d2047b6e6ef4de5ec6e041f8689b/packages/dev/docs/src/docs.css#L434-L436C2

looks like it's related to this line

reidbarber:
> do we also want to update the docs so that they say something about each aria attribute having an equivalent data attribute?

I don't think so. I think we can think of the ARIA attributes as an implementation detail, and the data attributes as a public API.

reidbarber (commented):
There are some example styles in the docs still using `aria-` selectors. They should be easy to catch by searching the codebase for `[aria-` (if we do want to remove them all).

yihuiliao:
> There are some example styles in the docs still using `aria-` selectors. They should be easy to catch by searching the codebase for `[aria-` (if we do want to remove them all).

oh good point, i forgot i could do that. thanks for reminding me

packages/react-aria-components/src/Collection.tsx
```diff
@@ -741,7 +741,7 @@ export interface ItemRenderProps {
   isPressed: boolean,
   /**
    * Whether the item is currently selected.
-   * @selector [aria-selected=true]
+   * @selector [data-selected=true]
```
devongovett, line 744: I think we can do just `[data-selected]` without the `=true` here since we only add the attribute if it is true. That would make it more consistent with our other data attributes.

packages/react-aria-components/src/Menu.tsx
```diff
@@ -172,7 +172,7 @@ function MenuSection<T>({section, className, style, ...otherProps}: MenuSectionP
 export interface MenuItemRenderProps extends ItemRenderProps {
   /**
    * Whether the item is currently selected.
-   * @selector [aria-checked=true]
+   * @selector [data-selected=true]
```
devongovett, line 175:
```suggestion
   * @selector [data-selected]
```

packages/react-aria-components/src/Meter.tsx
```diff
@@ -58,7 +58,12 @@ function Meter(props: MeterProps, ref: ForwardedRef<HTMLDivElement>) {
   });
 
   return (
-    <div {...meterProps} {...renderProps} ref={ref} slot={props.slot}>
+    <div 
+      {...meterProps} 
+      {...renderProps} 
+      ref={ref} 
+      slot={props.slot}
+      data-valuetext={meterProps['aria-valuetext']}>
```
devongovett, line 66: hmm I wonder what people would use this one for. It's not a boolean state so not sure? Also looks like ProgressBar has the same one.

packages/react-aria-components/src/Table.tsx
```diff
@@ -505,12 +505,12 @@ export interface ColumnRenderProps {
...
   /**
    * Whether the column allows sorting.
-   * @selector [aria-sort]
+   * @selector [data-sort]
    */
   allowsSorting: boolean,
   /**
    * The current sort direction.
-   * @selector [aria-sort="ascending | descending"]
+   * @selector [data-sort="ascending | descending"]
```
devongovett, line 513: Maybe we should do `data-sort-direction` to match `sortDirection`?

packages/react-aria-components/src/Table.tsx
```diff
@@ -505,12 +505,12 @@ export interface ColumnRenderProps {
   isFocusVisible: boolean,
   /**
    * Whether the column allows sorting.
-   * @selector [aria-sort]
+   * @selector [data-sort]
```
devongovett, line 508: `data-allows-sorting`?

packages/react-aria-components/src/Table.tsx
```diff
@@ -862,7 +862,9 @@ function TableColumnHeader<T>({column}: {column: GridNode<T>}) {
       ref={ref}
       data-focused={isFocused || undefined}
       data-focus-visible={isFocusVisible || undefined}
-      data-resizing={isResizing || undefined}>
+      data-resizing={isResizing || undefined}
+      data-sorting={column.props.allowsSorting || undefined}
```
devongovett, line 866: This one doesn't match the docs above. Maybe `data-allows-sorting` would be a better name?

packages/react-aria-components/docs/Menu.mdx
```diff
@@ -366,10 +366,10 @@ A custom `className` can also be specified on any component. This overrides the
 </Menu>
 ```
 
-In addition, some components support multiple UI states (e.g. pressed, hovered, etc.). React Aria components expose states using DOM attributes, which you can target in CSS selectors. These are ARIA attributes wherever possible, or data attributes when a relevant ARIA attribute does not exist. For example:
+In addition, some components support multiple UI states (e.g. pressed, hovered, etc.). React Aria components expose states using DOM attributes, which you can target in CSS selectors. These are data attributes wherever possible. For example:
 
 ```css
-.react-aria-Item[aria-checked=true] {
+.react-aria-Item[data-checked=true] {
```
devongovett, line 372: I think this is `data-selected`?

packages/react-aria-components/docs/GridList.mdx
```diff
@@ -367,10 +367,10 @@ A custom `className` can also be specified on any component. This overrides the
 </GridList>
 ```
 
-In addition, some components support multiple UI states (e.g. pressed, hovered, etc.). React Aria components expose states using DOM attributes, which you can target in CSS selectors. These are ARIA attributes wherever possible, or data attributes when a relevant ARIA attribute does not exist. For example:
+In addition, some components support multiple UI states (e.g. pressed, hovered, etc.). React Aria components expose states using DOM attributes, which you can target in CSS selectors. These are data attributes wherever possible. For example:
 
 ```css
-.react-aria-Item[aria-selected=true] {
+.react-aria-Item[data-selected=true] {
```
devongovett, line 373: can get rid of all the `=true`

packages/react-aria-components/src/ToggleButton.tsx
```diff
@@ -19,7 +19,7 @@ import {ToggleState, useToggleState} from 'react-stately';
 export interface ToggleButtonRenderProps extends ButtonRenderProps {
   /**
    * Whether the button is currently selected.
-   * @selector [aria-pressed=true]
+   * @selector data-pressed=true]
```
devongovett, line 22: I think `data-pressed` is already used for the mouse down state (extended from ButtonRenderProps). We should call this one `data-selected` to match the prop name.

packages/react-aria-components/docs/ToggleButton.mdx
```diff
@@ -167,10 +167,10 @@ A custom `className` can also be specified on any component. This overrides the
 </ToggleButton>
 ```
 
-In addition, some components support multiple UI states (e.g. focused, placeholder, readonly, etc.). React Aria components expose states using DOM attributes, which you can target in CSS selectors. These are ARIA attributes wherever possible, or data attributes when a relevant ARIA attribute does not exist. For example:
+In addition, some components support multiple UI states (e.g. focused, placeholder, readonly, etc.). React Aria components expose states using DOM attributes, which you can target in CSS selectors. These are data attributes wherever possible. For example:
 
 ```css
-.react-aria-ToggleButton[aria-pressed=true] {
+.react-aria-ToggleButton[data-pressed=true] {
```
devongovett, line 173:
```suggestion
.react-aria-ToggleButton[data-selected] {
```

devongovett:
Great work! A other general notes:

* I probably didn't cover this in the ticket, but I think we should also convert the CSS pseudo classes like `:focus`, `:focus-within`, and `:disabled` to data attributes as well. That way it's always consistent between components (e.g. we have `:disabled` in some, and `data-disabled` in others right now).
* We might also want to convert other HTML attributes like `[draggable]` to `[data-allows-dragging]`?
* We are currently a bit inconsistent between `[data-validation-state=invalid]` and `[data-invalid]`. We might want to audit that and make a decision.

packages/react-aria-components/src/Meter.tsx
yihuiliao: yeah i noticed that too. figured it wouldn't hurt to have it so i made it a data attribute but not opposed to changing it

yihuiliao:
> * We are currently a bit inconsistent between `[data-validation-state=invalid]` and `[data-invalid]`. We might want to audit that and make a decision.

I went ahead to see how many RAC components used `[data-validation-state=invalid]` vs. `[data-invalid]` and seems like more use `[data-validation-state]`. Not sure if that's really a means to pick one over another but might be nice to know.

packages/react-aria-components/src/Tabs.tsx
```diff
@@ -302,9 +311,10 @@ function TabPanel(props: TabPanelProps, forwardedRef: ForwardedRef<HTMLDivElemen
     <div
       {...domProps}
       ref={ref}
+      data-focused={isFocused || undefined}
       data-focus-visible={isFocusVisible || undefined}
       // @ts-ignore
-      inert={!isSelected ? 'true' : undefined} />
+      data-inert={!isSelected ? 'true' : undefined} />
```
devongovett, line 317: We still need the `inert` attribute for its native behavior, but we can document `data-inert` as the one to use for styling.

```suggestion
      inert={!isSelected ? 'true' : undefined}
      data-inert={!isSelected ? 'true' : undefined} />
```

packages/react-aria-components/src/Input.tsx
```diff
@@ -61,6 +61,10 @@ function Input(props: InputProps, ref: ForwardedRef<HTMLInputElement>) {
       {...mergeProps(props, focusProps, hoverProps)}
       {...renderProps}
       ref={ref}
+      data-focused={isFocused || undefined}
+      data-disabled={props.disabled || undefined}
+      data-readonly={props.readOnly || undefined}
+      data-invalid={!!props['aria-invalid'] || undefined}
```
devongovett, line 67: readonly and invalid don't seem to be documented above or included in the `useRenderProps` values.

devongovett (commented):
looks good. just a couple last things.

snowystinger (commented):
@devongovett is complete reliance on data-* going to mess with the tailwind sorting features?
I assume not because all those selectors still exist. We're just showcasing the data-* ones in our docs now.

devongovett:
Reid is also working on a tailwind plugin that should help with that.

devongovett (approved):

reidbarber (approved):

PR #4900, opened 2023-08-10
Add `isPending` prop to `Button` component which triggers a spinner.

iamwillpowell:
Closes [Pending button task](https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=34893644)

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Open a Button story in Storybook and toggle the isPending control. Also, there is a Pending spinner story that you can click on buttons to test the interaction. The story is set up to cancel the pending state after 5 seconds.

## 🧢 Your Project:

RSP


packages/@react-spectrum/button/src/Button.tsx
```diff
@@ -20,13 +20,18 @@ import {
...
-import React, {ElementType, ReactElement} from 'react';
+import {ProgressCircle} from '@react-spectrum/progress';
+import React, {ElementType, ReactElement, useEffect, useRef, useState} from 'react';
 import {SpectrumButtonProps} from '@react-types/button';
+import {SpectrumProgressCircleProps} from '@react-types/progress';
 import styles from '@adobe/spectrum-css-temp/components/button/vars.css';
 import {Text} from '@react-spectrum/text';
 import {useButton} from '@react-aria/button';
 import {useHover} from '@react-aria/interactions';
+import {useLocalizedStringFormatter} from 'react-aria';
```
devongovett, line 34: We should use the individual package here (`@react-aria/i18n`) instead of the monopackage. That way if someone uses the individual button Spectrum package they don't need to install all of react aria.

packages/@react-spectrum/button/src/Button.tsx
```diff
@@ -39,15 +44,44 @@ function Button<T extends ElementType = 'button'>(props: SpectrumButtonProps<T>,
...
+  };
   let hasLabel = useHasChild(`.${styles['spectrum-Button-label']}`, domRef);
   let hasIcon = useHasChild(`.${styles['spectrum-Icon']}`, domRef);
+  let timeout = useRef<ReturnType<typeof setTimeout>>();
+  let [showLoader, setShowLoader] = useState(false);
+
+  useEffect(() => {
+    if (isPending) {
+      // Start timer when isPending is set to true.
+      timeout.current = setTimeout(() => {
```
devongovett, line 73: I think you might not need a ref for this? A local variable within the effect should work since the cleanup function has access to it.

packages/@react-spectrum/button/src/Button.tsx
```diff
@@ -39,15 +44,44 @@ function Button<T extends ElementType = 'button'>(props: SpectrumButtonProps<T>,
...
+  let stringFormatter = useLocalizedStringFormatter(intlMessages);
   let {styleProps} = useStyleProps(otherProps);
+  let progressCircleProps: SpectrumProgressCircleProps = {
+    'aria-label': stringFormatter.format('loading'),
+    isIndeterminate: true,
+    size: 'S',
+    ...(
+      (variant === 'overBackground' && style !== 'fill') ||
+      ((variant === 'cta' || variant === 'accent') && style !== 'outline')
+    ) && {variant: 'overBackground'}
```
devongovett, line 63: We will need to double check with design, but it looks like the pending state should also cause the button to become disabled, i.e. gray background. So I guess the spinner would always be blue? https://spectrum.adobe.com/page/button/#Pending

packages/@react-spectrum/button/src/Button.tsx
```diff
@@ -39,15 +44,44 @@ function Button<T extends ElementType = 'button'>(props: SpectrumButtonProps<T>,
     style = variant === 'accent' || variant === 'cta' ? 'fill' : 'outline',
     staticColor,
     isDisabled,
+    isPending,
     autoFocus,
     ...otherProps
   } = props;
   let domRef = useFocusableRef(ref);
   let {buttonProps, isPressed} = useButton(props, domRef);
```
devongovett, line 52: According to design, I think the button should stop responding to clicks during the delay before the spinner appears, even if it still looks enabled. So you might need to do something like this:

```js
useButton({
  ...props,
  isDisabled: props.isDisabled || props.isPending
}, domRef)
```

https://spectrum.adobe.com/page/button/#Delay-before-pending-state

packages/@react-spectrum/button/src/Button.tsx
iamwillpowell: Good catch. I've pushed the change.

packages/@react-spectrum/button/src/Button.tsx
iamwillpowell: Thanks. That was a holdover from Dylan's PR. Since we don't need to pass the ref around, it isn't needed.

packages/@react-spectrum/button/src/Button.tsx
iamwillpowell: I based my changes on the Spectrum Site and an XD file that @dannify sent me. In the XD file the buttons are not in a visually disabled state.

image

packages/@react-spectrum/button/src/Button.tsx
iamwillpowell: I ended up writing a small hook to filter out the press and click events while pending since the disabled visual state is closely tied to the disabled attribute. I added aria-disabled as well.

packages/@react-spectrum/button/src/Button.tsx
```diff
@@ -65,15 +109,17 @@ function Button<T extends ElementType = 'button'>(props: SpectrumButtonProps<T>,
         data-variant={variant}
         data-style={style}
         data-static-color={staticColor || undefined}
+        aria-disabled={isPending}
```
snowystinger, line 112:
```suggestion
        aria-disabled={isPending || undefined}
```
just so that false doesn't interfere with the aria-disabled that might come from `useButton`'s returned properties
https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/button/src/useButton.ts#L82

ideally, this would be set in `useButton` as well, trying to decide if this is a Spectrum specific thing. it might coincide with 'focusable but disabled' https://github.com/adobe/react-spectrum/issues/3662

packages/@adobe/spectrum-css-temp/components/button/index.css
```diff
@@ -157,6 +157,17 @@ governing permissions and limitations under the License.
     min-inline-size: unset;
     padding: var(--spectrum-global-dimension-size-65);
   }
+
+  &.is-pending {
+    .spectrum-CircleLoader {
```
snowystinger, line 162:
```suggestion
  &.spectrum-Button--pending {
    .spectrum-Button-circleLoader {
```

just convention for naming, doubledash for states that aren't native, and since we use CSS modules, while copying the name will work, it can be difficult to tell what file it came from, so we use the namespace for the component followed by the subcomponent name

packages/@react-spectrum/button/stories/Button.stories.tsx
```diff
@@ -102,6 +106,11 @@ export const UserSelect: ButtonStory = {
   }
 };
 
+export const IsPending: ButtonStory = {
+  name: 'Pending spinner',
+  render: (args) => <Pending {...args} />
```
snowystinger, line 111: nit pick, render first, name second
you could also get rid of name if you called the exported const `PendingSpinner` instead

packages/@react-spectrum/button/stories/Button.stories.tsx
```diff
@@ -224,3 +233,77 @@ function Example() {
...
+          {...props}>
+          <Text>with onClick</Text>
+        </PendingButtonOnClickComponent>
+      </View>
+
+      <View padding={16}>
+        <Button variant="primary" isPending={props.isPending} onPress={() => {window.alert('use storybook control to change this button isPending prop');}}>
+          <Text>Controlled pending</Text>
+        </Button>
+      </View>
```
snowystinger, line 264: don't need this if we just hook up all of the buttons to the controls

packages/@react-spectrum/button/test/Button.test.js
```diff
@@ -273,6 +273,40 @@ describe('Button', function () {
...
+        </Button>
+      );
+    }
+    let {getByRole} = render(<TestComponent />);
+    let button = getByRole('button');
+    expect(button).toHaveAttribute('aria-disabled', 'false');
+    fireEvent.click(button);
+    expect(button).toHaveAttribute('aria-disabled', 'true');
+    // Multiple clicks shouldn't call onPressSpy
+    fireEvent.click(button);
```
snowystinger, line 294: use `triggerPress` instead of `fireEvent.click`, it's more accurate to the events that get fired in a real browser

packages/@react-spectrum/button/test/Button.test.js
```diff
@@ -273,6 +273,40 @@ describe('Button', function () {
...
+    act(jest.runAllTimers);
```
snowystinger, line 295:
```suggestion
    act(() => {jest.advanceTimersByTime(999);});
    triggerPress(button);
    expect(onPressSpy).toHaveBeenCalledTimes(1);
    expect(button).toHaveAttribute('aria-disabled', 'true');
    expect(queryByRole('progressBar')).not.toBeInTheDocument();
    act(() => {jest.advanceTimersByTime(1);}); // 1 full second since first press
```
I know your way works, but this will match our other uses and lint won't complain about the wrong return value from runAllTimers for act should we ever make this a ts file

in addition, I'd like the delay to be in a test so that it's very obvious if we should ever change the delay and I want to test the boundary of the timer so that we know it actually waited to show the spinner

packages/@react-spectrum/button/test/Button.test.js
```diff
@@ -273,6 +273,40 @@ describe('Button', function () {
     expect(eventDown.defaultPrevented).toBeFalsy();
   });
 
+  // is pending state
+  it('displays a spinner after a short delay when isPending prop is true', function () {
+    jest.useFakeTimers();
```
snowystinger, line 278: go ahead and put this in a beforeAll, I doubt it'll affect any other test in this file

if it does, then wrap this set of tests which is dependent on faketimers in a new describe block with a beforeAll which turns on fake timers, an after each which runs all timers, and an afterAll which restores us to real timers

snowystinger (changes requested):
Is the spinner in `static color`'s correct? I'm unsure

packages/@react-spectrum/button/src/Button.tsx
devongovett: Wouldn't this mean the pressed state would still apply even if the events are not triggered? Did you try getting rid of the `:disabled` selector and just using `is-disabled` for the visual state?

packages/@adobe/spectrum-css-temp/components/button/index.css
```diff
@@ -157,6 +157,17 @@ governing permissions and limitations under the License.
...
+      position: absolute;
+    }
+
+    .spectrum-Button-label,
+    .spectrum-Icon {
+      opacity: 0;
```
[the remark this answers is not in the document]
iamwillpowell: Makes sense. I changed it.

packages/@adobe/spectrum-css-temp/components/button/index.css
```diff
@@ -157,6 +157,17 @@ governing permissions and limitations under the License.
     min-inline-size: unset;
     padding: var(--spectrum-global-dimension-size-65);
   }
+
+  &.is-pending {
+    .spectrum-CircleLoader {
+      position: absolute;
```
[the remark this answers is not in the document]
iamwillpowell: Works great

packages/@react-spectrum/button/docs/Button.mdx
```diff
@@ -84,6 +84,31 @@ function Example() {
 }
 ```
 
+## Pending
+
+Buttons can indicate that a quick progress taking place (e.g., saving settings on a server). After a 1 second delay an indeterminate spinner will be displayed in place of the button label and icon. You can trigger this behavior with the `isPending` prop.
```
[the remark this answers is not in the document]
iamwillpowell: Done

packages/@react-spectrum/button/intl/en-US.json
```diff
@@ -0,0 +1,3 @@
+{
+  "loading": "Loading…"
```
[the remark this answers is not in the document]
iamwillpowell: Add it as another prop, `isPendingLabel`, possibly?

packages/@adobe/spectrum-css-temp/components/button/index.css
iamwillpowell: Basically SuitCSS style names. Done

packages/@react-spectrum/button/stories/Button.stories.tsx
iamwillpowell: All the examples are hooked up to controls for isPending except three of the ones in the Pending story. I wanted to have clickable examples but I didn't want anyone to miss the externally controlled version either.

packages/@react-spectrum/button/stories/Button.stories.tsx
```diff
@@ -224,3 +233,77 @@ function Example() {
...
+function PendingButtonOnClickComponent(props) {
+  let [isPending, setPending] = useState(false);
+
+  let handlePress = (e) => {
+    action('press')(e);
+    setPending(true);
+    setTimeout(() => {
+      setPending(false);
+    }, timerValue);
+  };
```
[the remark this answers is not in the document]
iamwillpowell: I think you are correct that this is a problem is unrelated to my changes. Also, since `onClick` is deprecated, I'm inclined not to worry much about it. `Press` events work fine with keyboard. How do others on the team feel?

packages/@react-spectrum/button/src/Button.tsx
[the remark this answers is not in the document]
iamwillpowell: Thanks.

packages/@react-spectrum/button/stories/Button.stories.tsx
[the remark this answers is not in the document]
snowystinger: It never has, onClick is specific to mouse, it never handled keyboard. The press events are the correct way to go.

packages/@react-spectrum/button/src/Button.tsx
```diff
@@ -20,34 +20,87 @@ import {
...
+  // Don't allow interaction while isPending is true
+  if (props.isPending) {
+    props.onPress = undefined;
+    props.onPressStart = undefined;
+    props.onPressEnd = undefined;
+    props.onPressChange = undefined;
+    props.onPressUp = undefined;
+    props.onFocus = undefined;
+    props.onBlur = undefined;
+    props.onFocusChange = undefined;
```
snowystinger, line 47: I think if we remove these, then people may have trouble with cleanup outside the component. It's still focusable, so I think these properties should stay.

On another related note. If someone set isPending during pressstart or keydown... should we still finish the rest of the events? I'm not sure the use case for this, but it seems like you could easily get into trouble if you did this.

should keydown/keyup be disabled as well? they aren't really press and Tab or Esc might be useful, or cleanup of modifier keys?

packages/@react-spectrum/button/src/Button.tsx
```diff
@@ -20,34 +20,87 @@ import {
...
+    isPending,
     autoFocus,
     ...otherProps
   } = props;
   let domRef = useFocusableRef(ref);
   let {buttonProps, isPressed} = useButton(props, domRef);
   let {hoverProps, isHovered} = useHover({isDisabled});
+  let stringFormatter = useLocalizedStringFormatter(intlMessages);
   let {styleProps} = useStyleProps(otherProps);
+  let progressCircleProps: SpectrumProgressCircleProps = {
```
snowystinger, line 76: any reason not to just put these directly on the progress circle? then we wouldn't need that extra variant spread, we'd just put the logic inside the prop itself

packages/@react-spectrum/button/test/Button.test.js
```diff
@@ -26,6 +26,14 @@ import React from 'react';
 describe('Button', function () {
   let onPressSpy = jest.fn();
 
+  beforeAll(() => {
+    jest.useFakeTimers();
+  });
+
+  afterAll(() => {
+    jest.useRealTimers();
+  });
```
snowystinger, line 35: don't need to call this if it's at the top level

packages/@react-spectrum/button/test/Button.test.js
```diff
@@ -273,6 +281,42 @@ describe('Button', function () {
...
+      );
+    }
+    let {getByRole} = render(<TestComponent />);
+    let button = getByRole('button');
+    expect(button).not.toHaveAttribute('aria-disabled');
+    triggerPress(button);
+    // Button is disabled immediately, but spinner visibility is delayed
+    expect(button).toHaveAttribute('aria-disabled', 'true');
+    // Multiple clicks shouldn't call onPressSpy
+    triggerPress(button);
```
snowystinger, line 303: any reason not to check that the spinner hasn't shown up yet?
https://github.com/adobe/react-spectrum/pull/4900#discussion_r1303657303

packages/@react-spectrum/button/src/Button.tsx
snowystinger: please bring this up in standup/testing as well

packages/@react-spectrum/button/stories/Button.stories.tsx
```diff
@@ -102,6 +106,10 @@ export const UserSelect: ButtonStory = {
   }
 };
 
+export const PendingSpinner: ButtonStory = {
```
reidbarber, line 109: For combination `style="outline"`  + `staticColor="black"`, the spinner color doesn't currently match the text/outline color.

Screenshot 2023-08-28 at 12 42 05 PM

packages/@react-spectrum/button/test/Button.test.js
```diff
@@ -273,6 +281,42 @@ describe('Button', function () {
     expect(eventDown.defaultPrevented).toBeFalsy();
   });
 
+  // iPending state
```
reidbarber, line 284: super-nit typo
```suggestion
  // isPending state
```

packages/@react-types/button/src/index.d.ts
```diff
@@ -74,6 +74,8 @@ export interface SpectrumButtonProps<T extends ElementType = 'button'> extends A
   style?: 'fill' | 'outline',
   /** The static color style to apply. Useful when the button appears over a color background. */
   staticColor?: 'white' | 'black',
+  /** Whether to disable clicks and display a loading spinner. */
```
reidbarber, line 77: We may want to mention something about the delay here as well.

packages/@react-spectrum/button/docs/Button.mdx
```diff
@@ -84,6 +84,31 @@ function Example() {
 }
 ```
 
+## Pending
```
reidbarber, line 87: This may be follow-up, but will other buttons support the pending state? (i.e. action button)

reidbarber (changes requested):
Looks great so far!

packages/@react-spectrum/button/stories/Button.stories.tsx
reidbarber: We could just not use the `overBackground` for this combination, or custom style it black via `UNSAFE_className`.

packages/@react-spectrum/button/src/Button.tsx
iamwillpowell: removed blur and focus events based on feedback

packages/@react-spectrum/button/stories/Button.stories.tsx
iamwillpowell: Will custom style in a separate PR

packages/@react-spectrum/button/intl/en-US.json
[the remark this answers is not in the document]
iamwillpowell: changing pending aria label to better account for use cases by appending "pending" to the button label.

packages/@react-stately/utils/test/useControlledState.test.js
```diff
@@ -270,7 +270,7 @@ describe('useControlledState tests', function () {
     }
 
     function TransitionButton({onClick}) {
-      let [isPending, startTransition] = React.useTransition();
+      let [UNSTABLE_isPending, startTransition] = React.useTransition();
```
snowystinger, line 273: I don't think you meant to edit this file

iamwillpowell: Oops. I searched to hard. Thanks for catching that.

snowystinger (approved):

reidbarber (approved):

PR #4913, opened 2023-08-15
Add support for SSR to RAC collections

### LINKED ISSUE #4633: When copy pasting Select and ComboBox examples of React-Aria Components I get an error
# 🐛 Bug Report 

I copy/pasted the Select components "Reusable Wrappers" example of React-Aria component in my project and I get the following error at runtime:

```
Unhandled Runtime Error
TypeError: Cannot destructure property 'arrowProps' of '(0 , react__WEBPACK_IMPORTED_MODULE_0__.useContext)(...)' as it is null.
```

Copy pasted the "Reusable Wrapper" exemple and then using it like this (I renamed `MyItem` to `MySelectItem`):

```
<MySelect label="Ice cream flavor">
       <MySelectItem>Chocolate</MySelectItem>
       <MySelectItem>Mint</MySelectItem>
        <MySelectItem>Strawberry</MySelectItem>
        <MySelectItem>Vanilla</MySelectItem>
</MySelect>
```

## 🤔 Expected Behavior

No error at runtime.

## 😯 Current Behavior

```
Unhandled Runtime Error
TypeError: Cannot destructure property 'arrowProps' of '(0 , react__WEBPACK_IMPORTED_MODULE_0__.useContext)(...)' as it is null.

Call Stack
$44f671af83e7d9e0$var$OverlayArrow
node_modules\.pnpm\react-aria-components@1.0.0-alpha.4_react-dom@18.2.0_react@18.2.0\node_modules\react-aria-components\dist\import.mjs (1954:10)
renderWithHooks
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (10648:0)
updateForwardRef
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (14677:0)
beginWork$1
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (17369:0)
beginWork
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (25650:0)
performUnitOfWork
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (24501:0)
workLoopSync
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (24217:0)
renderRootSync
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (24182:0)
recoverFromConcurrentError
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (23428:0)
performConcurrentWorkOnRoot
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\react-dom\cjs\react-dom.development.js (23334:0)
workLoop
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\scheduler\cjs\scheduler.development.js (261:0)
flushWork
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\scheduler\cjs\scheduler.development.js (230:0)
MessagePort.performWorkUntilDeadline
node_modules\.pnpm\next@13.4.4_react-dom@18.2.0_react@18.2.0\node_modules\next\dist\compiled\scheduler\cjs\scheduler.development.js (537:0)
```

## 💻 Code Sample

Here a Github repository with the code: https://github.com/patricklafrance/next13-react-aria/blob/main/app/react-aria/MySelect.tsx

Uncomment the following code in `app/react-aria/page.tsx`:

```
{/* <div>
    <h2>Select</h2>
    <MySelect label="Ice cream flavor">
        <MySelectItem>Chocolate</MySelectItem>
        <MySelectItem>Mint</MySelectItem>
        <MySelectItem>Strawberry</MySelectItem>
        <MySelectItem>Vanilla</MySelectItem>
    </MySelect>
</div> */}
```

## 🌍 Your Environment

| react-spectrum   | 3.27.0
| react-aria   | 3.25.0
| react-aria-components   | 1.0.0-alpha.4
| Next.js (with the App folder)   | 13.4.4
| Browser          |  Chrome 113
| Operating System | Windows

### LINKED ISSUE #4870: Table component does not render any rows on server

When server side rendering a basic Table component the thead and tbody are empty until the component loads on the client.

### 🤔 Expected Behavior?

The table should render all rows on the server.

### 😯 Current Behavior

The table does not render rows on the server.

### 💁 Possible Solution

Unsure.

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

```js
function () {
  return (
    <Table aria-label="Files">
      <TableHeader>
        <Column isRowHeader>Name</Column>
        <Column>Type</Column>
        <Column>Date Modified</Column>
      </TableHeader>
      <TableBody>
        <Row>
          <Cell>Games</Cell>
          <Cell>File folder</Cell>
          <Cell>6/7/2020</Cell>
        </Row>
      </TableBody>
    </Table>
  );
}
```

Output when rendering on the server:

```html
<table
  class="react-aria-Table"
  aria-label="Files"
  id="react-aria-:R6kl:"
  role="grid"
  tabindex="0"
  aria-describedby=""
>
  <thead role="rowgroup" class="react-aria-TableHeader"></thead>
  <tbody
    role="rowgroup"
    class="react-aria-TableBody"
    data-empty="true"
  ></tbody>
</table>
```

### Version

1.0.0-alpha.5

Chrome

### If other, please specify.

_No response_

MacOS

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

### LINKED ISSUE #4938: RAC: `Tabs` is renders late

I am using `Combobox` and `Tabs` components from react-aria-components with `"use client"` in `Next.js v13.4.18`. When the page loads, `Combobox` renders instantly but `Tabs` renders slower/later than other components.

### 🤔 Expected Behavior?

`Tabs` component needs to be rendered instantly because other components are rendered when the page loads. If `Tabs` is renders late, the layout is empty. It makes the UI look bad.

### 😯 Current Behavior

`Tabs` renders later/slower than other components.

### 💁 Possible Solution

_No response_

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

https://codesandbox.io/p/sandbox/react-aria-components-issue-54vk42

### Version

1.0.0-nightly.4081

Chrome

### If other, please specify.

_No response_

Windows 10 19045.3324

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
Fixes #4870, fixes #4633, fixes #4938

This adds support for SSR to RAC collections. Previously, nothing would be rendered on the server because React doesn't support portals during SSR and we rely on one to implement collections. Now, during SSR, we render the collection elements without a portal, however, each collection component returns `null` so that nothing is actually output into the HTML. We use a context to keep track of the current collection node, and append children to it during render. This would normally not be safe, however, because SSR occurs in a single pass, we know the children will be rendered in order. If we ensure that the collection "portal" is always rendered first, it will be updated before render to the real "DOM". After hydration, we switch to our existing strategy of using a portal to track updates. Credit for the idea of double rendering goes to @souporserious and the [reforest](https://github.com/souporserious/reforest) project.

For components like ComboBox and Select, the collection is deeper than the immediate children (e.g. in the nested ListBox within a Popover). So in order for us to get the collection before rendering, even when the popover is closed, we actually render the children twice (as two separate instances). The first instance is used to build the collection, and the second uses that collection to render the real DOM with the correct content (e.g. the value based on the selectedKey). Elements inside a `<Hidden>` tree know they are hidden via a context and return `null` to prevent rendering any HTML. However, that only works for our own components, so we render into a [`<template>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) element, which the browser will not display, and which will not affect the accessibility tree. After SSR, we use a portal into a `DocumentFragment` instead, which will not be exposed to the document at all.


packages/@react-aria/ssr/src/SSRProvider.tsx
```diff
@@ -71,7 +69,9 @@ function LegacySSRProvider(props: SSRProviderProps): JSX.Element {
 
   return (
     <SSRContext.Provider value={value}>
-      {props.children}
+      <IsSSRContext.Provider value={isSSR}>
```
devongovett, line 72: Apparently SSR was kinda broken in React 16/17. The problem was that the SSR context would reset whenever `isSSR` changed, and the children would not necessarily re-render in the same order, which could result in duplicate ids. The fix is to split that into two contexts so the ids don't change value.

packages/@react-stately/combobox/src/useComboBoxState.ts
```diff
@@ -219,11 +219,6 @@ export function useComboBoxState<T extends object>(props: ComboBoxStateOptions<T
       }
     }
 
-    // If it is the intial render and inputValue isn't controlled nor has an intial value, set input to match current selected key if any
-    if (isInitialRender.current && (props.inputValue === undefined && props.defaultInputValue === undefined)) {
-      resetInputValue();
-    }
```
devongovett, line 225: Changing this caused the tests to be updated above, but I discussed with Daniel and we think it's ok. This way the value is known during the initial render, which means the input value can be populated during ssr.

packages/react-aria-components/src/ComboBox.tsx
```diff
@@ -42,11 +42,7 @@ export interface ComboBoxRenderProps {
    * Whether the combobox is required.
    * @selector [data-required]
    */
-  isRequired: boolean,
-  /**
-   * State of the combobox.
-   */
-  state: Omit<ComboBoxState<unknown>, 'children' | 'setOpen' | 'toggle' | 'open' | 'close' | 'selectionManager' | 'setSelectedKey' | 'setFocused' | 'collection' | 'commit' | 'revert'>
```
devongovett, line 49: Decided to just remove state as a render prop from ComboBox, Select, and Tabs, which have a circular dependency on the children in order to build the state. We were previously omitting a lot of properties anyway, but now it is actually not possible to get the state before the first render.

If this comes up as a request we could potentially make a proxy object that has all of these properties, but uses getters so that the object can be passed in but the actual property values aren't available until after the render.

snowystinger (approved):
Appears to be behaving well for me now

packages/react-aria-components/src/Select.tsx
```diff
@@ -141,45 +122,56 @@ function Select<T extends object>(props: SelectProps<T>, ref: ForwardedRef<HTMLD
...
-        name={props.name} />
-    </Provider>
+    <>
+      {/* Render a hidden copy of the children so that we can build the collection even when the popover is not open.
+        * This should always come before the real DOM content so we have built the collection by the time it renders during SSR. */}
+      <Hidden>
+        <Provider
+          values={[
+            [InternalSelectContext, {state, valueProps, placeholder: props.placeholder}],
+            [ListBoxContext, {document}]
```
LFDanLu, line 132: Don't we have to provide `items` the ListBox here as well like we do for ComboBox?

LFDanLu (changes requested):
Just one thing I noticed with Select's `items` case but otherwise things look good to me

packages/react-aria-components/src/Select.tsx
devongovett: hmm interesting. I wonder if we should even support `items` as a prop on the top-level `Select` or if `items` on the `ListBox` is enough? For ComboBox we needed to support `defaultItems` which isn't a prop of `ListBox`. Should `Select` match ComboBox or do we treat that as a special case? Sorta weird to have `items` in two places.

LFDanLu: my gut feeling here is that we should get rid of `items` on the top level `Select` if possible, it feels more like a carry over from RSP's api that isn't needed here since the user provides the ListBox in RAC and I agree that having `items` in two places is pretty confusing.

LFDanLu (approved):
Pushed a small fix for the Select doc examples, LGTM now

snowystinger (approved):

souporserious:
Super cool to see this pattern make it's way into Spectrum! Just wanted to call out that the double render idea was originally inspired by @nihgwu [here](https://twitter.com/_neonie/status/1553807303794720768) 😊

PR #4931, opened 2023-08-17
Do not close overlays on blur if relatedTarget is null

### LINKED ISSUE #4130: Dialog popover loses focus when switching tabs in Chrome/FF
# 🐛 Bug Report 

In Chrome, when clicking on another browser tab with https://react-spectrum.adobe.com/react-spectrum/DialogTrigger.html open, dialog closes. This is being used in Unified Shell for the Help Center.

## 🤔 Expected Behavior
When clicking on another browser tab then returning to the original tab, the dialog should remain open. 

## 😯 Current Behavior

The popover dialog is closed when returning to the original tab, losing any info that was in there.

## 🌍 Your Environment

| react-spectrum   |
| Browser          |
| Operating System |

### LINKED ISSUE #4384: DatePicker/DateRangePicker: On pressing left/right arrow keys while focus is on next/previous buttons is closing the picker
# 🐛 Bug Report 

On DatePicker and RangePicker, if user presses left or right arrow keys while focus is on Previos/Next month buttons, the picker is closing. 

## 🤔 Expected Behavior

Expected Behavior: This unexpected closing of the picker should not happen. 

## 🔦 Context

Issue can be observed on [Storybook](https://reactspectrum.blob.core.windows.net/reactspectrum/41e26fa0f01ee9ab6e5c21edf6f7bb5d498bf40f/storybook/index.html?path=/story/date-and-time-datepicker--default&providerSwitcher-express=false&providerSwitcher-toastPosition=bottom) or [react-spectrum docs](https://react-spectrum.adobe.com/react-spectrum/DatePicker.html)

### LINKED ISSUE #4922: DatePicker is difficult/impossible to use with screen reader

When navigating in the DatePicker with screen reader (here VoiceOver in Chrome), the calendar closes unexpectedly.

### 🤔 Expected Behavior?

I expect that the calendar will not close totally unexpected when using it with the screen reader.

### 😯 Current Behavior

As explained in the summary, the calendar closes while navigating in it with VoiceOver in Chrome. The same does not happen in Safari, however. A screen recording of this is attached below.

I also experience the same when using the [useDatePicker](https://react-spectrum.adobe.com/react-aria/useDatePicker.html). From what I have seen while debugging, this happens due to `onFocus` in the `buttonProps`  used in the `CalendarCell`. It seems that changing the focused date somehow triggers the state to close. When excluding `onFocus` from the `buttonProps`, it works fine, but the actual focus is then not moved of course.

Could someone tell me why this happens and if it can be solved?

https://github.com/adobe/react-spectrum/assets/15145686/47f2716b-3917-46af-9774-eebabaa88ae2

### 💁 Possible Solution

It would be great if the calendar was not so unpredictable and worked fine using VoiceOver.

### 🔦 Context

I am trying to implement an accessible DatePicker in React for a large group of customers, and struggle doing so, which is a pity.

### 🖥️ Steps to Reproduce

1. Open [DatePicker](https://react-spectrum.adobe.com/react-spectrum/DatePicker.html in Chrome
2. Turn on VoiceOver
3. Open the calendar
4. Try to navigate in it with ctrl+option+arrows
 

### Version

3.19.0 of DatePicker and 3.27.0 of useDatePicker

Chrome

### If other, please specify.

_No response_

MacOS

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
In collaboration with @LFDanLu

Fixes #4922, fixes #4130, fixes #4384, fixes https://github.com/adobe/react-spectrum/issues/4533

When debugging #4922, we found a Chrome bug where extraneous blur events are fired with `relatedTarget = null` when navigating with the VoiceOver cursor (via Control + Option + Arrow keys). https://bugs.chromium.org/p/chromium/issues/detail?id=1473752 Note that this affects more than just DatePicker - it also affects any button contained within a Popover.

In addition, Chrome and Firefox fire blur events with `relatedTarget = null ` when switching tabs, which causes overlays to close.

This changes the logic in `useOverlay` to not close in those cases. We haven't thought of a case where `relatedTarget` was `null` where we _wouldn't_ want to close the overlay. Clicking on the body should already be handled by `useInteractOutside`.

Note: this will change the behavior of iframes as well. Popovers will no longer close when clicking outside an iframe.

### DIFF
--- a/packages/@react-aria/overlays/src/useOverlay.ts
+++ b/packages/@react-aria/overlays/src/useOverlay.ts
@@ -125,10 +125,16 @@ export function useOverlay(props: AriaOverlayProps, ref: RefObject<Element>): Ov
   let {focusWithinProps} = useFocusWithin({
     isDisabled: !shouldCloseOnBlur,
     onBlurWithin: (e) => {
+      // Do not close if relatedTarget is null, which means focus is lost to the body.
+      // That can happen when switching tabs, or due to a VoiceOver/Chrome bug with Control+Option+Arrow navigation.
+      // Clicking on the body to close the overlay should already be handled by useInteractOutside.
+      // https://github.com/adobe/react-spectrum/issues/4130
+      // https://github.com/adobe/react-spectrum/issues/4922
+      //
       // If focus is moving into a child focus scope (e.g. menu inside a dialog),
       // do not close the outer overlay. At this point, the active scope should
       // still be the outer overlay, since blur events run before focus.
-      if (e.relatedTarget && isElementInChildOfActiveScope(e.relatedTarget)) {
+      if (!e.relatedTarget || isElementInChildOfActiveScope(e.relatedTarget)) {
         return;
       }
 


dannify (approved):

LFDanLu (approved):
Did a brief sweep and things still seem to work as expected barring the following cases which we've discussed:
- tabbing away from a Popover to outside the storybook iframe/browser window doesn't close the Popover. Expected since we want to support Shell's use case of not closing the Dialog on browser tab change
-  cannot tab to adjacent tabbable elements next to the Popover's trigger. For others testing this PR/looking at this PR in the future, this is actually due to `inert` change made in https://github.com/adobe/react-spectrum/pull/4314. I will file a ticket in our backlog for this.

Also verified that this change fixes https://github.com/adobe/react-spectrum/issues/4533 as well, updated the description appropriately. Lets get this merged in for thorough testing next week!

snowystinger (approved):

PR #4966, opened 2023-08-23
Add a script to convert anatomy SVGs to use CSS variables

devongovett:
After you export an SVG from XD, run this script pointing at the file and hard coded colors will be replaced by the closest variable. Also fixes the font family syntax.


reidbarber (approved):
Nice!

snowystinger (approved):

PR #4976, opened 2023-08-25
(WIP Submenu support

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/@react-aria/menu/src/useMenu.ts
```diff
@@ -72,7 +75,7 @@ export function useMenu<T>(props: AriaMenuOptions<T>, state: TreeState<T>, ref:
   });
 
   return {
-    menuProps: mergeProps(domProps, {
+    menuProps: mergeProps(domProps, {onKeyDown}, {
       role: 'menu',
       // this forces AT to move their cursors into any open sub dialogs, the sub dialogs contain hidden close buttons in order to come back to this level of the menu
       'aria-hidden': state.expandedKeys.size > 0 ? true : undefined,
```
LFDanLu, line 81: One thing that might become necessary is having `useMenu` accept the `menuTreeState` created by `useMenuState` and its own `level` so that any given menu/submenu check if it has a expandedKey for the above. Alternatively, we would need to move `useSubMenuTrigger` or something up a level so that it can initialize the `level` for a given Menu and thus setup an `expandedKeys` set to pass to useTreeState

LFDanLu: For now I've made it so we update each menu/submenus's treestate expanded keys when a submenu is expanded and collapsed for the above. It felt like the easiest path forward since we already have access to the TreeState here. Open for discussion

packages/@react-aria/menu/src/useMenu.ts
```diff
@@ -30,7 +30,10 @@ export interface AriaMenuOptions<T> extends Omit<AriaMenuProps<T>, 'children'> {
    * An optional keyboard delegate implementation for type to select,
    * to override the default.
    */
-  keyboardDelegate?: KeyboardDelegate
+  keyboardDelegate?: KeyboardDelegate,
+  // TODO: Should we have an accompanying onKeyUp even though we don't use it and then just extend KeyboardEvents?
+  // Should this be UNSTABLE? That would mean the props from useSubMenuTrigger would be the UNSTABLE variants as well
+  onKeyDown?: (e: KeyboardEvent) => void
```
LFDanLu, line 36: Open question, this onKeyDown is provided by `useSubMenuTrigger` but wondering if it is weird just to have it by itself without a `onKeyUp` handler. Also would be good to decide what pattern we want for potentially unstable props when they are provided by alpha level hooks. Feels nice to have an alpha hook return the standard DOMAttributes but open to appending UNSTABLE to them

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -85,7 +83,18 @@ export interface AriaMenuItemProps {
...
+  // TODO: Add descriptions when these props are okayed
+  // Open it up fully to all pressProps + hoverEvents + keyboardEvents?
+  // Should these all be UNSTABLE? That would mean the return types from useSubMenuTrigger would be the UNSTABLE variants as well...
+  onPressStart?: (e: PressEvent) => void,
+  onPress?: (e: PressEvent) => void,
+  onHoverChange?: (isHovering: boolean) => void,
+  onKeyDown?: (e: KeyboardEvent) => void,
+  onBlur?: (e: FocusEvent<Element>) => void,
+  'aria-expanded'?: boolean | 'true' | 'false',
+  'aria-controls'?: string
```
LFDanLu, line 97: Same question as https://github.com/adobe/react-spectrum/pull/4976/files#r1333680477.

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -161,10 +151,15 @@ export function useMenuItem<T>(props: AriaMenuItemProps, state: TreeState<T>, re
     role,
     'aria-label': props['aria-label'],
     'aria-labelledby': labelId,
-    'aria-describedby': [descriptionId, keyboardId].filter(Boolean).join(' ') || undefined
+    'aria-describedby': [descriptionId, keyboardId].filter(Boolean).join(' ') || undefined,
+    // TODO: perhaps we should just expect that the user would spread these on the submenu trigger directly
+    // instead of just passing it to useMenuItem? Kinda want to follow useButton's pattern where we handle aria-attributes pass throughs
+    'aria-controls': isTrigger ? props['aria-controls'] : undefined,
+    'aria-haspopup': hasPopup,
+    'aria-expanded': isTrigger ? props['aria-expanded'] : undefined
```
LFDanLu, line 159: open to opinions

packages/@react-aria/menu/src/useSubMenuTrigger.ts
```diff
@@ -0,0 +1,244 @@
...
+import {useLocale} from '@react-aria/i18n';
+
+export interface AriaSubMenuTriggerProps {
+  /** An object representing the submenu trigger menu item. Contains all the relevant information that makes up the menu item. */
+  node: RSNode<unknown>,
+  /** Whether the submenu trigger is disabled. */
+  isDisabled?: boolean,
+  // TODO: naming. Also talk about if this should be customizable in this hook or if it belongs somewhere else
+  /** Type of the submenu being rendered. */
+  subMenuType?: 'dialog' | 'menu',
```
LFDanLu, line 30: Naming of this prop? I assume unavailable menu items is something we'll want to support in aria as well

packages/@react-aria/menu/src/useSubMenuTrigger.ts
```diff
@@ -0,0 +1,244 @@
...
+  subMenuTriggerProps: AriaMenuItemProps,
+  subMenuProps: AriaMenuOptions<T>,
+  popoverProps: Pick<AriaPopoverProps, 'isNonModal'>,
+  // TODO: descriptions
+  // TODO: perhaps cherry pick these props from existing types, comes from several different types though
+  overlayProps:
+  {
+    onExit: () => void,
+    disableFocusManagement: boolean,
+    shouldCloseOnInteractOutside: (element: Element) => boolean
```
LFDanLu, line 47: Leaning towards just redefining them. `onExit` is for the spectrum PopOver, `disableFocusManagement` is for aria Overlay, `shouldCloseOnInteractOutside` is for the `useOverlay` call made by `usePopover` (aka aria Popover)

packages/@react-aria/menu/src/useSubMenuTrigger.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+}
+
+// TODO description
+// TODO: debatable if we should have a useSubMenu hook for the submenu keyboard handlers. Feels better to have it here so we don't need to ferry around
+// things like parentMenuRef and the open timeout canceling since we centralized them here
```
LFDanLu, line 53: open for discussion

packages/@react-aria/menu/src/useSubMenuTrigger.ts
```diff
@@ -0,0 +1,244 @@
...
+      cancelOpenTimeout();
+    };
+  }, [cancelOpenTimeout]);
+
+
+  // TODO: problem with setting up the submenu close handlers here is that we don't get access to the onClose the user provides
+  // to the menu since this is too far up . Maybe just provide the subMenuTriggerState and have useMenu and useMenuItem handle the key handlers internally?
+  // Maybe we should just have the user pass something like onMenuClose to the trigger level since we have the same problem of being unable to call onClose if the user interacts outside?
+  // Maybe have Menu call onClose in an effect when it detects that it is unmounting? StrictMode is problematic since the cleanup fires even when the component isn't actually unmounting
+  // (not a problem if we wanna keep onClose to fire only if the user selects a item)
```
LFDanLu, line 88: Followup work captured in https://github.com/orgs/adobe/projects/19/views/20?pane=issue&itemId=39054342 for getting onClose to fire whenever a menu/submenu closes rather than only when the user selects an item. Ideally it could happen on Menu unmount but had some problems with StrictMode and effect cleanup behavior. Also would require useMenu to accept a new prop alongside `onClose` since we need `onClose` to still close all menus on item select but the user provided onClose should only fire in said effect cleanup or whatever method of calling it on unmount we go with

packages/@react-aria/menu/src/useSubMenuTrigger.ts
```diff
@@ -0,0 +1,244 @@
...
+        cancelOpenTimeout();
+      }
+    }
+  };
+
+  let onBlur = (e) => {
+    // TODO: getting rid of the (!isElementInChildOfActiveScope(e.relatedTarget) part of the check below fixes hovering to open a root unavailable menu item when focus is on a submenu
+    // but breaks the case where the user is hovering over a submenu's submenu trigger child item and then hovers a root menu item (eg hover lvl 1 item 2 -> hover lvl 2 item 3 -> hover lvl 1 item 3)
+    // Ideally we'd be able to track the full menu tree and then check if focus has moved to an element that isn't part of the current submenu tree and close it then
+    if (state.isOpen && parentMenuRef.current.contains(e.relatedTarget)) {
```
LFDanLu, line 185: captured in ticket, will want to eventually be able to track if any arbitrary click/focus movement is within a root menu's tree or outside the menu tree

packages/@react-aria/menu/src/useSubMenuTrigger.ts
LFDanLu: Also `onExit` is spectrum Popover specific, but I figure ideally we'd have support for a prop like that in our aria Popover

packages/@react-aria/overlays/src/usePopover.ts
```diff
@@ -46,7 +46,11 @@ export interface AriaPopoverProps extends Omit<AriaPositionProps, 'isOpen' | 'on
    *
    * @default false
    */
-  isKeyboardDismissDisabled?: boolean
+  isKeyboardDismissDisabled?: boolean,
+  // TODO: note new prop, essentially a passthrough to useOverlay, allows me to control when useOverlay will
+  // try to close the popover even if focus moves outside of the overlay (aka going from submenu to submenu trigger via mouse hover)
+  // TODO: update description to be popover specific
+  shouldCloseOnInteractOutside?: (element: Element) => boolean
```
LFDanLu, line 53: how do we feel about exposing this?

packages/@react-spectrum/menu/src/MenuTrigger.tsx
```diff
@@ -56,9 +56,12 @@ function MenuTrigger(props: SpectrumMenuTriggerProps, ref: DOMRef<HTMLElement>)
   }
 
   let isMobile = useIsMobileDevice();
+  let menuTreeState = UNSTABLE_useMenuState({}, state);
+  // TODO: override menuTriggerState's .close so we clear the expandedKeysStack when clicking on the underlay or whenever
+  // Popover's .close is triggered. Alternatively, we could potentially clear it in a cleanup effect when the root menu unmounts...
+  let close = menuTreeState.closeAll;
```
LFDanLu, line 62: I figure aria hook users would do the same when adding submenu support to their menus.

packages/@react-spectrum/menu/src/SubMenuTrigger.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+// TODO: Add shouldFlip and closeOnSelect if we feel like those should be customizable for SubMenuTrigger
+// Other MenuTriggerProps like onOpenChange and positioning stuff have been removed as per discussion
+interface SubMenuTriggerProps {
+  /**
+   * The contents of the SubMenuTrigger - a Item and a Menu.
+   */
+  children: ReactElement[],
+  targetKey: Key
+}
```
LFDanLu, line 30: Open to opinions, would we want `closeOnSelect` to be specific to each submenu? Not sure of a good use case there

packages/@react-spectrum/menu/src/SubMenuTrigger.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+export interface SpectrumSubMenuTriggerProps extends Omit<SubMenuTriggerProps, 'targetKey'> {}
+
+// TODO: got rid of user provided ref support since it doesn't really make sense for submenus IMO
+function SubMenuTrigger(props: SubMenuTriggerProps) {
```
LFDanLu, line 35: mirrors ContextualHelpTrigger

packages/@react-spectrum/menu/src/context.ts
```diff
@@ -10,18 +10,21 @@
...
   onClose?: () => void,
   closeOnSelect?: boolean,
   shouldFocusWrap?: boolean,
   autoFocus?: boolean | FocusStrategy,
   ref?: MutableRefObject<HTMLDivElement>,
-  state?: MenuTriggerState
+  state?: MenuTriggerState,
+  menuTreeState?: MenuTreeState,
+  onKeyDown?: (e: KeyboardEvent) => void,
+  onBackButtonPress?: () => void
```
LFDanLu, line 27: passed down through context just for MenuTrays so we can avoid adding it to the Menu props

packages/@react-spectrum/menu/stories/SubMenu.stories.tsx
```diff
@@ -0,0 +1,587 @@
...
+        <Item key={item.name}>{item.name}</Item>
+        <Menu items={item.children} onAction={action(`${item.name} onAction`)}>
+          {(item) => dynamicRenderFunc(item)}
+        </Menu>
+      </SubMenuTrigger>
+    );
+  } else {
+    return <Item key={item.name}>{item.name}</Item>;
+  }
+};
```
LFDanLu, line 94: Perhaps we can provide these render funcs (see line 194-214 as well) in the docs so users don't have to come up with this themselves

packages/@react-stately/menu/src/useMenuState.ts
```diff
@@ -0,0 +1,58 @@
...
+  // TODO descriptions
+  expandedKeysStack: Key[],
+  closeAll: () => void,
+  openSubMenu: (triggerKey: Key, level: number) => void,
+  closeSubMenu: (triggerKey: Key, level: number) => void
+}
+
+// TODO descriptions and naming. useMenuState or useMenuTreeState?
+// Added props here so we can handle adding props in the future if need be, otherwise it would be a breaking change
+export function UNSTABLE_useMenuState(props: {}, state: MenuTriggerState): MenuTreeState  {
```
LFDanLu, line 26: open to opinions, also added placeholder prop arg for future additions

packages/@react-spectrum/menu/src/SubMenuTrigger.tsx
```diff
@@ -0,0 +1,136 @@
...
+  let menuRef = useRef<HTMLDivElement>();
+  let {
+    children,
+    targetKey
+  } = props;
+
+  let [menuTrigger, menu] = React.Children.toArray(children);
+  let {popoverContainerRef, trayContainerRef, menu: parentMenuRef, menuTreeState, state} = useMenuStateContext();
+  let triggerNode = state.collection.getItem(targetKey);
+  let subMenuTriggerState = UNSTABLE_useSubMenuTriggerState({triggerKey: targetKey}, {...menuTreeState, ...state});
```
LFDanLu, line 46: Note how I combine the TreeState (from Menu's call of `useTreeState`) and MenuTreeState (from root MenuTrigger's call of `useMenuState`). Not sure if we have this pattern elsewhere but I need access to the tree collection and the open/close submenu handlers

packages/@react-stately/menu/src/useSubMenuTriggerState.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export interface SubMenuTriggerProps {
+  /** Key of the trigger item. */
+  triggerKey: Key
+}
+
+export interface SubMenuTriggerState extends OverlayTriggerState {
+  // TODO descriptions
+  // Need some form of isOpen and close for usePopover. Also need closeAll for onClose for menu items, would be kinda weird
+  // to pass in both menuTreeState and the state returned by useSubMenuTriggerState to useMenu.
```
LFDanLu, line 27: This state takes the role of tracking the trigger's submenu open state much like `useMenuTriggerState` would. It also provides some `closeAll` so that clicking on a menu item will close all menus. 
the comment should also read 
```suggestion
export interface SubMenuTriggerState extends OverlayTriggerState {
  // TODO descriptions
  // Need some form of isOpen and close for usePopover. Also need closeAll for onClose for menu items, would be kinda weird
  // to pass in both menuTreeState and the state returned by useSubMenuTriggerState to useSubMenuTrigger.
```

packages/@react-stately/menu/src/useSubMenuTriggerState.ts
```diff
@@ -0,0 +1,84 @@
...
+  isOpen: boolean,
+  focusStrategy: FocusStrategy | null,
+  open: (focusStrategy?: FocusStrategy | null) => void,
+  close: () => void,
+  closeAll: () => void,
+  // TODO: add setOpen and toggle for type parity with useOverlayTriggerState type that Tray and Popover expect. Perhaps call useOverlayTriggerState in
+  // this hook and just make it controlled? Then spread? Kinda weird since they are unused but needed for type parity.
+  toggle: (focusStrategy?: FocusStrategy | null) => void,
+  setOpen: () => void
```
LFDanLu, line 36: Since this state acts as the replacement state for Popover/Tray, it needs the same type parity as `useOverlayTriggerState` and thus has toggle and setOpen even though those never get used by Tray/Popover. Don't really wanna call useOverlayTriggerState in here if I can avoid it, so I provide a no-op for setOpen

packages/@react-stately/menu/src/useSubMenuTriggerState.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// TODO: ideally onClose would be passed in from the Trigger level rather than the menu level so that we could call it
+// TODO descriptions and naming
+// TODO: I'm passing in the parent menu's TreeState now so that we can set expanded keys so a menu can know if it has any sub menus open.
+// The alternative to using expandedKeys would involve making each menu track which "level" it belongs to and have it the look up that index in
+// the MenuTreeState's expandedKeyStack and see if its collection has a matching key. This feels arguably more correct but would then involve
+// having useMenu accept MenuTreeState instead which might not be the worst thing tbh. Open to discussion
+export function UNSTABLE_useSubMenuTriggerState<T>(props: SubMenuTriggerProps, state: MenuTreeState & TreeState<T>): SubMenuTriggerState  {
```
LFDanLu, line 45: Open to opinions here, `useMenu`/Menu need to know if the menu has any expandedKeys for things like FocusScope containment and `aria-hidden`

packages/@react-aria/menu/src/useMenu.ts
```diff
@@ -30,7 +30,10 @@ export interface AriaMenuOptions<T> extends Omit<AriaMenuProps<T>, 'children'> {
    * An optional keyboard delegate implementation for type to select,
    * to override the default.
    */
-  keyboardDelegate?: KeyboardDelegate
+  keyboardDelegate?: KeyboardDelegate,
+  // TODO: Should we have an accompanying onKeyUp even though we don't use it and then just extend KeyboardEvents?
```
LFDanLu, line 34: TODO: extend KeyboardEvents and call useKeyboard

packages/@react-aria/menu/src/useMenu.ts
LFDanLu: Don't add UNSTABLE_

packages/@react-aria/menu/src/useMenuItem.ts
LFDanLu: Extend pressProps/hoverEvents/keyboardEvents

packages/@react-aria/menu/src/useMenuItem.ts
LFDanLu: don't make conditional, put any required conditions into useSubMenuTrigger

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -288,7 +252,9 @@ export function useMenuItem<T>(props: AriaMenuItemProps, state: TreeState<T>, re
   return {
     menuItemProps: {
       ...ariaProps,
-      ...mergeProps(domProps, itemProps, pressProps, hoverProps, keyboardProps)
+      // TODO: Similar question as the aria attributes above, do we just expect the user to spread onKeyDown, onBlur, id and aria attributes directly?
+      // This current approach of handling the passthroughs feels more in line with how useMenuTrigger passes stuff to useButton though
+      ...mergeProps(domProps, itemProps, pressProps, hoverProps, keyboardProps, {onKeyDown, onBlur})
```
LFDanLu, line 257: pass through

packages/@react-aria/menu/src/useSubMenuTrigger.ts
```diff
@@ -0,0 +1,244 @@
...
+      closeTarget.current?.focus();
+      closeTarget.current = null;
+    } else if (!parentMenuRef.current.contains(document.activeElement)) {
+      // need to return focus to the trigger because hitting Esc causes focus to go to the subdialog, which is then unmounted
+      // this leads to blur never being fired nor focus on the body
+      parentMenuRef.current.focus();
+    }
+
+    cancelOpenTimeout();
+  };
```
LFDanLu, line 220: Try to get rid of this onExit by:
1. Add wrapping FocusScope w/ restoreFocus around ContextualHelpTrigger's contents. Might make the `else if` here unneeded
2. Get rid of the FocusScopes around each SubMenu, only have it for root menu. Then have Tab NOT move focus between the ContextualHelpTrigger and the parent trigger. Esc will suffice for now. For submenus, tab no longer needs to move between the two (maybe have it close the menus entirely).

LFDanLu: Might mean we can get rid of closeTarget ref

packages/@react-aria/menu/src/useSubMenuTrigger.ts
```diff
@@ -0,0 +1,244 @@
...
+    onExit: () => void,
+    disableFocusManagement: boolean,
+    shouldCloseOnInteractOutside: (element: Element) => boolean
+  }
+}
+
+// TODO description
+// TODO: debatable if we should have a useSubMenu hook for the submenu keyboard handlers. Feels better to have it here so we don't need to ferry around
+// things like parentMenuRef and the open timeout canceling since we centralized them here
+export function UNSTABLE_useSubMenuTrigger<T>(props: AriaSubMenuTriggerProps, state: SubMenuTriggerState, ref: RefObject<FocusableElement>): SubMenuTriggerAria<T> {
```
reidbarber, line 54: API nit: I think we should use `Submenu` and not `SubMenu` based on usage across the web.

packages/@react-spectrum/menu/package.json
```diff
@@ -39,6 +39,7 @@
     "@react-aria/focus": "^3.14.1",
     "@react-aria/i18n": "^3.8.2",
     "@react-aria/interactions": "^3.18.0",
+    "@react-aria/label": "^3.7.0",
```
LFDanLu, line 42: todo: remove

packages/@react-stately/menu/src/useMenuState.ts
LFDanLu: TODO: move this stuff into useMenuTriggerState and add UNSTABLE to the return stuff

packages/@react-spectrum/menu/src/MenuTrigger.tsx
LFDanLu: TODO: refactor after getting rid useMenuState.

packages/@react-stately/menu/src/useSubMenuTriggerState.ts
```diff
@@ -0,0 +1,84 @@
...
+  }, [close, open, isOpen]);
+
+  return useMemo(() => ({
+    focusStrategy,
+    isOpen,
+    open,
+    close,
+    closeAll,
+    setOpen: () => {},
+    toggle
```
LFDanLu, line 82: get rid of updating `state.expandedKeys` in favor of passing down level and having each menu use said level to look up if it has a matching triggerKey in the useMenuState expandedKeys stack to determine if it has a open sub menu

packages/@react-stately/menu/src/useSubMenuTriggerState.ts
LFDanLu: Keep as is for now, refactor if we make Popover accept isOpen/onOpenChange instead of the entire state object of OverlayTriggerState

packages/@react-aria/menu/src/useMenu.ts
LFDanLu: As discussed in the meeting, we will be getting rid of this line since it is a screenreader edge case. Alternative may be to use the level from SubMenuTriggerState to look up if a menu has a expanded submenu instead of using `expandedKeys` to track

packages/@react-aria/menu/src/useSubMenuTrigger.ts
LFDanLu: Will try to get rid of onExit via https://github.com/adobe/react-spectrum/pull/4976/files#r1336455224

packages/@react-aria/menu/src/useSubMenuTrigger.ts
LFDanLu: https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=39885795

packages/@react-aria/overlays/src/usePopover.ts
LFDanLu: talked about this, we are ok with exposing this, just need to be cognizant that this also affects RAC popover

packages/@react-aria/menu/src/useSubMenuTrigger.ts
LFDanLu: current attempt at getting rid of onExit is in https://github.com/adobe/react-spectrum/tree/submenu_tab_behavior. Running into a lot of issues with focus behavior getting highjacked by FocusScope or onSelectableCollection tab handling. If we want to have tab move to the preceeding/next focusable element adjacent to the trigger button, ContextualHelpTrigger would need to handle the Tab key for the Dialog as well...

packages/@react-aria/menu/src/useSubMenuTrigger.ts
LFDanLu: on second thought, the current api matches v2's naming so in the interest of ease of upgrading I'm going to keep as is.

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -212,33 +207,24 @@ export function useMenuItem<T>(props: AriaMenuItemProps, state: TreeState<T>, re
 
   let {pressProps, isPressed} = usePress({
     onPressStart,
+    onPress,
     onPressUp,
-    isDisabled: isDisabled || (isTrigger && state.expandedKeys.has(key))
+    onPressChange,
+    onPressEnd,
+    // TODO: verify we actually don't need isDisabled || (isTrigger && state.expandedKeys.has(key)
```
snowystinger, line 214: what do we need to check here?

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -262,33 +248,25 @@ export function useMenuItem<T>(props: AriaMenuItemProps, state: TreeState<T>, re
...
 
+  let {focusProps} = useFocus({onBlur, onFocus, onFocusChange});
   let domProps = filterDOMProps(item.props, {isLink: !!item?.props?.href});
   delete domProps.id;
+
   return {
     menuItemProps: {
       ...ariaProps,
-      ...mergeProps(domProps, itemProps, pressProps, hoverProps, keyboardProps)
+      // TODO: Key down from useSubmenuTrigger's subMenuTriggerProps not incorporated into useKeyboard here since we
```
snowystinger, line 266: still a todo?

packages/@react-aria/menu/src/useSubmenuTrigger.ts
```diff
@@ -0,0 +1,260 @@
...
+    }
+  };
+
+  let onHoverChange = (isHovered) => {
+    if (!isDisabled) {
+      if (isHovered && !state.isOpen) {
+        if (!openTimeout.current) {
+          openTimeout.current = setTimeout(() => {
+            onSubmenuOpen();
+          }, 200);
```
devongovett, line 188: Do we need the timeout anymore with the safe triangle?

packages/@react-aria/menu/src/useSubmenuTrigger.ts
```diff
@@ -0,0 +1,260 @@
...
+      closeTarget.current?.focus();
+      closeTarget.current = null;
+    } else if (!parentMenuRef.current.contains(document.activeElement)) {
+      // need to return focus to the trigger because hitting Esc causes focus to go to the subdialog, which is then unmounted
+      // this leads to blur never being fired nor focus on the body
+      parentMenuRef.current.focus();
+    }
+
+    cancelOpenTimeout();
+  };
```
devongovett, line 235: I'd like to understand better why this is needed. Shouldn't FocusScope handle restoring focus back to the trigger?

packages/@adobe/spectrum-css-temp/components/menu/index.css
```diff
@@ -159,6 +159,13 @@ governing permissions and limitations under the License.
   line-height: var(--spectrum-text-body-line-height);
 }
 
+.spectrum-Menu-section--noHeading {
+  &.spectrum-Menu-section--firstSection,
```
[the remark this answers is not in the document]
snowystinger: I believe i have implemented what this intended over here https://github.com/adobe/react-spectrum/pull/5276

packages/@adobe/spectrum-css-temp/components/menu/skin.css
```diff
@@ -73,6 +77,16 @@ governing permissions and limitations under the License.
 .spectrum-Menu-description {
   color: var(--spectrum-global-color-gray-700);
 }
+
+.spectrum-Submenu-popover {
+  filter: unset;
```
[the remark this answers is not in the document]
snowystinger: pulled out over here https://github.com/adobe/react-spectrum/pull/5278

packages/@react-aria/menu/src/useMenu.ts
```diff
@@ -72,11 +75,31 @@ export function useMenu<T>(props: AriaMenuOptions<T>, state: TreeState<T>, ref:
     onAction: props.onAction
   });
 
+  let {keyboardProps} = useKeyboard({
+    onKeyDown(e) {
+      // TODO: Let Tab propagate so FocusScope handle it. Revisit if we decide to close all menus
```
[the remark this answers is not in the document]
LFDanLu: kinda? More like a note to self marking a place in the code to revisit if we decide to change how Tab works when within a submenu. I'll edit it for clarity

packages/@react-aria/menu/src/useMenu.ts
```diff
@@ -72,11 +75,31 @@ export function useMenu<T>(props: AriaMenuOptions<T>, state: TreeState<T>, ref:
...
+      }
+
+      onKeyDown && onKeyDown(e);
+    },
+    onKeyUp(e) {
+      // Need to continue keyup propagation so MenuTrigger button's press state is properly reset, otherwise it gets stuck in pressed state
+      e.continuePropagation();
+      onKeyUp && onKeyUp(e);
+    }
+  });
```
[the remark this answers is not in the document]
LFDanLu: `useKeyboard` isn't strictly required persay, was just something we decided to add to ensure most other keydown events had `stopPropagation` by default. Happy to get rid of it and handle the stopPropagation in useSubMenuTrigger's `submenuKeyDown` so that we only stop propagation for keyboard events that we specifically handle

packages/@react-aria/menu/src/useMenu.ts
```diff
@@ -72,11 +75,31 @@ export function useMenu<T>(props: AriaMenuOptions<T>, state: TreeState<T>, ref:
...
+
   return {
-    menuProps: mergeProps(domProps, {
+    menuProps: mergeProps(domProps, keyboardProps, {
       role: 'menu',
+      // TODO: we no longer update treestate's expandedKeys so we can't do the below anymore. This affects Safari VO specifically, the VO cursor doesn't move into the newly opened
```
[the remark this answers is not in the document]
LFDanLu: yep this is still a TODO mainly for the team to discuss how bad the behavior is and how we'd like to mitigate it. Previously we were setting `aria-hidden` on a menu/submenu once it had an `expandedKey` to force the VO cursor to move to the newly opened submenu. This currently would require us to update each menu's treeState's `expandedKeys` which was something we wanted to move away from in favor of tracking a list of expanded keys in [useMenuTriggerState](https://github.com/adobe/react-spectrum/blob/5da908c1cbbfd868af749705a3cc4c81d6b7b58c/packages/%40react-stately/menu/src/useMenuTriggerState.ts#L51). Note that `state` here is the menu's TreeState rather than the state returned by `useMenuTriggerState`

For now I've gotten rid of the `aria-hidden` here since we deemed it a bit of a edge case in the meeting we had on 9/25, but I could bring it back without needing to use `state.expandedKeys.size` if I add another prop to useMenu for `level`/`hasOpenSubMenu` or if I pass the `subMenuTriggerState` to `useMenu` so I can look up `level` instead

packages/@react-aria/menu/src/useMenuItem.ts
LFDanLu: the `isDisabled` line here used to be `isDisabled: isDisabled || (isTrigger && state.expandedKeys.has(key))` but I'm fairly certain that we can get away with just `isDisabled` now. I'll remove this todo

packages/@react-aria/menu/src/useSubmenuTrigger.ts
LFDanLu: It still serves as a warmup timer so a submenu doesn't "flash" open when a user moves their pointer through the menu items quickly.

packages/@react-aria/menu/src/useSubmenuTrigger.ts
LFDanLu: So this is handling the following mouse hover case:
Case 1:
1. User has 3 levels of submenus open
2. User moves their mouse out of the last menu and in such a way that they are hovering a non-trigger menu item in the root menu.
3. We expect that all menus except the root menu would close and that focus would remain on the hovered menu item.

If we get rid of `onExit` and set `disableFocusManage=false`, then FocusScope will indeed handle restoring focus back to the trigger, breaking `Case 1` since focus would be forcibly moved to the original submenu menu trigger item.  If we only get rid of `onExit`, then `Case 1` would lose focus to the body instead.

Like I mention in the comment above, the ideal logic would be that focus should always move to the menu item that the user hovers and that FocusScope wouldn't then shift focus anywhere as menus are closed/opened as a result of the hover interaction the user performed. This `onExit` approach was the best I could come up with since it happens after menus close and after FocusScope does its focus shifting.

packages/@react-spectrum/menu/src/context.ts
```diff
@@ -30,26 +33,32 @@ export function useMenuContext(): MenuContextValue {
   return useContext(MenuContext);
 }
 
-export interface MenuDialogContextValue {
+export interface SubmenuTriggerContextValue extends DOMProps, Pick<PressEvents, 'onPressStart' | 'onPress'>, Pick<HoverEvents, 'onHoverChange'> {
```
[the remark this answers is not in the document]
LFDanLu: that works, will change. The `Pick` syntax is just to be careful about what I'm actually providing to the context here, don't want to blanket accept all PressEvents/HoverEvents if I'm not actually providing those ever

packages/@react-spectrum/menu/test/MenuTrigger.test.js
```diff
@@ -250,14 +250,14 @@ describe('MenuTrigger', function () {
       Name             | Component      | props
       ${'MenuTrigger'} | ${MenuTrigger} | ${{}}
     `('$Name autofocuses the selected item on menu open', function ({Component, props}) {
-      let tree = renderComponent(Component, props, {selectedKeys: ['Bar']});
+      let tree = renderComponent(Component, props, {selectedKeys: ['Bar'], selectionMode: 'single'});
```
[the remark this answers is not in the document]
LFDanLu: https://github.com/adobe/react-spectrum/commit/4e85079d656cec366056c2ff06f7731df07def44 for context. Essentially opening the first submenu in [that old story](https://reactspectrum.blob.core.windows.net/reactspectrum/82c23dbc122a23d05f2ba3e3968d45d36d82fc2b/storybook/index.html?path=/story/menutrigger-submenu--sub-menu-selection&providerSwitcher-express=false) via keyboard was autofocusing `Lvl 2 Item 3` since it is the only key in the selectedKey array, but it felt weird since it isn't actually a selectable key since it is a submenu trigger menu item. That led to the change in [useSelectableCollection](https://github.com/adobe/react-spectrum/commit/4e85079d656cec366056c2ff06f7731df07def44#diff-e356ae602508922357fd9cd1aa7896f74e24f87bcb2ee86a7bb4a486ea87a2f2R384-R388) where focus only get will get moved to an actually valid selected key, thus making it so I needed to change this test so that it actually supports selection or otherwise focus wouldn't get moved to "Bar"

packages/@react-stately/menu/src/useMenuTriggerState.ts
```diff
@@ -23,16 +23,52 @@ export interface MenuTriggerState extends OverlayTriggerState {
...
 export function useMenuTriggerState(props: MenuTriggerProps): MenuTriggerState  {
   let overlayTriggerState = useOverlayTriggerState(props);
   let [focusStrategy, setFocusStrategy] = useState<FocusStrategy>(null);
+  let [expandedKeysStack, setExpandedKeysStack] = useState<string[]>([]);
+  let closeAll = () => {
+    setExpandedKeysStack([]);
+    overlayTriggerState.close();
+  };
+
+  // TODO: if level > the length of the expandedKeyStack then this doesn't quite work.
```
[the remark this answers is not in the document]
LFDanLu: Theoretically it should never happen since it should never skip levels when opening a submenu (e.g. if I have 3 levels of submenus open, I can't then call "open level 7's sub menu"). I'll add a check here that guards against that case.

packages/@react-stately/menu/src/useSubmenuTriggerState.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Manages state for a submenu trigger. Tracks whether the submenu is currently open, the level of the submenu, and
+ * controls which item will receive focus when it opens.
+ */
+export function UNSTABLE_useSubmenuTriggerState(props: SubmenuTriggerProps, state: MenuTriggerState): SubmenuTriggerState  {
+  let {triggerKey} = props;
+  let {UNSTABLE_expandedKeysStack, UNSTABLE_openSubmenu, UNSTABLE_closeSubmenu, close: closeAll} = state;
+  let [level] = useState(UNSTABLE_expandedKeysStack?.length + 1);
+  let isOpen = useMemo(() => UNSTABLE_expandedKeysStack[level - 1] === triggerKey, [UNSTABLE_expandedKeysStack, triggerKey, level]);
```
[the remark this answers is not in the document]
LFDanLu: I made the level start from 1 since technically the root menu is level 0, so a submenu would be level 1.

packages/@react-aria/menu/src/index.ts
```diff
@@ -14,9 +14,12 @@ export {useMenuTrigger} from './useMenuTrigger';
 export {useMenu} from './useMenu';
 export {useMenuItem} from './useMenuItem';
 export {useMenuSection} from './useMenuSection';
+export {useSafelyMouseToSubmenu} from './useSafelyMouseToSubmenu';
```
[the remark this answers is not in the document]
reidbarber: Updated this to be built into `useSubmenuTrigger`.

packages/@react-aria/menu/src/useSubmenuTrigger.ts
LFDanLu: resolved after offline discussion. Made menuItems always have "tabIndex=-1" and got rid of FocusScope containment at every menu level

packages/@react-stately/menu/src/useSubmenuTriggerState.ts
```diff
@@ -0,0 +1,84 @@
...
+  /** Closes the submenu. */
+  close: () => void,
+  /** Closes all menus and submenus in the menu tree. */
+  closeAll: () => void,
+  /** The level of the submenu. */
+  level: number,
+  /** Toggles the submenu. */
+  toggle: (focusStrategy?: FocusStrategy | null) => void,
+  /** @private */
+  setOpen: () => void
```
LFDanLu, line 39: Relavant comment below: 
```
    // TODO: Placeholders that aren't used but give us parity with OverlayTriggerState so we can use this in Popover. Refactor if we update Popover via
    // https://github.com/adobe/react-spectrum/pull/4976#discussion_r1336472863
```
Open to opinions/feedback

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -85,7 +83,13 @@ export interface AriaMenuItemProps {
   onAction?: (key: Key) => void,
 
   /** What kind of popup the item opens. */
-  'aria-haspopup'?: 'menu' | 'dialog'
+  'aria-haspopup'?: 'menu' | 'dialog',
+
+  /** Indicates whether the menu item's popup element is expanded or collapsed. */
+  'aria-expanded'?: boolean | 'true' | 'false',
```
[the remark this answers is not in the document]
LFDanLu: Looks like the docs have a hard time expanding React's type for `aria-expanded` (doesn't show Booleanish in the prop table). I could make our own booleanish type, will dig further in case your react types PR helps the docs process this stuff

packages/@react-spectrum/menu/docs/Menu.mdx
```diff
@@ -366,6 +367,86 @@ import {Content} from '@react-spectrum/view';
...
+    return <Item key={item.name}>{item.name}</Item>;
+  }
+};
+
+<MenuTrigger>
+  <ActionButton>Actions</ActionButton>
+  <Menu items={items}>
+    {(item) => renderSubmenu(item)}
+  </Menu>
+</MenuTrigger>
```
snowystinger, line 447: You don't have to use an anonymous function for this, so you could shorten this example like so
```suggestion
<MenuTrigger>
  <ActionButton>Actions</ActionButton>
  <Menu items={items}>
    {function renderSubmenu(item) {
      if (item.children) {
        return (
          <SubmenuTrigger>
            <Item key={item.name}>{item.name}</Item>
            <Menu items={item.children}>
              {(item) => renderSubmenu(item)}
            </Menu>
          </SubmenuTrigger>
        );
      } else {
        return <Item key={item.name}>{item.name}</Item>;
      }
    }}
  </Menu>
</MenuTrigger>
```

packages/@react-spectrum/menu/docs/Menu.mdx
```diff
@@ -379,6 +460,13 @@ import {Content} from '@react-spectrum/view';
   }
 }} links={contextualHelpTriggerTypes.links} />
 
+### SubmenuTrigger Props
+<PropTable component={{
```
snowystinger, line 464: this is unusual, what was happening to the usual way of doing this?

snowystinger (commented):
started review, will continue tomorrow

packages/@adobe/spectrum-css-temp/components/menu/index.css
```diff
@@ -231,3 +232,33 @@ governing permissions and limitations under the License.
   /* keyboard shortcuts are always ASCII, so use base font */
   font-family: var(--spectrum-font-family-base);
 }
+
+.spectrum-Menu-chevron {
+  margin-inline-end: calc(var(--spectrum-global-dimension-size-50) * -1);
+  grid-area: chevron;
+  justify-self: end;
+  align-self: center;
```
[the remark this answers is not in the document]
LFDanLu: good point, it does look better like that when there is a keyboard shortcut, I'll change it
image
vs 
image

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -195,6 +188,8 @@ export function useMenuItem<T>(props: AriaMenuItemProps, state: TreeState<T>, re
         onClose();
       }
     }
+
+    pressUpProp && pressStartProp(e);
```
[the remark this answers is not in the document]
LFDanLu: ah good catch

packages/@react-spectrum/menu/docs/Menu.mdx
snowystinger: ah, should we split the type? i think i wound up doing that for contextual help trigger?

LFDanLu: thats the thing, I did split the types: https://github.com/adobe/react-spectrum/blob/98869934e42833a6c0dc626ab23482a6d401ef03/packages/%40react-spectrum/menu/src/SubmenuTrigger.tsx#L24-L34 
https://github.com/adobe/react-spectrum/blob/98869934e42833a6c0dc626ab23482a6d401ef03/packages/%40react-spectrum/menu/src/SubmenuTrigger.tsx#L152

so I was expecting 
```
<PropTable component={docs.exports.SubmenuTrigger} links={docs.links} />
```
to have the proper props table, but it still included the `targetKey` prop

snowystinger: What about
```
<PropTable component={docs.exports.SpectrumSubmenuTriggerProps} links={docs.links} />
```

packages/@react-stately/menu/src/useSubmenuTriggerState.ts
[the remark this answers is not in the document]
snowystinger: i still think it'd simplify the logic, off by one is just so common and we don't need to be

snowystinger (commented):
Screenshot 2023-11-03 at 2 50 40 PM

Just noticed some of the border corners look a little funny at normal zoom ^ looks double?

will continue again monday

packages/@react-spectrum/menu/src/ContextualHelpTrigger.tsx
```diff
@@ -58,62 +64,96 @@ function ContextualHelpTrigger(props: InternalMenuDialogTriggerProps): ReactElem
...
+    submenuTriggerState.close();
+    if (parentMenuRef.current && !parentMenuRef.current.contains(document.activeElement)) {
+      // Delay for the parent menu in the tray to no longer be display: none so focus can properly be moved to it
+      requestAnimationFrame(() => parentMenuRef.current.focus());
+    }
+  };
+
+  if (isMobile) {
+    delete submenuTriggerProps.onBlur;
+    delete submenuTriggerProps.onHoverChange;
```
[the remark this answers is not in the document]
LFDanLu: yep, we don't want to open ContextualHelpTrigger/SubMenuTrigger on hover in a tray experience. This would only happen if you have a device with mouse support + a very small screen so pretty edge casey but can be reproduced via Chrome dev tools + device emulation + desktop mode turned on.

packages/@react-spectrum/menu/src/ContextualHelpTrigger.tsx
```diff
@@ -58,62 +64,96 @@ function ContextualHelpTrigger(props: InternalMenuDialogTriggerProps): ReactElem
...
+      );
+
+      overlay = ReactDOM.createPortal(tray, trayContainerRef.current);
+    }
+  } else {
+    overlay = (
+      <Popover
+        UNSAFE_style={{clipPath: 'unset', overflow: 'visible', filter: 'unset', borderWidth: '0px'}}
+        {...popoverProps}
+        {...overlayProps}
```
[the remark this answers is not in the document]
LFDanLu: can def merge, good point

packages/@react-spectrum/menu/src/ContextualHelpTrigger.tsx
```diff
@@ -58,62 +64,96 @@ function ContextualHelpTrigger(props: InternalMenuDialogTriggerProps): ReactElem
...
       }
     }
   };
+
+  let overlay;
+  let tray;
+  let onBackButtonPress = () => {
+    submenuTriggerState.close();
+    if (parentMenuRef.current && !parentMenuRef.current.contains(document.activeElement)) {
+      // Delay for the parent menu in the tray to no longer be display: none so focus can properly be moved to it
```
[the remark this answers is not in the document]
LFDanLu: I think I tried doing this before but I forget the details so lemme try again.

packages/@react-spectrum/menu/src/Menu.tsx
```diff
@@ -78,13 +104,59 @@ function Menu<T extends object>(props: SpectrumMenuProps<T>, ref: DOMRef<HTMLDiv
               return menuItem;
             })}
           </div>
-        </div>
-        <div ref={scopedRef} />
+        </TrayHeaderWrapper>
+        {/* Make the portal container for submenus wide enough so that the submenu items can render as wide as they need to be */}
+        <div ref={popoverContainerRef} style={{width: '100vw', position: 'absolute', top: -5, ...leftOffset}} />
```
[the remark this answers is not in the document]
LFDanLu: This div extends the whole width of the screen so it doesn't matter if we are RTL or not. The goal is always to start the div at the farest `left` of the screen and extend to the opposite end. We extend the whole screen since menu flipping means we may need to position to either side of the button regardless of LTR or RTL

packages/@react-stately/menu/src/useSubmenuTriggerState.ts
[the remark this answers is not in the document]
LFDanLu: I guess that is true, I'll try it out and maybe rename level to submenu level to make it clear it doesn't follow the same level value as a tree collection would calculate

LFDanLu:
@snowystinger looks like Popover has a border width applied for the submenus but not the original menu, still digging

packages/@react-spectrum/overlays/src/Popover.tsx
```diff
@@ -35,7 +35,8 @@ interface PopoverProps extends Omit<AriaPopoverProps, 'popoverRef' | 'maxHeight'
   onExit?: () => void,
   container?: HTMLElement,
   disableFocusManagement?: boolean,
-  enableBothDismissButtons?: boolean
+  enableBothDismissButtons?: boolean,
+  onDismiss?: () => void
```
LFDanLu, line 39: Another change to consider, required to make sure focus is restored to the parent menu when a submenu popover is closed via the visually hidden dismiss button since we have disabled our overlay FocusScope management. At the moment this change applies this only to RSP Popover but we should consider adding it to aria Popover. Only thing that gave me pause was that the hidden dismiss buttons aren't really a part of the hooks persay (no concept of dismissButtonProps in usePopover since they are up to the user's discretion to add) but they are provided as a part of RAC Popover.

If we decide that we want to include it in `usePopover` I'll add that when I finish the aria example for submenu in https://github.com/adobe/react-spectrum/pull/5415/files

packages/@react-spectrum/menu/src/ContextualHelpTrigger.tsx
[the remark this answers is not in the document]
devongovett: wonder if we should move this into the hook at some point so others can make a tray version too

packages/@react-stately/selection/src/SelectionManager.ts
```diff
@@ -480,7 +480,7 @@ export class SelectionManager implements MultipleSelectionManager {
     }
 
     let item = this.collection.getItem(key);
-    if (!item || (item.type === 'cell' && !this.allowsCellSelection)) {
+    if (!item || (item.type === 'cell' && !this.allowsCellSelection) || (item.type === 'item' && item?.props?.isTrigger)) {
```
devongovett, line 483: What is this for? Would be good to avoid menu-specific code in here

devongovett (commented):
great work overall! many of these can be addressed later if you want.

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -262,33 +247,26 @@ export function useMenuItem<T>(props: AriaMenuItemProps, state: TreeState<T>, re
...
+
   return {
     menuItemProps: {
       ...ariaProps,
-      ...mergeProps(domProps, itemProps, pressProps, hoverProps, keyboardProps)
+      // Key down from useSubmenuTrigger's submenuTriggerProps is not incorporated into useKeyboard here since we
+      // cannot sufficiently distinguish when propagation should be continued or stopped in this section of code.
+      // It should only stop propagation if handling the ArrowRight case when the submenutrigger's menu is open and focus is on the trigger,
+      ...mergeProps(domProps, itemProps, pressProps, hoverProps, keyboardProps, focusProps, {onKeyDown, onKeyUp}),
+      tabIndex: -1
```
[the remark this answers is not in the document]
LFDanLu: ah good point, I'll go ahead and change this

packages/@react-aria/menu/src/useSubmenuTrigger.ts
```diff
@@ -0,0 +1,232 @@
...
+  submenuLevel: number
+}
+
+export interface SubmenuTriggerAria<T> {
+  /** Props for the submenu trigger menu item. */
+  submenuTriggerProps: SubmenuTriggerProps,
+  /** Props for the submenu controlled by the submenu trigger menu item. */
+  submenuProps: SubmenuProps<T>,
+  /** Props for the submenu's popover container. */
+  popoverProps: Pick<AriaPopoverProps, 'isNonModal'>,
```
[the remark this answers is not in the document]
LFDanLu: I only wanted the `isNonModal` prop definition from `AriaPopoverProps` since it has a couple required props that we don't provide from this hook. I could instead just redefine `isNonModal` here if we wanna get rid of the `Pick`, but figured it was the best way to maintain prop definition parity

packages/@react-aria/menu/src/useSubmenuTrigger.ts
```diff
@@ -0,0 +1,232 @@
...
+  /** Props for the submenu's popover overlay container. */
+  overlayProps: {
```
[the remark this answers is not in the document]
LFDanLu: true, looking back at this perhaps I made this a bit too complicated by splitting them up when they already get merged when passed to Popover right now, I'll merge em

packages/@react-aria/menu/src/useSubmenuTrigger.ts
```diff
@@ -0,0 +1,232 @@
...
+import {useLocale} from '@react-aria/i18n';
+import {useSafelyMouseToSubmenu} from './useSafelyMouseToSubmenu';
+
+export interface AriaSubmenuTriggerProps {
+  /** An object representing the submenu trigger menu item. Contains all the relevant information that makes up the menu item. */
+  node: RSNode<unknown>,
+  /** Whether the submenu trigger is disabled. */
+  isDisabled?: boolean,
+  /** Type of the submenu being rendered. */
+  submenuType?: 'dialog' | 'menu',
```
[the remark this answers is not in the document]
LFDanLu: def kinda weird, open to suggestions haha. Needed here so that contextual help dialogs don't get the same behavior as an actual submenu. Would we consider that Spectrum specific (didn't really feel like a uncommon use case though)?

 `suboverlayType` doesn't feel good

majornista:
Submenu Examples with many items no longer seem to scroll items into view on keyboard focus: [Submenu > dynamic submenu with many items](https://reactspectrum.blob.core.windows.net/reactspectrum/7f988e3230af75ede795251e4d499483b4026b40/storybook/index.html?path=/story/menutrigger-submenu--submenu-many-items)

packages/@react-stately/selection/src/SelectionManager.ts
LFDanLu: I needed a way to make SubMenuTriggers non-selectable but NOT disabled since they should be keyboard navigable. Without this code, if a user specified the key of a submenu trigger menu item in their "selectedKeys" array, it would be considered selected from the SelectionManager's point of view

LFDanLu:
@majornista good catch, must be from one of the latest css changes ugh...

packages/@react-spectrum/menu/src/ContextualHelpTrigger.tsx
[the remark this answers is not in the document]
LFDanLu: happy to do so, but isn't our definition of mobile pretty Spectrum specific? Would we need to add a prop for users to then define what they consider 'isMobile'?

packages/@react-stately/selection/src/SelectionManager.ts
LFDanLu: What do you think about a different prop name for this? Maybe something more generic like `props?.selectionDisabled`?

devongovett: can that logic be in `useMenuItem`? seems like programmer error to mark a trigger as selected, but we could just not display it as selected at least

packages/@react-aria/menu/src/useMenuItem.ts
```diff
@@ -262,33 +247,26 @@ export function useMenuItem<T>(props: AriaMenuItemProps, state: TreeState<T>, re
...
   let domProps = filterDOMProps(item.props, {isLink: !!item?.props?.href});
   delete domProps.id;
+
   return {
     menuItemProps: {
       ...ariaProps,
-      ...mergeProps(domProps, itemProps, pressProps, hoverProps, keyboardProps)
+      // Key down from useSubmenuTrigger's submenuTriggerProps is not incorporated into useKeyboard here since we
+      // cannot sufficiently distinguish when propagation should be continued or stopped in this section of code.
+      // It should only stop propagation if handling the ArrowRight case when the submenutrigger's menu is open and focus is on the trigger,
```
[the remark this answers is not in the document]
LFDanLu: Revisited and fixed this, thanks for calling it out. Just had to wrap the continuePropagation that useMenuItem was calling by default in a isTrigger check.

packages/@react-aria/menu/src/useSubmenuTrigger.ts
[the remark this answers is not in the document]
devongovett: just `type`? or is that not descriptive enough you think? We have something similar in `useMenuTrigger`

packages/@react-stately/selection/src/SelectionManager.ts
LFDanLu: off the top of my head, I think we'd have to add a prop or something to useSelectableItem since it currently uses `manager.canSelectItem` as its check for `allowsSelection`. I'll dig

packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -268,7 +268,9 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
           // in the collection, so that the browser default behavior will apply starting from that element
           // rather than the currently focused one.
           if (e.shiftKey) {
-            ref.current.focus();
+            // Needs flushSync to make sure expandedKeys is updated before focus is moved out of the containing Menu's FocusScope. Without this
+            // the menu's contain isn't adjusted in time resulting in FocusScope contain keydown handlers not being cleaned up. Caught in test, doesn't happen in browser cuz lucky
+            flushSync(() => ref.current.focus());
```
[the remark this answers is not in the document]
LFDanLu: Ah we actually don't need this anymore since we let focus move to the preceeding/next element adjacent to the menu button rather than shifting focus between the submenutrigger and its submenu, I'll remove

packages/@react-stately/selection/src/SelectionManager.ts
LFDanLu: One alternative is that I don't use the item props (bar the `onFocus` to maintain focusedKey tracking when a user presses on a trigger in a submenu Tray)  from `useSelectableItem` in `useMenuItem` if it is a submenu trigger, but it still exposes the submenu trigger item as a valid selectable key from the SelectionManager's point of view which feels risky since we have various checks all over the place that use `manager.isSelected(key)`.

packages/@react-spectrum/menu/src/ContextualHelpTrigger.tsx
```diff
@@ -141,8 +148,8 @@ function ContextualHelpTrigger(props: InternalMenuDialogTriggerProps): ReactElem
         triggerRef={triggerRef}
         placement="end top"
         containerPadding={0}
-        crossOffset={-5}
-        offset={-10}
+        crossOffset={offset - 1}
```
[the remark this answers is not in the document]
LFDanLu: So I was actually off by 1 in the offset case (messed up the alignment in the XD ugh), they should be the same (border radius + 1). Will update

packages/@adobe/spectrum-css-temp/components/menu/skin.css
```diff
@@ -29,6 +29,8 @@ governing permissions and limitations under the License.
 .spectrum-Menu {
   --spectrum-heading-subtitle3-text-color: var(--spectrum-global-color-gray-900);
   background-color: var(--spectrum-selectlist-background-color);
+  /* Define this var so it can be read from JS */
+  --spectrum-submenu-offset-distance-x: var(--spectrum-global-dimension-size-50);
```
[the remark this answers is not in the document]
LFDanLu: Should be border radius + 1 (size-65) which we don't have. I'll update the naming, but I didn't want to do the `calc` in the CSS because getComputedSize + getPropertyValue will return the string rather than the calculated version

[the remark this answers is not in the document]
snowystinger: that's interesting. can you put in a comment with the calc so we know what it's derived from?

packages/@react-aria/menu/src/useSafelyMouseToSubmenu.ts
```diff
@@ -131,7 +132,7 @@ export function useSafelyMouseToSubmenu(options: SafelyMouseToSubmenuOptions) {
       if (isMovingTowardsSubmenu) {
         timeout.current = setTimeout(() => {
           reset();
-          setTimeout(() => {
+          autoCloseTimeout.current = setTimeout(() => {
```
snowystinger, line 135: if this timeout is already running, we should either cancel it, or cancel the new request
same with `timeout`

LFDanLu: So `timeout` is canceled here: https://github.com/adobe/react-spectrum/blob/af97396b952675c528c19506e568173246eb3342/packages/%40react-aria/menu/src/useSafelyMouseToSubmenu.ts#L80 on `pointermove` so it should clear the `autoCloseTimeout` too. I guess I could clear the autoCloseTimeout there as well in case the user moves the mouse after the `timeout` triggers but before the `autoCloseTimeout` triggers.

snowystinger: yep, too long of a function, i didn't notice. thanks, the additional clearing I think is correct

PR #4981, opened 2023-08-28
Bailout early if Virtualizer is bouncing between two rect sizes constantly

LFDanLu:
we have had customer issues where tableview has been crashing due to an infinite rerender but havent been able to reproduce it locally but this change seems to fix it. This PR is to start a discussion on how best we want to handle this, code comes from https://github.com/adobe/react-spectrum/pull/4846/files/858512a3cbdff8f07bae43c1d7fd5b0b1ba23ee7.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test virtualizer components and double check that you don't ever get React error 185 issue crash (maximum update depth)

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-aria/virtualizer/src/Virtualizer.tsx
+++ b/packages/@react-aria/virtualizer/src/Virtualizer.tsx
@@ -181,9 +181,14 @@ export function useVirtualizer<T extends object, V extends ReactNode, W>(props: 
   // Handle scrolling, and call onLoadMore when nearing the bottom.
   let isLoadingRef = useRef(isLoading);
   let prevProps = useRef(props);
-  let onVisibleRectChange = useCallback((rect: Rect) => {
-    state.setVisibleRect(rect);
+  let rectTracker = useRef({
+    // Track the last two visible rect sizes
+    lastTwoRects: [],
+    // Number of times the newest visible rect size has matched the last two sizes
+    sameSizeCounter: 0
+  }).current;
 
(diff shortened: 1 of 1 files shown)


packages/@react-aria/virtualizer/src/Virtualizer.tsx
```diff
@@ -181,17 +181,40 @@ export function useVirtualizer<T extends object, V extends ReactNode, W>(props:
...
         onLoadMore();
       }
     }
-  }, [onLoadMore, state]);
+
+    if (rectTracker.lastTwoRects.length > 1) {
+      if (rectTracker.lastTwoRects.some(oldRect => rect.equals(oldRect))) {
+        ++rectTracker.sameSizeCounter;
+        if (rectTracker.sameSizeCounter > 5) {
+          return;
```
reidbarber, line 204: Should it get reset before early returning? Just thinking what would happen if another real resize happens.

LFDanLu: If onVisibleRectChange is triggered by a real resize and returns a `rect` that matches one of the two old rects I don't think we wanna reset the counter since we can't really distinguish that from the failure case right? I figure a user causing a resize of the virtualizer will typically hit a rect that is different from the last two rects and thus it will reset the counter then

reidbarber (approved):

PR #4999, opened 2023-08-31
Reorganize React Aria Components docs

devongovett:
* Adds pseudo code example to the anatomy section of each component. This is meant to show where all of the possible parts of a component go, including sub-components and slots, without any content or props. It should help map the visual shown in the anatomy diagram to the JSX structure needed in code. Some anatomy diagrams were also updated to reflect differences in naming in RAC vs hooks.
* Add cards with illustrations for the "Concepts" links under Anatomy, including Collections, Selection, and @internationalized/date.
* Moves the Props and Styling sections to the end of the page, after the usage examples and before advanced customization. Feedback was that this required a lot of scrolling to get by in order to see the examples. As more of a reference section, it makes more sense at the end and this matches what we have done in the React Spectrum docs.
* Gets rid of the "Usage" heading, and up levels the sub-sections. This allows there to be more of a hierarchy. It is now much more similar to how the React Spectrum docs are organized.
* Group pages in the left side nav by type: Components, Hooks, Patterns, Interactions, and Utilities. This was already there under feature flag, but disabled for alpha. I think it's time to enable, to avoid so much scrolling.
* Give more padding to h3 and h4 headings in the docs. I was noticing that it felt a bit cramped.

**Note**: for review, the diff is mostly useless because so much got moved around. Might be easier to just look at the built docs.

## Open questions

- Is usage of components like `MyComboBox` which are defined on the page clear enough? You sort of have to map this back to the real API in your head, but if we repeated the whole thing for every example it might get really long. I undid this for some examples like the help text ones, because they discussed "slots" but showed props and this seemed weird.
- Drag and drop is currently the only thing under Patterns. This was meant as higher level than interactions like press and hover but lower than component hooks. But maybe weird to just have one? Is it too hidden? Should we merge into another?


snowystinger (approved):
I think I'd merge the current patterns into interactions but keep their sub heading/section

If we get more pattern hooks, then we can always split it out again now that we have search

I really like the new "makes use of these [x] concepts" sections

reidbarber (approved):
Nothing blocking from me, but here are some thoughts:
- Maybe this can be a followup before RC, but I think we can update lots of the wording on the "Getting Started" and "Why React Aria?" pages to be less oriented towards Hooks.
- I'll use this as another opportunity to vouch for removing the hero image :)
- We may want to remove references to virtualized scrolling for now.
- Don't forget to update the badge to beta
- I think the Drag and Drop section of the navigation menu could go under Interactions instead of Patterns.

devongovett:
Going to merge this one and do some followups in another PR.

devongovett:
> Maybe this can be a followup before RC, but I think we can update lots of the wording on the "Getting Started" and "Why React Aria?" pages to be less oriented towards Hooks.

Yeah I have this on the list for RC/GA.

> I'll use this as another opportunity to vouch for removing the hero image :)

Also want to make a new homepage for GA.

> We may want to remove references to virtualized scrolling for now.
> Don't forget to update the badge to beta
> I think the Drag and Drop section of the navigation menu could go under Interactions instead of Patterns.

Will do in followup PR.

Issue #4885, opened 2023-08-09
Unable to style the `Breadcrumbs` component with Tailwind.


polc:
Hi ! The `Breadcrumbs` component creates a `nav` element containing an `ol` element as its children. Unfortunately, we can't assign a class name to the `ol` element, even though it's quite important !

```
<Breadcrumbs className="flex items-center">
</Breadcrumbs>
```

I'm not sure how it's done in other components, is there any precedent ?

reidbarber:
We have a TODO for this here:

https://github.com/adobe/react-spectrum/blob/8e6cefdb5f13aa61eab9fcf5c91f2eaac1d807b5/packages/react-aria-components/src/Breadcrumbs.tsx#L41-L42

I think the concern was balancing having too many containers, but also allowing easy customization of every element. 

Specific to Tailwind, but for now you could use `[&> ol]:` on the parent.

polc:
Oh I was not aware that these selectors had been added to Tailwind. This looks good enough to me. Thank you!

PR #642, opened 2020-06-15
(RSP-1814) Update breadcrumb docs to reflect new api

LFDanLu:
Closes https://jira.corp.adobe.com/browse/RSP-1814

removes mention of using Heading in the breadcrumb and updates docs to reflect removal of maxVisibleItems

## ✅ Pull Request Checklist:

- [ ] Included link to corresponding [Issue](https://github.com/adobe-private/react-spectrum-v3/issues).
- [ ] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [ ] Filled out test instructions.
- [ ] Updated documentation (if it already exists for this component).
- [ ] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/TR/wai-aria-practices-1.1/)

PR #5014, opened 2023-09-01
Update RAC breadcrumbs API

devongovett:
* Remove default `<nav>` landmark. Not all breadcrumbs should be landmarks. Instead, I added an example in the docs explaining how to create a landmark. This also solved the problem of `Breadcrumbs` rendering two elements, making it hard to style. #4885
* Remove support for breadcrumbs as a heading. We removed this from React Spectrum many years ago (https://github.com/adobe/react-spectrum/pull/642), but never removed it from React Aria. The accessibility doesn't seem correct anyway (e.g. `aria-current` on a heading element?). I also removed the example from the hooks docs.
* Switch from using the `<Item>` component to a `<Breadcrumb>` component. Item had too many props and data attributes that didn't apply to breadcrumbs.

### DIFF
--- a/packages/react-aria-components/docs/Breadcrumbs.mdx
+++ b/packages/react-aria-components/docs/Breadcrumbs.mdx
@@ -337,7 +337,7 @@ The states, selectors, and render props for all components used in `Breadcrumbs`
 
 ### Breadcrumbs
 
-`Breadcrumbs` can be targed with the `.react-aria-Breadcrumbs` CSS selector, or by overriding with a custom `className`. It is rendered `<ol>` element representing the list of items.
+`Breadcrumbs` can be targed with the `.react-aria-Breadcrumbs` CSS selector, or by overriding with a custom `className`. It is rendered as an `<ol>` element representing the list of items.
 
 ### Breadcrumb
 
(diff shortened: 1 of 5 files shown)


packages/react-aria-components/docs/Breadcrumbs.mdx
```diff
@@ -334,19 +337,11 @@ The states, selectors, and render props for all components used in `Breadcrumbs`
...
-`Breadcrumbs` can be targed with the `.react-aria-Breadcrumbs` CSS selector, or by overriding with a custom `className`. It is rendered as a `<nav>` element, and contains an `<ol>` element representing the list of items. This additional element can be targed with the `ol` selector.
-
-```css render=false
-.react-aria-Breadcrumbs {
-  & ol {
-    /* ... */
-  }
-}
-```
+`Breadcrumbs` can be targed with the `.react-aria-Breadcrumbs` CSS selector, or by overriding with a custom `className`. It is rendered `<ol>` element representing the list of items.
```
LFDanLu, line 340:
```suggestion
`Breadcrumbs` can be targed with the `.react-aria-Breadcrumbs` CSS selector, or by overriding with a custom `className`. It is rendered as an `<ol>` element representing the list of items.
```

LFDanLu (approved):
LGTM, verified that only the landmark example was picked up as a landmark by screenreaders, api changes look sound to me.

snowystinger (approved):

PR #5018, opened 2023-09-02
Add data-selection-mode attribute

devongovett:
Exposes the `selectionMode` render prop as a data attribute. Mainly to fix the Menu selection examples which used to use `aria-checked` to reserve space for the checkmark even when not selected, but this broke after the move to data attributes.

### DIFF
--- a/packages/react-aria-components/docs/Table.mdx
+++ b/packages/react-aria-components/docs/Table.mdx
@@ -1311,22 +1311,6 @@ function Example() {
   box-sizing: border-box;
   outline: none;
 
-  .react-aria-Section:not(:first-child) {
-    margin-top: 12px;
-  }
-
-  .react-aria-Header {
-    font-size: 1.143rem;
-    font-weight: bold;
-    padding: 0 0.714rem;
-  }
-
-  [role=separator] {
-    height: 1px;
-    background: var(--separator-color);
-    margin: 2px 4px;
-  }
-
   .react-aria-Item {
     margin: 2px;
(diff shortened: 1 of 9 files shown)


packages/react-aria-components/src/Menu.tsx
```diff
@@ -169,14 +169,6 @@ function MenuSection<T>({section, className, style, ...otherProps}: MenuSectionP
   );
 }
 
-export interface MenuItemRenderProps extends ItemRenderProps {
-  /**
-   * Whether the item is currently selected.
-   * @selector [data-selected]
-   */
-   isSelected: boolean
-}
```
devongovett, line 178: Now unnecessary - ItemRenderProps already covers this.

packages/react-aria-components/docs/Table.mdx
```diff
@@ -1346,45 +1330,6 @@ function Example() {
...
-        left: 4px;
-        font-weight: 600;
-      }
-
-      &[role=menuitemradio][data-checked=true]::before {
-        content: '●';
-        content: '●' / '';
-        transform: scale(0.7)
-      }
-    }
```
devongovett, line 1387: unused in this example

snowystinger (approved):

reidbarber (changes requested):
Can you add this to the tailwind plugin as well?

reidbarber (approved):
Approving since my comment https://github.com/adobe/react-spectrum/pull/5018#pullrequestreview-1611214752 got handled in #5027

PR #5027, opened 2023-09-05
Add missing variants to Tailwind plugin

devongovett:
* Added `group-` and `peer-` modifiers for each variant. Tailwind doesn't do this by default: https://tailwindcss.com/docs/plugins#parent-and-sibling-states
* Audited for missing states
* Fixed some modifiers that conflict with builtin Tailwind ones

### DIFF
--- a/packages/tailwindcss-react-aria-components/src/index.js
+++ b/packages/tailwindcss-react-aria-components/src/index.js
@@ -5,37 +5,60 @@ const attributes = {
     ['hover', 'hovered'],
     ['focus', 'focused'],
     'focus-visible',
+    'focus-within',
     'pressed',
     'disabled',
     'drop-target',
     'dragging',
     'empty',
+    'allows-dragging',
     'allows-removing',
     'allows-sorting',
     'placeholder',
     'selected',
     'indeterminate',
-    'readonly',
+    ['read-only', 'readonly'],
     'required',
     'entering',
     'exiting',
     'open',
     'unavailable',
+    'outside-month',
+    'outside-visible-range',
+    'selection-start',
+    'selection-end',
     'current',
-    'invalid'
+    'invalid',
+    'resizing'
   ],
   enum: {
     placement: ['left', 'right', 'top', 'bottom'],
     type: ['literal', 'year', 'month', 'day'],
     layout: ['grid', 'stack'],
-    orientation: ['horizontal', 'vertical']
+    orientation: ['horizontal', 'vertical'],
+    'selection-mode': ['single', 'multiple'],
+    'resizable-direction': ['right', 'left', 'both'],
+    'sort-direction': ['ascending', 'descending']
   }
+};
+
+const shortNames = {
+  'selection-mode': 'selection',
+  'resizable-direction': 'resizable',
+  'sort-direction': 'sort'
 };
 
 // Variants we use that are already defined by Tailwind:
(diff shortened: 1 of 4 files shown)


packages/tailwindcss-react-aria-components/src/index.js
```diff
@@ -5,37 +5,60 @@ const attributes = {
...
     'dragging',
     'empty',
+    'allows-dragging',
     'allows-removing',
     'allows-sorting',
     'placeholder',
     'selected',
     'indeterminate',
-    'readonly',
+    ['read-only', 'readonly'],
```
devongovett, line 20: I thought it would be confusing to have both the builtin `read-only:` as well as the RAC `readonly:` so I merged them like we did for `hover` and `focus`

packages/tailwindcss-react-aria-components/src/index.js
```diff
@@ -5,37 +5,60 @@ const attributes = {
...
+    'resizable-direction': ['right', 'left', 'both'],
+    'sort-direction': ['ascending', 'descending']
   }
 };
 
+const shortNames = {
+  'selection-mode': 'selection',
+  'resizable-direction': 'resizable',
+  'sort-direction': 'sort'
+};
```
devongovett, line 49: Writing `resizeable-direction-right:` seemed a bit long, so this lets you write `resizable-right:`. Bad idea?

packages/tailwindcss-react-aria-components/src/index.js
```diff
@@ -5,37 +5,60 @@ const attributes = {
...
-const nativeVariants = ['indeterminate', 'required', 'invalid', 'empty', 'focus-visible', 'disabled'];
-const nativeVariantSelectors = new Map([...nativeVariants.map((variant) => [variant, `:${variant}`]), ['hovered', ':hover'], ['focused', ':focus'], ['open', '[open]']]);
+const nativeVariants = ['indeterminate', 'required', 'invalid', 'empty', 'focus-visible', 'focus-within', 'disabled'];
+const nativeVariantSelectors = new Map([
+  ...nativeVariants.map((variant) => [variant, `:${variant}`]),
+  ['hovered', ':hover'],
+  ['focused', ':focus'],
+  ['readonly', ':read-only'],
+  ['open', '[open]'],
+  ['placeholder', '::placeholder']
```
devongovett, line 60: The `placeholder:` modifier that is built into tailwind actually targets a pseudo element not a pseudo class. There is also `placeholder-shown:` which is a bit closer. Should we match that or keep `placeholder:`?

packages/tailwindcss-react-aria-components/src/index.js
reidbarber: All the shortened ones look good to me.

hosmelq:
Hello everyone, I want to express my appreciation for your hard work. I have been testing the plugin and I need to style a child element depending on its parent's state. Will you support group variants?

hosmelq:
Hello, everyone; I am grateful for your hard work on this fantastic library. I have been testing this plugin, and it works incredibly well. As a result, I was able to remove a lot of custom selectors for RAC. It's incredible that group variants will be shipped soon. Thank you once again for your fantastic work.

packages/tailwindcss-react-aria-components/src/index.js
reidbarber: Since we're matching tailwind's built-in variants for the others, my vote would be to use `placeholder-shown`

snowystinger (approved):

snowystinger (approved):

reidbarber (approved):

PR #5052, opened 2023-09-08
Refactor internal contexts to expose state publicly

### LINKED ISSUE #4986: Ability to properly clear value in useDateRangePickerState

We're implementing a custom `DateRangePicker` with all related hooks incl. `useDateRangePickerState`.

It seems to be hard to clear the state in a good way.

What we have done for our `DatePicker` is already a little hacky, as we repeatedly call `fieldState.clearSegment` for each segment.

It's not possible to do it like this when having to call `clearSegment` for both `startFieldState` and `endFieldState` though, as the `value`, used when `clearSegment` via `setDateTime` sets the `useDateRangePickerState` `value`, is stale.

Meaning that when running `endFieldState.clearSegment` it will set the `{ start: ... }` part back to the last selected value, because `endFieldState.clearSegment` is not aware that it has just been reset by `startFieldState.clearSegment`.

We could of course just do `dateRangePickerState.setValue({ start: null!, end: null! })` (though null is not accepted by TS).
But `clearSegment` is doing other useful stuff like saving last selected `value` to be used for a re-selection.

What we've found would work is if the `value` was saved as a `ref` (which [you've done before](https://github.com/adobe/react-spectrum/pull/2510/commits/ca696af45e1f4548e1ace06e5de6d4f297f46eeb?diff=split&w=0#diff-eb983dc43959711e7dea6f33bcbbe97aa0bab850f24441753cdbd885125f3b71R50-R51)) for the latest state to be used in `endFieldState.clearSegment` call:
```typescript
  // useDateRangePickerState.ts

  let value = controlledValue || placeholderValue;
  let valueRef = useRef(value);
  valueRef.current = value;

  let setValue = (value: DateRange) => {
    valueRef.current = value;
    ...
  }

  setDateTime(part, dateTime) {
    setValue({ ...valueRef.current, [part]: dateTime });
  },
```

But even better if a new method like `clearAllSegments` would be made available to avoid the repeated calls as well.

Are we overlooking something? Is there another way to achieve this clearing properly?
 

### 🤔 Expected Behavior?

Calling `startFieldState.clearSegment` then `endFieldState.clearSegment` should fully clear value of `useDateRangePickerState`.

### 😯 Current Behavior

Calling `startFieldState.clearSegment` then `endFieldState.clearSegment` sets start-part of `value` of `useDateRangePickerState` back to last selected value.

### 💁 Possible Solution

```typescript 
  // useDateRangePickerState.ts

  let value = controlledValue || placeholderValue;
  let valueRef = useRef(value);
  valueRef.current = value;

  let setValue = (value: DateRange) => {
    valueRef.current = value;
    ...
  }

  setDateTime(part, dateTime) {
    setValue({ ...valueRef.current, [part]: dateTime });
  },
```

Even better if a new method like `clearAllSegments` would be made available to avoid the repeated calls as well.

### 🔦 Context

This is how we work around it to both make clearSegment do the extra stuff it does (even though only taking effect for the end field) and also end up with a fully cleared state value:
```typescript
  const clearFields = () => {
    segmentTypes.forEach((segmentType) => {
      startFieldState.clearSegment(segmentType)
      endFieldState.clearSegment(segmentType)
    });
    
    dateRangePickerState.setValue({ start: null!, end: null! });
  };
```

### 💻 Examples

_No response_

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
Fixes #4986

Most React Aria Components compose other standalone components in their children to build larger patterns. However, some components are made up of more tightly coupled children. For example, Calendar includes children such as CalendarGrid and CalendarCell that cannot be used standalone, and must appear within a Calendar or RangeCalendar. Previously it was impossible to customize these without rewriting the all of the components from scratch.

This refactors our internal contexts so that state can be accessed using publicly exported state contexts. The state objects are already publicly documented in react-stately, so we are not exposing anything we didn't already in other places. The rest of the internal contexts have been refactored using our existing pattern of sending props via other publicly exported contexts for each component.

The state objects can be used to replace an existing sub-component (as shown in the Calendar hooks example), or to create new components that use the state. For example, a clear button in a combobox, or a time zone help text component in a time field. I have added a bunch of examples in the documentation for each component. There were a few that I couldn't think of a good one so those just list the state object for now.

For now I've decided to omit documentation for pure collection components (GridList, ListBox, Menu, and Table). Due to the way collections are implemented, it's not currently possible to replace one of the sub-children (e.g. Item) yet. Also the collections are not accessible within item wrappers (e.g. `MyItem`) either because there is a circular dependency on building the collection before the state can be constructed. There might be a way to solve at least one of these problems in the future but for now the states will remain undocumented (though they are public).

There will be further documentation on this in the Advanced Customization overview page I'm working on separately.


hosmelq:
Hello @devongovett,

I am in the process of creating a custom Menu component by reusing all the base components. I created a wrapper for each component to receive props through context. With the changes made in this pull request, I am wondering if it would be possible to create a specific component besides the Menu component using the collection in the menu state, similar to what you are doing [here](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/Menu.tsx#L84), or if I should stick with my current implementation.

Thank You.

```tsx
<Provider
  values={[
    [HeaderContext, {className: headerClassName()}],
    [IconContext, {className: iconClassName(), fixedWidth: true}],
    [ItemContext, {className: itemClassName()}],
    [SectionContext, {className: sectionClassName()}],
    [
      TextContext,
      {
        slots: {
          description: {
            className: descriptionClassName(),
          },
          label: {
            className: labelClassName(),
          },
        },
      },
    ],
  ]}
>
  <RAMenu
    {...otherProps}
    className={baseClassName({className})}
    ref={ref}
  >
    {children}
  </RAMenu>
</Provider>
```

devongovett:
@hosmelq what you're trying to do there might not work because those contexts will get overwritten deeper inside the tree. For example, the `Section` component provides a value for `HeaderContext`, and the `Item` component provides a value for `TextContext`. That will overwrite the value you provided in the outer `Menu` - the contexts are not merged with each other. You might need to create your own contexts for that instead.

packages/react-aria-components/docs/Calendar.mdx
```diff
@@ -880,14 +914,12 @@ function WeekCalendarGrid({state, ...props}: WeekCalendarGridProps) {
...
 ```tsx example
 <Calendar visibleDuration={{weeks: 1}} defaultValue={today(getLocalTimeZone())}>
-  {({state}) => (
-    <div className="week">
-      <Heading />
-      <Button slot="previous">◀</Button>
-      <WeekCalendarGrid state={state} />
-      <Button slot="next">▶</Button>
-    </div>
-  )}
```
LFDanLu, line 890: Now that we can access the component state via a public context, what do we think about the existence of state in renderProps? I guess it can come in handy if the user wants to style something/do something simple without needing to make a wrapper component and access state via the state context, but feels like it is largely unneeded now.

packages/react-aria-components/docs/Menu.mdx
LFDanLu, line 1: for others who got confused why the menu wasn't opening, it isn't a context issue, the example specifically only opens if the alt key is pressed haha

packages/react-aria-components/docs/Tabs.mdx
```diff
@@ -603,6 +662,48 @@ import {Link} from 'react-aria-components';
...
+  let state = React.useContext(TabListStateContext);
+  /*- end highlight -*/
+  let prevKey = state?.collection.getKeyBefore(state.selectedKey);
+  let nextKey = state?.collection.getKeyAfter(state.selectedKey);
+  let onPrev = prevKey != null ? () => state.setSelectedKey(prevKey) : null;
+  let onNext = nextKey != null ? () => state.setSelectedKey(nextKey) : null;
+  return (
+    <div className="button-group">
+      <Button aria-label="Previous tab" onPress={onPrev}>⬅</Button>
+      <Button aria-label="Next tab" onPress={onNext}>➡</Button>
```
LFDanLu, line 685: was gonna comment a fix but no idea why the emoji turns out a different size...

LFDanLu (dismissed):
Did a sweep of the docs/stories and things seem to work as expected. Just had one question about state in renderProps

packages/react-aria-components/docs/Calendar.mdx
devongovett: I guess there's no reason not to expose it. Like you said it could be useful for one-offs that don't need to be a whole new reusable component.

devongovett:
@hosmelq I think this is the part I mentioned in the PR description as not being possible yet. Those MenuSection and MenuItem components are indeed coupled with Menu at the moment. You can create a wrapper around `<Item>` but you cannot replace the implementation of it currently. I have some ideas for this but will be tackled separately from this PR.

packages/react-aria-components/src/Breadcrumbs.tsx
```diff
@@ -50,7 +50,7 @@ function BreadcrumbsInner<T extends object>({props, collection, breadcrumbsRef:
     <ol
       ref={ref}
       {...filterDOMProps(props, {labelable: true})}
-      slot={props.slot}
+      slot={props.slot || undefined}
```
snowystinger, line 53: what is setting props.slot to `null`? or what falsey value is coming through that isn't already `undefined`?

snowystinger (dismissed):
i'll re-approve after the merge conflicts are resolved

packages/react-aria-components/src/Breadcrumbs.tsx
devongovett: This was due to the type change, where I added support for passing `null` explicitly to make local props override context completely rather than merging them. But the TS types for HTML elements only allow `string | undefined` and not `null`. 😞

snowystinger: Thanks, I've run into this a few times in the TS strict upgrade, just wanted to make sure

reidbarber (approved):

LFDanLu (approved):

PR #5064, opened 2023-09-12
Initialize RAC string translations

reidbarber:
Adds translation file for remaining RAC strings that need translations.

## ✅ Pull Request Checklist:


snowystinger (approved):
I guess an Intl Provider was already required for translations in our hooks. So this should be fine.

The only other comment is that all RAC users, regardless of if they use Breadcrumbs, will now get breadcrumb strings. There aren't many strings now, but should we split it out anyways?

yihuiliao (approved):

reidbarber:
> The only other comment is that all RAC users, regardless of if they use Breadcrumbs, will now get breadcrumb strings. There aren't many strings now, but should we split it out anyways?

@snowystinger I think this ok for now, and we aren't anticipating a lot of strings in RAC that wouldn't be in the hooks. We can split them up in the future if we need to and it will be non-breaking.

PR #5074, opened 2023-09-13
Update RAC Link API

devongovett:
Closes https://github.com/orgs/adobe/projects/19/views/32?pane=issue&itemId=25877582

Depends on #4993

This updates the RAC `Link` API to just have `href` and other link props directly on it rather than cloning its child. With the `RouterProvider` added in #4993 this will support client side routers the same way as our other components, meaning there is no need to support wrapping a router Link component.

### DIFF
--- a/packages/react-aria-components/docs/Breadcrumbs.mdx
+++ b/packages/react-aria-components/docs/Breadcrumbs.mdx
@@ -46,8 +46,8 @@ type: component
 import {Breadcrumbs, Breadcrumb, Link} from 'react-aria-components';
 
 <Breadcrumbs>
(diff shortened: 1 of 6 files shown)


snowystinger (dismissed):
looks good!

snowystinger (approved):

yihuiliao (approved):

LFDanLu (approved):
LGTM, will add a PR for the outdated copy in https://reactspectrum.blob.core.windows.net/reactspectrum/7e8cb425b66c1f2cca44be86d67de442de008de3/docs/react-aria/Breadcrumbs.html#link-1 that mentions the cloning

devongovett:
Oh good catch!

PR #5084, opened 2023-09-15
Add LocalizedStringProvider to inject i18n strings during SSR

devongovett:
This is an extension of the work done in #4838.

To recap, #4838 adds a build plugin that allows apps to configure which languages to include in the JS bundle. This is a good start for apps that don't support all 30+ languages we do, but doesn't entirely solve the problem because users still need to download all supported languages even though they won't use them. For example, English speaking users might need to download the French and German translations as well.

However, by default you only get localized strings for the components you actually use in your app. Originally I thought this tradeoff was worth it. If you only used one component, you'd need to download the strings for all supported languages for only that one component, but not all of the others. However, upon further research, it turns out that it may actually be smaller to download the strings for all components but only in one language than it is to download the strings for only one component but in all languages. We do not have a high number of strings, so the total size of a single language for all components is around 2KB gzipped, whereas the size of a single component across all languages may be much larger. Depends on how many languages you support and how many components you use.

Server rendering gives us an opportunity to solve this problem. The server knows what language the user will need, either via an http header, a URL param, a database setting, etc. It can inject the strings for only the user's language into the initial HTML so that the client rendering code has access to it. The previously mentioned build plugin can be used to remove _all_ localized strings from the JS bundle, and instead, the JS relies on the strings injected by the server.

Theoretically this could also be done in client rendered apps, using a dynamic import to load the strings for the user's language prior to rendering the app. However, that requires an extra HTTP request compared with server rendering where the strings can be injected into the initial HTML. You could also bundle the React Aria translations together with other translations you might have. It all depends on the tradeoffs for your individual app.

This approach gives a number of different options to applications to deal with our translations. By default, we optimize for ease of getting started - all translations are included in the bundle. When you're ready to optimize, you can reduce the set by adding a build plugin, and if using SSR, optimize even further using this SSR injection approach.

## Implementation details

Each monopackage has a new `i18n` folder in its published package. Within this folder are JS files for each individual language exporting the strings for each package in a map. (These are JS files not JSON because strings with interpolations are compiled to functions.)

```js
export default {
  '@react-aria/calendar': {
    stringIdentifier: 'value'
  },
  // ...
};
```

These individual language files can be imported directly (e.g. for dynamic import in the client rendering case). The i18n folder also contains an index file that exports a `LocalizedStringDictionary` containing all of these languages, which accepts a locale identifier and maps it to the closest set of strings (e.g. if the language is `en-GB`, it would return the `en-US` strings).

There is also a convenience component `LocalizedStringProvider`, which looks up the languages using the dictionary and renders a `<script>` element with a serialized version of the strings. This should be rendered in the initial html (e.g. in `_document.js` or a React server component in Next.js).

```js
import {LocalizedStringProvider} from '@adobe/react-spectrum/i18n';

<LocalizedStringProvider locale="en" />
```

The script tag looks something like this:

```html
<script>
window[Symbol.for('react-aria.i18n.locale')] = 'en';
window[Symbol.for('react-aria.i18n.strings')] = {
  '@react-aria/calendar': {
    stringIdentifier: 'value'
  },
  // ...
};
</script>
```

This is similar to how Next.js and other SSR implementations inject data needed during hydration into the initial HTML.

The `useLocalizedStringFormatter` hook will look for this global variable using the same symbol and use those strings if available. It accepts a new (optional) parameter for the package name so it can choose the strings for the correct package.

```js
// existing code
import i18nMessages from '../intl/*.json';

let formatter = useLocalizedStringFormatter(i18nMessages, '@react-aria/calendar');
```

When using the build plugin configured to include no locales, `i18nMessages` here will be empty, and the strings from the global will be used instead.

By default, `LocalizedStringProvider` for each monopackage includes all strings that would be needed. For example, in RSP, it includes the strings for all RSP components plus all aria hooks. However, if you want, you can do even more manual optimization by including only the packages you need in your app. This is advanced because you would need to know all of the dependencies and manually update the list of included packages. The `PackageLocalizationProvider` component from `@react-aria/i18n/server` is used internally by `LocalizedStringProvider` and accepts a list of strings, so you can pass that your custom subset.

```js
import en from '@adobe/react-spectrum/i18n/en';
import fr from '@adobe/react-spectrum/i18n/fr';
import {LocalizedStringDictionary} from '@internationalized/string';
import {PackageLocalizationProvider} from '@react-aria/i18n/server';

// Only include calendar strings.
let strings = new LocalizedStringDictionary({
  'en-US': {
    '@react-spectrum/calendar': en['@react-spectrum/calendar']
    '@react-aria/calendar': en['@react-aria/calendar']
  },
  'fr-FR': {
    '@react-spectrum/calendar': fr['@react-spectrum/calendar']
    '@react-aria/calendar': fr['@react-aria/calendar']
  }
});

// In initial html...
<PackageLocalizationProvider locale={locale} strings={strings.getStringsForLocale(locale)} />
```

This is a pretty advanced use case but it is possible. We'll need to have docs for all of this and how to configure for various frameworks too.


examples/rsp-next-ts/pages/_document.tsx
```diff
@@ -0,0 +1,15 @@
+import { Html, Head, Main, NextScript } from 'next/document'
+import {LocalizedStringProvider} from '@adobe/react-spectrum/i18n';
+
+export default function Document(props) {
+  return (
+    <Html lang="en">
+      <Head />
+      <body>
+        <Main />
+        <LocalizedStringProvider locale="en" />
```
devongovett, line 10: One thing to consider is what to do if you had both RSP and RAC in your app since there is overlap between the two in terms of translations, and the script tags would overwrite each other. You might need something more manual using `PackageLocalizationProvider` in that case.

snowystinger (approved):
Having trouble finding the old PR for it, but I think this should open us to doing the color names finally.

snowystinger (approved):

LFDanLu (approved):

PR #5109, opened 2023-09-20
fix #5019 Popover prevents scroll only when it's opened

### LINKED ISSUE #5019: usePreventScroll behaves differently between usePopover and useModalOverlay

If you create a Popover component that calls `usePopover` and this component takes care itself of `state.isOpen` to mount or not the html, the call of `usePopover` will automatically blocks the scroll on mount.

The same way of working is possible with a Modal component and doesn't have this bug.

The reason is that `usePopover` has
https://github.com/adobe/react-spectrum/blob/2600c965036eb66e9f4af90a8925b1f7fd165b24/packages/%40react-aria/overlays/src/usePopover.ts#L96C25-L96C25
while `useModalOverlay` stops the scroll only when opened
https://github.com/adobe/react-spectrum/blob/2600c965036eb66e9f4af90a8925b1f7fd165b24/packages/%40react-aria/overlays/src/useModalOverlay.ts#L54 

### 🤔 Expected Behavior?

calling `usePopover` should not block the scroll as long as `state.isOpen` is not `true`.

### 😯 Current Behavior

The scrolling is block as long as you mount a component that calls `usePopover`

### 💁 Possible Solution

in `usePopover`:
```js
usePreventScroll({
    isDisabled: isNonModal || !state.isOpen
});
```

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

Works as expected: (imagine the triggersProps, etc...)

```jsx
<button>open</button>
<Modal><Dialog /><Modal>
```

Blocks the scroll
```jsx
<button>open</button>
<PopOver><Dialog /><Popover>
```

In both `Modal` and `Popover` we have
```jsp
return state.isOpen ? <content /> : null
```
No need to do
```jsp
<button>open</button>
{state.isOpen && <Popover><Dialog /></Popover>}
```
which is more convenient for the devs and works well with `Modal`

### Version

latest

Safari

### If other, please specify.

_No response_

macOS

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

cedeber:
Closes #5019

## ✅ Pull Request Checklist:


reidbarber:
GET_BUILD

snowystinger (approved):

LFDanLu (approved):
Thanks!

PR #5116, opened 2023-09-21
Update useModalOverlay types

doytch:
Extended the type of the `useModalOverlay()` props from `useOverlay()`'s props.

I want to be able to pass `shouldCloseOnInteractOutside` to `useModalOverlay` so that I can prevent closing in certain cases. 

The values are already being passed through to `useOverlay()` so I went ahead and simply extended `AriaOverlayProps`  instead of adding more and more repeating variables. I'm not sure if there's a particular reason that approach wasn't taken already, so feel free to recommend the alternative.

It seems like any docs for this should be updated automatically but lemme know if I'm missing anything here.

## ✅ Pull Request Checklist:


doytch:
Sorry for the close/re-open spam - was trying to shake the CLABot out of its slumber 😆

LFDanLu:
I've actually had to do something similar when implementing SubMenus for `usePopover`, will let you know what the team decides about exposing `shouldCloseOnInteractOutside`.

doytch:
Right on, thanks.

In case it's useful, the specific use case here came up when with our design system's Dialog component. When a Dialog contained a legacy Select component, attempting to select an entry from the dropdown would close the dialog. 

[`useInteractOutside`](https://github.com/adobe/react-spectrum/blob/f65c181868be342a9238e5c26c12a1a0d346ca2f/packages/%40react-aria/interactions/src/useInteractOutside.ts#L33) triggers the closing, since the select component would render the dropdown in a React-Portal-based overlay.

packages/@react-aria/overlays/src/useModalOverlay.ts
```diff
@@ -11,26 +11,15 @@
...
-   * @default false
-   */
-  isDismissable?: boolean,
-  /**
-   * Whether pressing the escape key to close the modal should be disabled.
-   * @default false
-   */
-  isKeyboardDismissDisabled?: boolean
-}
+export interface AriaModalOverlayProps extends AriaOverlayProps {}
```
LFDanLu, line 22: We discussed adding `shouldCloseOnInteractOutside` to `usePopover` for submenus today, leaning towards being fine adding it there and for `useModalOverlay` as well as you've done here. I'd personally would prefer just adding `shouldCloseOnInteractOutside` only here instead of all the overlay props until we get use cases for the rest of those props but happy for the rest of the team to weigh in.

Would you mind adding a basic test for this? You can probably add a new test file for useModalOverlay or for the RAC Modal since it extends from AriaModalOverlayProps. https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/overlays/test/useModal.test.js or any of the tests in https://github.com/adobe/react-spectrum/tree/main/packages/react-aria-components/test could serve as a guide

LFDanLu:
This is now supported via https://github.com/adobe/react-spectrum/pull/5986.

PR #5125, opened 2023-09-23
Static progress followup

iamwillpowell:
Fixes a couple of High Contrast Mode issues.

## ✅ Pull Request Checklist:


packages/@adobe/spectrum-css-temp/components/barloader/skin.css
```diff
@@ -84,6 +86,9 @@ governing permissions and limitations under the License.
     forced-color-adjust: none;
     --spectrum-barloader-large-track-fill-color: ButtonText;
     --spectrum-barloader-large-over-background-track-fill-color: ButtonText;
+    --spectrum-barloader-large-over-background-track-color: ButtonText
```
snowystinger, line 89:
```suggestion
    --spectrum-barloader-large-over-background-track-color: ButtonText;
```

snowystinger:
A couple combinations don't quite look right still

Screenshot 2023-09-25 at 10 29 05 AM
Screenshot 2023-09-25 at 10 30 06 AM

In addition, I think all the HCM pending buttons should have the green (disabled) outline to indicate disabled. It also makes it clear that it's where the button was, instead of a spinner floating in space.

iamwillpowell:
> A couple combinations don't quite look right still
> 
> Screenshot 2023-09-25 at 10 29 05 AM Screenshot 2023-09-25 at 10 30 06 AM
> In addition, I think all the HCM pending buttons should have the green (disabled, colors may change depending on system settings)) outline to indicate disabled. It also makes it clear that it's where the button was, instead of a spinner floating in space.

 I pushed fixes for the combinations that don't look right. I also found that HCM light theme was pretty bad the progress track and fill were almost indistinguishable. I decided the track color should be gone for HCM.

Will add a new ticket to add an outline for HCM pending buttons.

snowystinger (approved):

snowystinger (changes requested):
Screenshot 2023-09-26 at 12 33 58 PM

Actually, went back and checked on this one because I was worried about the case of a progress with 0 or very little progress. It's impossible to tell its there, or to see the ratio of completed to uncompleted. I think we have to maintain some sort of track. It might be that we do something like what we do in https://react-spectrum.adobe.com/react-aria/ProgressBar.html with two borders and then a fill inside them instead of track and fill being the same space. I realize this is more work, so bring it up in standup/planning tomorrow.

packages/@adobe/spectrum-css-temp/components/button/skin.css
```diff
@@ -420,6 +420,13 @@ governing permissions and limitations under the License.
   }
 }
 
+@media (forced-colors: active) {
```
iamwillpowell, line 423: I tried to just update the custom property value for HCM, but there were a couple of problems, so I needed to add a whole rule inside the media query.

snowystinger: any reason this is in a separate media block from the other forced-colors block just above it?

Just notes for a future discussion of disabled buttons in HCM - I don't think it needs to be addressed in this PR:
I think all buttons with outline should get the green ring, floating green text isn't all that helpful either.

Then we wouldn't be dependent on the isPending state either, so it'd simplify this little bit

iamwillpowell:
> Screenshot 2023-09-26 at 12 33 58 PM
> Actually, went back and checked on this one because I was worried about the case of a progress with 0 or very little progress. It's impossible to tell its there, or to see the ratio of completed to uncompleted. I think we have to maintain some sort of track. It might be that we do something like what we do in https://react-spectrum.adobe.com/react-aria/ProgressBar.html with two borders and then a fill inside them instead of track and fill being the same space. I realize this is more work, so bring it up in standup/planning tomorrow.

I added a separate task to add the outline to the track: https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=40002851

iamwillpowell:
@snowystinger, I added the outline to pending buttons in HCM.

snowystinger (approved):
Thanks for the updates

packages/@adobe/spectrum-css-temp/components/button/skin.css
iamwillpowell: I just wanted it to be separated from the variable settings block. Could go either way.

packages/@react-spectrum/progress/docs/ProgressBar.mdx
LFDanLu, line 1: There are a couple of mentions of the `overBackground` variant in the ProgressBar/ProgressCircle accessbility docs section that should be replaced by `staticColor`

LFDanLu: image

LFDanLu (changes requested):
just one small change in the docs, but otherwise looks good to me

packages/@react-spectrum/progress/docs/ProgressBar.mdx
iamwillpowell: Good catch. Updated

packages/@react-spectrum/progress/docs/ProgressBar.mdx
```diff
@@ -112,7 +112,7 @@ By default, the value label is formatted as a percentage. This can be customized
 
 Depending on the visualization being used (i.e. depending on the `showValueLabel` prop), a `label`, `aria-label`, or `aria-labelledby` prop is required.
 
-If using the over background variant, ensure the background offers enough contrast for the ProgressBar to be legible and meets color contrast guidelines.
+If using the `staticColor`` prop, ensure the background offers enough contrast for the ProgressBar to be legible and meets color contrast guidelines.
```
LFDanLu, line 115:
```suggestion
If using the `staticColor` prop, ensure the background offers enough contrast for the ProgressBar to be legible and meets color contrast guidelines.
```

LFDanLu (approved):

devongovett (approved):

PR #5147, opened 2023-09-27
Fix specificity of tailwind generated selectors

devongovett:
I noticed that some states like `focus-visible` were overriding others like `selected`. That is because of the extra `[data-rac]` part of the selector, which raises a rule like `.focus-visible\:bg-red[data-rac][data-focus-visible]` to a specificity of (0, 3, 0) whereas `.selected\:bg-red[data-selected]` only has (0, 2, 0).

This PR uses `:where([data-rac])` instead to ensure that is not taken into account in the specificity.


snowystinger (approved):

reidbarber (approved):
LGTM

PR #5174, opened 2023-09-29
Fix overlay positioning when portal container is not body

devongovett:
Fixes popover positioning with contextual help trigger, and left/right placement.


LFDanLu (approved):
Tested docs popovers for RSP, React Aria, and RAC LGTM. Storybook stories look good at various offsets/placement options as well. Chromatic ran clean as well

PR #5196, opened 2023-10-04
Copy browserslist into prod website build

devongovett:
Hopefully fixes media query compilation on older browsers that is breaking layout on small devices in Safari


LFDanLu (approved):
LGTM, tested on iPhone Safari, Android Chrome, and desktop browsers

yihuiliao (approved):
also tested on iphone safari (16.7) and desktop browsers

devongovett:
Thanks for testing, but this would only change the prod docs so we will need to test once verdaccio builds. 😄

PR #5211, opened 2023-10-06
Add isRootDropTarget parameter to renderEmptyState in GridList component

### LINKED ISSUE #5210: Add `isRootDropTarget` parameter to `renderEmptyState` in `GridList` component

Add `isRootDropTarget` parameter to `renderEmptyState` in `GridList` component.

This might be useful when you want to change the content of a placeholder depending on whether there is a dragging element over the grid list root or not similar to `DropIndicator` where the flag `isDropTarget` returned in a `className`.

```
<DropIndicator target={target} className={({ isDropTarget }) => ...} />
```

Currently as far as I understand there is no way to check it using `GridList` from `react-aria-components` although it is possible using low-level hooks.

### 🤔 Expected Behavior?

I would expect `isRootDropTarget` boolean to be returned as a param for `renderEmptyState` in a `GridList`.

Something like that:

```
<GridList
      ...
      renderEmptyState={(isRootDropTarget) => {
        return (
          <span style={{ color: isRootDropTarget ? "red" : "black" }}>
            Drop items here
          </span>
        );
      }}
      dragAndDropHooks={dragAndDropHooks}
      ...
```

### 😯 Current Behavior

Currently I couldn't find any solution to check if the root is a drop target or not with `react-aria-components`. So we've been using CSS for that accessing grid list with `data-drop-target` attribute:

```
.gridList[data-drop-target] .rootDrop ...
```

### 💁 Possible Solution

WIP version here: #5211

### 🔦 Context

It becomes an issue when we needed to check whether the root is currently a drop target or not and just using CSS selectors is not enough. 

### 💻 Examples

_No response_

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

gennadiipel:
Add `isRootDropTarget` parameter to `renderEmptyState` in `GridList` component

Closes #5210

## ✅ Pull Request Checklist:


packages/react-aria-components/src/GridList.tsx
```diff
@@ -54,7 +54,7 @@ export interface GridListProps<T> extends Omit<AriaGridListProps<T>, 'children'>
   /** The drag and drop hooks returned by `useDragAndDrop` used to enable drag and drop behavior for the GridList. */
   dragAndDropHooks?: DragAndDropHooks,
   /** Provides content to display when there are no items in the list. */
-  renderEmptyState?: () => ReactNode
+  renderEmptyState?: (isRootDropTarget: boolean) => ReactNode
```
reidbarber, line 57:
```suggestion
  renderEmptyState?: (props: {isDropTarget: boolean, isFocused: boolean, isFocusVisible: boolean, state: ListState<unknown>}) => ReactNode
```

reidbarber (changes requested):
I like this idea a lot!

I say we go ahead and pass all the render props in (except for `isEmpty` since it is implied). And we can use one argument with everything.

reidbarber:
GET_BUILD

reidbarber (approved):

packages/react-aria-components/src/GridList.tsx
```diff
@@ -176,7 +176,12 @@ function GridListInner<T extends object>({props, collection, gridListRef: ref}:
     // they don't affect the layout of the children. However, WebKit currently has
     // a bug that makes grid elements with display: contents hidden to screen readers.
     // https://bugs.webkit.org/show_bug.cgi?id=239479
-    let content = props.renderEmptyState();
+    let content = props.renderEmptyState({
```
snowystinger, line 179: seems like a fine approach to me
we will need this in ListBox, Table, and TagGroup as well in order to be consistent
also, some form of a test/storybook example would be appreciated if you do not mind https://react-spectrum.adobe.com/contribute.html#pull-requests

gennadiipel:
@snowystinger @reidbarber Hello, I updated `Table`, `ListBox` and `TagGroup`. Also updated docs a little and added example of reorderable tables including PR's changes to storybook. Please take a look.

snowystinger:
GET_BUILD

packages/react-aria-components/docs/ListBox.mdx
```diff
@@ -850,7 +850,7 @@ Use the `renderEmptyState` prop to customize what the `ListBox` will display if
 <ListBox
   aria-label="Search results"
   /*- begin highlight -*/
-  renderEmptyState={() => 'No results found.'}
+  renderEmptyState={({isDropTarget, state, isFocused, isFocusVisible}) => <span style={{color: isDropTarget ? 'red' : 'black'}}>No results found.</span>}
```
snowystinger, line 853: these are great, though in dark mode, it's a little hard to see
Screenshot 2023-10-10 at 1 59 16 PM

Instead, we should use one of the variables available to us in our docs, `--spectrum-alias-text-color`
There is another issue, though: I cannot drop anything on this example, so it never goes to red.

As a result, I think it should be in the example a little further down https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/docs/ListBox.mdx#L1101
And instead of changing colors (esp to red which is negative), let's swap the string. Something like
```
isDropTarget ? 'This will not be uploaded' : 'Drop items here'
```

gennadiipel: Yep, updated docs like that in all of the cases and also added for `GridList` (forgot this one)

reidbarber:
GET_BUILD

reidbarber (dismissed):
LGTM, thanks!!

snowystinger (approved):

devongovett (approved):

snowystinger (approved):

PR #5226, opened 2023-10-09
Add styled examples for React Aria Components

devongovett:
This adds an examples page that includes fully styled examples integrating various libraries. Each example is on a separate page within the docs, and they are also linked to from the relevant component pages. The examples include the tailwind examples from our example app, plus some that I had already built in codesandbox. Moving them into the docs makes them more discoverable, and easier for us to test and maintain.

I have some additional examples in the works which will come later. This initial batch works without any code changes so we should be able to deploy before the next release.


packages/react-aria-components/docs/examples/stock-table.mdx
```diff
@@ -0,0 +1,188 @@
...
+      if (sortDescriptor.direction === 'descending') {
+        cmp *= -1;
+      }
+      return cmp;
+    })
+  }, [sortDescriptor]);
+
+  return (
+    <div className="bg-gradient-to-r from-indigo-500 to-violet-500 p-8 rounded-lg flex items-center justify-center md:col-span-2">
+      <ResizableTableContainer className="max-h-[280px] w-full overflow-auto relative bg-white rounded-lg shadow text-gray-600">
```
devongovett, line 109: Seems like we have a bug in column resizing where the columns jump to a different size the first time you mouse down. Should I pull this example until we get that fixed?

.parcelrc
```diff
@@ -9,6 +9,7 @@
     "bundle-text:*.svg": ["@parcel/transformer-svg", "@parcel/transformer-inline-string"],
     "*.{md,mdx}": ["parcel-transformer-mdx-docs"],
     "*.svg": ["@parcel/transformer-svg-react"],
+    "*.global.css": ["parcel-transformer-css-global", "..."],
```
devongovett, line 12: This hack will go away once https://github.com/parcel-bundler/parcel/pull/9301 lands.

packages/react-aria-components/docs/examples/stock-table.mdx
devongovett: I have a potential fix locally. I think this is RAC specific because the collection isn't filled until the second render. Looks like we are missing some logic for when the columns change (ie added/removed).

packages/dev/docs/src/docs.css
```diff
@@ -532,6 +536,31 @@ footer li:last-child:after {
...
+}
+
+:global(.flip) ~ pre:global(.example) {
+  border-radius: 0 0 4px 4px;
+  border-top: none;
+  border-bottom: 1px solid var(--spectrum-gray-200);
+  margin-top: 0;
+}
+
+:global(.standalone) > .example {
```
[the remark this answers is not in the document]
devongovett: this is confusing. there are actually two separate classes called `example`: one global and one module. 🤷

reidbarber (dismissed):
LGTM, my only input is that I think the Examples page grid should fill the width, or be centered.

reidbarber (dismissed):

snowystinger (approved):

PR #5322, opened 2023-10-28
Make RAC GridList TagGroup items easier to style

devongovett:
GridList and TagGroup items use the ARIA grid pattern, which means they need to have both a `row` and a `gridcell` element. However in our API, it is only exposed as a single component. That made it harder to style because you couldn't pass a class/style attribute to the cell, and had to know about this implementation detail. This mainly affected things like layout, where placing `flex`/`grid` on the row would not affect the children as expected.

This PR adds `display: contents` to the `gridcell` element so that it does not affect layout/styling. Previously we couldn't do this because of a [WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=239479) that made it behave like `aria-hidden`. This is now fixed in recent versions of Safari. Normally we wouldn't bump the minimum browser version so quickly, but since RAC is still in beta, I don't want to get locked into supporting a worse styling API forever (after we GA) when the browser bug is already fixed. If needed, this can be worked around temporarily by overriding the `display: contents` with CSS.


snowystinger (approved):
I don't recall if it was always like this, but the only way on here or on main for me to navigate the tags is to Ctrl+Opt+DownArrow. Or is this also a new OS/Safari version issue?

Other than that, announcements seem to work as expected.

reidbarber (approved):

PR #5323, opened 2023-10-28
Use explicit "title" slot for RAC dialogs, and improve aria labelling

devongovett:
The `<Heading>` component can be used inside a `<Dialog>` to set the aria label for the dialog. However, if you had multiple headings inside the dialog contents they would all receive the aria labelling props from the dialog. This changes the API to require an explicit `slot="title"` prop to be added to the instance of the heading you want to use as the title. In addition it adds a warning if you have a dialog with no title slot or explicit aria labelling props.

After adding the warning I found that we had a bunch of unlabeled dialogs in our examples/tests, mostly in popovers. To fix these, I made dialogs within a `<DialogTrigger>` be labeled by the trigger element as a fallback, in case no title slot or aria labelling props are provided. Standalone popovers will need to do this manually, and will get a warning.

### DIFF
--- a/packages/react-aria-components/docs/Popover.mdx
+++ b/packages/react-aria-components/docs/Popover.mdx
@@ -42,16 +42,26 @@ type: component
 ## Example
 
 ```tsx example
-import {DialogTrigger, Popover, Dialog, Button, OverlayArrow} from 'react-aria-components';
+import {DialogTrigger, Popover, Dialog, Button, OverlayArrow, Heading, Switch} from 'react-aria-components';
 
 <DialogTrigger>
-  <Button>Open popover</Button>
+  <Button>Settings</Button>
   <Popover>
     <OverlayArrow>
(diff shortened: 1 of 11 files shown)


packages/react-aria-components/docs/Popover.mdx
```diff
@@ -42,16 +42,26 @@ type: component
...
 <DialogTrigger>
-  <Button>Open popover</Button>
+  <Button>Settings</Button>
   <Popover>
     <OverlayArrow>
       <svg width={12} height={12} viewBox="0 0 12 12"><path d="M0 0 L6 6 L12 0" /></svg>
     </OverlayArrow>
     <Dialog>
-      This is a popover.
+      <div className="flex-col">
```
snowystinger, line 54: this example announces really oddly in Chrome with VO
Screenshot 2023-10-30 at 12 26 05 PM

snowystinger (approved):
In general behaves well, only one comment about some Chrome weirdness

reidbarber (approved):

PR #5326, opened 2023-10-29
fix package.json#types field in tailwind plugin

stefanprobst:
in the tailwind plugin package, point `package.json#types` to https://github.com/adobe/react-spectrum/blob/main/packages/tailwindcss-react-aria-components/src/index.d.ts


reidbarber (approved):
Good catch, thanks!

yihuiliao (approved):

PR #5344, opened 2023-11-02
Auto focus the first invalid field when submitting a form

devongovett:
This adds back a UX enhancement for forms where after attempting to submit invalid data, focus is automatically moved to the first invalid field. This is the default for native HTML forms, but our `preventDefault` in the `invalid` event to hide the browser's default error message UI also prevents this, so we have to add it back manually.

There is also a way to prevent this if you don't want this behavior: add an `onInvalid` event handler to the form and call `preventDefault` yourself. We will respect `defaultPrevented` on the event and not move focus. Then it's up to you to do it yourself.

A common case for that would be to add an `InlineAlert` at the top of a form to aggregate multiple errors. This adds an `autoFocus` prop to `InlineAlert` that makes it focus when it first appears. See the added docs example on the Form page.

### DIFF
--- a/packages/@adobe/spectrum-css-temp/components/inlinealert/index.css
+++ b/packages/@adobe/spectrum-css-temp/components/inlinealert/index.css
@@ -48,6 +48,12 @@ governing permissions and limitations under the License.
 }
 
 .spectrum-InLineAlert {
+  composes: spectrum-FocusRing;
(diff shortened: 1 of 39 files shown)


snowystinger (dismissed):
Looks like this is working well and as expected, tested with VO as well

reidbarber (approved):
LGTM

PR #3780, opened 2022-11-22
Enable horizontal scrolling in usePreventScroll

jhyj0521:
#3779

On iOS when using `usePreventScroll`, The code in the onTouchMove event handler prevents the internal element from scrolling horizontally.

This fix enable horizontal scrolling of the `usePreventScroll` internal element.

## ✅ Pull Request Checklist:

- [x] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- [x] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [x] Filled out test instructions.
- [x] Updated documentation (if it already exists for this component).
- [x] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/TR/wai-aria-practices-1.1/)

### before
https://codesandbox.io/s/xenodochial-grass-3onoeo?file=/src/index.js

### after
https://codesandbox.io/s/clever-sea-b2iuzi?file=/src/index.js

Run before and after in iOS to compare.
Support for horizontal scrolling without side effects to existing code.

PR #5346, opened 2023-11-02
Improve behavior of usePreventScroll on iOS

### LINKED ISSUE #5019: usePreventScroll behaves differently between usePopover and useModalOverlay

devongovett:
Closes #5019

Currently in many of the [RAC examples](https://react-spectrum.adobe.com/react-aria/examples/) with overlays on iOS, the entry/exit animations are weird/glitchy. This is due to `usePreventScroll` on iOS messing up the overlay positioning due to the hacks to undo scrolling with margin. However this is only really necessary to deal with the software keyboard coming up and trying to center an input in the viewport. This PR therefore only applies those changes when the keyboard is about to come up, and not while entry animations are happening, making the UX less janky. Need to test in a few places to ensure it still works well.


snowystinger (commented):
Ok, i see how to test this now. You can't be at the top of the page for most of them or they appear normal.

This PR introduces behavior in DatePicker where I can scroll the page while it's open. If I touch and drag from inside the overlay while it's open.

devongovett:
I think that's related to https://github.com/adobe/react-spectrum/pull/3780 That PR seemed to break preventing page scroll in general, including vertical when you start from inside the horizontally scrollable element. The story that was added just doesn't have much page content. Not sure how we can allow horizontal scrolling while preventing page scrolling though. I was hoping [overscroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior) might work but seems like not.

devongovett:
One tradeoff could be to change it to 

```js
let bottom = scrollable.scrollHeight - scrollable.clientHeight;
let right = scrollable.scrollWidth - scrollable.clientWidth;

if (bottom === 0 && right > 0) {
  return;
}
```

That way we only allow it when the target is actually horizontally scrollable. But if it was you wouldn't get any scroll locking in either direction.

devongovett:
Also related: #5144

devongovett:
Found another approach. We can use the relatively new [overscroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior) CSS property to prevent scroll chaining of inner scrollable elements within an overlay from causing page scrolling. This mostly replaces our manual `preventDefault` in the `touchmove` event, except for one [bug](https://bugs.webkit.org/show_bug.cgi?id=243452) that prevents it from working when the content doesn't actually overflow. In the horizontal scrolling only case, the best solution is to add `touch-action: pan-x` to the scrollable element to only allow scrolling in the x direction. We can't add that automatically because it must be added _before_ the `touchstart` event fires.

The changes in this PR also allow us to bring back #5109 and stop preventing scroll immediately when `isOpen` is set to `false` without waiting for animations. We previously reverted in #5144 due to flickering on iOS, which now won't happen because we don't apply the page margin styles unless an input was focused. That case is far less likely. cc. @cedeber.

LFDanLu (approved):
Behavior looks pretty good, confirmed that the exit/entry animations looks good in the RAC examples and scrolling/preventing scrolling works well in other components. I was still able to reproduce the behavior @snowystinger reported here: https://github.com/adobe/react-spectrum/pull/5346#pullrequestreview-1711758334 but it wasn't very consistent (might have been a micro pinch zoom or something). This was in https://reactspectrum.blob.core.windows.net/reactspectrum/3167d15ba9cdd8b9a36ffa8ad91a2a360c33ee86/docs/react-aria/examples/notifications-popover.html.

devongovett:
On iPhone? I haven't been able to reproduce...

yihuiliao (approved):
I also haven't been able to reproduce the thing that Rob showed on my iPhone but I did notice for one of the examples that it flashes when you open it. Not sure if this is something that was supposed to be fix by this PR but the other popover examples look good to me. 


https://github.com/adobe/react-spectrum/assets/44729383/8c0ec3a5-76fe-41ed-b1c4-3057f6016769

LFDanLu:
> On iPhone? I haven't been able to reproduce...

yep, on iPhone, iOS 16.7.2. Still haven't found solid reproduction steps

ktabors:
> * View the documentation

ktabors:
> yep, on iPhone, iOS 16.7.2. Still haven't found solid reproduction steps

To verify I understand the issue I found that if you select a RAC DatePicker segment (opens virtual keyboard) then click on the open popover (closes virtual keyboard) that causes this every time. iOS 16.6.1

With RAC Notification popover example I only see if render then shift up a few pixels.

With the current changes the Notification popover doesn't shift a few pixels. The RAC DatePicker example behaves the same, tries to render above, then renders below. The RAC Datepicker is fixed, no partial render, just a single full render.

ktabors (approved):

snowystinger (approved):
iOS 17 I cannot reproduce anymore
new features on horizontal scrolling look good, question, should we put that pan-x in an example somewhere in docs?

In general I'd be against it because at 400% zoom people need to eliminate most horizontal scrolling. but there are some exceptions, namely diagrams/maps are probably most likely.

I do find the horizontal scrolling dialog to be very difficult to use with both scroll directions at once. I tried adding 5paragraphs to the content, this worked well on `main`.

That said, nested scroll regions inside the Dialog seem to work better. I can actually scroll to the bottom of the nested region, whereas in `main`, I never can quite bring it into view.

For reference, this is what I added to the horizontal scrolling Dialog story.
```
              <div style={{height: '250px', overflow: 'auto'}}>
                {fiveParagraphs()}
              </div>
```

devongovett:
> should we put that pan-x in an example somewhere in docs?

maybe but it isn't a common scenario, and the example _works_ without it. it's just an enhancement.

> I tried adding 5paragraphs to the content, this worked well on main.

well that's not quite fair because on main the whole page scrolls as well 😉

cedeber:
Nice, thanks.

PR #5354, opened 2023-11-03
Fix ComboBox re-opening after selection

### LINKED ISSUE #5120: [RAC] ComboBox stays open after selection

https://github.com/adobe/react-spectrum/assets/7543552/eb7f698f-d16e-4356-8593-3139fcd23e59

### 🤔 Expected Behavior?

ComboBox popover closes after selecting an item.

### 😯 Current Behavior

ComboBox popover stays open after selecting an item.

### 💁 Possible Solution

Workaround: control the input value via `inputValue` and `onInputChange`, then this bug doesn't happen.

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

I haven't been able to reproduce the issue in [this CodeSandbox](https://codesandbox.io/s/bitter-morning-nchy6t?file=/src/App.js) yet, so any pointers on where to look for the bug would be appreciated. But the code in the sandbox matches what we have, minus the styling related code.

### Version

Reproduced in 1.0.0-beta.0 and 1.0.0-nightly.4153

Chrome

### If other, please specify.

_No response_

macOS

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
Fixes #5120

This fixes ComboBox re-opening after selecting an option in the RAC docs. The problem stems from how React batches updates, and how our collection implementation works. Our collection implementation uses React's `useSyncExternalStore` hook to trigger updates. Inside ComboBox, we have a `lastValue` ref, and an `inputValue` state, which are updated simultaneously. Then, in the next render, we assume that `lastValue.current` will equal `inputValue`. However, this assumption is not true when some ancestor component uses `useSyncExternalStore`, which triggers a _synchronous_ update. Essentially, `setState` is asynchronous, but a synchronous update can preempt it. In that case, the ref will already be updated but the state will not. This [CodeSandbox](https://codesandbox.io/s/happy-goodall-w2lyhw?file=/src/App.tsx:463-814) has a minimal reproduction.

The solution is to move `lastValue` to state instead of a ref. There are probably a lot more instances of this pattern that we should audit throughout the codebase, but this is the first real world case where this assumption turned out to be incorrect.

This started appearing in our docs due to the change from `ReactDOM.render` to `ReactDOM.createRoot`. `render` behaves like React 17, so state updates run synchronously. `createRoot` opts-into concurrent features.


snowystinger (approved):

LFDanLu (approved):
Tested the RAC docs and storybook examples and things seem to still work, fingers crossed

PR #5358, opened 2023-11-03
Add isEntering and isExiting props to control animation from outside

### LINKED ISSUE #4888: RAC Select RenderProps prevents Popover from rendering

when using the Select renderprops "isOpen" variable to manually control the appearance of the popover (use case -> framer motions AnimatePresence requires this to effectively handle its child components animation lifecycle, the popover in this instance), the popover looses all functionality. 

### 🤔 Expected Behavior?

The popover should work with or without the option of using the Select fields renderprops to conditionally control the state of its children. 

### 😯 Current Behavior

At first i thought it was the AnimatePresence blocking the rendering, however it appears that attempting to conditionally control the render state of the popover prevents it from being rendered at all.

### 💁 Possible Solution

Without looking at the underlying code i suspect a ref or id is attempting to be passed to a popover wrapper component on initial render, and with a predicate in the form of Selects isOpen prop preventing it from being passed, it can never be controlled.

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

codesandbox available [here](https://codesandbox.io/s/patient-surf-c6t4h6?file=/src/App.js)

This sandbox has popover wrapped in an animate presence as it was going to be my initial bug submission, however if you remove any semblance of framer motion and your just left with the isOpen predicate, you can see that

a. the popover doesn't appear and
b. the predicate state being rendered in the button stays false. 

### Version

3.25.1

Chrome

### If other, please specify.

_No response_

Windows 11

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
Closes #4888

This adds `isEntering` and `isExiting` props to RAC Modal, Popover, and Tooltip so animation state can be controlled from the outside just like we do by default for CSS animations. This allows us to avoid problems with Framer Motion's `AnimatePresence` which fully removes the element from the React tree. Sometimes we need to keep it there even when not open, e.g. to register Select/ComboBox items. Updated the example in the styling docs to show how to do this.

Also fixed another issue where the popover underlay would not unmount until after the animation completed, resulting in hover styles being delayed and flickering occurring (you can see this on the RAC Action Menu example).

Finally, this merges `style` passed from context with the local props so that `--trigger-width` still gets set on popovers (via context) even if a `style` prop is present (e.g. via framer motion). I did not add this to `mergeProps` because it might be considered a breaking change? Not sure.

Ideally we'd have a live example but not sure where to put it.


packages/react-aria-components/docs/styling.mdx
```diff
@@ -288,46 +288,61 @@ This enables using props like [animate](https://www.framer.com/motion/animation/
...
+          {/* ... */}
+        </MotionModal>
+      </MotionModalOverlay>
+    </DialogTrigger>
   );
 }
 ```
 
-`AnimatePresence` also allows you to animate when items are added or removed in collection components. Use `array.map` to create children, and make sure each child has a unique `key` in addition to an `id` to ensure Framer Motion can track it.
+**Note**: Framer Motion's `AnimatePresence` component may not work with React Aria overlays in all cases, so the example shown above is the recommended approach for exit animations.
```
devongovett, line 343: Wasn't really sure where to put this. I thought a note about why we aren't using AnimatePresence might be important though? Sorta weird that we say this and then immediately start talking about it for collections below though.

reidbarber (approved):
LGTM

packages/react-aria-components/docs/styling.mdx
LFDanLu: Yeah the flow here does feel a bit odd, but I think its fine since it serves as a forewarning

LFDanLu (approved):

PR #5372, opened 2023-11-06
Do not crash in useFormValidation if ref is undefined

devongovett:
Fixes useDateField/useDatePicker examples in the docs, which do not render a native `<input>` and therefore have no `inputRef`. They still call `useFormValidation` and then crash inside there.


snowystinger (approved):

LFDanLu (approved):
verified the useDatefield and useDatePicker examples don't crash anymore

PR #5374, opened 2023-11-07
Docs css fixes

devongovett:
From testing

### DIFF
--- a/packages/react-aria-components/docs/Table.mdx
+++ b/packages/react-aria-components/docs/Table.mdx
@@ -136,7 +136,11 @@ import {Table, TableHeader, TableBody, Column, Row, Cell} from 'react-aria-compo
   }
 
   .react-aria-Row {
-    border-radius: 6px;
+    --radius-top: 6px;
+    --radius-bottom: 6px;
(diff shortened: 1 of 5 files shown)


packages/react-aria-components/docs/Table.mdx
```diff
@@ -149,15 +153,13 @@ import {Table, TableHeader, TableBody, Column, Row, Cell} from 'react-aria-compo
...
       background: var(--spectrum-global-color-gray-200);
     }
 
     &[data-selected] {
-      .react-aria-Cell {
-        background: var(--highlight-background);
-        color: var(--highlight-foreground);
-      }
+      background: var(--highlight-background);
+      color: var(--highlight-foreground);
```
devongovett, line 163: Looks like this was was actually broken in production for a while as well. The problem was that the outline was on the row, but we put the background on the cells which covered it. This moves the background back to the row instead. However, Firefox seems not to support border radius when rendering the background of a table row. Adding the `clip-path` as well works around this.

LFDanLu (approved):
LGTM, verified that the issues found in testing don't reproduce in the docs anymore

snowystinger (approved):

PR #5379, opened 2023-11-07
Bump lightningcss

### LINKED ISSUE #5310: Multiple empty :is() selectors in generated CSS

After creating an empty `react-ts` project with `vite`, installing the latest @adobe/react-spectrum in it, and changing `App.tsx` to this:

```js
import { Button } from '@adobe/react-spectrum'

function App() {
  return <Button variant="accent">Save</Button>;
}

export default App
```

after a production build we get CSS with several empty `:is()` and `:-webkit-any()` selectors like these:

```css
:is() .uFrYBa_spectrum-BarLoader-percentage{
    color:var(--spectrum-barloader-large-over-background-track-fill-color,var(--spectrum-global-color-static-white))
}
:-webkit-any() .uFrYBa_spectrum-BarLoader-track{
    background-color:var(--spectrum-barloader-large-over-background-track-color,var(--spectrum-alias-track-color-over-background))
}
```
Same issue with create-react-app, in fact these are already present in the files distributed in react-spectrum packages, e.g., `@react-spectrum/button/dist/main.css`

### 🤔 Expected Behavior?

No such empty rules are emitted/present.

### 😯 Current Behavior

Rules like that are created which probably cannot match any elements.

### 💁 Possible Solution

_No response_

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

https://codesandbox.io/p/sandbox/wonderful-diffie-w8pfwl

### Version

3.31.1

### If other, please specify.

_No response_

Linux

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
Fixes #5310

Upstream fix: https://github.com/parcel-bundler/lightningcss/commit/b4bbe4726aa82612d6208485f710a151dd96facc

This also bumped the caniuse/mdn compat data. Looks like a bunch of logical properties are no longer compiled as well. Overall compiled + minified CSS size dropped by ~600 KB.

### DIFF
--- a/packages/@adobe/spectrum-css-temp/components/textfield/index.css
+++ b/packages/@adobe/spectrum-css-temp/components/textfield/index.css
@@ -263,7 +263,7 @@ governing permissions and limitations under the License.
 }
 
 /* styles the textfield properly if the left icon is provided */
-.spectrum-Textfield-inputIcon {
+.spectrum-Textfield-input.spectrum-Textfield-inputIcon {
   /* Use padding instead of text-indent so long strings don't overlap the icon */
   /* These values have real names in a spectrum-css update, when we update, use those. */
   padding-inline-start: calc(
--- a/packages/@react-spectrum/text/src/Keyboard.tsx
+++ b/packages/@react-spectrum/text/src/Keyboard.tsx
@@ -26,7 +26,7 @@ function Keyboard(props: KeyboardProps, ref: DOMRef) {
   let domRef = useDOMRef(ref);
 
   return (
-    <kbd {...filterDOMProps(otherProps)} {...styleProps} dir="ltr" ref={domRef}>
+    <kbd {...filterDOMProps(otherProps)} {...styleProps} ref={domRef}>
       {children}
     </kbd>
   );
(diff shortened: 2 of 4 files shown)


packages/@react-spectrum/text/src/Keyboard.tsx
```diff
@@ -26,7 +26,7 @@ function Keyboard(props: KeyboardProps, ref: DOMRef) {
   let domRef = useDOMRef(ref);
 
   return (
-    <kbd {...filterDOMProps(otherProps)} {...styleProps} dir="ltr" ref={domRef}>
+    <kbd {...filterDOMProps(otherProps)} {...styleProps} ref={domRef}>
```
devongovett, line 29: This is because with real logical properties, `dir` will affect which side `padding-inline-start` applies to as well as the text itself. Changed to use `unicode-bidi: plaintext` instead.

packages/@adobe/spectrum-css-temp/components/textfield/index.css
```diff
@@ -263,7 +263,7 @@ governing permissions and limitations under the License.
 }
 
 /* styles the textfield properly if the left icon is provided */
-.spectrum-Textfield-inputIcon {
+.spectrum-Textfield-input.spectrum-Textfield-inputIcon {
```
devongovett, line 266: Specificity has to win over `.spectrum-Textfield-multiLine .spectrum-Textfield-input`, or the `padding` shorthand there will override the `padding-inline-start` here. Didn't see this before because of the extra specificity of the old logical properties compilation.

snowystinger (approved):

LFDanLu (approved):

PR #5392, opened 2023-11-09
Fix Next.js and Remix form examples

devongovett:
Tested the examples locally and found some issues. I used the Zod API incorrectly, Next.js server actions must be async functions, and had the wrong type for remix actions.


snowystinger (approved):

yihuiliao (approved):

devongovett:
TS build won't work until I actually hit publish on main.

PR #5419, opened 2023-11-15
Extract CSS vars from RAC docs into reusable package

devongovett:
This updates the examples in the RAC docs to use a shared set of CSS variables rather than each component defining their own. This will also be published as an example theme on npm so that copy/pasting the example code works without any Spectrum variables, and it's easy to start by modifying the theme/styles.

The base colors were generated with Leonardo to ensure contrast ratios in light and dark mode, with semantic variables defined on top of those. This also allows us to centralize all of the Windows high contrast mode logic into the theme, which reduces a lot of code in the individual docs examples.

The example theme package uses media queries for dark mode, but in our docs we also have a manual theme switcher. We make this work using a small lightningcss plugin to convert the media query to a selector that looks for the `color-scheme: dark` style on the root element in our docs.

Lastly, I included a CSS change to migrate from using box-shadow for focus rings to using outline now that all browsers actually respect the border radius. This also simplifies WHCM because outline renders by default whereas box-shadow does not.


snowystinger (approved):
did a quick sweep in both dark and hcm, looks good so far
lets get this in for testing

reidbarber (approved):

PR #5429, opened 2023-11-16
Add data-trigger to Popover to style based on trigger component

devongovett:
Several of our docs examples have overrides for popovers and things within popovers like listbox that are specific to the triggering component, e.g. combobox or select. These are currently using `@layer` and just overriding for the whole page, but if this CSS were reused on a page with another component containing a popover the overrides would clash.

This introduces a new `trigger` prop to the `Popover` component, which is reflected as the `data-trigger` attribute on the DOM element. This allows style overrides to be scoped based on the trigger component so they don't clash. Normally this would be done by nesting selectors but due to portals that's not possible. The `data-trigger` attribute allows that to be emulated.


reidbarber (approved):
LGTM

PR #5440, opened 2023-11-18
Generate a starter storybook from the docs

devongovett:
This adds a script that generates a starter kit containing all of the "Reusable wrappers" from the RAC docs, along with the corresponding CSS. It also comes with storybook examples extracted from the docs, which also include auto-generated docs and controls. We'll provide a zip download link in the docs for this, which will let people bootstrap a new project really quickly so they can play around with all of the components locally. It'll deploy automatically when we deploy the website to production.


devongovett:
* [Built storybook](https://reactspectrum.blob.core.windows.net/reactspectrum/757681ffe9a86dd15e08012e23648108b48db2e6/verdaccio/starter-storybook/index.html?path=/docs/breadcrumbs--docs)
* [Zip file](https://reactspectrum.blob.core.windows.net/reactspectrum/757681ffe9a86dd15e08012e23648108b48db2e6/verdaccio/docs/react-aria-starter.87104bfcf.zip)

.circleci/comment.js
```diff
@@ -11,7 +11,7 @@ async function run() {
   let pr;
   // If we aren't running on a PR commit, double check if this is a branch created for a fork. If so, we'll need to
   // comment the build link on the fork.
-  if (!process.env.CIRCLE_PULL_REQUEST) {
+  if (true) {
```
devongovett, line 14: will be reverting

snowystinger (approved):
Anything I find can be followup, generally looks really good though

PR #5444, opened 2023-11-20
Add directory support for FileTrigger

ryo-manba:
Closes https://github.com/adobe/react-spectrum/issues/4963

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Verified that only directories are selected when the UNSAFE_directory property is applied.
Added test to verify that the webkitdirectory property is applied.


packages/react-aria-components/src/FileTrigger.tsx
```diff
@@ -62,7 +67,9 @@ function FileTrigger(props: FileTriggerProps, ref: ForwardedRef<HTMLInputElement
         accept={acceptedFileTypes?.toString()}
         onChange={(e) => onSelect?.(e.target.files)}
         capture={defaultCamera}
-        multiple={allowsMultiple} />
+        multiple={allowsMultiple} 
+        // @ts-expect-error
+        webkitdirectory={UNSAFE_directory ? '' : undefined} />
```
ryo-manba, line 72: this implementation uses an empty string for the webkitdirectory attribute instead of a boolean value because it does not function correctly with boolean. 

ref: https://github.com/facebook/react/issues/3468#issuecomment-448336672

packages/react-aria-components/docs/FileTrigger.mdx
```diff
@@ -120,6 +120,20 @@ A file trigger can accept multiple files by passsing the `allowsMultiple` proper
 </FileTrigger>
 ```
 
+## Directory Selection
+
+To enable selecting directories instead of files, use the `UNSAFE_directory` property.
```
reidbarber, line 125: Since it looks like this is well-supported with browsers we support, I think we can just use `directory`.

packages/react-aria-components/docs/FileTrigger.mdx
```diff
@@ -120,6 +120,20 @@ A file trigger can accept multiple files by passsing the `allowsMultiple` proper
 </FileTrigger>
 ```
 
+## Directory Selection
+
+To enable selecting directories instead of files, use the `UNSAFE_directory` property. 
+
+This reflects the webkitdirectory HTML attribute and allows users to select directories and their contents.
```
reidbarber, line 127: Feel free to make `webkitdirectory` link to https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory

reidbarber (changes requested):
Thanks for the PR!

packages/react-aria-components/docs/FileTrigger.mdx
ryo-manba: Fixed by the following commit.

[refac: Remove unsafe prefix](https://github.com/adobe/react-spectrum/pull/5444/commits/10b8984dd21818df134a315c507b010d44184a65)

packages/react-aria-components/docs/FileTrigger.mdx
ryo-manba: Thanks for the comment. Link added.
[refac: Remove unsafe prefix](https://github.com/adobe/react-spectrum/pull/5444/commits/10b8984dd21818df134a315c507b010d44184a65)

yihuiliao (dismissed):
Thanks for the PR!

snowystinger (commented):
GET_BUILD

reidbarber:
GET_BUILD

reidbarber (approved):
LGTM

yihuiliao (approved):

PR #5472, opened 2023-11-23
Add capitalization settings support for useTextField

ryo-manba:
Closes https://github.com/adobe/react-spectrum/issues/4832

For now, I've implemented this only for textField. As mentioned in the issue comments, if this implementation seems fine, I'm thinking of extending it to others as well.

> Seems like a reasonable addition to me. I think if we do this, we should also make this apply to other input fields as well (SearchField, Combobox, SearchAutocomplete, etc.)

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

To test this pull request, please provide the `autoCapitalize` prop in the `useTextField` storybook and observe its reflection in the attribute. For comprehensive testing, it is recommended to try this on iOS Safari to confirm the behavior aligns with expectations.


packages/@react-types/form/src/index.d.ts
```diff
@@ -60,7 +60,7 @@ export interface FormProps extends AriaLabelingProps {
    * A nonstandard attribute used by iOS Safari that controls how textual form elements should be automatically capitalized.
    * See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#autocapitalize).
    */
-  autoCapitalize?: 'off' | 'on',
+  autoCapitalize?: 'off' | 'none' | 'on' | 'sentences' | 'words' | 'characters',
```
ryo-manba, line 63: I changed props type.
but, I'm not confident this is correct and would like to hear opinions on it.

ref: 
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#autocapitalize
- https://html.spec.whatwg.org/multipage/interaction.html#autocapitalization

packages/@react-aria/textfield/src/useTextField.ts
```diff
@@ -79,7 +79,12 @@ export interface AriaTextFieldOptions<T extends TextFieldIntrinsicElements> exte
    * For example, [`type`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-type).
    * @default 'input'
    */
-  inputElementType?: T
+  inputElementType?: T,
+  /**
+   * A nonstandard attribute used by iOS Safari that controls how textual form elements should be automatically capitalized.
+   * See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#autocapitalize).
```
snowystinger, line 85: i did notice this link is wrong, should go here
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#autocapitalize

which I guess would lead this to accepting the type
`'none' | 'sentences' | 'words' | 'characters'`
instead of what form accepted

ryo-manba: @snowystinger 
The link was wrong and will be corrected.
However, regarding type, according to the [HTML Standard](https://html.spec.whatwg.org/multipage/interaction.html#autocapitalization), I feel it is correct to accept on and off as keywords as well, what do you think?

image

snowystinger: It's probably worth it to do some testing to see which ones are actually supported. The spec is a good guideline, and if we find gaps in coverage, we perhaps could map the missing values.
Would you mind getting us started on that?

ryo-manba: Sure thing, I'll get started on the testing and report back with the results.

ryo-manba: @snowystinger 
I tested autocapitalize on iOS Safari 16.6 using the link below and found that the 'off' setting behaves identically to 'none', and 'on' mirrors 'sentences'.
https://codepen.io/ryo-manba/pen/rNPrwWM

Based on these findings, it appears that our current PR approach is accurate.

snowystinger: Thanks! I checked iOS 17.1, and my coworker checked on an Android with the same results
appears Mozilla might be a little outdated on usage

snowystinger (dismissed):

yihuiliao:
I hope you don't mind me adding an extra story for easier testing. Otherwise, thanks for the PR!

yihuiliao:
GET_BUILD

snowystinger (approved):

yihuiliao (approved):

ryo-manba:
I wanted to share that I've submitted a correction to MDN, and it has been successfully merged.
https://github.com/mdn/content/pull/30645.

PR #5513, opened 2023-12-02
Fix user's ability to scroll menu with sections

LFDanLu:
Closes https://github.com/orgs/adobe/projects/19/views/4?filterQuery=scroll&pane=issue&itemId=45329096

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test that menus with sections are scrollable on iPad. Also test that other aspects of prevent scroll behavior still work as expected. In particular, try the top level RAC docs examples

## 🧢 Your Project:

RSP


packages/@react-aria/overlays/src/usePreventScroll.ts
LFDanLu, line 1: alternatively, a similar change could be made here: https://github.com/adobe/react-spectrum/blob/ee403ec3ba5dba7c66903e3454b8f510c628a8f9/packages/%40react-aria/utils/src/getScrollParent.ts#L25-L28 where we only return true if the node has any overflow. Not sure of the ramifications yet if applied there though since that func is indirectly used in many more places...

snowystinger (commented):
It's more expensive to make the change directly in getScrollParent. the change as is makes sense. maybe we do it here and see where else this would be useful?

LFDanLu:
I'm fine going either way, but IMO since getScrollParent's new flow only happens if you pass `checkForOverflow` + it only happens here in `usePreventScroll` on iOS on touch start, it shouldn't be that much more expensive.

snowystinger (approved):

snowystinger (commented):
Couldn't test on ipad, but looks like it's working in iPhone and Desktop Safari

PR #5525, opened 2023-12-05
Normalize press handling for Checkbox, Radio, and Switch labels

devongovett:
Built this animation for the new landing page to show how our press interactions work, but then realized that it didn't work on iOS. Seems like Safari doesn't trigger a click event if you drag off and then back onto the label like it works on desktop.

https://github.com/adobe/react-spectrum/assets/19409/9d6d8230-2528-4755-ba51-ab30ee00ebdc

This adds `labelProps` as a return value to `useCheckbox`, `useRadio`, and `useSwitch`, which uses `usePress` to toggle the value and prevents the native `onClick`. This has the side benefit of allowing us to remove the press handling from RAC, which we use for the visible press state. This is not a breaking change: if `labelProps` is unused, it will have no effect.

### DIFF
--- a/packages/react-aria-components/src/Checkbox.tsx
+++ b/packages/react-aria-components/src/Checkbox.tsx
@@ -9,13 +9,13 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-import {AriaCheckboxGroupProps, AriaCheckboxProps, HoverEvents, mergeProps, useCheckbox, useCheckboxGroup, useCheckboxGroupItem, useFocusRing, useHover, usePress, VisuallyHidden} from 'react-aria';
+import {AriaCheckboxGroupProps, AriaCheckboxProps, HoverEvents, mergeProps, useCheckbox, useCheckboxGroup, useCheckboxGroupItem, useFocusRing, useHover, VisuallyHidden} from 'react-aria';
 import {CheckboxGroupState, useCheckboxGroupState, useToggleState} from 'react-stately';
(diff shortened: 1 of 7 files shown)


snowystinger (dismissed):
LGTM, also, i think you might've solved https://github.com/adobe/react-spectrum/issues/3675

packages/react-aria-components/src/Checkbox.tsx
LFDanLu, line 1: shouldn't we also make the Spectrum components use the labelProps returned by the hooks as well? Probably should update the hooks docs as well to do the same, happy to help with these if you'd like

LFDanLu (changes requested):

packages/react-aria-components/src/Checkbox.tsx
devongovett: yeah but I'm on a tight schedule and not sure I'll have time. mainly wanted this because it was very visible for this demo. seems fine to add to the backlog for RSP if we don't get to it, but if you have time feel free!

LFDanLu: makes sense, I'll see about getting something out for it

LFDanLu (dismissed):

packages/react-aria-components/src/Checkbox.tsx
LFDanLu: Ah I see, it isn't as trivial as using the labelProps since RSP/hooks example have an actual pressable input (and thus block the press + release on label) element vs RAC which hides it and relies on the label for pointer interactions. I'll file something

PR #5529, opened 2023-12-05
Fix ActionGroup icon padding when overflow="collapse" and buttonLabelBehavior="hide"

LFDanLu:
Closes https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=46694245

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Try the actiongroup stories with overflow="collapse" and buttonLabelBehavior="hide"

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@adobe/spectrum-css-temp/components/actiongroup/index.css
+++ b/packages/@adobe/spectrum-css-temp/components/actiongroup/index.css
@@ -54,7 +54,7 @@ governing permissions and limitations under the License.
       padding-inline-end: var(--spectrum-actionbutton-icon-padding-x);
     }
 
-    .spectrum-ActionGroup-item-icon {
+    .spectrum-ActionGroup-itemIcon {
       padding-inline-end: 0;
     }
   }
--- a/packages/@adobe/spectrum-css-temp/components/actiongroup/index.css
+++ b/packages/@adobe/spectrum-css-temp/components/actiongroup/index.css
@@ -49,10 +49,14 @@ governing permissions and limitations under the License.
 
   .spectrum-ActionGroup-item {
     flex-shrink: 0;
-  }
 
-  .spectrum-ActionGroup-item--iconOnly {
-    padding-inline-end: var(--spectrum-actionbutton-icon-padding-x);
+    &.spectrum-ActionGroup-item--iconOnly {
+      padding-inline-end: var(--spectrum-actionbutton-icon-padding-x);
+    }
+
+    .spectrum-ActionGroup-item-icon {
+      padding-inline-end: 0;
+    }
   }
 
   &:focus {
--- a/packages/@adobe/spectrum-css-temp/components/actiongroup/index.css
+++ b/packages/@adobe/spectrum-css-temp/components/actiongroup/index.css
@@ -51,6 +51,10 @@ governing permissions and limitations under the License.
     flex-shrink: 0;
   }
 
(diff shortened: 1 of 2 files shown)


packages/@adobe/spectrum-css-temp/components/actiongroup/index.css
LFDanLu, line 1: Note: specificity issue doesn't happen locally or in Chromatic hence why this wasn't caught earlier. The change bumps up the specificity up 1 so it always is prioritized over ActionButton's. 

As an aside, you may notice the following bug:
image

The bottom behavior is for ActionGroups with icons only, looks to be the behavior since at least 20 minor versions ago: https://codesandbox.io/s/broken-shadow-w7y3gx?file=/src/App.js:596-622. Think it shouldn't actually get this iconOnly style...

LFDanLu: EDIT: fixed the 2nd issue too

reidbarber (dismissed):
LGTM, both issues look fixed

packages/@adobe/spectrum-css-temp/components/actiongroup/index.css
snowystinger: technically you've bumped the specificity up by 2 ;)

LFDanLu: haha, originally it was one level up, but moved it in some more with the fix for the 2nd issue :P

packages/@react-spectrum/actiongroup/src/ActionGroup.tsx
```diff
@@ -308,6 +308,9 @@ function ActionGroupItem<T>({item, state, isDisabled, isEmphasized, staticColor,
             text: {
               id: hideButtonText ? textId : null,
               isHidden: hideButtonText
+            },
+            icon: {
+              UNSAFE_className: hideButtonText ? classNames(styles, 'spectrum-ActionGroup-item-icon') : null
```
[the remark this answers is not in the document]
LFDanLu: the conditional is needed because the Icon has some padding already applied for the icon only ActionButton case:
image
which we don't want to override with `spectrum-ActionGroup-itemIcon` unless our `spectrum-ActionGroup-item--iconOnly` is handling applying this padding already

Here are the cases:
1. We are in a default ActionGroup case. `spectrum-ActionGroup-item--iconOnly` doesn't get applied anywhere because we aren't in the "buttonLabelBehavior=hide" case. All ActionButtons within the ActionGroup have ActionButton specific classes that already handle creating the padding on the left/right to make the button properly sized, regardless if the ActionButton with or without accompanying text. We don't want to apply our `spectrum-ActionGroup-itemIcon' class either since everything is handled already.
 
2. We are in a ActionGroup with "buttonLabelBehavior=hide" case. `spectrum-ActionGroup-item--iconOnly` is applied, handling the `padding-inline-end` for icon only and icon + text ActionButtons. Since we are handling it there, we need to override the ActionButton's icon only styling shown at the very top so we don't double dip on `padding-inline-end`

I'll go ahead and change the className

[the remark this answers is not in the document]
snowystinger: ooo, and useHasChild doesn't update because we're not removing the text, just hiding it
that's annoying

PR #5533, opened 2023-12-06
improving the docs breadcrumb example links

ktabors:
just fix it

These links are mostly all valid depending on if you are on a local build or production, this just makes them more consistent and less likely to cause issues. 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

confirm these links are valid by checking them all

Currently if you right click on the RSP Links example "Breadcrumbs" which is `<Item href="/react-spectrum/Breadcrumbs">Breadcrumbs</Item>` and open it in a new tab/window on the production, it gives a 404 error. This will be fixed.

## 🧢 Your Project:
RSP


ktabors:
> > This will be fixed.
> 
> as in in a future PR or future commit? because I checked the useBreadcrumb example, and I still see the 404 blob not found

Production will be fixed by this.

Locally I get a forward not a 404 like production. Since the behavior is different in these two places confirming is an assertion not a proof. I didn't investigate this issue on build docs. I see that on docs builds they break because they don't include the commit part of the URL.

LFDanLu (approved):

PR #5538, opened 2023-12-06
Prevent flicker from all items to filtered collection when combobox is animating closed

devongovett:
Issue can be seen in the docs or storybook. If you select an item, and then click the button to open the listbox repeatedly, the collection changes while the menu is animating closed causing a flicker. It can even cause the popover to flip depending on the scenario. Original fix was in #4060 but this extends it to use the original collection rather than the filtered collection if `showAllItems` is true.


snowystinger (approved):
Tested in a variety of controlled/uncontrolled comboboxes, this seems to make sense

ktabors (approved):
The flickering is gone. If you click really fast on the button to open the popover, it currently decides you want the Combobox closed. The old behavior had a lot more open and close and weird flickering. This is better.

PR #5541, opened 2023-12-07
Fix `tailwindcss-react-aria-components` has incorrect `import` instead of `require`

tom-fletcher:
## ✅ Pull Request Checklist:

- n/a Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- n/a Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- n/a Filled out test instructions.
- n/a Updated documentation (if it already exists for this component).
- n/a Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/TR/wai-aria-practices-1.1/)

## 📝 Test Instructions:

n/a

## 🧢 Your Project:

The `tailwindcss-react-aria-components` plugin uses an ESM `import` in a CJS file that uses`module.exports`. While it looks like most bundlers do just ignore this, it can cause issues with webpack in certain scenarios - for instance in the following example:

- Using `Storybook` with the `Next.js` framework setting (which uses webpack)
- Importing the `tailwind.config.ts` file into the `Storybook` `preview.ts`, for example to parse the theme colors and use these for `Storybook` backgrounds
- This will cause the following error when attempting to view any components in `Storybook`:

```
ES Modules may not assign module.exports or exports.*, Use ESM export syntax, instead: ../../../node_modules/.pnpm/tailwindcss-react-aria-components@1.0.0-rc.0_tailwindcss@3.3.5/node_modules/tailwindcss-react-aria-components/src/index.js
```

The error message points out that two different module systems are trying to be used at the same time. This PR just corrects the `import` to a `require`, which makes everything CJS and removes the incompatibility. This also brings the plugin in line with the [Tailwind plugin documentation](https://tailwindcss.com/docs/plugins#static-utilities).


reidbarber:
GET_BUILD

reidbarber (approved):

snowystinger (commented):
Just making sure, this isn't that weird bug that @devongovett found with the testing of the string 'module' anywhere in the file is it?

tom-fletcher:
> Just making sure, this isn't that weird bug that @devongovett found with the testing of the string 'module' anywhere in the file is it?
> 
> [vercel/next.js#57649](https://github.com/vercel/next.js/issues/57649)

I don't believe this is related - that issue looks suspiciously similar to [this one](https://github.com/antfu/shikiji/issues/13#issuecomment-1749588964) which was caused by `Next.js` not loading `.mjs` files (and apparently fixed in `Next v14.0.4` which just released).

ktabors (approved):

reidbarber (approved):

kylemh:
hey @tom-fletcher didn't want to open an issue for it because it's just a question, but I'm curious how I'm supposed to use the plugin with a prefix if my TS config is a TypeScript file. I tried:

```ts
const config = {
  plugins: [
    (pluginAPI) => reactAriaComponentClasses({ prefix: 'rac' })(pluginAPI),
  ],
}
```

```ts
const config = {
  plugins: [
    plugin((pluginArgs) => reactAriaComponentClasses({ prefix: 'rac' })(pluginArgs)),
  ],
}
```

but both yield type and build errors.

PR #5565, opened 2023-12-12
Escape strings used in query selectors

reidbarber:
[Escapes](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static) our strings used in query selectors. Keys may be strings that are invalid as selectors, so we want to avoid errors like: `"Failed to execute 'querySelector' on 'Element': '[data-key="Apple""]' is not a valid selector."`

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Smoke test affected components.


snowystinger (approved):
Looks good.

At some point we should really write a test which can fuzz the inputs. We could use the same fast check which we used for NumberField. https://github.com/adobe/react-spectrum/blob/main/packages/%40internationalized/number/test/NumberParser.test.js#L168 and instead of round trips, we could just check for a couple basic interactions such as keyboard navigate and select.

ktabors (approved):

PR #5604, opened 2023-12-20
Fix prod/verdaccio docs build

devongovett:
Some missing dependencies for the new homepage. I think this is everything but I didn't run verdaccio locally.


LFDanLu (approved):
🤞

snowystinger (approved):

PR #5622, opened 2023-12-22
#5621 [a11y] ResizableTableContainer: focused row obscured by table header with position: sticky

### LINKED ISSUE #5621: [a11y] ResizableTableContainer: focused row obscured by table header with position: sticky

When a table within a ResizableTableContainer is scrolled, and its `thead` has `position: sticky`, when a row receives focus, it can be obscured by the sticky header.

### 🤔 Expected Behavior?

Focused elements should not be obscured by other content. See https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum

### 😯 Current Behavior

The RAC table example has a fixed header that can obscure focused content in the row, when the table is scrolled.

### 💁 Possible Solution

Use [`scroll-padding`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-top) to fix the issue with CSS, or with JavaScript, add a listener as follows:
```ts
// Ensure that the row containing focus is not obscured by a thead with position: sticky.
  const onFocusCapture = useCallback((e: React.FocusEvent<HTMLDivElement>) => {
    const target = e.target as HTMLElement;
    const {scrollTop, scrollLeft} = e.currentTarget;
    const row = target.tagName === 'TR' || target.getAttribute('role') === 'row' ? target : target.closest('tr, [role="row"]') as HTMLElement;
    const thead = target.closest('table, [role="grid"], [role="table"]').querySelector('thead, [role="rowgroup"]') as HTMLElement;
    const top = row.offsetTop - thead.offsetHeight;
    if (
      scrollTop > top &&
      getComputedStyle(thead).position === 'sticky'
    ) {
      e.currentTarget.scrollTo({
        top,
        left: scrollLeft,
        behavior: 'smooth'
      });
    }
  }, []);
```
to the `ResizableTableContainer` at https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/Table.tsx#L211-L245

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

Open https://react-spectrum.adobe.com/react-aria/

1. Navigate to the Example App
2. Scroll the Table 
3. Tab or use arrow keys to focus rows or cells within the table
4. Notice that focused content can become obscured by the sticky thead.

https://github.com/adobe/react-spectrum/assets/154077/67310457-05ef-4d4c-b641-b1e5fab5b1d3

### Version

RAC 1.0

### If other, please specify.

_No response_

macOS

### 🧢 Your Company/Team

Adobe/Accessibility

### 🕷 Tracking Issue

_No response_

majornista:
Note that this commit includes both Tailwind CSS and Javascript solutions to the problem.

Closes #5621

## ✅ Pull Request Checklist:

- [X] Included link to corresponding [React Spectrum GitHub Issue 5621](https://github.com/adobe/react-spectrum/issue/5621).

## 🧢 Your Project:

Adobe/Accessibility


ktabors (approved):
Behaves well in Chrome on Desktop

devongovett:
Do we need both solutions? I thought the CSS one would be enough. Or were you just demonstrating?

packages/react-aria-components/src/Table.tsx
```diff
@@ -230,13 +230,62 @@ function ResizableTableContainer(props: ResizableTableContainerProps, ref: Forwa
...
+          }
+        }
+      });
+    }
+    return () => {
+      if (requestRef.current !== undefined) {
+        cancelAnimationFrame(requestRef.current);
+      }
+    };
+  }, [objectRef, isSticky, scrollPaddingTop]);
```
majornista, line 260: > Do we need both solutions? I thought the CSS one would be enough. Or were you just demonstrating?
I think the javascript solution, or css added with javascript as done here, would be more robust. With users bringing their own CSS, I'm not sure we can count on them adding `scroll-padding` to the container when they add `sticky` to the `thead`, otherwise we would never have discovered this bug. We may not need the `onFocusCapture` fallback with adequate browser support for `scroll-padding`.

majornista: Note: I removed `onFocusCapture` fallback.

devongovett: Didn't realize this was in the component itself. I was thinking of just solving the immediate issue in the example. I'm not sure what the side effects of us adding scroll-padding automatically would be. If we did do that we'd probably want to do it somewhere deeper as this would affect more than just table (e.g. if you have sticky section headers in a menu or another collection).

majornista: I can remove the change to the component itself and just make the change in the documentation example, but my concern about the potential for error with sticky still stands.

LFDanLu (approved):

ktabors (approved):

devongovett (approved):

Issue #5609, opened 2023-12-20
Ability to conditionally select if Table supports DnD


jaknas:
### Provide a general summary of the issue here

I'd like to decide when Table component supports DnD or not. Let's say that in my app I have a feature like `editMode`, which when enabled should turn on DnD feature for all Tables on the page.

Currently, I have to reach out to conditionally mount 2 different tables, one with `dragAndDropHooks` and other without. This causes some performance issues since I basically tear down whole table and re-create it.

### 🤔 Expected Behavior?

I'd like to be able to conditonally pass if a table supports DnD

### 😯 Current Behavior

If I remove `dragAndDropHooks`, Table errors because hooks cannot be called conditionally. 

### 💁 Possible Solution

I would be OK with always passing `dragAndDropHooks`, but have a way to disable dragging and dropping rows for the user. 

### 🖥️ Steps to Reproduce

Will add later, but hopefully you know what I mean.

### Version

1.0.0-rc.0

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### What operating system are you using?

MacOS Sonoma

LFDanLu:
I think this should be possible if we introduce a `isDisabled` prop to `dragAndDropHooks`. This would also require some changes to `useDraggableItem` so accept `isDisabled` from the user so that it doesn't just look at the selectionManager to determine if a item is disabled for dragging. For the drop side of things, we could probably just override the `acceptedDropTypes` to return an empty array. 

Depending on your current use case you might be able implement this behavior already by conditionally disabling all keys in your Table and modifying the `acceptedDropTypes` provided to the `useDragAndDrop` hook when you want to disable DnD for your table. Not 100% if there would be any bugs with that though

PR #5641, opened 2023-12-29
Add conditional DnD support to Table component

ryo-manba:
Closes https://github.com/adobe/react-spectrum/issues/5609

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Go to http://localhost:9003/?path=/story/react-aria-components--dnd-table-example&args=isDisabledFirstTable:true;isDisabledSecondTable:true&providerSwitcher-express=false&strict=true
2. Open the Controls tab in Storybook and set isDisabledFirstTable to true.
3. Verify that the table on the left side cannot perform drag and drop actions.


packages/@react-aria/dnd/src/useDraggableItem.ts
```diff
@@ -65,7 +67,7 @@ const MESSAGES = {
  */
 export function useDraggableItem(props: DraggableItemProps, state: DraggableCollectionState): DraggableItemResult {
   let stringFormatter = useLocalizedStringFormatter(intlMessages, '@react-aria/dnd');
-  let isDisabled = state.selectionManager.isDisabled(props.key);
+  let isDisabled = props.isDisabled || state.selectionManager.isDisabled(props.key);
```
LFDanLu, line 70: Discussed with the team some more about this, and it might be better if we disabled all the draggable items via the state. The reason for this is because by adding this prop someone could disable any arbitrary draggable item, resulting in cases where a user might select multiple items (including the disabled draggable items) -> drag the items -> wonder why some items aren't being dragged. Disabling dragging still makes sense in the case where ALL items should have their draggability disabled though

ryo-manba: Thank you for your comment. You're right, there might indeed be a chance of users accidentally misoperating. I'll try to see if it can be fixed.

packages/@react-stately/dnd/src/useDraggableCollectionState.ts
```diff
@@ -62,6 +65,9 @@ export function useDraggableCollectionState(props: DraggableCollectionStateOptio
     preview,
     getAllowedDropOperations
   } = props;
+  if (isDisabled) {
+    selectionManager.isDisabled = () => true;
+  }
```
LFDanLu, line 70: Noticed that doing this makes it so clicking on the Table row itself doesn't cause the row to be selected if drag and drop is disabled. Perhaps useDraggableState should just return `isDisabled` and useDraggableItem should use `state.isDisabled ||  state.selectionManager.isDisabled(props.key)` as its `isDisabled` check instead? Will think about this some more and ask the team for opinions.

ryo-manba: I'm sorry for the late response. 
I've made some adjustments to the functionality and added new tests!

- [modified the condition of disabled flag](https://github.com/adobe/react-spectrum/pull/5641/commits/123aaeb048694d93ab8da660d541e04d2803999b)
- [Add test for DraggableTable with selection](https://github.com/adobe/react-spectrum/pull/5641/commits/08bcdacb63d7c6a04c62a5f2655198811609b062)

snowystinger:
GET_BUILD

packages/@react-stately/dnd/src/useDraggableCollectionState.ts
```diff
@@ -30,6 +32,8 @@ export interface DraggableCollectionState {
   draggedKey: Key | null,
   /** The keys of the items that are currently being dragged. */
   draggingKeys: Set<Key>,
+  /** Whether drag events are disabled. */
+  isDisabled?: boolean,
```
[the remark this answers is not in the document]
ryo-manba: I think just returning cancel with defaultGetDropOperation should be fine.
I've also updated the Storybook to match. Thanks for your thoughtful comments!

[Add isDisabled prop support to useDroppableCollectionState](https://github.com/adobe/react-spectrum/pull/5641/commits/b5b279f9d51067db710626215e0612ecc224bacb)

packages/react-aria-components/src/Table.tsx
```diff
@@ -416,7 +416,7 @@ function Table(props: TableProps, ref: ForwardedRef<HTMLTableElement>) {
             ref={ref}
             slot={props.slot || undefined}
             onScroll={props.onScroll}
-            data-allows-dragging={isListDraggable || undefined}
+            data-allows-dragging={(isListDraggable && !dragAndDropHooks?.isDisabled) || undefined}
```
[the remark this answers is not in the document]
ryo-manba: oh, this was not well. I've  fix it! 
[Refactor draggable list naming for clarity](https://github.com/adobe/react-spectrum/pull/5641/commits/9433c63630b900651cb8aa0bba72b033373bffe6)

packages/react-aria-components/src/useDragAndDrop.tsx
```diff
@@ -63,7 +63,13 @@ interface DropHooks {
...
 }
 
-export type DragAndDropHooks = DragHooks & DropHooks;
+export type DragAndDropHooks = DragHooks & DropHooks & {
+  /**
+   * Determines whether drag and drop functionality is disabled for the collection.
+   * If set to true, both dragging items out of and dropping items into the collection are disabled.
+   */
+  isDisabled?: boolean
+};
```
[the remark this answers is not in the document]
ryo-manba: Sure, an option to disable drag-only or drop-only might be useful. However, since this PR is getting bigger, it might be more appropriate to address it separately. what do you think?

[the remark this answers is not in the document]
LFDanLu: I agree, we can handle it separately. I will file a separate issue if/when this PR goes in!

packages/react-aria-components/src/Table.tsx
```diff
@@ -373,13 +373,16 @@ function Table(props: TableProps, ref: ForwardedRef<HTMLTableElement>) {
...
+  let isListDroppable = !dragAndDropHooks?.isDisabled && hasDropHooks;
+
   let {selectionBehavior, selectionMode, disallowEmptySelection} = state.selectionManager;
   let ctx = useMemo(() => ({
     selectionBehavior: selectionMode === 'none' ? null : selectionBehavior,
     selectionMode,
     disallowEmptySelection,
-    allowsDragging: isListDraggable
-  }), [selectionBehavior, selectionMode, disallowEmptySelection, isListDraggable]);
+    allowsDragging: hasDragHooks
```
LFDanLu, line 384: Perhaps `allowsDragging` strange name now that `hasDragHooks` doesn't necessarily mean that the table allows dragging depending if `isDisabled` is true but I think we are stuck with the name or it would be a breaking change. Might be useful to have another value on the context here called `isDragDisabled` perhaps?

Commenting so others on the team can weigh in, no need to implement in this PR as of now until we decide on the api here.

packages/react-aria-components/src/Table.tsx
```diff
@@ -373,13 +373,16 @@ function Table(props: TableProps, ref: ForwardedRef<HTMLTableElement>) {
     }
   });
 
+  let isListDraggable = !dragAndDropHooks?.isDisabled && hasDragHooks;
+  let isListDroppable = !dragAndDropHooks?.isDisabled && hasDropHooks;
```
LFDanLu, line 377: Perhaps this should be 
```suggestion
  let isListDraggable = hasDragHooks && dragState.isDisabled;
  let isListDroppable = hasDragHooks && dropState.isDisabled;
```
That way we use the `isDisabled` set on the state objects themselves and we can have `useDragAndDrop` only return the hooks still?

packages/react-aria-components/src/useDragAndDrop.tsx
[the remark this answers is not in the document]
LFDanLu: Ah actually, I think we can get rid of this `isDisabled` from the return type since the draggable/droppable collection state objects should already have `isDisabled` as a property already

packages/react-aria-components/src/useDragAndDrop.tsx
```diff
@@ -105,13 +113,18 @@ export function useDragAndDrop(options: DragAndDropOptions): DragAndDrop {
...
       renderDropIndicator,
-      dropTargetDelegate
-     } = options;
+      dropTargetDelegate,
+      isDisabled
+    } = options;
+
+    if (isDisabled) {
+      options.acceptedDragTypes = [];
+    }
```
LFDanLu, line 122: We can get rid of this now that useDroppableCollectionState handles returning 'cancel` in the `defaultDropOperation` for us when `isDisabled = true` right?

packages/react-aria-components/src/useDragAndDrop.tsx
```diff
@@ -135,6 +148,7 @@ export function useDragAndDrop(options: DragAndDropOptions): DragAndDrop {
       hooks.dropTargetDelegate = dropTargetDelegate;
       hooks.ListDropTargetDelegate = ListDropTargetDelegate;
     }
+    hooks.isDisabled = isDisabled;
```
LFDanLu, line 151: Can get rid of this as well if we rely on the state objects to tell us if drag or drop is disabled

packages/@react-stately/dnd/src/useDroppableCollectionState.ts
```diff
@@ -101,7 +110,7 @@ export function useDroppableCollectionState(props: DroppableCollectionStateOptio
     }
 
     return 'cancel';
-  }, [acceptedDragTypes, getDropOperation, onInsert, onRootDrop, onItemDrop, shouldAcceptItemDrop, onReorder, onDrop]);
+  }, [isDisabled, acceptedDragTypes, getDropOperation, onInsert, onRootDrop, onItemDrop, shouldAcceptItemDrop, onReorder, onDrop]);
 
   return {
```
LFDanLu, line 115: We should return `isDisabled` in this return object too for consistency with `useDraggableCollection` state I think.

LFDanLu (changes requested):
Some small changes for consistency, let me know if they make sense! Hopefully the last batch of changes and then we can see about getting some more eyes on this.

LFDanLu (approved):
LGTM, thanks for the updates!

LFDanLu:
GET_BUILD

packages/react-aria-components/src/Table.tsx
snowystinger: I think isDragDisabled would be useful. It might come up when we implement edit mode as well.

LFDanLu: Yeah I imagine it would be common api across several components. We can tackle that later as a followup IMO

reidbarber:
GET_BUILD

reidbarber (approved):
LGTM

packages/react-aria-components/src/Table.tsx
pnicolli: +1 for having a way of knowing whether the drag&drop is disabled. I am currently building a table that has drag&drop disabled in certain situations and I would really like to only render the drag button when enabled, or at least properly disable the drag button. Is there some other way to achieve this?

LFDanLu: Mind sharing your setup (perhaps in a github discussion for greater visibility)? I think ideally we'd add `isDragDisabled` or something along those lines to the table options context, but for the time being you can probably just create your own context wrapper around your table implementation that informs your Row if dragging is disabled

PR #5646, opened 2024-01-01
allow passing readonly array to LocalesPlugin

stefanprobst:
this pr updates types to allow passing a readonly array to the optimize-locales plugin options, e.g.:

```ts
const locales = ["de", "en"] as const
// ...
optimizeLocales.webpack({ locales })
```


snowystinger (approved):
Looks good to me, thanks

reidbarber (approved):

PR #5648, opened 2024-01-03
RAC: Submenu support

reidbarber:
Implements submenus in RAC via the SubmenuTrigger component.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test new stories and docs.

## 🧢 Your Project:

RSP


majornista:
I'm having some issues when navigating back to the previous menu from a submenu with VoiceOver on iOS within the Storybook iframe.

1. Open https://reactspectrum.blob.core.windows.net/reactspectrum/a11f5882bc1afce7ad4e08ea7d5a8894e3e91089/storybook/index.html?path=/story/menutrigger-submenu--submenu-static on iOS in portrait mode with VoiceOver running.
2. Navigate to the Menu Button and double tap to open the menu in a tray.
3. Focus appears on the first menu item, Lvl 1, Item 1.
4. Swipe right to navigate to the next, collapsed menu item Lvl 1, Item 2.
5. Double-tap to expand the submenu, "selected,  Lvl 1, Item 2, Menu Item, Menu Pop up, Expanded"
6. Swipe right to navigate to the next, collapsed menu item Lvl 2, Item 3.
7. **Problem 1:** Focus first goes to a Dismiss button that seems to be at the top of the tray,  which would be fine, except if you swipe right again to navigate forward, focus seems to announce Storybook's Actions menu, which is collapsed in this view, and then immediately restores focus to the same Dismiss button. This seems to be because, when the contents of the tray are replaced, Safari/VoiceOver loses focus and focus winds up at the end of the menu rather than at the beginning. If I swipe left to navigate backwards, I can reach the collapsed menu item Lvl 2, Item 3, but it takes some trial and error to get the lay of the land when the tray content is replaced by the submenu content.
8. With focus now on Lvl 2, Item 3, double tap to expand the submenu.
9. With **Problem 1** in mind, swipe left to navigate backward to focus the  "Return to Lvl 2 Item 3" button at the top of the tray.
10. Double tap to close the Lvl 3 submenu and return to Lvl 2 Item 3.
11. **Problem 2:** The menu closes entirely. This seems to happen within the Storybook iframe, but may happen elsewhere.

If you repeat steps 1-10 using the Storybook example alone: https://reactspectrum.blob.core.windows.net/reactspectrum/a11f5882bc1afce7ad4e08ea7d5a8894e3e91089/storybook/iframe.html?scrolling=false&providerSwitcher-locale=&providerSwitcher-theme=&providerSwitcher-scale=&providerSwitcher-express=false&id=menutrigger-submenu--submenu-static&viewMode=story, it is possible to navigate to the previous submenu, but rather than focus the previous menu item in the submenu, focus always seems to go to the first focusable descendant, either the "Return to" button or the first menu item.

majornista:
> I'm having some issues when navigating back to the previous menu from a submenu with VoiceOver on iOS within the Storybook iframe.
> 
> 1. Open https://reactspectrum.blob.core.windows.net/reactspectrum/a11f5882bc1afce7ad4e08ea7d5a8894e3e91089/storybook/index.html?path=/story/menutrigger-submenu--submenu-static on iOS in portrait mode with VoiceOver running.
> 2. Navigate to the Menu Button and double tap to open the menu in a tray.
> 3. Focus appears on the first menu item, Lvl 1, Item 1.
> 4. Swipe right to navigate to the next, collapsed menu item Lvl 1, Item 2.
> 5. Double-tap to expand the submenu, "selected,  Lvl 1, Item 2, Menu Item, Menu Pop up, Expanded"
> 6. Swipe right to navigate to the next, collapsed menu item Lvl 2, Item 3.
> 7. **Problem 1:** Focus first goes to a Dismiss button that seems to be at the top of the tray,  which would be fine, except if you swipe right again to navigate forward, focus seems to announce Storybook's Actions menu, which is collapsed in this view, and then immediately restores focus to the same Dismiss button. This seems to be because, when the contents of the tray are replaced, Safari/VoiceOver loses focus and focus winds up at the end of the menu rather than at the beginning. If I swipe left to navigate backwards, I can reach the collapsed menu item Lvl 2, Item 3, but it takes some trial and error to get the lay of the land when the tray content is replaced by the submenu content.
> 8. With focus now on Lvl 2, Item 3, double tap to expand the submenu.
> 9. With **Problem 1** in mind, swipe left to navigate backward to focus the  "Return to Lvl 2 Item 3" button at the top of the tray.
> 10. Double tap to close the Lvl 3 submenu and return to Lvl 2 Item 3.
> 11. **Problem 2:** The menu closes entirely. This seems to happen within the Storybook iframe, but may happen elsewhere.
> 
> If you repeat steps 1-10 using the Storybook example alone: https://reactspectrum.blob.core.windows.net/reactspectrum/a11f5882bc1afce7ad4e08ea7d5a8894e3e91089/storybook/iframe.html?scrolling=false&providerSwitcher-locale=&providerSwitcher-theme=&providerSwitcher-scale=&providerSwitcher-express=false&id=menutrigger-submenu--submenu-static&viewMode=story, it is possible to navigate to the previous submenu, but rather than focus the previous menu item in the submenu, focus always seems to go to the first focusable descendant, either the "Return to" button or the first menu item.

Just realized these issues relate to the React-Spectrum version of submenus, but I'm seeing similar instability in the RAC examples, with submenus, or the entire menu, closing unexpectedly, and in the Submenus example here: https://reactspectrum.blob.core.windows.net/reactspectrum/a11f5882bc1afce7ad4e08ea7d5a8894e3e91089/docs/react-aria/Menu.html#submenus, not being able to access menu items after the first one in the expanded submenu.

reidbarber:
@majornista Thanks for finding these. We do have a ticket for RSP Submenu follow-up bugs here: https://github.com/orgs/adobe/projects/19/views/4?filterQuery=submenu&pane=issue&itemId=39054342

And it does look like the iOS VO behavior does need more attention. I suppose we didn't see some of this due to the RSP Submenu using a tray experience.

packages/react-aria-components/src/Menu.tsx
```diff
@@ -112,29 +161,35 @@ function MenuInner<T extends object>({props, collection, menuRef: ref}: MenuInne
...
+        <Provider
+          values={[
+            [MenuStateContext, state],
+            [SeparatorContext, {elementType: 'div'}],
+            [SubmenuContext, {popoverContainerRef}]
+          ]}>
+          {children}
+        </Provider>
+      </div>
+      {rootMenuTriggerState?.isOpen && <div ref={popoverContainerRef} style={{width: '100vw', position: 'absolute', top: -5, ...leftOffset}} /> }
```
LFDanLu, line 191: This `-5` for top was added to Spectrum to help with some of the offsetting/positioning of the submenu, but for RAC do we want to support the end user specifying some kind of offset at a higher level? I think the Popover allows for specifying the offsets but it might still be bounded by the positioning of this div... Might be hard, not sure

packages/react-aria-components/src/Menu.tsx
```diff
@@ -268,3 +328,86 @@ function MenuItemInner<T>({item}: MenuItemInnerProps<T>) {
...
+  let {isFocusVisible, focusProps} = useFocusRing();
+  let isSubmenuOpen = submenuTriggerState.isOpen;
+  let renderProps = useRenderProps({
+    ...props,
+    id: undefined,
+    children: item.rendered,
+    defaultClassName: 'react-aria-MenuItem',
+    values: {
+      ...states,
+      isHovered: states.isFocused,
```
LFDanLu, line 364: I forget, do we want isHovered to be true on keyboard focus? I have a faint recollection that we changed this before the break for some other components but that there was some complications. I see that we have the same for MenuItem as well

LFDanLu (changes requested):
Just some small comments, just wanna make sure though, was disabling the submenutrigger/menu specific onAction/etc tested? I _think_ everything is hooked up properly but would be nice to double check

packages/react-aria-components/src/Menu.tsx
```diff
@@ -268,3 +328,86 @@ function MenuItemInner<T>({item}: MenuItemInnerProps<T>) {
...
+      isFocusVisible,
+      selectionMode: state.selectionManager.selectionMode,
+      selectionBehavior: state.selectionManager.selectionBehavior,
+      hasSubmenu: true,
+      isSubmenuOpen
+    }
+  });
+
+  let ElementType: React.ElementType = props.href ? 'a' : 'div';
```
snowystinger, line 373: can a submenu trigger be a link? seems like we shouldn't allow that

packages/react-aria-components/src/Popover.tsx
```diff
@@ -150,9 +150,13 @@ function PopoverInner({state, isExiting, UNSTABLE_portalContainer, ...props}: Po
 
   let style = {...popoverProps.style, ...renderProps.style};
 
+  let onPointerDown = () => {
+    state.close();
```
snowystinger, line 154: won't this affect a lot more than menus?
popovers were already closing without this, I assume that was interact outside?

all popovers are dismissable, so I don't think this will hurt other things, but I'd like to have this specifically called out for testing

additionally, should we split this into pointerDown for mouse but pointerUp for touch? and do we need the mouse/touch fall backs? keeping in mind it'd really be onPress for touch, the event sequence would need to start and end on the underlay so that a swipe or scroll on the popover doesn't accidentally close it.

I'm fine for the touch stuff to be followup, but we do need to test this

packages/react-aria-components/test/Menu.test.js
```diff
@@ -316,4 +345,240 @@ describe('Menu', () => {
...
+      expect(submenu).toBeInTheDocument();
+      
+      let submenuPopover = submenu.closest('.react-aria-Popover');
+      expect(submenuPopover).toBeInTheDocument();
+      expect(submenuPopover).toHaveAttribute('data-trigger', 'SubmenuTrigger');
+
+      // Click a submenu item
+      await user.click(getAllByRole('menuitem')[5]);
+      expect(onAction).toHaveBeenLastCalledWith('email');
+      expect(submenu).not.toBeInTheDocument();
```
snowystinger, line 409: also expect menu not to be in the document

packages/react-aria-components/test/Menu.test.js
```diff
@@ -316,4 +345,240 @@ describe('Menu', () => {
...
+      expect(nestedTriggerItem).toHaveAttribute('data-hovered', 'true');
+      expect(nestedTriggerItem).toHaveAttribute('aria-expanded', 'true');
+      let nestedSubmenu = getAllByRole('menu')[1];
+      expect(nestedSubmenu).toBeInTheDocument();
+
+      let nestedSubmenuPopover = nestedSubmenu.closest('.react-aria-Popover');
+      expect(nestedSubmenuPopover).toBeInTheDocument();
+      expect(nestedSubmenuPopover).toHaveAttribute('data-trigger', 'SubmenuTrigger');
+
+      await user.click(document.body);
```
snowystinger, line 579: did you mean to click the underlay? this doesn't appear to match the code you changed above adding a pointerDown listener to the underlay

snowystinger (commented):
it's working so much better, only a couple additions to daniel's review

packages/react-aria-components/src/Menu.tsx
reidbarber: I feel like we changed this somewhere else. I'll try to find where this was discussed.

reidbarber: Looks like we made a change in listbox regarding this: https://github.com/adobe/react-spectrum/commit/2b05b6d984e1b37d4ac76f0bb305433d8f57d988#diff-9da69535256428710b8d6d7b6ce51163e15105e3084c7bc06d2ba612e1ba9dbf

I'll update these to use the hover state from useHover instead.

packages/react-aria-components/src/Popover.tsx
reidbarber: We do this for RSP Popovers:

https://github.com/adobe/react-spectrum/blob/ee51290b98fdc7cad31957ed2dd0a486eb0c1427/packages/%40react-spectrum/overlays/src/Popover.tsx#L120-L122

I think everything you mentioned makes sense. I'll bring this up with the team soon.

packages/react-aria-components/src/Menu.tsx
reidbarber: Removed the -5. Users can set the popover offset, but not sure if they would need to adjust the container's.

devongovett:
Still getting the behavior in the homepage example where if you open the submenu, then move your mouse from Twitter back to Share it doesn't get the hover state again. Did you look into this one yet?

reidbarber:
> Still getting the behavior in the homepage example where if you open the submenu, then move your mouse from Twitter back to Share it doesn't get the hover state again. Did you look into this one yet?

@devongovett Looking into this now. Looks to be unrelated to `useSafelyMouseToSubmenu` and doesn't happen in RSP submenus.

devongovett:
We should also add the new state(s) to the tailwind plugin.

packages/react-aria-components/docs/Menu.mdx
```diff
@@ -658,12 +673,106 @@ function Example() {
...
+    {id: 'twitter', name: 'Twitter'},
+    {id: 'email', name: 'Email', children: [
+      {id: 'work', name: 'Work'},
+      {id: 'personal', name: 'Personal'},
+    ]}
+  ]}
+];
+
+<MyMenuButton label="Actions" items={items}>
+    {function renderSubmenu(item) {
```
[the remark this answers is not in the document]
reidbarber: Can an arrow function be recursive?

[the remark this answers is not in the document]
devongovett: oh interesting, I missed that. hmm this is different from our usual collection implementation which would handle the recursion for you. I guess it's fine.

packages/react-aria-components/src/Menu.tsx
```diff
@@ -268,3 +329,86 @@ function MenuItemInner<T>({item}: MenuItemInnerProps<T>) {
...
+      ref={ref}
+      data-disabled={states.isDisabled || undefined}
+      data-hovered={isHovered || undefined}
+      data-focused={states.isFocused || undefined}
+      data-focus-visible={isFocusVisible || undefined}
+      data-pressed={states.isPressed || undefined}
+      data-selected={states.isSelected || undefined}
+      data-selection-mode={state.selectionManager.selectionMode === 'none' ? undefined : state.selectionManager.selectionMode}
+      data-has-submenu
+      data-is-submenu-open={isSubmenuOpen || undefined}>
```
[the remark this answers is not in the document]
devongovett: yeah seems fine as long as the docs say "submenu" in them.

packages/react-aria-components/src/Menu.tsx
```diff
@@ -100,8 +139,18 @@ function MenuInner<T extends object>({props, collection, menuRef: ref}: MenuInne
...
+  let popoverContainerRef = useRef<HTMLDivElement>(null);
   let {menuProps} = useMenu(props, state, ref);
 
+  let [leftOffset, setLeftOffset] = useState({left: 0});
+  useEffect(() => {
+    if (popoverContainerRef.current) {
+      let {left} = popoverContainerRef.current.getBoundingClientRect();
+      setLeftOffset({left: -1 * left});
+    }
+  }, []);
```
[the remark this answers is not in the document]
reidbarber: This was taken from the RSP submenu, but I don't think its needed here now that I look at it.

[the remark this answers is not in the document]
LFDanLu: This was to position the submenu's container so that it is the full width of the page, used to properly calculate if it needs to flip horizontally. I think you'll still need it?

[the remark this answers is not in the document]
reidbarber: Seems to work without it, so I've removed it for now.

packages/@react-aria/menu/src/useSubmenuTrigger.ts
```diff
@@ -106,6 +106,7 @@ export function UNSTABLE_useSubmenuTrigger<T>(props: AriaSubmenuTriggerProps, st
       case 'Escape':
         e.stopPropagation();
         state.closeAll();
+        ref.current.focus();
```
majornista, line 109: Restoring focus on Escape fails beyond submenu level 2 in the [Submenu Nested Example](https://reactspectrum.blob.core.windows.net/reactspectrum/d4d305826cd9a57226a151185b9be1451237b3ac/storybook/index.html?path=/story/react-aria-components--submenu-nested-example) and [Submenu Many Items Example](https://reactspectrum.blob.core.windows.net/reactspectrum/d4d305826cd9a57226a151185b9be1451237b3ac/storybook/index.html?path=/story/react-aria-components--submenu-many-items-example).

majornista (changes requested):

packages/@react-aria/menu/src/useSubmenuTrigger.ts
reidbarber: @LFDanLu This helps restore focus for the case of one submenu, but doesn't work for the nested case.

majornista: Would it be easier to manage focus if Escape only closed the top menu, rather than `state.closeAll()`?

reidbarber: I've fixed the behavior to restore focus in the nested case. I'll still bring this up with the team on whether we want to close all or one. I think it would be nice if the hooks+RAC supported both cases (i.e. a prop for `shouldCloseAllOnEscape`).

reidbarber: After bringing it up with the team, we decided to go with closing all for now. In the future, if the use cases comes up for supporting closing one submenu (via a prop), we can consider it.

LFDanLu (commented):
Provisional approval for the code changes, will wait for the new build to test it fully. Will do a full approval after than, pulling it out of the main page example/starter example can be followup

LFDanLu (approved):
LGTM, lets get this in for some final testing

packages/react-aria-components/src/Menu.tsx
```diff
@@ -21,13 +21,17 @@ import {Key, LinkDOMProps} from '@react-types/shared';
...
+import React, {createContext, ForwardedRef, forwardRef, ReactElement, ReactNode, RefObject, useCallback, useContext, useRef, useState} from 'react';
+import {RootMenuTriggerState, UNSTABLE_useSubmenuTriggerState} from '@react-stately/menu';
 import {Separator, SeparatorContext} from './Separator';
 import {TextContext} from './Text';
+import {UNSTABLE_useSubmenuTrigger} from '@react-aria/menu';
 
 export const MenuContext = createContext<ContextValue<MenuProps<any>, HTMLDivElement>>(null);
 export const MenuStateContext = createContext<TreeState<unknown> | null>(null);
+export const RootMenuTriggerStateContext = createContext<RootMenuTriggerState | null>(null);
+export const SubmenuContext = createContext<{popoverContainerRef: RefObject<Element>} | null>(null);
```
devongovett, line 34: Do we need to create a separate SubmenuContext for this, or could we have Menu provide PopoverContext with UNSTABLE_portalContainer directly?

reidbarber: That should work. We'll get a reading refs during render lint warning, but I guess we can ignore.

PR #5660, opened 2024-01-05
Make submenu properly size itself to fill available room and improve overlay positioning when pinch zoomed

LFDanLu:
Closes https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=43393367 and https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=45014178

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test the Submenu stories and make sure that the submenu properly moves itself to display all of its contents if possible. It should mimic MacOS's menu behavior. In addition, Popover that are top oriented should have their max height properly.
In addition test using the "Popover triggers on edges" story and in/out of the storybook iframe. Various browser should also be tests (especially Safari)

Overlays should have proper positioning and max height set when opened after pinch zooming into the screen. Test this outside the storybook iframe or in the built docs. Note that "freezing" the size/position of the overlay is handled in a different PR   

## 🧢 Your Project:

RSP


packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -149,12 +152,13 @@ function getDelta(
...
-  padding: number
+  padding: number,
+  containerOffsetWithBoundary: Offset
 ) {
   let containerScroll = containerDimensions.scroll[axis];
   let boundaryHeight = boundaryDimensions[AXIS_SIZE[axis]];
-  let startEdgeOffset = offset - padding - containerScroll;
-  let endEdgeOffset = offset + padding - containerScroll + size;
+  let startEdgeOffset = offset - padding - containerScroll + containerOffsetWithBoundary[axis];
+  let endEdgeOffset = offset + padding - containerScroll + size + containerOffsetWithBoundary[axis];
```
LFDanLu, line 161: In the case where the menu is portaled to a container that isn't the body, we need to adjust the offset calculated to fit the same coordinate system that the boundary uses hence the addition of the `containerOffsetWIthBounday[axis]` here. This fixes the `getDelta` for submenus since they are positioned with respect to their parent menu, not the page body.

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -222,7 +226,11 @@ function computePosition(
   }/* else {
     the overlay top should match the button top
   } */
-  // add the crossOffset from props
+
+  // TODO: crossAxis behaves differently from offset. A positive offset value will always shift the overlay in the "direction" it has
+  // (aka a overlay positioned "top" will be shifted upwards when a positive offset is applied and a overlay positioned "bottom" will be shifted downwards with the same offset)
+  // However, a positive crossAxis value always shifts the overlay up/to the right regardless of the cross axis direction... This means that the cross axis we apply to the submenu
+  // is incorrect if the submenu needs to flip upwards since it will shift it upwards instead of downwards... Changing this behavior is a breaking change though...
```
LFDanLu, line 233: Just something to note, was more pertinent when I was trying to make the submenus flip along the axis formed by their parent menu item trigger. The current behavior of `crossOffset` is different from `offset` since a positive `offset` always moves the overlay in the direction of the main axis but `crossOffset` will move the overlay "up" or to the "right" regardless of the `crossAxis` direction. I'll get rid of this comment later but just wanted to point it out for potential discussions over desired behavior

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -242,31 +250,41 @@ function computePosition(
...
+  padding: number,
+  overlayHeight: number,
+  heightGrowthDirection: HeightGrowthDirection
 ) {
-  return position.top != null
+  // TODO: mention the change to the logic in this code, in paticular with how it affectted the tests in calculatePosition
+  // In particular the "left bottom" test case with no viewport offset was calculating a max height of 400 previously because the
+  // overlay's "height growth" was judged to be the downwards direction even though the placement was "left bottom" aligns the bottom of the overlay with the
+  // bottom of the trigger and thus the overlay grows upwards
+  const containerHeight = (isContainerPositioned ? containerOffsetWithBoundary.height : boundaryDimensions[TOTAL_SIZE.height]);
```
LFDanLu, line 270: I found the logic in here to be pretty confusing since it made some assumptions that a overlay with `position.top` specified meant that the overlay was positioned below the trigger and thus expands downwards, something that wasn't true when I had the flip behavior added for submenus. I changed this here to use a `heightGrowthDirection` so that we know in what direction the overlay can grow. I also modified the latter part of the calculation to calculate the overlay's max height from the bottom of the overlay to the top of the boundary instead of from the trigger's top to the top of the boundary since that felt like it made more sense.

Lemme know if I misunderstood some part of this logic here.

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -337,16 +355,34 @@ export function calculatePositionInternal(
...
+    } else if (placementInfo.placement === 'bottom') {
+      heightGrowthDirection = 'bottom';
+    }
+  } else if (placementInfo.crossAxis === 'top') {
+    if (placementInfo.crossPlacement === 'top') {
+      heightGrowthDirection = 'bottom';
+    } else if (placementInfo.crossPlacement === 'bottom') {
+      heightGrowthDirection = 'top';
+    }
+  }
```
LFDanLu, line 372: note that a `crossAxis=top` and `crossPlacement=top` means that the overlay's top edge is aligned with the trigger's top edge, hence why the height should grow downwards.

packages/@react-aria/overlays/test/calculatePosition.test.ts
```diff
@@ -216,13 +219,13 @@ describe('calculatePosition', function () {
...
-      crossAxisOffsetNegative: [50, 140, undefined, 196, 410],
-      mainAxisOffset: [40, 150, undefined, 196, 400],
-      arrowBoundaryOffset: [50, 322, undefined, 176, 228]
+      noOffset: [50, 150, undefined, 196, 300],
+      offsetBefore: [-200, 50, undefined, 50, 200],
+      offsetAfter: [300, 350, undefined, 196, 500],
+      crossAxisOffsetPositive: [50, 160, undefined, 196, 310],
+      crossAxisOffsetNegative: [50, 140, undefined, 196, 290],
+      mainAxisOffset: [40, 150, undefined, 196, 300],
+      arrowBoundaryOffset: [50, 322, undefined, 176, 472]
```
LFDanLu, line 228: with the previous logic, a `left bottom` overlay had its max height calculated as if it should grow downwards but that positioning has the bottom edge of the of the overlay line up with the bottom of the trigger and so it should grow upwards.

packages/@react-aria/overlays/test/calculatePosition.test.ts
```diff
@@ -231,7 +234,7 @@ describe('calculatePosition', function () {
       offsetAfter: [350, 300, 196, undefined, 450],
       crossAxisOffsetPositive: [210, 50, 196, undefined, 200],
       crossAxisOffsetNegative: [190, 50, 196, undefined, 200],
-      mainAxisOffset: [200, 40, 196, undefined, 200],
+      mainAxisOffset: [200, 40, 196, undefined, 190],
```
LFDanLu, line 237: This seemed like a bug, max height didn't change if the `top` positioned overlay was pushed upwards via `offset`

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -149,17 +152,24 @@ function getDelta(
...
-    return -startEdgeOffset;
-  } else if (endEdgeOffset > boundaryHeight) {
-    return Math.max(boundaryHeight - endEdgeOffset, -startEdgeOffset);
+  // The height/width of the boundary. Matches the axis along which we are adjusting the overlay position
+  let boundarySize = boundaryDimensions[AXIS_SIZE[axis]];
+  // Calculate the edges of the boundary (accomodating for the boundary padding) and the edges of the overlay.
+  let boundaryStartEdge = boundaryDimensions.scroll[AXIS[axis]] + padding;
+  let boundaryEndEdge = boundarySize + boundaryDimensions.scroll[AXIS[axis]] - padding;
+  let startEdgeOffset = offset - containerScroll + containerOffsetWithBoundary[axis];
+  let endEdgeOffset = offset - containerScroll + size + containerOffsetWithBoundary[axis];
```
LFDanLu, line 165: In the case where the menu is portalled to a container that isn't the body, we need to adjust the offset calculated to fit the same coordinate system that the boundary uses hence the addition of the containerOffsetWIthBounday[axis] here. This fixes the getDelta for submenus since they are positioned with respect to their parent menu, not the page body.

Also moved the padding accommodation to boundary edge calculations since it felt more intuitive to put it there

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -149,17 +152,24 @@ function getDelta(
...
+
+  // If any of the overlay edges falls outside of the boundary, shift the overlay the required amount required to align one of the overlay's
+  // edges with the closest boundary edge.
+  if (startEdgeOffset < boundaryStartEdge) {
+    return boundaryStartEdge - startEdgeOffset;
+  } else if (endEdgeOffset > boundaryEndEdge) {
+    return Math.max(boundaryEndEdge - endEdgeOffset, boundaryStartEdge - startEdgeOffset);
```
LFDanLu, line 172: Note that the second calculation here will prefer the minimum amount of movement of the overlay since the ifs will ensure the compared movement values are both negative (aka would shift the overlay "upwards"). I didn't add the same consideration to the first calculation because we don't have the same guarantee that the movement values are both in the same direction. Could do some calculations to ensure that but at the end of the day `getMaxHeight` will ensure that the overlay fits in the boundary anyways so it isn't really necessary anyways

ktabors (changes requested):
I can't get any of the bad behavior in Safari. Works in Chrome. iPhone seems good too. Partial review since the boundary padding needs improvement.

packages/@react-aria/overlays/src/calculatePosition.ts
LFDanLu: This is potentially a case for introducing a new prop. The container for the submenu isn't actually the body but a div within its parent menu that the submenu gets portalled into. Said container matches the width of the viewport but doesn't actually have a height and so we apply a container padding of 0 for the submenu. In this case we could consider adding a `boundaryPadding` prop to `useOverlayPosition`/`calculatePosition` and all the other hooks that call them. 

In many places in `calculatePosition`, `padding` seems to be equivalent to `boundaryPadding` already, but `useOverlayPosition` passes `containerPadding` as `padding` since we most likely considered container === boundary?

packages/@react-aria/overlays/src/calculatePosition.ts
LFDanLu: good catch on this one, its an issue with the maxHeight calculation that needed to account for the padding, will push a fix once I figure out some unrelated Safari positioning issues.

packages/@react-aria/overlays/src/calculatePosition.ts
LFDanLu: I'll handle this as a followup, will write a issue to track

ktabors:
While testing with the popover edge story, this change breaks most of the bottom popovers. It's rendering them somewhere outside of the viewable area. https://reactspectrum.blob.core.windows.net/reactspectrum/0458b7f6f9bae2ceecb11181c05051b907b639c2/storybook/index.html?path=/story/dialogtrigger--triggers-on-edges&providerSwitcher-express=false

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -425,7 +477,14 @@ export function calculatePosition(opts: PositionOpts): PositionResult {
...
   let boundaryDimensions = getContainerDimensions(boundaryElement);
   let containerDimensions = getContainerDimensions(container);
+  // If the container is the HTML element wrapping the body element, the retrieved scrollTop/scrollLeft will be equal to the
+  // body element's scroll. Set the container's scroll values to 0 since the overlay's edge position value in getDelta don't then need to be further offset
+  // by the container scroll since they are essentially the same containing element and thus in the same coordinate system
   let containerOffsetWithBoundary: Offset = boundaryElement.tagName === 'BODY' ? getOffset(container) : getPosition(container, boundaryElement);
+  if (container.tagName === 'HTML' && boundaryElement.tagName === 'BODY') {
+    containerDimensions.scroll.top = 0;
+    containerDimensions.scroll.left = 0;
+  }
```
LFDanLu, line 487: This feels pretty gross, I was operating under the assumption that the boundary element would contain the container with all the changes in this code to calculate overlay/boundary edges but the above makes it a bit iffy. I'm not 100% sure this covers all the possible cases where perhaps the body is offset from the HTML element but the cases I tested with absolute positioning/applying a margin on the body seem to work alright. If anything it really feels like we should move over to floating-ui haha

devongovett: Is this related the the webkit issue you mentioned above? I didn't quite follow the description...

devongovett (approved):

snowystinger (approved):

ktabors (dismissed):

ktabors:
There are still some edge cases, but I'm not sure it's worth fixing or blocking this PR.
Renders with no container padding.
image
Not showing this one in the viewable area.
image

packages/@react-aria/overlays/src/calculatePosition.ts
LFDanLu: Not quite, the webkit one is the following in getContainerDimensions:
```
  if (isWebKit() && (containerNode.tagName === 'BODY' || containerNode.tagName === 'HTML') && isPinchZoomedIn) {
    // Safari will report a non-zero scrollTop/Left for the non-scrolling body/HTML element when pinch zoomed in unlike other browsers.
    // Set to zero for parity calculations so we get consistent positioning of overlays across all browsers.
    // Also switch to visualViewport.pageTop/pageLeft so that we still accomodate for scroll positioning for body/HTML elements that are actually scrollable
    // before pinch zoom happens
    scroll.top = 0;
    scroll.left = 0;
    top = visualViewport.pageTop;
    left = visualViewport.pageLeft;
  }
```

this one is mainly highlighting that the base case where boundary = 'body' and container = 'html' is different from the envisioned typical setup of an arbitrary boundary element that contains an arbitrary container element which has the trigger element since the container in the former case is actually around the boundary element instead AND typically is of the same dimensions. Since those two are typically the same "box" their coordinate systems should be the same (aka 0,0 refers to the same point from the view point of the boundary and the container) and thus the overlay's top/left values is already in the same coordinate system as the boundary and thus shouldn't need to be offset by the container's scroll. This means we need `containerDimensions` scroll value to be 0 in that case so that the following code in `getDelta` doesn't calculate the wrong `startEdgeOffset`/`endEdgeOffset` values
```
  let containerScroll = containerDimensions.scroll[axis];
  // The height/width of the boundary. Matches the axis along which we are adjusting the overlay position
  let boundarySize = boundaryDimensions[AXIS_SIZE[axis]];
  // Calculate the edges of the boundary (accomodating for the boundary padding) and the edges of the overlay.
  // Note that these values are with respect to the visual viewport (aka 0,0 is the top left of the viewport)
  let boundaryStartEdge = boundaryDimensions.scroll[AXIS[axis]] + padding;
  let boundaryEndEdge = boundarySize + boundaryDimensions.scroll[AXIS[axis]] - padding;
// THE CONTAINER SCROLL BELOW SHOULD BE 0 IN THE CASE HIGHLIGHTED ABOVE SINCE OFFSET IS ALREADY IN THE SAME COORDINATE SYSTEM AS THE BOUNDARY VALUES.
  let startEdgeOffset = offset - containerScroll + containerOffsetWithBoundary[axis] - boundaryDimensions[AXIS[axis]];
  let endEdgeOffset = offset - containerScroll + size + containerOffsetWithBoundary[axis] - boundaryDimensions[AXIS[axis]];
```

Note that `startEdgeOffset = offset - containerScroll + containerOffsetWithBoundary[axis] - boundaryDimensions[AXIS[axis]];` should work as expected if the scroll of the `containerElement` is separate from the scroll of the `boundaryElement`

snowystinger (approved):

devongovett (approved):

PR #5689, opened 2024-01-12
Fix ComboBox async loading runtime error

### LINKED ISSUE #5680: ComboBox with useAsyncList breaks on down arrow key press

Typing into the ComboBox filters the list as expected but when I type a value that does not match anything and then quickly press the down arrow twice I get a TypeError `Cannot read properties of null (reading 'type')`.

### 🤔 Expected Behavior?

When I press the down arrow quickly after filtering the list when there are no matching items, nothing happens.

### 😯 Current Behavior

When I press the down arrow quickly after filtering the list when there are no matching items, a `TypeError` occurs.

### 💁 Possible Solution

Screenshot 2024-01-10 at 16 03 57

The `TypeError` occurs in the [ListKeyboardDelegate class](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/selection/src/ListKeyboardDelegate.ts#L173) and is being caused by the key existing in the collection but then not being able to find the item in the collection based on that key.

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

Code Sandbox: https://codesandbox.io/p/sandbox/brave-williamson-l4nz2v?file=%2Fsrc%2FApp.js

**Code Sandbox: Error Reproduction**

https://github.com/adobe/react-spectrum/assets/135844822/1d2c419b-d0e3-43f1-9db0-2dcf75e08d35

**Async Example in React Aria Docs: Error Reproduction**

https://github.com/adobe/react-spectrum/assets/135844822/7b49a431-f563-422f-805e-9288a8c45f0a

### Version

react-aria-components@1.0.0

### If other, please specify.

_No response_

macOS Ventura 13.13.1

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

sookmax:
Closes #5680

Hello, this PR includes a minimal fix for #5680 and a ComboBox story for an async-loading example.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Visit http://localhost:9003/?path=/story/react-aria-components--combo-box-async-loading-example&providerSwitcher-express=false&strict=true
2. Test the behavior described in the issue #5680.

### DIFF
--- a/packages/@react-aria/selection/src/ListKeyboardDelegate.ts
+++ b/packages/@react-aria/selection/src/ListKeyboardDelegate.ts
@@ -170,7 +170,7 @@ export class ListKeyboardDelegate<T> implements KeyboardDelegate {
     let key = this.collection.getFirstKey();
     while (key != null) {
       let item = this.collection.getItem(key);
-      if (item.type === 'item' && !this.disabledKeys.has(key)) {
+      if (item?.type === 'item' && !this.disabledKeys.has(key)) {
         return key;
       }
 
--- a/packages/react-aria-components/stories/ComboBox.stories.tsx
+++ b/packages/react-aria-components/stories/ComboBox.stories.tsx
@@ -14,6 +14,7 @@ import {Button, ComboBox, Input, Label, ListBox, Popover} from 'react-aria-compo
 import {MyListBoxItem} from './utils';
 import React from 'react';
 import styles from '../example/index.css';
+import {useAsyncList} from 'react-stately';
 
 export default {
   title: 'React Aria Components'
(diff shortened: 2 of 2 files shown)


Haeven (commented):
LGTM, good catch on this. ⭐

richardbann:
Hi, I think the root cause of this issue is here: https://github.com/adobe/react-spectrum/blob/4ae2831f9487958511c88b9ebe672b5650042d7e/packages/react-aria-components/src/Collection.tsx#L599-L621
When all nodes in the collection are deleted, `this.mutatedNodes.size = 0` so the commit does not run and `firstKey`, `lastKey` remains unchanged. Also `frozen` remains `false`.

As an experimental fix I added the line `this.getMutableNode(node);` to Document's removeNode:
https://github.com/richardbann/react-spectrum/blob/e1adabd3b2684cd96150dfc345d565c550e098bc/packages/react-aria-components/src/Collection.tsx#L579-L588. This adds the deleted node to `mutatedNodes` and the commit runs.

What do you think?

LFDanLu:
@richardbann Thank you for the investigation, I believe you are correct. Did some digging myself and it looks like the Document is properly updating its own tracked [firstChild](https://github.com/adobe/react-spectrum/blob/ee51290b98fdc7cad31957ed2dd0a486eb0c1427/packages/react-aria-components/src/Collection.tsx#L248-L250) (aka it knows there is nothing rendered in the fake DOM), but just isn't updating its internal collection as you noticed. I wonder if we can get into a similar non-updated state if a new node was added before the current first node, I think not because of [this](https://github.com/adobe/react-spectrum/blob/ee51290b98fdc7cad31957ed2dd0a486eb0c1427/packages/react-aria-components/src/Collection.tsx#L600-L604) which will call `updateNode` -> `getMutableNode` for each node so long as it is "connected" to the document. 

I think your solution is fine or we could potentially perform a update on every dirty node (aka remove the [isConnected](https://github.com/adobe/react-spectrum/blob/ee51290b98fdc7cad31957ed2dd0a486eb0c1427/packages/react-aria-components/src/Collection.tsx#L601) check here), will have to test some more though

sookmax:
@LFDanLu I don't think I have a proper understanding on the above discussion.

I'm happy to close this PR in favor of @richardbann's if it's going to contain more of a fundamental fix for more general/broader cases.

LFDanLu (approved):
LGTM, thanks! Noticed a strange flicker when using up/down arrow keys when the async combobox doesn't have any items but that is unrelated to this change

richardbann:
Issue & PR: #5734 #5735

packages/@react-aria/selection/src/ListKeyboardDelegate.ts
```diff
@@ -170,7 +170,7 @@ export class ListKeyboardDelegate<T> implements KeyboardDelegate {
     let key = this.collection.getFirstKey();
     while (key != null) {
       let item = this.collection.getItem(key);
-      if (item.type === 'item' && !this.disabledKeys.has(key)) {
+      if (item?.type === 'item' && !this.disabledKeys.has(key)) {
```
devongovett, line 173: Do we need this check in other methods like `getLastKey`, `getPreviousKey`, and `getNextKey` which have similar code?

devongovett (approved):

packages/@react-aria/selection/src/ListKeyboardDelegate.ts
sookmax: Well yes I agree, and that's why I [suggested](https://github.com/adobe/react-spectrum/issues/5680#issuecomment-1887307935) to convert the whole directory `packages/@react-aria/selection` to strict TS.

Maybe I should work on that next.

PR #5735, opened 2024-01-22
Ensure Collection commits its updated information when nodes are removed

### LINKED ISSUE #5734: RAC Collection not updated properly when all items removed

In certain situations RAC collection's `firstKey`, `lastKey` and `frozen` attributes are not updated correctly. See discussin in #5689.

### 🤔 Expected Behavior?

In situation discussed below in case the items are deleted from the collection `firstKey` and `lastKey` should be `null`. 

### 😯 Current Behavior

Rendering a collection components with items:
```jsx
let items = ['a', 'b', 'c'];
return (
  <Listbox >
    {items.map((item) => <ListBoxItem key={item} />)}
  </Listbox>
);
```
and later rerendering with `items = []` leaves the internal collection with `frozen = false`, `firstKey = 'c'`, `lastKey = 'c'` which potentially leads to issues like #5680.
When the collection is initially created with `items = []` the internal collection's `frozen` attribute is `false`.

### 💁 Possible Solution

I am filing a pull request with a possible solution. #5735

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

See issue description in #5680.
Jest test suite:

```js
import {ListBoxItem} from '../src/ListBox';
import React from 'react';
import {render} from '@testing-library/react';
import {useCollection} from '../src/Collection';


const CollectionTest = (props) => {
  const result = useCollection(props);
  props.spyConnection.current = result.collection;
  return <>{result.portal}</>;
};

const renderItems = (items, spyConnection) => (
  <CollectionTest spyConnection={spyConnection}>
    {items.map((item) => <ListBoxItem key={item} />)}
  </CollectionTest>
);

describe('Collection', () => {
  it('should be frozen even in case of empty initial collection', () => {
    let spyConnection = {};
    render(renderItems([], spyConnection));
    expect(spyConnection.current.frozen).toBe(true);
  });

  it('should have correct firstKey, lastKey and should be frozen after all items are deleted', () => {
    let spyConnection = {};
    const {rerender} = render(renderItems(['a'], spyConnection));
    rerender(renderItems([], spyConnection));
    expect(spyConnection.current.frozen).toBe(true);
    expect(spyConnection.current.firstKey).toBe(null);
    expect(spyConnection.current.lastKey).toBe(null);
  });
});
```

### Version

RAC 1.0.1

Other

### If other, please specify.

not browser specific

linux

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

richardbann:
Closes #5734

## ✅ Pull Request Checklist:

- [ x] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- [ x] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [ x] Filled out test instructions.

## 📝 Test Instructions:
`yarn jest "/Collection.test.js"`


LFDanLu:
GET_BUILD

LFDanLu (dismissed):
Verified that this fix works and seems to be pretty lightweight. Not sure what other situations could cause this to happen though, wondering if we should always commit the collection whenever there are dirty nodes but that might be overly excessive. I'll defer to @devongovett here since he would be more knowledgable about this Collection implementation.

richardbann:
Went through the logic, the missing commits can be guaranteed by checking if `collectionMutated` is `true`. To use `dirtyNodes` a bit more code reorg would be needed.

sookmax:
@richardbann this fix is very cool! I now just realized how significant this fix is in terms of preventing issues like #5680 since it makes `Collection.getFirstKey()` to properly return `null` when there's no item in `Collection`.

Great stuff 👍

LFDanLu:
GET_BUILD

LFDanLu (approved):
> Went through the logic, the missing commits can be guaranteed by checking if collectionMutated is true. To use dirtyNodes a bit more code reorg would be needed.

Ah, yep that would do it. Looks good to me locally, will get a second reviewer to look at this soon

devongovett (approved):

PR #5775, opened 2024-01-30
Add isWheelDisabled prop to useNumberField and useColorField

ryo-manba:
Closes https://github.com/adobe/react-spectrum/issues/5676

## ✅ Pull Request Checklist:

### NumberField

1. Visit http://localhost:9003/?path=/story/react-aria-components--number-field-example&args=isScrollingDisabled:true&providerSwitcher-express=false&strict=true
2. `isWheelDisabled` prop set to `true` and verify that scrolling does not change the value.
3. Prop set to `false` and ensure default behavior.

### ColorField

1. Visit http://localhost:9003/?path=/story/colorfield--default
2. `isWheelDisabled` prop set to `true` and verify that scrolling does not change the value.
3. Prop set to `false` and ensure default behavior.


mstllc:
We are looking for this change as well, subscribing!

LFDanLu:
GET_BUILD

LFDanLu (commented):
IMO it is fine to add this prop since several people have been running this problem but will open this up for the team to discuss

packages/@react-aria/numberfield/src/useNumberField.ts
```diff
@@ -75,6 +75,7 @@ export function useNumberField(props: AriaNumberFieldProps, state: NumberFieldSt
     onKeyUp,
     description,
     errorMessage,
+    isScrollingDisabled,
```
LFDanLu, line 78: Talked with the team, we think that perhaps the name `isWheelDisabled` might be better since it better captures that it is really a wheel event that is modifying the numberfield. Also, it feels like this should also be applied to useColorField as well for parity, would you like to add that as well? No worries if not, the team can handle the update if need be since it would be nice to get this in for the upcoming release, just let us know!

ryo-manba: Okay, thanks! I'll try changing the variable name to isWheelDisabled. Also, I'll work on the useColorField support as well!

ryo-manba: @LFDanLu 
I've fixed it with the following commit. Looking forward to your review, thank you.
- [rename: isWheelDisabled props](https://github.com/adobe/react-spectrum/pull/5775/commits/ddc5f4f5d68aac4baad21eafa622cf0b4e9799bc)
- [Add isWheelDisabled prop to useColorField](https://github.com/adobe/react-spectrum/pull/5775/commits/e1ed7edccc4535db79273934a02c78176968f4f7)

ryo-manba: The test that is failing in CI consistently fails when run individually in a local environment as well. I have tested it with the latest version of the main branch, and it still fails.
https://app.circleci.com/pipelines/github/adobe/react-spectrum/18624/workflows/83a16daa-3db9-46a8-9dd2-147902647300/jobs/266647

The source code for the failing test can be found here:
https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/test/useFocusVisible.test.js#L111

However, when running the entire test suite that this test is part of (`describe('useFocusVisible', function () {...})`), it appears to pass.
https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/test/useFocusVisible.test.js#L51

It's possible that there might be an issue with the setup or cleanup not being handled correctly.
The cause has not been determined yet but I'll share it with you.

ryo-manba: Since this issue seems to be unrelated to this change, I don't see the need to include a correction in this PR. What do you think?

LFDanLu (approved):
LGTM, thanks for the PR! Only small thing would be that we'd move that new prop to the AriaProps interface since it affect interaction and thus belongs in that layer (as opposed to the base props which are also provided to the stately hooks as well), but I'd like to get this merged for testing and we can handle the props move on our end

PR #5785, opened 2024-01-31
Fix focusVisible & calendar issue #3975

### LINKED ISSUE #3975: Multiple focus rings sometimes appear when navigating by keyboard
# 🐛 Bug Report 

When navigating calendar grids using the keyboard, multiple focus rings sometimes appear. This generally happens when moving to a month that was previously not rendered. Confirmed the be behavior in Chrome and Brave. I can not reproduce the issue using Safari or Firefox.

## 🤔 Expected Behavior

Only one element should have a visible focus ring at a time

## 😯 Current Behavior

I hope this is as easy to reproduce for you as it is for me (using update-to-date Chrome on MacOS). This generally happens when moving to a month that was previously not rendered. It most often happens with disabled dates, but not always ... I've seen the the "ghost focus" item sometimes be a non-disabled date.

This example is using Devon's [Code Sandbox for useCalendar](https://codesandbox.io/s/objective-shape-8r4utm?file=/src/App.js:213-1267) - styled example with Tailwind:

focus-bug-demo

In the gif, I refresh the page, click on the preview area then tab to the calendar. Then I use the arrow keys to navigate back and forth from Feb. to March. As I do so, one or more days that do *not* have focus will show with a focus ring. These "ghost focus" days are most often in early Jan. outside the calendar range, and thus disabled. But you'll see I eventually get a day in March to behave the same).

At the end of the gif, I click on the preview area to hide the visual focus ring(s), then hit the tab key to bring them back. You'll see the multiple days in the calendar are still showing as if they have focus. 

Not seen in the gif, but if you can move the focus to one of the "ghost focus" days, then the problem for that day will correct itself.

## 💁 Possible Solution

(tldr; I have no solution, only guesses at what is happening)

I do not think this is a calendar-related bug, but an issue with useFocusRing. I've seen similar visual focus ring bugs occasionally pop up working with other components built using useFocusRing ... but never as bad as this.

You'll notice that the January "ghost focus" days tend to match the previous real focus position in March. So for example, if you focus the fourth column in the first row of March and then hit the "up" key, the fourth column in the first row of January will often get the "ghost focus". I don't know how the focus management logic works in useFocusRing, but seems like it's trying to guess based on similar html structure ... Focus was on the element in the first table in the first row in the fourth cell, but the DOM changed so I'm going to try to re-select that same first-table-in-the-first-row-in-the-fourth-cell element.

## 🔦 Context

Discovered while evaluating the useCalendar hooks to build a replacement for an aging fork of Airbnb's react-dates on a government website where accessibility/Section 508 compliance is critical. 

## 💻 Code Sample

As mentioned above, reproducible here: https://codesandbox.io/s/objective-shape-8r4utm?file=/src/App.js:213-1267

## 🌍 Your Environment

| react-aria/focus   | 3.10.1
| Browser          | Chrome, Brave, Edge
| Operating System | MacOS 13.1

I can not reproduce the issue using Safari or Firefox.

richardbann:
Closes #3975

## ✅ Pull Request Checklist:

- [ x] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- [ x] Filled out test instructions.

## 📝 Test Instructions:

Check the storybook link locally: http://localhost:9003/?path=/story/react-aria-components--calendar-example&providerSwitcher-express=false&strict=true, select date 2024-JAN-3, open dev toolbar and inspect the selected div element. Navigate to DEC-27 __using the "up" arrow key__. With the fix applied, the element does not have `data-focus-visible`. Without the fix, the attribute is there even if the date is disabled.

### DIFF
--- a/packages/react-aria-components/src/Calendar.tsx
+++ b/packages/react-aria-components/src/Calendar.tsx
@@ -478,6 +478,7 @@ function CalendarCell({date, ...otherProps}: CalendarCellProps, ref: ForwardedRe
 
   let {hoverProps, isHovered} = useHover({...otherProps, isDisabled: states.isDisabled});
   let {focusProps, isFocusVisible} = useFocusRing();
+  isFocusVisible &&= states.isFocused;
   let isOutsideMonth = !isSameMonth(currentMonth, date);
   let isSelectionStart = false;
   let isSelectionEnd = false;


richardbann:
@LFDanLu https://github.com/adobe/react-spectrum/issues/3975#issuecomment-1409134125 Spectrum does not seem to have these data attributes, but calendar does exactly the same thing, that is why multiple focus rings are not visible: https://github.com/adobe/react-spectrum/blob/bd477d20344f491144dc34b9bd114070c2e9b10b/packages/%40react-spectrum/calendar/src/CalendarCell.tsx#L67C38-L67C52

LFDanLu:
@richardbann Thank you for the fix, at a glance I believe you are correct but I'll take a closer look soon!

LFDanLu:
GET_BUILD

LFDanLu:
strange, I get the same test failures locally on main, but the last build passed the tests, will need to dig

LFDanLu:
GET_BUILD

reidbarber (approved):
LGTM, verified the previous failure case was fixed.

LFDanLu (approved):
I think this fix is fine for now, but we may want to revisit this in the future and perhaps have something like https://github.com/adobe/react-spectrum/blob/7c1da9f8a5e6aa7a6169b7c9ebe8725169ce80d0/packages/%40react-aria/interactions/src/utils.ts#L88-L106 handle this perhaps. This seems like a similar class of issue and we might be able to try to detect if the tabindex gets removed/changed in a way that a blur would happen.

We encountered something similar in our TableView here but the fix was to retain the tab index there: https://github.com/adobe/react-spectrum/issues/4219

Issue #4111, opened 2023-02-23
ColorWheel: Labelling is not useful for screen reader user.


jnurthen:
# 📝 Feedback

Labeling the color in the color wheel with the number of degrees is not an equitable experience for screen reader users. I understand that the color wheel is not often used on its own - but there should be a way to read a more useful label than the number of degrees on the color wheel.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| [Package name]   | ColorWheel
| Browser          | All
| Operating System | All (w/ screen readers)

devongovett:
The color components are not yet finished from an accessibility point of view, which is why they are still in beta. We are planning on using a list of color names, translated into each locale, to find the closest hue name with modifiers like light/dark etc. This should be much more descriptive and understandable to a user. However, it is currently blocked on figuring out how to distribute these translated strings in a way that doesn't bloat the bundle size (i.e. loading languages on demand). Some prior discussion here: https://github.com/adobe/react-spectrum/pull/1732#issuecomment-818311480.

PR #5792, opened 2024-02-02
#4110 #4111 Color hue/color name accessibility/i18n

### LINKED ISSUE #4110: ColorArea: Labelling seems confusing when using on iOS w/ VO
# 📝 Feedback

Labelling seems confusing when using on iOS
It is not clear which of the elements is going to be adjusted for each slider.
Suggest fixing in the following ways:
(1) only include the value that is being changed in the label after it is modified.
(2) add the word "Adjust" before the value that is being changed

In example would have: Adjust Saturation: 92 %, Lightness 43% when focused, then when adjusted only repeat the saturation value rather than both.
## 🌍 Your Environment

| [Package name]   | ColorArea
| Browser          | Safari
| Operating System | iOS w/ VO

majornista:
Closes #4110 and #4111. 

This PR implements the color naming strategy discussed in https://github.com/adobe/react-spectrum/pull/1732#issuecomment-818311480 for further review and consideration.

## ✅ Pull Request Checklist:

- [x] Included link to corresponding React Spectrum GitHub Issues: #4110 and #4111.

## 🧢 Your Project:

Adobe/Accessibility


packages/@react-aria/color/src/useColorArea.ts
```diff
@@ -427,14 +443,17 @@ export function useColorArea(props: AriaColorAreaOptions, state: ColorAreaState)
...
     },
     yInputProps: {
       ...yInputLabellingProps,
       ...visuallyHiddenProps,
       ...yInputFocusProps,
+      style: {
+        ...visuallyHiddenProps.style,
+        WebkitAppearance: 'slider-vertical'
+      },
+      orient: 'vertical',
```
devongovett, line 456: What were these added for? I see a deprecation warning in the chrome devtools for -webkit-appearance, and mdn says orient is also non-standard. Do we need them if we already have `aria-orientation` and these are visually hidden anyway?

majornista: Trying to provide a better distinction between the `x` and `y` axis when interacting with the ColorArea using a screen reader, per [#4110](https://github.com/adobe/react-spectrum/issues/4110).

Issue #5795, opened 2024-02-03
RAC: setting aria-label on MenuItem not working


hebergonza719:
### Provide a general summary of the issue here

Setting `aria-label` on `MenuItem` doesn't apply the attribute. Property isn't displayed on the DOM.

### 🤔 Expected Behavior?

`aria-label` attribute should be applied on the element.

### 😯 Current Behavior

`aria-label` attribute isn't applied on the element.

### 🖥️ Steps to Reproduce

https://stackblitz.com/edit/stackblitz-starters-7a36ep?file=src%2FApp.tsx

### Version

"react-aria-components": "1.0.0"

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

macOS 13.6.4

LFDanLu:
Looks like we forgot to pass the `aria-label` to the hooks for ListBoxItem and MenuItem, opened a PR for this. Will have to audit the rest of the components for parity. Thanks for reporting this!

PR #5796, opened 2024-02-03
Add missing aria-labels to ListBox, Menu, Tabs items in RAC

LFDanLu:
Closes https://github.com/adobe/react-spectrum/issues/5795

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/react-aria-components/src/GridList.tsx
```diff
@@ -228,9 +228,7 @@ export interface GridListItemProps<T = object> extends RenderProps<GridListItemR
   /** The object value that this item represents. When using dynamic collections, this is set automatically. */
   value?: T,
   /** A string representation of the item's contents, used for features like typeahead. */
-  textValue?: string,
-  /** An accessibility label for this item. */
-  'aria-label'?: string
```
LFDanLu, line 233: Removed because the grid item hooks doesn't use the provided `aria-label` anyways, instead opting to use the `textValue`. I don't quite remember why this differs with the menu and listbox item hooks, feels inconsistent

ktabors (approved):

packages/react-aria-components/src/GridList.tsx
ktabors: Won't this still be an API change?

snowystinger (approved):

packages/react-aria-components/src/GridList.tsx
LFDanLu: indeed it will be, but seeing as providing an `aria-label` doesn't do anything I figured we could remove it as a bug

PR #5808, opened 2024-02-05
Fix RAC submenu flipping

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/react-aria-components/src/Menu.tsx
```diff
@@ -399,6 +399,7 @@ function MenuItemTriggerInner<T>({item, popover, parentMenuRef, delay}: MenuItem
           triggerRef: ref,
           placement: 'end top',
           UNSTABLE_portalContainer: submenuContext.popoverContainerRef.current || undefined,
+          containerPadding: 0,
```
LFDanLu, line 402: Open question about this one, do we want to just hard code this to 0 or let the user configure it? This is what it looks like when using the default (12px):
image

Note the offset applied to the 3rd menu, this is because we are offsetting by that additional amount

This is with the above applied:
image

LFDanLu: Depending on the way the menus are styled (such as if there is additional margins and stuff pushing the menu up/down), this could still look a bit off...

snowystinger: in the first one you mean it's actually at 24px?

LFDanLu: hmm, actually on a closer look, they are both offset by 12px, its something with how the menus are styled in the storybook. I'll revert, but perhaps `containerPadding` is something to still consider providing to the user.

snowystinger (approved):
behaving well for me, tested in chrome/safari/ff on macos

reidbarber (approved):
LGTM

PR #5826, opened 2024-02-06
Remove "use client" and mark packages as client-only instead

### LINKED ISSUE #5639: Large bundle size when using Button component
### Provide your feedback here.

I noticed a ~175kb jump in bundle size after adding a `Button` (`react-aria-components`) in a fresh Next.js project. I ran a bundle analyzer check and it seems like it bundles the complete lib (seeing `dnd`, `datepicker` and others), so I'm guessing tree shaking is not working as expected. Seems to be the same problem for `Tabs`.

As mentioned in https://github.com/adobe/react-spectrum/issues/4356, I also tested the build with `swcMinify: false`, which does not have a positive effect 

Screenshot 2023-12-29 at 10 00 57

### 🔦 Context

_No response_

### 💻 Code Sample

It's easily reproduceable by creating a fresh Next.js project (`npx create-next-app@latest`), accept al the defaults and add this in a new page:

```tsx
import { Button } from 'react-aria-components'

export default function Page() {
  return (
    <Button>
      Test
    </Button>
  )
}
```

### Version

react-aria-components v1.0.0 & v1.0.1-nightly.4385

_No response_

### If other, please specify

_No response_

Mac OS 14.1.1

devongovett:
Closes #5639

This reverts #5498 where we added "use client" to our packages to support Next.js server components. I now believe this was a mistake. "use client" indicates a boundary between a server component and a client component, so the idea was to enable RSP/RAC components to be used directly from server components without needing to create your own client component wrapper. However in reality this did not work well because all of the props sent from a server component to a client component must be serializable. Therefore, things like event handlers could not be sent. Most of our components require some kind of event, so you'd end up needing to move the component that used RSP/RAC to your own client component anyway. In addition, some bundling issues with next.js caused code splitting not to work properly, resulting in massive bundles (#5639).

This PR instead uses `import "client-only"` to enforce that RAC and RSP can only be imported from a client component. Next.js will throw a build time error when this package is imported from a server component, providing a better error message than the one you get by default (either something deep inside RSP/RAC trying to import `useState` or another React hook not available in server components, or when trying to pass an event handler from a server component to a client component). Now the error is something like "this package can only be imported in a client component". Outside of Next.js this package is an empty file and has no effect.


snowystinger (approved):
Just checking, we only want to import this client only for our mono packages, we don't want to use it for any of the individual packages?

devongovett:
ehhhh we could, but it seemed like a lot of effort and I don't think many people use the individual packages (especially with next.js).

reidbarber (approved):

stevegreco:
I am curious what makes `use client` more limiting than `import 'client-only'`.  As a library maintainer that is using React Aria to build out our own component library, we have been getting questions about RSC support.  People seem to be split between the use of `client-only` and `use client`

PR #5852, opened 2024-02-12
combobox: Fix onSelectionChange type declaration

acr92:
onSelectionChange can be invoked with a null key if the user clears the input.

https://github.com/adobe/react-spectrum/blob/c11afb257a8bc409ed23272a167372005adf4bcc/packages/%40react-stately/combobox/src/useComboBoxState.ts#L225-L229

This in turns triggers onSelectionChange with a null value, even though the type contract doesn't allow that.

Here's my stacktrace:
image

SelectUser is my component, which sets `defaultItems`, `selectedKey` and `onSelectionChange`.

## ✅ Pull Request Checklist:

- N/A Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- N/A - Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- N/A - Filled out test instructions.
- N/A - Updated documentation (if it already exists for this component).
- N/A - Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/WAI/ARIA/apg/)

N/A for all as this is just a correction of the type declaration.


reidbarber:
Thanks for the PR! If you [sign the CLA](https://github.com/adobe/react-spectrum/pull/5852/checks?check_run_id=21477501401), then close and re-open this PR, it should trigger a re-build that passes CI.

snowystinger (approved):
Thanks for the PR
Initially I thought this should be moved up to SingleSelection. However, Picker and StepList both extend it as well, and those have selectionChanges which cannot be null.

This falls into our TS Strict https://github.com/adobe/react-spectrum/issues/3183 as well

reidbarber (approved):

Issue #5476, opened 2023-11-24
React Aria Components: rendering as custom components


chaance:
### Provide a general summary of the feature here

We should have a simpler way to render components other than what is returned from the React Aria component by default. This would be either some sort of polymorphic component rendering or a higher level hook.

In the `MenuItem` component (and possibly some others), the current way to render a link is to pass an `href` prop. This simply renders an `a` link instead of a `div`. For a library like React Aria, which is typically used as a primitive to create design-system level components, this is quite limiting. There should be some API for rendering another component without dropping down to the lowest-level APIs with React Aria hooks.

### 🤔 Expected Behavior?

TBD (exact behavior depends on the solution)

### 😯 Current Behavior

Currently I can't get the functionality of a component like `MenuItem` without either rendering that component or recreating it entirely with the much lower-level React Aria hooks.

There are a few challenges with the current implementation:

**TypeScript.** Particularly when using `forwardRef`, the `ref`'s type is always assumed to be `HTMLDivElement`. This is a common theme in pretty much every component lib that embraces the "polymorphic" component pattern, and it's difficult to solve without introducing a number of other tradeoffs.

**Custom link components.** React Aria basically assumes you either want a regular `a` for external links or you simply want to navigate using your app's client-side router, in which case it uses the navigator provided to `RouterProvider`. I think this leaves a lot to be desired, as router and/or framework link components often bake in other features we lose if rendered in the context of a RA component.

### 💁 Possible Solution

1. Allow for some sort of "slot" based solution that allows you to directly render a specific component, similar to `asChild` in Radix UI. This is probably not a great one for React Aria, since its interaction event props (`onPress` and family) heavily abstract and hide the underlying DOM props which a lot of components (including RR links) rely on. Leaving React Aria to decide how to resolve likely conflicts is probably going to be a mess.
2. Utility hooks with a higher level of abstraction. I think this would basically look like the lower level hooks in React Aria, but you get all of the props that React Aria would otherwise pass to its own components. This would allow consumers to decide how to resolve any conflicts and render whatever they'd like. This would be a sharp knife potentially but would allow us to drop down to a lower level without having to completely rewrite all of the component's functionality using the base React Aria hooks.
  ```ts
  function MenuItemLink({ to, prefetch, ...props }) {
  	// this would give you all of the props that are
  	// passed into `ElementType` in your `MenuItemInner`
  	// component, but leaves rendering to the consumer.
  	const ref = React.useRef();
[...]

PR #5864, opened 2024-02-13
Add support for routerOptions and useHref

### LINKED ISSUE #5335: Links inside RouterProvider with basename have inconsistent href handling

devongovett:
Closes #5395, closes #5335, relates to #5476

This improves integration with client side routers by introducing two new APIs:

* A `routerOptions` prop on each component that supports links. This is an opaque object that React Aria simply passes through to the router. It can be used for options like controlling scroll behavior, using replaceState instead of pushState, or any other features offered by a router. It is passed to the `navigate` function provided to `RouterProvider` as a second argument in addition to the `href`.
* A `useHref` prop for `RouterProvider` which is used to convert an `href` provided to a link component to a native `href`. For example, a router might accept hrefs relative to a base path, or offer additional custom ways of specifying link destinations. The original `href` specified on the link is passed to the `navigate` function of the `RouterProvider`, and `useHref` is used to generate the full native href to put on the actual DOM element.

Both of these APIs work with TypeScript to enable autocomplete and type checking for the router options and also the href. To configure it, you can use TypeScript module augmentation to specify the `RouterOptions` type:

```ts
declare module "react-aria-components" {
  interface RouterOptions {
    href: MyRouterHref,
    routerOptions: MyRouterOptions
  }
}
```

Do that once in your app (maybe where you setup RouterProvider), and all link components in React Aria/RSP will use those options.

## To do

- [x] Docs


packages/@react-stately/selection/src/SelectionManager.ts
```diff
@@ -494,4 +494,8 @@ export class SelectionManager implements MultipleSelectionManager {
   isLink(key: Key) {
     return !!this.collection.getItem(key)?.props?.href;
   }
+
+  getItemProps(key: Key) {
```
devongovett, line 498: Adding this to selection manager feels a bit weird, but not sure where else to put it. `useSelectableCollection` and `useSelectableItem` don't currently receive the collection as props, except via `SelectionManager`...

snowystinger: agreed, it's a bit odd, however, we already have 'isLink' in here, which is almost the same thing and is only tangentially related to selection

would it make more sense to just return the collection and then handle this specific logic outside?

snowystinger (approved):
can keep discussing, but approving so we can get nightlies out sooner rather than later

packages/@react-stately/selection/src/SelectionManager.ts
devongovett: well, the original idea was that the `MultipleSelectionManager` interface could have multiple implementations, perhaps not even using collections under the hood. So if we expose the collection from the interface that will force other implementations to also use collections.

snowystinger: that makes sense :-/

LFDanLu (approved):
Docs/code look good to me, will test the apps once they get built on main. Does the remix one get built somewhere? I tried running it locally but got a ` Invalid URL` error. Otherwise just one small comment but happy to get this merged as is

packages/@react-aria/utils/src/openLink.tsx
```diff
@@ -152,3 +156,15 @@ export function getSyntheticLinkProps(props: LinkDOMProps) {
...
+  let router = useRouter();
+  return {
+    href: props?.href ? router.useHref(props?.href) : undefined,
+    target: props?.target,
+    rel: props?.rel,
+    download: props?.download,
+    ping: props?.ping,
+    referrerPolicy: props?.referrerPolicy
+  };
+}
```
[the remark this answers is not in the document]
devongovett: because filterDOMProps isn't a hook so it cannot call other hooks... :/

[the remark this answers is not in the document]
LFDanLu: ahhh, right ugh

levrik:
Any timeline on these changes? I just started a new project with React Aria Components and wanted to use TanStack Router but one of its main features (link autocompletion) simply does not work atm.

LFDanLu (dismissed):

reidbarber (approved):

snowystinger (approved):

PR #5866, opened 2024-02-13
Fix submenu safe area edge case

reidbarber:
For RSP and RAC submenus.

There was an edge case where the submenu may close unexpectedly. To reproduce it, you could:
1. Move your pointer over the trigger item, then down to the submenu
2. Move your pointer back to the trigger item, but immediately back out to the submenu when the trigger was entered.

A stale previous pointer position state was being used, so the pointer movement angle used was inaccurate, so the submenu closed. Now that we properly reset the previous pointer position, this doesn't occur.

Previous (broken) behavior:

https://github.com/adobe/react-spectrum/assets/8961049/2382e1ce-cdaa-4440-9d9b-98ee4dfdca95

New (working) behavior:

https://github.com/adobe/react-spectrum/assets/8961049/d853dde6-78e9-4c69-9375-7c92c8564cda

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Try the reproduction steps above in RAC and RSP submenu stories.


yihuiliao (approved):
lgtm

PR #5878, opened 2024-02-14
fix(#5877): Picker in TableView creates a focus trap with arrow key navigation

### LINKED ISSUE #5877: Picker in TableView creates a focus trap with arrow key navigation

- A Picker in a TableView or ListView creates a focus trap when using ArrowLeft for grid navigation.
- ArrowRight will change the value of a Picker even when focus has moved to another cell or element within the grid.

### 🤔 Expected Behavior?

It should be possible to navigate from a cell containing a Picker to the previous or next cell in the row using ArrowLeft or ArrowRight:

1. without focus getting trapped on the Picker.
2. without the value of the Picker changing unexpectedly when navigating to a new cell.

### 😯 Current Behavior

1. Focus remains on the Picker when trying to use ArrowLeft to navigate to the previous cell in a row.
2. The value of the Picker changes as the focus moves to the next cell using ArrowRight.

### 💁 Possible Solution

1. In `useSelect`, wait a frame, and verify that focus remains on the Picker before changing the value on ArrowRight or ArrowLeft.
2. In `HiddenSelect`, focusing the `input` should verify that the `relatedTarget` is not the `triggerRef.current`, before marshalling focus to the `triggerRef.current`.

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

https://codesandbox.io/p/sandbox/pedantic-euler-csggfh?file=%2Fsrc%2FApp.tsx

1. Open the example app at https://csggfh.csb.app/
7. Press ArrowLeft to try to move focus to the previous cell in the row.
8. Notice that focus remains on the Picker when trying to use ArrowLeft.
9. Press ArrowRight to move focus to the next cell, containing the word "Three."
10. Notice that the value of the Picker changes as the focus moves to the next cell.

### Version

3.34.0

### If other, please specify.

_No response_

macOS 14.4 Beta (23E5191e)

### 🧢 Your Company/Team

Adobe/Accessibility

### 🕷 Tracking Issue

_No response_

majornista:
Closes #5877 

## ✅ Pull Request Checklist:

- [x] Included link to corresponding [React Spectrum GitHub Issue 5877](https://github.com/adobe/react-spectrum/issue/5877).

## 📝 Test Instructions:

1. Open the Storybook example at https://reactspectrum.blob.core.windows.net/reactspectrum/5ea877f566df0d8fbd9543963778c4c21bfa0f10/storybook/index.html?path=/story/tableview--focusable-cells&providerSwitcher-express=false&strict=true
7. Press ArrowRight to move focus to the next cell, containing the word "Three."
8. The value of the Picker does not change as the focus moves to the next cell.
9. Press ArrowLeft to focus the Picker in the second cell again.
10. Press ArrowLeft again to move focus to the previous cell.
11. Focus moves to the Switch in the first cell, and the value of the Picker does not change.

## 🧢 Your Project:

Adobe/Accessibility

### DIFF
--- a/packages/@react-aria/select/src/useSelect.ts
+++ b/packages/@react-aria/select/src/useSelect.ts
@@ -13,9 +13,9 @@
 import {AriaButtonProps} from '@react-types/button';
 import {AriaListBoxOptions} from '@react-aria/listbox';
 import {AriaSelectProps} from '@react-types/select';
-import {chain, filterDOMProps, mergeProps, useId} from '@react-aria/utils';
-import {DOMAttributes, FocusableElement, KeyboardDelegate, ValidationResult} from '@react-types/shared';
-import {FocusEvent, RefObject, useMemo} from 'react';
+import {chain, filterDOMProps, getOwnerDocument, mergeProps, useId} from '@react-aria/utils';
+import {DOMAttributes, FocusableElement, Key, KeyboardDelegate, ValidationResult} from '@react-types/shared';
+import {FocusEvent, KeyboardEvent, RefObject, useEffect, useMemo, useRef} from 'react';
 import {ListKeyboardDelegate, useTypeSelect} from '@react-aria/selection';
 import {SelectState} from '@react-stately/select';
 import {setInteractionModality} from '@react-aria/interactions';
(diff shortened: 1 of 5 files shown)


packages/@react-aria/select/src/useSelect.ts
```diff
@@ -89,31 +89,45 @@ export function useSelect<T>(props: AriaSelectOptions<T>, state: SelectState<T>,
...
+
+    if (key) {
+      // Wait until the next render to change the selection, so that any other key handlers are done running.
+      requestRef.current = requestAnimationFrame(() => {
+        // If focus is no longer on the event target, don't change the selection.
+        if (target !== getOwnerDocument(target as Element).activeElement) {
+          return;
+        }
+        state.setSelectedKey(key);
+      });
```
LFDanLu, line 123: was this RAF the only way to get this to work? Did you try having table capture/handle all keyboard events (at least arrow key navigation)?

majornista: TLDR; I tried a bunch of things to try to correct the order of operations so that a keyboard event on the table wouldn't propagate to the picker in the cell, but nothing worked.

`useGridCell` handles keyboard events using `onKeyDownCapture` and calls `stopPropagation`: 
https://github.com/adobe/react-spectrum/blob/fd3f09c798da23bf2148ba11f65e8856f8af9de6/packages/%40react-aria/grid/src/useGridCell.ts#L107C1-L245C6

`useSelectableCollection` handles the event in the bubble phase and does not `stopPropagation`: https://github.com/adobe/react-spectrum/blob/fd3f09c798da23bf2148ba11f65e8856f8af9de6/packages/%40react-aria/selection/src/useSelectableCollection.ts#L123-L291

`useSelect` seems to move events returned from the `useTypeSelect` from the capture phase to the bubble phase [here](https://github.com/adobe/react-spectrum/pull/5878/files/5ea877f566df0d8fbd9543963778c4c21bfa0f10#diff-7e599b9da08764665cb17b548a888c3caaa9a9fb472d331de37619fec7f3318eL133-L134): 
  ```ts
  typeSelectProps.onKeyDown = typeSelectProps.onKeyDownCapture;
  delete typeSelectProps.onKeyDownCapture;
  ```

LFDanLu: urgh, yup this is bringing back memories of edit mode pain points as well with all the differences in when components each handle the keydown handlers. I'll try to think on this some more, rather not have the RAF if at all possible, maybe we should pick up edit mode again...

devongovett:
I have an alternative solution that is less specific to picker. Basically the same thing as we already do for the up/down arrows. #6116

Issue #4789, opened 2023-07-19
Definition of the number of weeks in a month


atl31:
# 🙋 Feature Request

I haven't found a way to set a minimum and maximum number of weeks per month. In some cases we need to have a calendar whose height does not change, to meet this need I would find it good to be able to define a minimum and maximum number of weeks that we wish to have in the month.

## 🤔 Expected Behavior

Have props that allow you to define the minimum number of weeks in a month and another field for the maximum value

## 😯 Current Behavior

The react-aria defines the weeks per month, the height of the calendar is not predictable and changes according to the months and the year selected

## 💁 Possible Solution

For technical solutions, I don't know how it works internally of the lib, sorry :/

## 🔦 Context

I have a time picker next to my calendar (custom implementation) which needs to be the same size as my calendar, I couldn't find in react aria/datepicker a way to tell my calendar to take 7 weeks for example.
So I had to do with js, observing the height of the calendar to define the height of my time picker, as the latter is not predictable, I have to adapt accordingly. The other problem this causes in my case is that visually it's not consistent with what we're currently doing, the time picker is getting bigger and the popup containing the calendar too, I think it's not specifically accepted by everyone.
I tried to play with the visibleDuration property but this does not impact the number of weeks in a month but just disables them which exceeds, in my case, I would like to say does not show less than 6 weeks and at most 6 weeks.

I think it would be really a big added value because it's the basic operation of the browser's datetime locale input type and also the only headless react library competing with yours, so there are real applications in the real world.

I'm afraid that for this little detail, I'll have to switch to the other headless and I'll be really annoyed

## 💻 Examples

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local => by default
https://github.com/rehookify/datepicker/tree/main **=> static mode**

reidbarber:
Would adding a fixed or min height to your calendar solve this? (of cell height * max weeks)

atl31:
Thank you for your quick response, it is a solution indeed but in my opinion it is more a circumvention than a real solution. Setting a fixed height solves the problem of the popup changing dimensions but adds a new visual problem.
I have a footer below my calendar and I find myself in some cases (current months) with a large margin between my calendar and my footer (see screenshot).
Also I always have to calculate the height of the total weeks of each month to have an effect aligned with the timepicker. For me, this kind of case cannot be handled by style, you have to have a fixed number of weeks per month to have a coherent design.
[...]

Issue #5840, opened 2024-02-08
 Inconsistent Calendar Display in Date Picker: UX Issue with Month Navigation.


YakuBrangJa:
### Provide a general summary of the issue here

The date picker's calendar currently only pads the remaining days of the first and last week of a month. This results in a UX problem when switching between months using buttons. The calendar's next/prev button displacing up and down causes inconvenience, since some months span five weeks while others span up to six weeks.

### 🤔 Expected Behavior?

The date picker's calendar should maintain a consistent layout and size when switching between months, regardless of the number of weeks each month spans.

### 😯 Current Behavior

![](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExMnN6cHo5em15dHM3ZWlnOTBmbjgxdHNmZXhqMHNkamd1cHAzMWQ3eiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/M6rAXmBmusbeNpQLNZ/giphy.gif)

### 💁 Possible Solution

Setting a fix height to calendar's is the only solution for now. Which does leave a bottom blank space for certain months.
A possible solution is to pad an extra week at the end, if the total weeks of the current month are only five. 

### 🖥️ Steps to Reproduce

![](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExMnN6cHo5em15dHM3ZWlnOTBmbjgxdHNmZXhqMHNkamd1cHAzMWQ3eiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/M6rAXmBmusbeNpQLNZ/giphy.gif)

### Version

react-aria-components: ^1.0.0-rc.0

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### What operating system are you using?

Window, MacOs

ryo-manba:
@YakuBrangJa 
I've tested the following links and it appears that the positions of the year and month are fixed.
- https://react-spectrum.adobe.com/react-aria/examples/datepicker.html
- https://react-spectrum.adobe.com/react-aria/examples/datepicker.html

I tested the following link and it appears that the year and month are fixed.
- https://react-spectrum.adobe.com/react-aria/examples/datepicker.html
- https://react-spectrum.adobe.com/react-aria/examples/datepicker.html

Does this not solve the problem?

[...]

Issue #3257, opened 2022-06-22
useCalendarCell equates "outside of visible range" with "disabled".


rglazebrook:
# 🐛 Bug Report 

`useCalendarCell` returns true for `isDisabled` when calendar dates are outside the visible date range (i.e. `isOutsideVisibleRange` is true), but are still within `minValue` and `maxValue`, and are not passed an `isDisabled` prop. This is preventing click/press events on those dates.

## 🤔 Expected Behavior

You should be able to click on dates that are outside the "visible" date range of the current month, but aren't otherwise marked as unavailable or disabled. This would mirror the behavior in the native HTML5 datepicker. 

## 😯 Current Behavior

Dates outside of the visible date range are being marked as isDisabled internally. This prevents any of the click/press events from firing. Based on the description of isDisabled , I believe this could be a bug.

E.g., If I'm looking at June 2022, July 1st can be made visible, but cannot be made clickable (because it is being marked internally as isDisabled). However, July 1st is enabled and clickable from the July 2022 calendar.

## 💁 Possible Solution

Dates outside of the current month should not be marked as isDisabled. From a styling perspective, we still have `isOutsideVisibleRange` to work with, and the dates would still be properly disabled if they fell outside of the min/max values.

If this is expected behavior and changing this would constitute a breaking change, adding a calendar-level prop something like `disableOutsideVisibleRange` and defaulting it to true would provide users a viable workaround.

## 🔦 Context

I am trying to replicate behavior found in the standard HTML5 datepicker, wherein selecting a date from the previous/next month, if it is visible in the calendar, is possible. My goal is to not lose features when switching over.

## 💻 ~~Code~~ GIF Sample

Here's a GIF of my implementation of `useDatepicker`, `useCalendar`, etc:
![aria-dateinput-cant-click](https://user-images.githubusercontent.com/2517006/175111446-eafaf8be-aa2e-4a09-9b4c-f673f773aab2.gif)

And here's an example of the same behavior with a standard HTML5 `<input type="date" />`:
![html-dateinput-clicking](https://user-images.githubusercontent.com/2517006/175111533-6a67a59e-c03c-41e4-b746-b8241001f91e.gif)

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| react-spectrum   | 3.0.0
| Browser          | Chrome
| Operating System | macOS Big Sur

## 🧢 Your Company/Team

N/A

## 🕷 Tracking Issue (optional)

N/A

[...]

PR #5910, opened 2024-02-20
Add support for fixed week display in Calendar

ryo-manba:
Closes https://github.com/adobe/react-spectrum/issues/4789, https://github.com/adobe/react-spectrum/issues/5840

I added props to the Calendar to enable the selection of a fixed 6-week view.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Visit the [Date Picker example](http://localhost:9003/?path=/story/react-aria-components--date-picker-example&providerSwitcher-express=false&strict=true) in your local Storybook.
2. Go to the Controls tab and set `isFixedWeeks` to true.
3. Observe that the calendar now consistently displays 6 weeks.
4. Confirm the 6-week display at the [DateRangePicker example](http://localhost:9003/?path=/story/react-aria-components--date-range-picker-example) as well.

### DIFF
--- a/packages/@react-types/calendar/src/index.d.ts
+++ b/packages/@react-types/calendar/src/index.d.ts
@@ -62,7 +62,12 @@ export interface CalendarPropsBase {
    * Controls the behavior of paging. Pagination either works by advancing the visible page by visibleDuration (default) or one unit of visibleDuration.
    * @default visible
    */
-  pageBehavior?: PageBehavior
+  pageBehavior?: PageBehavior,
+  /**
+   * Whether the calendar should always display 6 weeks.
+   * @default false
+   */
+  isFixedWeeks?: boolean
 }
 
 export type DateRange = RangeValue<DateValue>;
--- a/packages/react-aria-components/stories/DatePicker.stories.tsx
+++ b/packages/react-aria-components/stories/DatePicker.stories.tsx
@@ -16,10 +16,14 @@ import React from 'react';
 import styles from '../example/index.css';
 
 export default {
-  title: 'React Aria Components'
+  title: 'React Aria Components',
+  argTypes: {
+    isFixedWeeks: {control: 'boolean'}
+  }
 };
 
-export const DatePickerExample = () => (
+
+export const DatePickerExample = (args) => (
   <DatePicker data-testid="date-picker-example">
     <Label style={{display: 'block'}}>Date</Label>
     <Group style={{display: 'inline-flex'}}>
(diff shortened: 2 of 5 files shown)


packages/@react-types/calendar/src/index.d.ts
```diff
@@ -62,7 +62,12 @@ export interface CalendarPropsBase {
    * Controls the behavior of paging. Pagination either works by advancing the visible page by visibleDuration (default) or one unit of visibleDuration.
    * @default visible
    */
-  pageBehavior?: PageBehavior
+  pageBehavior?: PageBehavior,
+  /**
+   * Whether the calendar should always display 6 weeks.
+   * @default false
+   */
+  isFixedWeeks?: boolean
```
LFDanLu, line 70: The team is currently discussing whether this should be supported for the React Spectrum side of things as well but perhaps this should be moved to the RAC Calendar props specifically for now

packages/react-aria-components/stories/DatePicker.stories.tsx
```diff
@@ -16,10 +16,13 @@ import React from 'react';
 import styles from '../example/index.css';
 
 export default {
-  title: 'React Aria Components'
+  title: 'React Aria Components',
+  argTypes: {
+    isFixedWeeks: {control: 'boolean'}
+  }
```
LFDanLu, line 22: I noticed in the DatePicker stories that the dates outside the current month seem to be non-interactive (aka clicking on them doesn't select the date). This seems to be because the dates that fall outside the month are judged as disabled via https://github.com/adobe/react-spectrum/blob/fd3f09c798da23bf2148ba11f65e8856f8af9de6/packages/%40react-stately/calendar/src/useCalendarState.ts#L307. For this case, we should probably make those days considered as valid dates if `isFixedWeeks` is true but will bring this up to the team

LFDanLu (changes requested):
note that this feedback might change depending on team discussion

packages/react-aria-components/stories/DatePicker.stories.tsx
ryo-manba: It seems that in previous discussions, it was concluded that enabling dates outside the month/visible range would introduce too many UX challenges.

> The team discussed this again at length today, and we decided that there are too many UX problems with enabling dates outside the month / visible range.

https://github.com/adobe/react-spectrum/issues/3257#issuecomment-1644757484

LFDanLu: thanks for the reminder, I forgot about that investigation we did last year. In that case I'm not super thrilled about adding this behavior as a whole then, doesn't really feel very useful if those dates would always be disabled + dimmed. I guess it wouldn't hurt since it could stabilize the number of cells shown but I feel like users may then want to have those dates be interactive...

ryo-manba: You're right, making the displayed dates interactive could indeed be more useful. Perhaps focusing on a single datepicker, rather than a range picker for selecting multiple dates, would be a better approach this time.

LFDanLu:
Discussed this again recently, the team is wondering if the amount of weeks should be based on the calendar system or if a user should be able to freely customize it instead. Will need more discussion to decide what direction we want to go in, but for now closing until we come to a decision.

Issue #5921, opened 2024-02-22
useOverlay: IMEs escape key behavior not taken into account


rktyt:
### Provide a general summary of the issue here

Related to #5904.
This issue is about Escape.

Firefox's behavior is correct and other browsers' behavior is incorrect.
As for Firefox, I have checked with Developer Edition.

### 🤔 Expected Behavior?

Pressing the Escape key does not close dialogs while operating with the IME.
The Escape key cancels the selection of input candidates.

### 😯 Current Behavior

Dialogs can be closed with the Escape key even while operating with the IME.

### 💁 Possible Solution

https://github.com/adobe/react-spectrum/blob/%40react-aria/overlays%403.21.0/packages/%40react-aria/overlays/src/useOverlay.ts#L115

```patch
-        if (e.key === "Escape" && !isKeyboardDismissDisabled) {
+        if (e.key === "Escape" && !isKeyboardDismissDisabled && !e.nativeEvent.isComposing) {
```

----

Please someone open a PR. (I can't)

### 🖥️ Steps to Reproduce

Updated!

It may not be possible to confirm unless #5909 is reflected.
Therefore [the storybook of #5909](https://reactspectrum.blob.core.windows.net/reactspectrum/56a8390f3825e1d31c2e138fad4a4da0af600cce/storybook/index.html?path=/story/dialog--form&providerSwitcher-express=false) to check.

1. Open OS keyboard settings
2. Add Chinese Pinyin - Simplified
3. Type "ni", a set of suggestions should appear
4. Press Tab 3x
5. Press Escape and see the state of the dialog

### Version

react-aria@3.31.1

### What browsers are you seeing the problem on?

Chrome, Safari, Microsoft Edge

### What operating system are you using?

MacOS

PR #5923, opened 2024-02-22
Fix escape key to close IME in useOverlay

ryo-manba:
Closes https://github.com/adobe/react-spectrum/issues/5921

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Open OS keyboard settings
2. Add Chinese Pinyin - Simplified
3. Visit http://localhost:9003/?path=/story/dialog--form&providerSwitcher-express=false&strict=true
4. Type "ni", a set of suggestions should appear
5. Press Tab 3x
6. Press Escape and see the state of the dialog

### DIFF
--- a/packages/@react-aria/overlays/src/useOverlay.ts
+++ b/packages/@react-aria/overlays/src/useOverlay.ts
@@ -112,7 +112,7 @@ export function useOverlay(props: AriaOverlayProps, ref: RefObject<Element>): Ov
 
   // Handle the escape key
   let onKeyDown = (e) => {
-    if (e.key === 'Escape' && !isKeyboardDismissDisabled) {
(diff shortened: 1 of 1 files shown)


LFDanLu:
GET_BUILD

LFDanLu (approved):
Tested locally and confirmed that this works. For others who may test this, note that exiting the IME dropdown with ESC doesn't remove you from the composing state. You have to hit ESC again which will clear the field. This seems to be the correct behavior, I'm more used to the Windows IME dropdown which doesn't close until you hit ESC twice before closing the dropdown which then clears the field (first escape reverts to whatever raw input you had typed -> second clears and closes the dropdown) which is pretty much the same behavior except for when the dropdown closes

snowystinger (approved):

PR #5933, opened 2024-02-23
Show all toasts

ktabors:
Closes

Do we want to add additional animations to this?

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Goto a Toast story and add a Toast.
- Toast should be added from the bottom.
- Multiple toasts can be shown at the same time.
  - Toasts are sorted by order added.
- Toast has enter and exit animations.
- Toast should exit by timeout if defined or by close button (unless there is an action).

## 🧢 Your Project:
RSP

### DIFF
--- a/packages/@react-spectrum/toast/docs/Toast.mdx
+++ b/packages/@react-spectrum/toast/docs/Toast.mdx
@@ -58,7 +58,7 @@ Then, queue a toast from anywhere:
 
 Toasts are triggered using one of the methods of <TypeLink links={docs.links} type={docs.exports.ToastQueue} />. A &lt;<TypeLink links={docs.links} type={docs.exports.ToastContainer} />&gt; element must be rendered at the root of your app in order to display the queued toasts.
 
-Toasts are shown according to the order they are added from the bottommost position. Since all opens Toasts are displayed until dismissed, please be mindful of how often toasts are triggered, see [Spectrum design docs](https://spectrum.corp.adobe.com/page/toast/#Too-many-toasts).
+Toasts are shown according to the order they are added, with the most recent toast appearing at the bottom of the stack. Please use Toasts sparingly, see [Spectrum design docs](https://spectrum.corp.adobe.com/page/toast/#Too-many-toasts).
 
 ```tsx example
 <ButtonGroup>
--- a/packages/@react-spectrum/toast/docs/Toast.mdx
+++ b/packages/@react-spectrum/toast/docs/Toast.mdx
@@ -58,7 +58,7 @@ Then, queue a toast from anywhere:
 
 Toasts are triggered using one of the methods of <TypeLink links={docs.links} type={docs.exports.ToastQueue} />. A &lt;<TypeLink links={docs.links} type={docs.exports.ToastContainer} />&gt; element must be rendered at the root of your app in order to display the queued toasts.
 
-Toasts are shown according to a priority queue, depending on their variant. Actionable toasts are prioritized over non-actionable toasts, and errors are prioritized over other types of notifications. Only one toast is displayed at a time. See the [Spectrum design docs](https://spectrum.adobe.com/page/toast/#Priority-queue) for full information on toast priorities.
+Toasts are shown according to the order they are added from the bottommost position. Since all opens Toasts are displayed until dismissed, please be mindful of how often toasts are triggered, see [Spectrum design docs](https://spectrum.corp.adobe.com/page/toast/#Too-many-toasts).
 
 ```tsx example
 <ButtonGroup>
--- a/packages/@react-stately/toast/src/useToastState.ts
+++ b/packages/@react-stately/toast/src/useToastState.ts
@@ -134,7 +134,11 @@ export class ToastQueue<T> {
       }
     }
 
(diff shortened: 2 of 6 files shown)


packages/@react-spectrum/toast/docs/Toast.mdx
```diff
@@ -58,7 +58,7 @@ Then, queue a toast from anywhere:
 
 Toasts are triggered using one of the methods of <TypeLink links={docs.links} type={docs.exports.ToastQueue} />. A &lt;<TypeLink links={docs.links} type={docs.exports.ToastContainer} />&gt; element must be rendered at the root of your app in order to display the queued toasts.
 
-Toasts are shown according to a priority queue, depending on their variant. Actionable toasts are prioritized over non-actionable toasts, and errors are prioritized over other types of notifications. Only one toast is displayed at a time. See the [Spectrum design docs](https://spectrum.adobe.com/page/toast/#Priority-queue) for full information on toast priorities.
+Toasts are shown according to the order they are added from the bottommost position. Since all opens Toasts are displayed until dismissed, please be mindful of how often toasts are triggered, see [Spectrum design docs](https://spectrum.corp.adobe.com/page/toast/#Too-many-toasts).
```
LFDanLu, line 61:
```suggestion
Toasts are shown according to the order they are added, with the most recent toast appearing at the bottom of the stack. Since all some Toasts may be displayed until dismissed, please be mindful of how often toasts are triggered, see [Spectrum design docs](https://spectrum.adobe.com/page/toast/#Too-many-toasts).
```
Some small changes for clarity with auto dismissing toasts in mind

packages/@react-stately/toast/src/useToastState.ts
```diff
@@ -157,23 +161,28 @@ export class ToastQueue<T> {
...
-        .filter(t => !toasts.some(t2 => t.key === t2.key))
-        .map(t => ({...t, animation: 'exiting'}));
+          .filter(t => !toasts.some(t2 => t.key === t2.key))
+          .map(t => ({...t, animation: 'exiting'}));
+
+      if (oldIndex !== -1) {
+        toasts.splice(oldIndex, 0, prevToasts?.[0]);
+      }
 
-      this.visibleToasts = prevToasts.concat(toasts).sort((a, b) => b.priority - a.priority);
```
LFDanLu, line 176: So technically this updateVisibleToast function would work even if multiple toasts were being added/removed at the same time but now we use an index value so it assumes that we are only adding/removing one at a time right? I was a bit concerned but I guess the add/remove have always only supported adding one at a time anyways. Lemme know if I'm following these changes correctly

LFDanLu (changes requested):
Just one comment on the docs and a clarification question. I was surprised that we couldn't just keep everything the same on main and just set the priority of the toasts being added by the ToastContainer to the same value but I see you have changes here to flip it so that non prioritized toasts are added to the bottom instead of the top of the stack (hence your question at standup haha). Didn't realize it would involve changes in useToastState though, but it seems like an aria implementation with toasts with priorities set still work, though I guess a non-prioritized toast stack would now follow our bottom to top stack order now...

packages/@react-stately/toast/src/useToastState.ts
ktabors: The stately API currently exposes add, remove, close of one item at a time. When I add a whole bunch with five second timeouts quickly or try to close them all with the close button this change works.

The reason this code was modified is because it was moving closed and removed Toasts to the bottom of the visible queue then doing the exit animation. When you have one item showing and being removed that works, but with multiple visible items and one is removed, the removed item jumps to the bottom of the visible list. We need to keep the item being removed in the current position.

LFDanLu (approved):
LGTM

reidbarber (approved):
LGTM. Would be nice if we could have a transition on the toasts that are already in the stack, but we can handle that separately.

ktabors:
> LGTM. Would be nice if we could have a transition on the toasts that are already in the stack, but we can handle that separately.

I expected requests for additional animations to come in via this PR review or in a testing session.

packages/@react-spectrum/toast/docs/Toast.mdx
```diff
@@ -58,7 +58,7 @@ Then, queue a toast from anywhere:
 
 Toasts are triggered using one of the methods of <TypeLink links={docs.links} type={docs.exports.ToastQueue} />. A &lt;<TypeLink links={docs.links} type={docs.exports.ToastContainer} />&gt; element must be rendered at the root of your app in order to display the queued toasts.
 
-Toasts are shown according to a priority queue, depending on their variant. Actionable toasts are prioritized over non-actionable toasts, and errors are prioritized over other types of notifications. Only one toast is displayed at a time. See the [Spectrum design docs](https://spectrum.adobe.com/page/toast/#Priority-queue) for full information on toast priorities.
+Toasts are shown according to the order they are added, with the most recent toast appearing at the bottom of the stack. Please use Toasts sparingly, see [Spectrum design docs](https://spectrum.corp.adobe.com/page/toast/#Too-many-toasts).
```
devongovett, line 61: can't put corp links in public docs.

PR #5959, opened 2024-02-27
Fix submit button being focused on an implicit form submission (useButton)

sookmax:
Closes https://github.com/adobe/react-spectrum/issues/5940

I'm not sure this is the direction @snowystinger wanted in https://github.com/adobe/react-spectrum/issues/5940#issuecomment-1963071411, but I found it rather difficult to implement the fix inside the `usePress()` hook. I ended up placing the fix in `useButton()` instead.

I think it's a reasonable fix, but I might be missing something so any comments, suggestions, and opinions are welcome!

P.S. I wanted to add a unit test on this implementation, but realized `jsdom` didn't support `HTMLFormElement.prototype.requestSubmit`, which is needed for the implicit form submission test.

## ✅ Pull Request Checklist:

### Story
http://localhost:9003/?path=/story/usebutton--submit-type-button&providerSwitcher-express=false&strict=true

https://github.com/adobe/react-spectrum/assets/71210554/f2d0c67e-2749-44a5-aebc-199de8a3b2b2


packages/@react-aria/button/src/useButton.ts
```diff
@@ -101,6 +106,31 @@ export function useButton(props: AriaButtonOptions<ElementType>, ref: RefObject<
...
+        setPreventFocusOnPress(true);
+      }
+      setClickByUser(false);
+    }
+    if (ref?.current && ref.current instanceof HTMLButtonElement) {
+      let button =  ref.current;
+      button.addEventListener('pointerdown', onPointerDown);
+      button.addEventListener('click', onClick);
+      return () => {
+        button.removeEventListener('pointerdown', onPointerDown);
```
snowystinger, line 129: instead of needing a useEffect to hook up the events, I think we should be able to just pass them into the mergeProps down below, there's already a 'click' listener there

snowystinger (commented):
I think you're correct to make the fix in `useButton`, I didn't particularly like it in `usePress`, it was too specific

packages/@react-aria/button/src/useButton.ts
sookmax: Ha this is weird.. 

But first of all, thanks for the comment! I also think passing these callbacks in `mergeProps()` in the return statement is a better idea, and so I went ahead and tried to do just that, only to find myself puzzled by this weird behavior.. that I'm about to describe:

The implicit form submission (a.k.a, pressing the Enter key) is not detected and corrected in first try; by first try I mean when the input first gains (or re-gains) the focus. On the other hand, any subsequent tries (after the first failure) seem to work fine. Here's a video:

https://github.com/adobe/react-spectrum/assets/71210554/5a4e583d-a2eb-40dc-926d-656925bc00e4

In code, it looks something like this:
```ts
// In useButton()

...

let [clickByUser, setClickByUser] = useState(false);
let [preventFocusOnPress, setPreventFocusOnPress] = useState(false);

let detectClickByUser = useCallback((e: PointerEvent) => {
  if (e.currentTarget instanceof HTMLButtonElement) {
    setClickByUser(true);
    setPreventFocusOnPress(false);
  }
}, []);

let detectImplicitFormSubmission = useCallback((e: MouseEvent) => {
  if (e.currentTarget instanceof HTMLButtonElement) {
    let button = e.currentTarget;
    // this if-statement is to test whether the click event was generated by an implicit form submission triggered by the user pressing `Enter` key
    // more info on this: https://html.spec.whatwg.org/#implicit-submission
    if (button.type === 'submit' && button.form && !clickByUser) {
      setPreventFocusOnPress(true);
    }
    setClickByUser(false);
  }
}, [clickByUser]);

let {pressProps, isPressed} = usePress({
  onPressStart,
  onPressEnd,
  onPressChange,
  onPress,
  onPressUp,
  isDisabled,
  preventFocusOnPress: preventFocusOnPressProps !== undefined ? preventFocusOnPressProps : preventFocusOnPress,
  ref
});

...

return {
  isPressed, // Used to indicate press state for visual
  buttonProps: mergeProps(
    {
      onPointerDown: detectClickByUser,
      onClick: detectImplicitFormSubmission
    },
    additionalProps, 
    buttonProps, {
      'aria-haspopup': props['aria-haspopup'],
      'aria-expanded': props['aria-expanded'],
      'aria-controls': props['aria-controls'],
      'aria-pressed': props['aria-pressed'],
      onClick: (e) => {
        if (deprecatedOnClick) {
          deprecatedOnClick(e);
          console.warn('onClick is deprecated, please use onPress');
        }
      }
    }
  )
};
```

I've also tried placing state updaters (`setClickByUser()` and `setPreventFocusOnPress()` inside `flushSync()`, but to no avail. 

Also placing the callbacks as the first argument to `mergeProps()` (since `chain()` calls the callbacks in the order of their appearance) doesn't seem to have any effect as well.

I'm honestly clueless as to why this is happening with the above change.. so any help here to shed light on the root cause of this issue would be much appreciated 🙏.

sookmax: This is mostly a wild speculation, but I suspect the above code doesn't work because we're trying to _update_ (in `detectImplicitFormSubmission()`) and _access_ (in `pressProps.onClick()`) the state value (`preventFocusOnPress`) in a single rendering path. 

On the other hand, if we register our own handlers (bypassing React's event-managing system) in `useEffect()` then when an event occurs somehow the click handler in the `useEffect()` is called first, which results in a state update, React detects the update and updates the handler to have most current state value before it fires an event to the handler it manages (i.e., `pressProps.onClick()`).

Not sure if there's a way around not resorting to using `useEffect()` in this case.

snowystinger (commented):
Apologies for the delay, I've been mulling this over for a little bit now. I'm no longer so sure that the fix belongs in useButton. This is a thing that usePress is imposing and now we're writing useButton code that is accounting for it.

Ideally, virtual clicks from VO or input labels could be distinguished from implicit form submission. There may be others that I'm unaware of.

I do have a workaround though. If you provide a second submit button with 'display: none' on it, and this second button is placed earlier in the dom, then the bug is "fixed". This is due to how the browser determines which submit button is the primary one. And it addresses the focus issue because the display none element can't receive focus.

Because this is such a wide touching piece of code, I'd prefer to hold off until we've got a better understanding of the implications of this PR.

another idea I had:
- Add a useForm hook which would move focus back to an input after form submission. We could listen for the 'Enter' key on any element inside the form, if it's not the submit button, record that element as the place to return focus.

I'll see about getting this into a team discussion, but I think it can be lower priority with the workaround. Thanks for getting the discussion started and providing something to work off of. I'll comment the workaround in the Issue as well.

sookmax:
@snowystinger Thanks for the detailed comment. I like the workaround you suggested!

I also like the idea of `useForm()`, and if I'm not mistaken it sounds like something that can also be implemented in the client code right? 

So, you're talking about something like the following?
```tsx
<form
  onKeyUp={e => {
    if (e.key === 'Enter' && e.target !== submitButtonRef.current) {
      (e.target as HTMLElement).focus();
    }
}}>
  <input ref={setInputRef} type="text" />
  <button {...buttonProps} ref={ref}>
    Submit
  </button>
</form>
```

Hm, a quick test on the story shows this wouldn't work because by the time `onKeyUp()` on the form catches the event, it seems like `e.target` has already been assigned to the submit button even though the 'Enter' was pressed while the input was being focused (weird). Also it looks like attaching an event listener to a non-interactive element like `<form>` is considered bad practice? I'm seeing [jsx-a11y/no-noninteractive-element-interactions](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/0d5321a5457c5f0da0ca216053cc5b4f571b53ae/docs/rules/no-noninteractive-element-interactions.md#jsx-a11yno-noninteractive-element-interactions) lint warning as soon as I added `onKeyUp` to the form.

> Ideally, virtual clicks from VO or input labels could be distinguished from implicit form submission. There may be others that I'm unaware of.

I think you probably already know about this, but I've found clicking the button with a trackpad (on Macbook) is considered as 'virtual' (`state.pointerType`) while clicking the button with a normal mouse would produce `state.pointerType` of `null`.

snowystinger:
That warning should be a moot point if we pass it as we usually do with hooks.
I was thinking more along these lines:
```
let lastFocused = useRef();
<form
  onSubmit={() => lastFocused.current?.focus();}
  onKeyUpCapture={e => {
    // could also listen to the onFocus/onBlur to track the lastFocused instead of the key events
    if (e.key === 'Enter' && e.target !== submitButtonRef.current) {
      lastFocused.current = e.target
    }
}}>
```

> I think you probably already know about this, but I've found clicking the button with a trackpad (on Macbook) is considered as 'virtual' (state.pointerType) while clicking the button with a normal mouse would produce state.pointerType of null.

Huh, I was not seeing this, my trackpad returns mouse. I'm not sure what settings would change that

sookmax:
Hm, seems like using `onKeyCapture` to capture `e.target` that is not the submit button also doesn't work when I did a quick test on the story. By the time the event handler is called, `e.target` is, again, already the submit button even though the 'Enter' was pressed on the input.

> Huh, I was not seeing this, my trackpad returns mouse. I'm not sure what settings would change that

Haha sorry about that. I always turn on 'tap to click' setting whenever I use a Macbook:
tap-to-click

And that 'tap' as a click is recognized as a 'virtual' `state.pointerType`.

yihuiliao:
I just wanted to let you know that the team will be slow to review as we address some other priorities at the moment. We appreciate your understanding and apologize for the wait.

sookmax:
@yihuiliao haha it's okay! Thanks for letting me know 😄.

LFDanLu:
Fixed by https://github.com/adobe/react-spectrum/pull/7542

PR #6011, opened 2024-03-06
Toast focus improvements

ktabors:
Focus issues found during testing

This behaves slightly better than v2 Toasts. When removing Toasts it behaves the same as far as not focusing another toast, but staying within the container. With the last one we now restore focus to the last focused item outside the ToastContainer.

This isn't doing what the display one Toast at a time did, which is focus Toasts loading from the queue or focusing the container when the last toast is removed.

If we like this, I'll remove all the focus code from the container and focusing an actual Toast not the close button.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Add and remove several Toasts and see where focus goes.

In Toast storybook example at: https://reactspectrum.blob.core.windows.net/reactspectrum/b849db005f61cfeed14f599878c5085e14d1d154/storybook/index.html?path=/story/toast--with-action&args=shouldCloseOnAction:true&providerSwitcher-express=false

With a screen reader running:

Toast live region announcement should include,

- The icon alt text for severity level, "Success", "Error" or "Info"
- The toast message text
- The button label for the action button, if an action button is present.

Toast live region announcement should not include the close button.

The Toast landmark region should labelled "Notification," by default and should politely announce the number of notifications when focus moves into the landmark region, “# notifications.”

Focus management use cases:

- User triggers a toast through some action
  - Toast is announced, see above. Focus remains on the trigger that caused Toast.
- User does landmark navigation (F6) to switch to Toasts landmark region.
  - Focus moves to Toast landmark and announces itself, see above.
  - Only possible if focus hasn't previous visited any of the open Toasts.
- User keyboard navigates from application to Toasts.
  - If there are more than one Toast we navigate to the bottom Toast with is the most recent Toast.
  - Toast is focused and announces, see above.
  - User can tab navigate to the toast button(s) before navigating to another Toast or to leave region.
- User is on Toast landmark region or a Toast and switches landmark regions.
  - Focus moves to the next region or returns to the last focused item prior to navigating to a Toast or Toast landmark region.
- User deletes a Toast
  - If it is the last Toast in the Toast Container, focus returns to last focused item outside of Toast landmark region.
  - If there are additional Toasts remaining in the Toast Container:
    - Focus the Toast above it.
    - If Toast being removed is the topmost toast, then move focus to the Toast below it.
  - When a Toast is removed on action, focus should be restored to an adjacent Toast as well if visible Toasts remain: (see: https://github.com/adobe/react-spectrum/pull/6011#discussion_r1536923620)
    
Toast are organized within an ordered list within the landmark region, we should not need to announce the number of notifications except when moving focus to the landmark region. Depending on the screen reader, the new count for the number of items in the list might be announced automatically when the list count changes.

Focus can move to the Notifications landmark region:

-  by using F6 or Shift+F6 to cycle into the Notifications landmark,
- by tabbing into the Notifications landmark following the DOM order,
- or by using the mouse to focus a Toast.

Regardless of how focus entered the Notifications landmark, F6 should eventually cycle back to the last focused element in the main landmark, and focus management when a Toast is removed should behave the same regardless of the manner in which the Toast received focus. For example, if I were to remove a Toast using a mouse click on the close button, focus should move to the adjacent Toast, rather than be lost to the document.body.

## 🧢 Your Project:
RSP

### DIFF
--- a/packages/@react-aria/toast/intl/en-US.json
+++ b/packages/@react-aria/toast/intl/en-US.json
@@ -1,4 +1,5 @@
 {
   "close": "Close",
-  "notifications": "Notifications"
+  "notifications": "Notifications",
+  "countAnnouncement": "{count, plural, one {# notification} other {# notifications}}."
 }
--- a/packages/@react-aria/toast/src/useToast.ts
+++ b/packages/@react-aria/toast/src/useToast.ts
@@ -26,8 +26,10 @@ export interface AriaToastProps<T> extends AriaLabelingProps {
 }
 
 export interface ToastAria {
-  /** Props for the toast container element. */
+  /** Props for the toast container, non-modal dialog element. */
   toastProps: DOMAttributes,
+  /** Props for the toast content alert message. */
+  contentProps: DOMAttributes,
   /** Props for the toast title element. */
   titleProps: DOMAttributes,
   /** Props for the toast description element, if any. */
@@ -94,16 +96,18 @@ export function useToast<T>(props: AriaToastProps<T>, state: ToastState<T>, ref:
     // Only move focus when there is another Toast. At this point,
     // state.visibleToasts still includes Toast being removed.
     if (state.visibleToasts?.length > 1) {
-      let nextItemFocused = focusManager.focusNext();
+      const from = document.activeElement?.closest('[role="alertdialog"]') || document.activeElement;
+      const accept = (node:Element) => node.getAttribute('role') === 'alertdialog';
+      let nextItemFocused = focusManager.focusNext({from, accept});
       if (!nextItemFocused || Object.keys(nextItemFocused).length === 0) {
-        focusManager.focusPrevious();
+        focusManager.focusPrevious({from, accept});
       }
     }
   };
 
   return {
     toastProps: {
-      role: 'alert',
+      role: 'alertdialog',
       'aria-label': props['aria-label'],
       'aria-labelledby': props['aria-labelledby'] || titleId,
       'aria-describedby': props['aria-describedby'] || descriptionId,
@@ -111,6 +115,11 @@ export function useToast<T>(props: AriaToastProps<T>, state: ToastState<T>, ref:
       // Hide toasts that are animating out so VoiceOver doesn't announce them.
       'aria-hidden': animation === 'exiting' ? 'true' : undefined,
       tabIndex: 0
+    },
+    contentProps: {
+      role: 'alert',
+      'aria-atomic': 'true',
+      'aria-relevant': 'additions'
     },
     titleProps: {
       id: titleId
--- a/packages/@react-aria/toast/src/useToast.ts
+++ b/packages/@react-aria/toast/src/useToast.ts
@@ -15,7 +15,7 @@ import {AriaLabelingProps, DOMAttributes, FocusableElement} from '@react-types/s
 // @ts-ignore
 import intlMessages from '../intl/*.json';
 import {QueuedToast, ToastState} from '@react-stately/toast';
-import {RefObject, useEffect, useRef} from 'react';
+import {RefObject, useEffect, useRef, useState} from 'react';
 import {useFocusManager} from '@react-aria/focus';
 import {useId, useLayoutEffect, useSlotId} from '@react-aria/utils';
 import {useLocalizedStringFormatter} from '@react-aria/i18n';
@@ -78,14 +78,23 @@ export function useToast<T>(props: AriaToastProps<T>, state: ToastState<T>, ref:
     };
   }, [ref, state.visibleToasts]);
 
+  const [isBusy, setIsBusy] = useState(true);
+
   // eslint-disable-next-line
   useEffect(() => {
+    let timeoutId:ReturnType<typeof setTimeout>;
+    if (isBusy) {
+      timeoutId = setTimeout(() => setIsBusy(false), 300);
+    }
     return () => {
+      if (timeoutId) {
+        clearTimeout(timeoutId);
+      }
       if (focusOnUnmount.current) {
         focusOnUnmount.current.focus();
       }
     };
-  }, [ref]);
+  }, [ref, isBusy]);
 
   let titleId = useId();
   let descriptionId = useSlotId();
@@ -119,7 +128,7 @@ export function useToast<T>(props: AriaToastProps<T>, state: ToastState<T>, ref:
     contentProps: {
       role: 'alert',
       'aria-atomic': 'true',
-      'aria-relevant': 'additions'
+      'aria-busy': isBusy ?? undefined
     },
     titleProps: {
       id: titleId
--- a/packages/@react-aria/toast/src/useToast.ts
+++ b/packages/@react-aria/toast/src/useToast.ts
@@ -17,6 +17,7 @@ import intlMessages from '../intl/*.json';
 import {QueuedToast, ToastState} from '@react-stately/toast';
 import {RefObject, useEffect, useRef} from 'react';
 import {useId, useLayoutEffect, useSlotId} from '@react-aria/utils';
+import {useFocusManager} from '@react-aria/focus';
 import {useLocalizedStringFormatter} from '@react-aria/i18n';
 
 export interface AriaToastProps<T> extends AriaLabelingProps {
@@ -46,6 +47,7 @@ export function useToast<T>(props: AriaToastProps<T>, state: ToastState<T>, ref:
     timeout,
     animation
   } = props.toast;
+  let focusManager = useFocusManager();
 
   useEffect(() => {
     if (!timer) {
@@ -87,6 +89,18 @@ export function useToast<T>(props: AriaToastProps<T>, state: ToastState<T>, ref:
   let descriptionId = useSlotId();
   let stringFormatter = useLocalizedStringFormatter(intlMessages, '@react-aria/toast');
 
+  let onClosePress = (e) => {
+    state.close(key);
+    // Only move focus when there is another Toast. At this point,
+    // state.visibleToasts still includes Toast being removed.
+    if (state.visibleToasts.length > 1) {
+      let nextItemFocused = focusManager.focusNext();
+      if (!nextItemFocused || Object.keys(nextItemFocused).length === 0) {
+        focusManager.focusPrevious();
+      }
+    }
+  }
+
   return {
     toastProps: {
       role: 'alert',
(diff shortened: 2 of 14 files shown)


packages/@react-spectrum/toast/test/ToastContainer.test.js
```diff
@@ -247,7 +247,7 @@ describe('Toast Provider and Container', function () {
     triggerPress(closeButton);
     fireAnimationEnd(toast);
 
-    expect(document.activeElement).toBe(getByRole('region'));
+    expect(document.activeElement).toBe(document.body);
```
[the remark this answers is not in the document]
ktabors: This is one of the items of discussion. In v2 and this version it goes to the body after removing a toast and not a sibling. If you then press tab you focus another Toast's close button. In other components where we manage focus after delete, we have SelectionManager which we use to handle adjusting the focus to another item.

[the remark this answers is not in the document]
snowystinger: losing it to the body is definitely wrong. we'll need to move it in a meaningful way.
If you place a FocusScope in the ToastContainer, then you'll have access to a focusManager, and you can move focus to a non-deleted toast that way.

packages/@react-aria/toast/intl/en-US.json
```diff
@@ -1,4 +1,5 @@
 {
   "close": "Close",
-  "notifications": "Notifications"
+  "notifications": "Notifications",
+  "countAnnouncement": "{count, plural, one {# notification} other {# notifications}}."
```
snowystinger, line 4: I'm never seeing this announcement, when should it be happening?

snowystinger (commented):
If I remove/clear the top most toast in a stack of > 1, focus is lost to the body.

If I use the storybook controls to set `shouldCloseOnAction` true, and then use an action on a toast, the focus is lost to the body.

Open question, what should happen with timeout and focus on a particular toast, vs on the container, vs on a different toast? Right now, they all freeze and none of them dismiss no matter which toast you have focused.

packages/@react-aria/toast/intl/en-US.json
ktabors: F6 to switch landmark regions is the easiest way.

Regarding your general comment it looks like the addition of focusing Toasts broke the focus management improvements. Waiting for feedback from Michael before fixing.

packages/@react-aria/toast/src/useToast.ts
```diff
@@ -87,15 +91,35 @@ export function useToast<T>(props: AriaToastProps<T>, state: ToastState<T>, ref:
   let descriptionId = useSlotId();
   let stringFormatter = useLocalizedStringFormatter(intlMessages, '@react-aria/toast');
 
+  let onClosePress = () => {
```
snowystinger, line 94: We're still not handling the case where shouldCloseOnAction is clicked instead of the close button

packages/@react-aria/toast/intl/en-US.json
snowystinger: I'm not seeing this announcement in the latest built version on Safari. @majornista how do I get this notification?

snowystinger (commented):
Toasts no longer announce when they appear?

packages/@react-spectrum/toast/src/Toast.tsx
```diff
@@ -78,13 +78,13 @@ function Toast(props: SpectrumToastProps, ref: DOMRef<HTMLDivElement>) {
...
 
-  const handleAction = () => {
+  const handleAction = (evt: PressEvent) => {
     if (onAction) {
       onAction();
     }
 
     if (shouldCloseOnAction) {
-      state.close(key);
+      closeButtonProps.onPress(evt);
```
snowystinger, line 87: but we aren't pressing the close button, so this is a false event we're firing?

majornista: I don't want to duplicate the focus management logic from here: https://github.com/adobe/react-spectrum/pull/6011/files#diff-a1ee157fac4bb14b926cd23ba0263284fa555fbbc8168ed4514858ddda8b583bR103-R115

ktabors:
Michael, to confirm the testing strategy, can we write up a specification for what we're trying to implement? Below are the cases I'm aware of. Can you please update and/or fill in the blanks? This is to help with testing.

Toast announcement
- Message
- Button label
- Severity (matches icon)
 
Toast landmark region
- “# notifications”

Use cases
- User triggers a toast through some action
  - Toast is announced, see above. Focus remains on the trigger that caused Toast.
- User does landmark navigation (F6) to switch to Toasts landmark region.
  - Focus moves to Toast landmark and announces itself, see above.
  - Only possible if there are open Toasts.
- User keyboard navigates from application to Toasts.
  - If there are more than one Toast we navigate to the bottom Toast with is the most recent Toast.
  - Toast is focused and announces, see above.
  - User can tab navigate to the toast button(s) before navigating to another Toast or to leave region.
- User is on Toast landmark region or a Toast and switches landmark regions.
  - Focus moves to the next region or returns to the last focused item prior to navigating to a Toast or Toast landmark region.
- User deletes a Toast
  - If it is the last Toast in the Toast Container, focus returns to last focused item outside of Toast landmark region.
  - If there are additional Toasts remaining in the Toast Container:
    - Focus the Toast above it.
    - If Toast being removed is the topmost toast, then move focus to the Toast below it.

Questions
- Do we announce the number of notifications anytime besides when they navigate to the Toast landmark region?
- Is the only way to navigate to the Toast landmark region via F6 key event? I know there are ways to focus the Toast landmark region, like clicking on the text of a Toast can cause focus to move the landmark region, but that isn’t a keyboard navigation experience and should be considered an outlier.

packages/@react-aria/toast/stories/Example.tsx
```diff
@@ -43,13 +43,13 @@ function ToastRegion() {
...
   let ref = useRef(null);
-  let {toastProps, titleProps, closeButtonProps} = useToast(props, state, ref);
+  let {toastProps, contentProps, titleProps, closeButtonProps} = useToast(props, state, ref);
   let buttonRef = useRef();
   let {buttonProps} = useButton(closeButtonProps, buttonRef);
 
   return (
     <div {...toastProps} ref={ref} style={{margin: 20, display: 'flex', gap: 5}}>
-      <div {...titleProps}>{props.toast.content}</div>
+      <div {...contentProps} {...titleProps}>{props.toast.content}</div>
```
ktabors, line 52: I guess you're not wanting to breaking change the API by renaming `titleProps` by adding the values of `contentProps` to it, but I'm not liking having two sets of props from the same origin go to the same element.

ktabors (commented):
Seems like duplicate messages when triggering a Toast because it announces the Toast (message, button, and status) then announces the Toast status and message. 

When I navigated to the landmark region where I've previously focused a toast, it landed on a toast, anounced the toast, then announced the region (number of notifications) twice.

I noticed we have `list 2 items` and `2 notifications` being announced together, seems redundant.

packages/@react-aria/toast/stories/Example.tsx
majornista: I can wrap the `titleProps` within a separate `contentProps` div.

packages/@react-aria/toast/src/useToast.ts
```diff
@@ -83,19 +86,55 @@ export function useToast<T>(props: AriaToastProps<T>, state: ToastState<T>, ref:
...
       'aria-label': props['aria-label'],
       'aria-labelledby': props['aria-labelledby'] || titleId,
       'aria-describedby': props['aria-describedby'] || descriptionId,
       'aria-details': props['aria-details'],
       // Hide toasts that are animating out so VoiceOver doesn't announce them.
-      'aria-hidden': animation === 'exiting' ? 'true' : undefined
+      'aria-hidden': animation === 'exiting' ? 'true' : undefined,
+      tabIndex: 0
+    },
+    contentProps: {
```
snowystinger, line 134: we'll need to keep this in mind for announcement purposes, it might be argued that it's a breaking change

packages/@react-aria/toast/src/useToastRegion.ts
```diff
@@ -38,10 +39,20 @@ export function useToastRegion<T>(props: AriaToastRegionProps, state: ToastState
...
       state.pauseAll();
       lastFocused.current = e.relatedTarget;
+      if (ref.current?.contains(document.activeElement)) {
+        let count: number = state.visibleToasts.length || 0;
+        let announcement = stringFormatter.format('countAnnouncement', {count});
+        if (timeoutId) {
+          clearTimeout(timeoutId);
+          timeoutId = undefined;
+        }
+        timeoutId = setTimeout(() => announce(announcement, 'polite'), 750);
```
snowystinger, line 54: where'd this timeout value from from?

packages/@react-spectrum/toast/src/Toast.tsx
snowystinger: Can we handle the check/move focus on Toast unmount instead? Other design systems may decide to have two or more actions or will maybe close a toast for some other reason while focus is within it.

packages/@react-spectrum/toast/src/Toast.tsx
```diff
@@ -111,24 +119,27 @@ function Toast(props: SpectrumToastProps, ref: DOMRef<HTMLDivElement>) {
...
+      <div {...contentProps} className={classNames(toastContainerStyles, 'spectrum-Toast-contentWrapper')}>
+        {Icon &&
+          <Icon
+            id={iconId}
+            aria-label={iconLabel}
+            UNSAFE_className={classNames(styles, 'spectrum-Toast-typeIcon')} />
         }
+        <div className={classNames(styles, 'spectrum-Toast-body')} role="presentation">
+          <div className={classNames(styles, 'spectrum-Toast-content')} role="presentation" {...titleProps}>{children}</div>
+          {actionLabel &&
```
snowystinger, line 131: should the action button actually be outside the aria-atomic region? I guess it makes sense if you want someone to know that the alert has something you can do with it

packages/@react-aria/toast/src/useToastRegion.ts
majornista: Probably trial and error in an effort to allow ToastRegion to announce on focus before the number of notifications within it.

snowystinger:
closing in favor of https://github.com/adobe/react-spectrum/pull/6223

PR #6024, opened 2024-03-07
Change useMenuTrigger onPressStart to state.open

### LINKED ISSUE #5988: Select menu closes unexpectedly while holding mouse button and cursor leaving then returning

Hello!

My users report that the behavior of the `Select` is surprising, some think it is broken. I was originally going to open this as Feedback but the more I look at it, the less I think this is intentional and the more I think it is a bug.

The summary: if you press a `Button` to open a `Select`, hold the mouse button down, and move the cursor out of the `Button` and then back in, the menu will close.

### 🤔 Expected Behavior?

The Select menu should not close in this scenario. A user should be able to hold the original press without having to avoid the button.

### 😯 Current Behavior

As you'll see in the [example](https://react-spectrum.adobe.com/react-aria/Select.html#example):

* Click the menu and hold the mouse button down
* Without releasing the button, move the cursor over the menu and then move it back over the button
* The menu will close immediately
* Still holding the button, move the cursor out of the button and then back over top
* The menu will open again on its own

This is a strange experience that differs significantly from the HTML select element, as seen [here](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select).

Users are reporting that they think the Select is broken because it closes on its own before they can select. This will happen if someone is used to the HTML select behavior: pressDown to open menu, release over your option. If they happen to move their hand just right (or wrong, I guess) the menu closes.

### 💁 Possible Solution

What appears to happen is this:

* On first press, the `onPressStart` and `onPressChange (true)` events fire.
* While holding the button down, the `onPressEnd` and `onPressChange (false)` events both fire as soon as the cursor leaves the button.
* The `onPressStart` and `onPressChange (true)` fire again as soon as the cursor returns -- it sees it as a new click, which toggles the menu closed. It does not have the context to know that the menu is still 

When someone clicks and releases on the button, you get all four events in one go: `onPressStart`, `onPressChange (true)`, `onPressEnd`, `onPressChange (false)`. This does not close the menu. I have not looked at the code but the behavior suggests that `onPressStart` or `onPressChange (true)` toggles the menu state through `!currentState`; instead, it should probably use `setOpen(true)` to make it explicitly open.

### 🔦 Context

This disrupts users and hurts their confidence in my product because it deviates from their expectations. It's especially harmful on a product checkout page, where selecting a country or state is more difficult than it needs to be.

### 🖥️ Steps to Reproduce

Visit the docs [example](https://react-spectrum.adobe.com/react-aria/Select.html#example).

* Click the menu and hold the mouse button down
* Without releasing the button, move the cursor over the menu and then move it back over the button
* The menu will close immediately
* Still holding the button, move the cursor out of the button and then back over top
* The menu will open again on its own

### Version

latest

Firefox, Chrome, Safari

### If other, please specify.

_No response_

Mac OS 14.3.1

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

subvertallchris:
Closes #5988 

## ✅ Pull Request Checklist:

- [x] Link to issue https://github.com/adobe/react-spectrum/issues/5988
- [NA] Updated documentation (if it already exists for this component).
- [NA] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/WAI/ARIA/apg/)

## 📝 Test Instructions:

1. Review the problem as described at https://github.com/adobe/react-spectrum/issues/5988
2. Open Storybook and browse to `MenuTrigger`
3. Try to replicate the bug as originally described

I tried to write a new test covering the original condition and failed. I don't have the fluency in react-aria to know how to simulate the behavior. A test did cover the basic behavior, though, and it was updated to reflect the use of `open` instead of `toggle`.


subvertallchris:
@LFDanLu thanks for the advice and recommendation!

My CLA is signed, FYI.

LFDanLu:
@subvertallchris Thanks for the speedy turnaround! I'll see if I can figure out a way to test this flow when I get some spare time

subvertallchris:
Happy to help! Do you think more tests are necessary prior to merge or is this mergeable as is now?

LFDanLu:
GET_BUILD

snowystinger (approved):
Thanks for the PR!

reidbarber:
GET_BUILD

PR #6067, opened 2024-03-16
Allow passing inputRef to RAC `<Radio>`

sookmax:
Closes https://github.com/adobe/react-spectrum/issues/6054, https://github.com/adobe/react-spectrum/issues/6123

~~To implement the behavior described in https://github.com/adobe/react-spectrum/issues/6054#issuecomment-2000575049, I needed to pass `tabbable: true` to `getFocusableTreeWalker()`. This behavior is tested by the unit test, but hard to tell in the story.~~

~~Not sure if we still need to expose individual `inputRef`s for each `<Radio />` inside the group as done in https://github.com/adobe/react-spectrum/pull/5967.~~

~~Oh and should we document this behavior somewhere in the docs?~~

~~Let me know what you guys think!~~

This PR exposes `inputRef` through props in `<Radio>`, practically the same as https://github.com/adobe/react-spectrum/pull/5967.

## ✅ Pull Request Checklist:

### Docs
- Visit http://localhost:1234/react-aria/RadioGroup.html#radio
- See if `inputRef` is in the prop table with the description "A ref for the HTML input element."

### DIFF
--- a/packages/react-aria-components/src/RadioGroup.tsx
+++ b/packages/react-aria-components/src/RadioGroup.tsx
@@ -13,14 +13,19 @@
 import {AriaRadioGroupProps, AriaRadioProps, HoverEvents, Orientation, useFocusRing, useHover, useRadio, useRadioGroup, VisuallyHidden} from 'react-aria';
 import {ContextValue, forwardRefType, Provider, RACValidation, removeDataAttributes, RenderProps, SlotProps, useContextProps, useRenderProps, useSlot} from './utils';
 import {FieldErrorContext} from './FieldError';
-import {filterDOMProps, mergeProps} from '@react-aria/utils';
+import {filterDOMProps, mergeProps, mergeRefs, useObjectRef} from '@react-aria/utils';
 import {LabelContext} from './Label';
 import {RadioGroupState, useRadioGroupState} from 'react-stately';
-import React, {createContext, ForwardedRef, forwardRef, useRef} from 'react';
+import React, {createContext, ForwardedRef, forwardRef, MutableRefObject} from 'react';
 import {TextContext} from './Text';
 
 export interface RadioGroupProps extends Omit<AriaRadioGroupProps, 'children' | 'label' | 'description' | 'errorMessage' | 'validationState' | 'validationBehavior'>, RACValidation, RenderProps<RadioGroupRenderProps>, SlotProps {}
-export interface RadioProps extends Omit<AriaRadioProps, 'children'>, HoverEvents, RenderProps<RadioRenderProps>, SlotProps {}
+export interface RadioProps extends Omit<AriaRadioProps, 'children'>, HoverEvents, RenderProps<RadioRenderProps>, SlotProps {
+  /**
+   * A ref for the HTML input element.
+   */
+  inputRef?: MutableRefObject<HTMLInputElement>
+}
 
 export interface RadioGroupRenderProps {
   /**
@@ -163,9 +168,13 @@ function RadioGroup(props: RadioGroupProps, ref: ForwardedRef<HTMLDivElement>) {
 }
 
 function Radio(props: RadioProps, ref: ForwardedRef<HTMLLabelElement>) {
-  [props, ref] = useContextProps(props, ref, RadioContext);
+  let {
+    inputRef: userProvidedInputRef = null,
+    ...otherProps
+  } = props;
+  [props, ref] = useContextProps(otherProps, ref, RadioContext);
   let state = React.useContext(RadioGroupStateContext)!;
-  let inputRef = useRef<HTMLInputElement>(null);
+  let inputRef = useObjectRef(mergeRefs(userProvidedInputRef, props.inputRef !== undefined ? props.inputRef : null));
   let {labelProps, inputProps, isSelected, isDisabled, isPressed} = useRadio({
     ...removeDataAttributes<RadioProps>(props),
     // ReactNode type doesn't allow function children.
--- a/packages/react-aria-components/src/RadioGroup.tsx
+++ b/packages/react-aria-components/src/RadioGroup.tsx
@@ -14,10 +14,9 @@ import {AriaRadioGroupProps, AriaRadioProps, HoverEvents, Orientation, useFocusR
 import {ContextValue, forwardRefType, Provider, RACValidation, removeDataAttributes, RenderProps, SlotProps, useContextProps, useRenderProps, useSlot} from './utils';
 import {FieldErrorContext} from './FieldError';
 import {filterDOMProps, mergeProps} from '@react-aria/utils';
-import {getFocusableTreeWalker} from '@react-aria/focus';
 import {LabelContext} from './Label';
 import {RadioGroupState, useRadioGroupState} from 'react-stately';
-import React, {createContext, ForwardedRef, forwardRef, useImperativeHandle, useRef} from 'react';
+import React, {createContext, ForwardedRef, forwardRef, useRef} from 'react';
 import {TextContext} from './Text';
 
 export interface RadioGroupProps extends Omit<AriaRadioGroupProps, 'children' | 'label' | 'description' | 'errorMessage' | 'validationState' | 'validationBehavior'>, RACValidation, RenderProps<RadioGroupRenderProps>, SlotProps {}
@@ -109,7 +108,6 @@ export const RadioGroupStateContext = createContext<RadioGroupState | null>(null
 
 function RadioGroup(props: RadioGroupProps, ref: ForwardedRef<HTMLDivElement>) {
   [props, ref] = useContextProps(props, ref, RadioGroupContext);
-  let domRef = useRef<HTMLDivElement>(null);
   let state = useRadioGroupState({
     ...props,
     validationBehavior: props.validationBehavior ?? 'native'
@@ -135,25 +133,11 @@ function RadioGroup(props: RadioGroupProps, ref: ForwardedRef<HTMLDivElement>) {
     defaultClassName: 'react-aria-RadioGroup'
   });
 
-  useImperativeHandle<HTMLDivElement|null, HTMLDivElement|null>(ref, () => {
-    const radioGroup = domRef.current;
-    if (radioGroup) {
-      radioGroup.focus = () => {
-        let walker = getFocusableTreeWalker(radioGroup, {
-          tabbable: true, 
-          accept: (node) => node.getAttribute('type') === 'radio'
-        });
-        (walker.firstChild() as HTMLElement).focus();
-      };
-    }
-    return radioGroup;
-  });
-
   return (
     <div
       {...radioGroupProps}
       {...renderProps}
-      ref={domRef}
+      ref={ref}
       slot={props.slot || undefined}
       data-orientation={props.orientation || 'vertical'}
       data-invalid={state.isInvalid || undefined}
--- a/packages/react-aria-components/src/RadioGroup.tsx
+++ b/packages/react-aria-components/src/RadioGroup.tsx
@@ -14,9 +14,10 @@ import {AriaRadioGroupProps, AriaRadioProps, HoverEvents, Orientation, useFocusR
 import {ContextValue, forwardRefType, Provider, RACValidation, removeDataAttributes, RenderProps, SlotProps, useContextProps, useRenderProps, useSlot} from './utils';
 import {FieldErrorContext} from './FieldError';
 import {filterDOMProps, mergeProps} from '@react-aria/utils';
+import {getFocusableTreeWalker} from '@react-aria/focus';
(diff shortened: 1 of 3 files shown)


packages/react-aria-components/stories/RadioGroup.stories.tsx
```diff
@@ -96,3 +96,36 @@ export const RadioGroupInDialogExample = () => {
...
+  return (
+    <div
+      style={{
+        height: '200vh',
+        padding: '80px',
+        display: 'flex', 
+        flexDirection: 'column', 
+        justifyContent: 'space-between'
+      }}>
+      <button onClick={onClick}>Focus radio group</button>
```
snowystinger, line 117: Can you use a RAC Button here?

lets add a description to the story as well (if you search the code base for "parameters: {description" you should find some examples in other stories)
I think it'd be best to say something along these lines:
> Use keyboard interactions for a visible focus ring or use the NerdeFocus plugin to track focus. Try both clicking on the extra button when no radios are selected and clicking on the extra button when any radio is selected except the first.
Expectation is that focus goes to the first radio is the group when none are selected, and focus goes to the selected radio if one is selected.

snowystinger (commented):
Looks pretty good, couple small comments.
To your question about documentation. We'll need to change the type on RadioGroup to handle this. In React Spectrum side, we have a FocusableRef type which we can base a new type on over here. We don't want to actually use that type because it obscures access to the DOM Node itself in a way we don't want to do over here.

sookmax:
@snowystinger thanks for the feedback! I updated the story to include what you requested.

Let me know if there's anything else I should add or change!


https://github.com/adobe/react-spectrum/assets/71210554/517fcd20-3bf8-40f9-b6fd-e5acd653cac3

sookmax:
@Mrtly 

> I am not sure if all browsers automatically scroll on focus. thoughts on this?

It seems like the default behavior per [HTMLElement: focus() method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus):

"By default the browser will scroll the element into view after focusing it, and it may also provide visible indication of the focused element (typically by displaying a "focus ring" around the element)."

And so I thought calling `focus()` on the ref would be enough to bring the radio group that may possibly be out of screen into view, without calling `scrollIntoView()` separately again. But calling `scrollIntoView()` will still work if you choose to call it, it'll just scroll to the radio group's div. I'm just not sure if we want to implement the same behavior for `ref.scrollIntoView()` as we did for `ref.focus()`. We need opinions from the core team members for this, I think.

> Would it be possible to add data-focus-visible when focusing on the radio label programmatically?

I think this is impossible though I'm not 100% sure. From my understanding, the `focus-visible` state is controlled only by keyboard interactions, and we can't manipulate it with `HTMLElement.focus()`. But again, I might be missing something, so It would be better to get the answer from one of the core members!

packages/react-aria-components/src/RadioGroup.tsx
```diff
@@ -133,11 +135,23 @@ function RadioGroup(props: RadioGroupProps, ref: ForwardedRef<HTMLDivElement>) {
     defaultClassName: 'react-aria-RadioGroup'
   });
 
+  useImperativeHandle<HTMLDivElement|null, HTMLDivElement|null>(ref, () => {
+    if (domRef.current) {
+      domRef.current.focus = () => {
+        if (domRef.current) {
+          let walker = getFocusableTreeWalker(domRef.current, {tabbable: true});
+          (walker.firstChild() as HTMLElement).focus();
```
nwidynski, line 143: @sookmax After taking a brief look at [`useFocusVisible()`](https://github.com/adobe/react-spectrum/blob/1697a7fef3516f7a194f43b75edb858589ef6535/packages/%40react-aria/interactions/src/useFocusVisible.ts#L302), i think it might be okay to dispatch a keyboard event manually before the focus occurs. This triggers isFocusVisible to be activated, but might be good to get more input here from core members.

```tsx
let radio = getFocusableTreeWalker(domRef.current, {tabbable: true}).firstChild();

radio.dispatchEvent(new KeyboardEvent('keydown'));
radio.focus();
```

packages/react-aria-components/src/RadioGroup.tsx
nwidynski, line 1: nit: Probably out of scope, but `useRadioGroup()` should probably return `groupProps` instead of `radioGroupProps`, for consistency with `useCheckboxGroup()`.

snowystinger: We can't change this without it being a breaking change unfortunately. So won't be changing in the near future. But thanks for raising

packages/react-aria-components/src/RadioGroup.tsx
snowystinger: One consideration would be if native controls do this across all three browsers. Someone could check this with a jsfiddle or something pretty quickly.
Otherwise, I'd prefer to match Checkbox and Switch and avoid firing an extra event.

sookmax: By all three you mean Chrome, FireFox, and Safari?

If so, I checked them with [this jsfiddle](https://jsfiddle.net/sookmax/f0wqdj1m/1), and Safari is the only one that doesn't add a visual indicator properly.

### Chrome
native-radio-chrome

### FireFox
native-radio-firefox

### Safari
native-radio-safari

---
Safari is a little weird in that when you open the inspector and select one of the radio inputs and apply both pseudo-classes `:focus` and `:focus-visible`, you can see the styled outline being applied, and _from that moment on_.. if you re-run the fiddle, it will start to show the outline correctly. _However_, if you refresh Safari, it stops working. (it's even weirder to explain this in plain English 😂)

Screenshot 2024-03-21 at 11 40 15 PM

snowystinger: thank you very much for checking that, looks like we'll have to decide if we follow this

sookmax: If I have a say, then I'd like to support this feature unless firing a fake keyboard event incurs a severe side effect. If consistency with Checkbox and Switch is a concern, then I can also take a look at them as well.

I think having the focus-visible state when the user calls ref.focus() enhances the user experience after all.

snowystinger: Hey, I finally got some time to look at this a little more. Thanks again for the jsfiddle example. Unfortunately, the behavior changes depending on when/how you call the 'focus'. See this fiddle https://jsfiddle.net/re7d3gum/2/ where focus is called after clicking a button. None of the browsers show focus as visible.

Focus is visible if you press the button with keyboard across all three browsers as well, which makes sense.

Therefore, I think we do not need to fire an extra keyboard event to show the focus visible.

packages/react-aria-components/src/RadioGroup.tsx
```diff
@@ -133,11 +135,23 @@ function RadioGroup(props: RadioGroupProps, ref: ForwardedRef<HTMLDivElement>) {
     defaultClassName: 'react-aria-RadioGroup'
   });
 
+  useImperativeHandle<HTMLDivElement|null, HTMLDivElement|null>(ref, () => {
+    if (domRef.current) {
+      domRef.current.focus = () => {
+        if (domRef.current) {
+          let walker = getFocusableTreeWalker(domRef.current, {tabbable: true});
```
snowystinger, line 142: I think we should add an accept check to filter just radios
```
{tabbable: true, accept: (node) => node.getAttribute('type') === 'radio'}
```

packages/react-aria-components/src/RadioGroup.tsx
sookmax: Ha, it's interesting how the browser chooses not to give a focus indicator when a button is interacted with a mouse! Thanks for verifying this.

But I still cannot help wondering about a way for the user to style the focused (via `ref.focus()`) radio.. well I guess, at least `ref.focus()` gives the `<label>` `data-focused` state, but if you try to style a focused radio generically like this:
```css
.radio {
  &:before {
      content: '';
      display: block;
      width: 1.286rem;
      height: 1.286rem;
      box-sizing: border-box;
      border: 0.143rem solid var(--deselected-color);
      background: var(--background-color);
      border-radius: 1.286rem;
      transition: all 200ms;
    }
  
  &[data-focused]:before {
      box-shadow: 0 0 0 2px canvas, 0 0 0 4px var(--selected-color);
    }
}
```

Then (other than the intended use case of when `ref.focus()` is called) each radio will get focused every time it is clicked by a pointing device.

Would there be a better way to target (stylistically) this specific use case (`ref.focus()`)?

packages/react-aria-components/src/RadioGroup.tsx
sookmax: Thanks for the suggestion! [af94a00](https://github.com/adobe/react-spectrum/pull/6067/commits/af94a0039a484e0864d050317e2c5fda16b541a9) contains the update.

packages/react-aria-components/src/RadioGroup.tsx
```diff
@@ -133,11 +135,25 @@ function RadioGroup(props: RadioGroupProps, ref: ForwardedRef<HTMLDivElement>) {
     defaultClassName: 'react-aria-RadioGroup'
   });
 
+  useImperativeHandle<HTMLDivElement|null, HTMLDivElement|null>(ref, () => {
+    const radioGroup = domRef.current;
+    if (radioGroup) {
+      radioGroup.focus = () => {
```
devongovett, line 141: I'm not sure that patching the native `focus` function on a DOM element is a good idea. Could be unexpected. You could do something like this from the outside though:

```js
radioGroupRef.current.querySelector('input').focus();
```

We should probably add `inputRef` to individual Radios as requested in #6123 though.

sookmax: Hey, thanks for the feedback!

I might be misinterpreting your intent so correct me if I'm wrong here.

I was working on this PR under the premise that what we prefer for `RadioGroup` is the group level focus management using [a roving tabindex](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex), rather than exposing individual radio's input refs. And that's why I chose to overwrite the radio group dom's focus function&mdash;I thought It would be okay to do so since `<div>` is not a focusable element by default (no tabIndex attr), I believe?

Anyways, I was mostly following @snowystinger's idea about the `RadioGroup` focus management that is detailed in https://github.com/adobe/react-spectrum/issues/6054, but you might be out of context so I'll paste some of his comments on the issue page here:

> RadioGroup doesn't expose a focusable ref, we could consider doing that.
It looks like we don't expose the inputRef to the Radio ref either. So that could also be an option.
I think using useImperitiveHandle on the RadioGroup though makes the most sense since it's aware of the roving tab index already

> ... only the currently selected radio should be focusable, so the focusabletreewalker should only find that one. There is the case where none of them have been selected yet, in which case, it should find the very first radio as all of them will be technically focusable.

> I think RadioGroup and CheckboxGroup are pretty different in terms of focus behavior. One acts as a single focus stop, the other is a set of focus stops. As a result, we don't want people to call focus on a single Radio, I think it makes more sense at the RadioGroup. This, I feel, is different than the behavior for a CheckboxGroup where any of them could receive focus and calling it on the Group is meaningless.

> If we need to give access to the individual input's per radio, then those would be done the same as we did switch and checkbox, on each individual control. But if we can I think we should hold off on that, due to the roving tabIndex, it could get you into unexpected situations. For instance you could focus an unselected one when a selection already exists. So I want to be more careful in considering exposing Radio input refs

snowystinger: @sookmax sorry, timezones messed up some internal conversations. You've done all the things I asked and very quickly. We then had some discussions internally now that we can see this, and as a result we changed direction a little.
We want to align on the work that was done in the other PR for checkbox and switch and expose an inputRef on radios.
We also no longer want to expose the imperative handler, as @devongovett pointed out, it may be unexpected.

As for the issue I brought up earlier about the roving tabIndex, we decided it was a non-issue. This is because it is recommended that RadioGroups have a default and do not usually participate in constraint validation https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#validation
So it should be unusual that focus gets sent back to a radio group. If it does, the roving tab index will sort itself out when one of the other options is selected.

So we're opting to keep things simpler.

sookmax: Ah I see. Thanks for clarifying the situation!

And yeah alright then, I'll go ahead and update this PR similar to https://github.com/adobe/react-spectrum/pull/5967.

snowystinger (dismissed):
thank you for the quick pivot!

snowystinger (approved):

reidbarber (approved):

PR #6139, opened 2024-04-02
fix(#6130): Ignore HiddenSelect when tree walking

### LINKED ISSUE #6130: isElementVisible: remove unnecessary attribute check
Per https://github.com/adobe/react-spectrum/pull/6116#discussion_r1544017444

> won't this conflate hiding things from the axe plugin vs implementation?

https://github.com/adobe/react-spectrum/blob/784737effd44b9d5e2b1316e690da44555eafd7e/packages/%40react-aria/focus/src/isElementVisible.ts#L47

majornista:
We should skip HiddenSelect element in Picker when tree walking to determine the next or previous focusable element. However, we should not use [data-a11y-ignore], which is used to exclude elements from aXe automated accessibility tests.

Per comments: https://github.com/adobe/react-spectrum/pull/6116#discussion_r1544017444

Closes #6130 

## ✅ Pull Request Checklist:

- [x] Included link to corresponding [React Spectrum GitHub Issue 6130](https://github.com/adobe/react-spectrum/issues/6130).

## 📝 Test Instructions:

1. Open TableView focusable cells example: https://reactspectrum.blob.core.windows.net/reactspectrum/b783832dd2ff7bf0a85ee30f1233ef9ce370b661/storybook/iframe.html?scrolling=false&providerSwitcher-locale=&providerSwitcher-theme=&providerSwitcher-scale=&providerSwitcher-express=false&args=&id=tableview--focusable-cells&viewMode=story
7. Press ArrowRight to move focus to the next cell, containing the word "Three."
8. The value of the Picker does not change as the focus moves to the next cell.
9. Press ArrowLeft to focus the Picker in the second cell again.
10. Press ArrowLeft again to move focus to the previous cell.
11. Focus moves to the Switch in the first cell, and the value of the Picker does not change.

Related Issue and PR: #5877, #5878, #6116

## 🧢 Your Project:

Adobe/Accessibility

### DIFF
--- a/packages/@react-aria/focus/src/isElementVisible.ts
+++ b/packages/@react-aria/focus/src/isElementVisible.ts
@@ -44,7 +44,7 @@ function isAttributeVisible(element: Element, childElement?: Element) {
   return (
     !element.hasAttribute('hidden') &&
     // Ignore HiddenSelect when tree walking.
-    !(element.hasAttribute('data-a11y-ignore') && element.getAttribute('aria-hidden') === 'true') &&
+    !(element.hasAttribute('data-hidden-select-ignore') && element.getAttribute('aria-hidden') === 'true') &&
     (element.nodeName === 'DETAILS' &&
       childElement &&
       childElement.nodeName !== 'SUMMARY'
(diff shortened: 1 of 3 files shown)


snowystinger (commented):
I'm happier with this. I also feel like `data-hidden` might be a more common than 0 chance, so I'm happier with the longer name for it as well. I think we can still change the name down the line as well, but at least now it's separated from tooling

snowystinger (approved):

LFDanLu (approved):

ashfaque-pixelotech:
what is the actual solution of this problem i am facing issue with Input, combobox inside table i am using react-aria-components

LFDanLu:
@ashfaque-pixelotech Are you running into issues with keyboard navigation and combobox inside the table?

PR #6145, opened 2024-04-03
Add backgroundColor prop to Table Cell

dsabsay:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

See new story.

## 🧢 Your Project:

Adobe

### DIFF
--- a/packages/@react-spectrum/table/src/TableViewBase.tsx
+++ b/packages/@react-spectrum/table/src/TableViewBase.tsx
@@ -21,6 +21,7 @@ import {
   useDOMRef,
   useFocusableRef,
   useStyleProps,
+  tableCellStyleProps,
   useUnwrapDOMRef
 } from '@react-spectrum/utils';
 import {ColumnSize, SpectrumColumnProps} from '@react-types/table';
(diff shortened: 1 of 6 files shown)


snowystinger:
Hi, thanks for the PR; however, we do not support background colors on table cells because we ensure a certain level of contrast for accessibility. If you need to indicate the status of a row or cell, please use a StatusLight along with a visible label, as color alone is not enough for some users.

dsabsay:
@snowystinger Thanks for the context. I understand the accessibility requirement. The color I intend for my use case is entirely supplemental to the content of the cell and not required for understanding of the content. What do you think about allowing cell coloring while also ensuring a certain contrast level (method TBD)?

dsabsay:
For additional context, my use case is probably somewhat uncommon in that almost ALL Cells will receive a background color. I would use a `View` embedded in the `Cell`; however, there is a significant amount of padding applied directly in the `Cell` which is undesired:

Screen Shot 2024-04-03 at 2 03 33 PM

Would the option to remove `Cell` padding be acceptable? (And by padding I mean padding and/or margin; I have not looked at how it is implemented)

reidbarber:
Even if you're successfully meeting contrast requirements, this design isn't supported by Spectrum for TableView. I recommend reaching out the Spectrum design team internally for an alternative solution.

dsabsay:
@reidbarber Understood, thanks!

Issue #6184, opened 2024-04-11
I18nProvider re-renders unnecessarily due to a new dict value being re-created every time


IgnusG:
### Provide a general summary of the issue here

Some hooks like `useCalendarGrid` use the `useLocale` hook to fetch locale information. However if the `I18nProvider` is rendered (eg. as part of a parent update somewhere in the tree) even if `locale` stays the same it re-creates a new dictionary object forcing all child hooks to re-render (even if their parent components don't have to).

### 🤔 Expected Behavior?

The hook should memo the dict value to avoid causing renders down the tree or `I18nContext` should be exported to allow for custom adjustments to this behavior.

### 😯 Current Behavior

Only the provider is exported which causes this issue. It's not possible to access the context itself so it's not possible to resolve this problem in userland. 

### 🖥️ Steps to Reproduce

1. Create a calendar structure as per examples wrapping it in `I18nProvider` somewhere close to the top of the app tree
2. Memo the calendar component so it doesn't update when parents do
3. Force a render to the app causing a re-render to the `I18nProvider` - observe the dict now changes reference causing re-renders everywhere its used (including inside the calendar - even if we don't change anything in it)

### Version

3.32.1

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### What operating system are you using?

macOS

PR #6185, opened 2024-04-11
Cache locale dict lest child hooks always re-render

IgnusG:
Closes https://github.com/adobe/react-spectrum/issues/6184

The child hooks like `useCalendarGrid` etc. use `useLocale` to fetch their locale data. 

As such whenever this context value changes these hooks (and the component they are part of are forced to re-render) even if nothing else is changed. Even if the locale doesn't change as long as `I18nProvider` renders (eg. as part of a parent component update) it creates a new dict which forces a re-render.

## ✅ Pull Request Checklist:

- [ ] ~~Added/updated unit tests and storybook for this change (for new code or code which already has tests).~~
- [ ] ~~Updated documentation (if it already exists for this component).~~
- [X] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/WAI/ARIA/apg/)

## 📝 Test Instructions:

Perform the steps in https://github.com/adobe/react-spectrum/issues/6184 and observe the rendering of `I18nProvider` when locale doesn't change no longer causes renders inside of components using the calendar hooks.


snowystinger (dismissed):
thanks the pr

snowystinger (approved):

yihuiliao (approved):

PR #6194, opened 2024-04-11
Fix RAC Submenu jittering when rendered close to edge of container

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test the RAC submenu stories, sizing the window so the final submenu should just barely fit without flipping. There shouldn't be any jittering as shown here

https://github.com/adobe/react-spectrum/assets/9661127/87fb7f58-efc5-4e88-9c64-cb53fa03e05d

## 🧢 Your Project:

RSP


packages/react-aria-components/src/Menu.tsx
```diff
@@ -179,6 +179,16 @@ function MenuInner<T extends object>({props, collection, menuRef: ref}: MenuInne
...
+  let prevPopoverContainer = useRef<HTMLDivElement | null>(null) ;
+  let [leftOffset, setLeftOffset] = useState({left: 0});
+  useEffect(() => {
+    if (popoverContainer && prevPopoverContainer.current !== popoverContainer && leftOffset.left === 0) {
+      prevPopoverContainer.current = popoverContainer;
+      let {left} = popoverContainer.getBoundingClientRect();
+      setLeftOffset({left: -1 * left});
+    }
+  }, [leftOffset, popoverContainer]);
+
```
LFDanLu, line 191: The div that serves as the popoverContainer has `width: 100vw` (aka full page width) but its origin is set as the left edge of its parent menu, resulting in a div that extends far off the page depending on the parent menu's positioning. The above copies the same logic from RSP Submenu which calculates how much said container needs to be offset in order to make it start from the left most edge of the page instead, meaning it will be flush with the boundaries of the page. This give the positioning code the proper container "area" in which it can calculate if the submenu should be repositioned/flipped/sized.

snowystinger (approved):

ktabors (approved):
I can't reproduce the error anymore. 😄

PR #6199, opened 2024-04-12
Add ColorArea, ColorField, ColorSlider, and ColorWheel to RAC

### LINKED ISSUE #5484: useColorField: Controlled version does not allow free input

When using the controlled version of `useColorField` it will jump to a value when you are not done typing yet. Strangely this does not seem to happen when you start without a `#`

### 🤔 Expected Behavior?

Allowing you to freely type

### 😯 Current Behavior

Examples: 
- Type: `#334` -> It will jump to `#333344`
- Type: `#125` -> it will jump to `#112255`
Doing either of these without a `#` will allow you to keep typing

Similiarly, when there is a complete value and you only backspace the last character it will keep jumping to complete values.

### 💁 Possible Solution

_No response_

### 🔦 Context

We have a custom `ColorPicker` component that makes use of `useColorField` together with `useColorSlider` and `useColorArea` to allow users to type as well as drag to the color they want. This requires us to keep the state synced between the 3 parts.

### 🖥️ Steps to Reproduce

https://react-spectrum.adobe.com/react-aria/useColorField.html#controlled

### Version

3.0.0-beta.27

Firefox, Chrome

### If other, please specify.

_No response_

MacOS Sonoma, Linux Mint, Windows 11

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
Fixes #5484

This adds React Aria Components for our existing color picker hooks, including ColorArea, ColorField, ColorSlider, and ColorWheel. It also fixes some bugs and simplifies the API, especially for ColorArea. See inline comments.

Because the color components often return inline styles, e.g. the background gradients, and users might need to override/combine these with their own, I've also added support for `defaultClassName`, `defaultStyle`, and `defaultChildren` as render props across all components. This allows you to access the defaults in a render prop function, and modify/merge them with your own values. For example, in ColorSlider, one might want to have a checkerboard pattern in the background behind the gradient when partially transparent. This can be done by concatenating `defaultStyle.background` with your own. You could also add extra wrappers around the default children of an element, or combine our default class name with a custom one instead of overriding.

One problem with inline styles is that you need to use `!important` if you want to override them in regular CSS, for example, to override the background with a gray color when disabled. An alternative to the current approach would be to expose the background as a custom property (e.g. `--color-slider-background`). Then the user would have to set `background` themselves using that var, and could merge with the checkerboard or override. However, then it would be _required_ to use that var and there would be no background by default, which might be unexpected/harder to use. Another point is that there are several other inline styles currently, and not sure we'd want to have a lot of custom properties for each of these.  Opening this point for debate. I think exposing `defaultStyle`/`defaultChildren`/`defaultClassName` is useful independent of this though.


packages/@react-aria/color/src/useColorArea.ts
```diff
@@ -413,10 +410,6 @@ export function useColorArea(props: AriaColorAreaOptions, state: ColorAreaState)
       ...colorAreaStyleProps,
       role: 'group'
     },
-    gradientProps: {
-      ...gradientStyleProps,
-      role: 'presentation'
-    },
```
devongovett, line 419: Removed the need for an additional gradient element inside the color area. Now both gradients are applied to the root color area element.

packages/@react-aria/color/src/useColorAreaGradient.ts
```diff
@@ -10,205 +10,89 @@
...
+        if (zChannel === 'hue') {
+          bg.push(value.toString('css'));
         }
+    
+        colorAreaStyles = {
+          background: bg.join(', ')
+        };
+        break;
       }
     }
```
devongovett, line 97: This refactors the way the gradients are generated to be both simpler and more correct. For RGB, we use `background-blend-mode` to combine the channel values rather than a transparency mask. This produces a more correct result than before, where the pixels at a given x/y position are closer to the actual value you get when you click. Some differences may appear in Chromatic for this.

For HSL and HSB, transparency is still used for saturation and lightness/brightness. But we had some bugs especially in HSB that are now fixed. For example, in the hue/saturation configuration, when lowering the brightness, the result shown in the thumb is much grayer than shown in the gradient. Same in the hue/brightness configuration with the saturation lowered.

image 
image

These are both now fixed.

image
image

packages/@react-aria/color/src/useColorField.ts
```diff
@@ -100,17 +104,17 @@ export function useColorField(
...
-      value: inputValue,
-      defaultValue: undefined,
-      validate: undefined,
-      [privateValidationStateProp]: state,
-      type: 'text',
-      autoComplete: 'off',
-      onChange
-    }), state, ref);
+  let {inputProps, ...otherProps} = useFormattedTextField({
+    ...props,
```
devongovett, line 108: `mergeProps` resulted in `onChange` being called twice, with both the raw text value of the field, and with the parsed color. It should only be called with a color.

packages/@react-aria/color/src/useColorSlider.ts
```diff
@@ -109,6 +112,15 @@ export function useColorSlider(props: AriaColorSliderOptions, state: ColorSlider
     inputProps['aria-valuetext'] += `, ${value.getColorName(locale)}`;
   }
 
+  let {visuallyHiddenProps} = useVisuallyHidden({
```
devongovett, line 115: Building visually hidden into the hook for color slider and color wheel like we do for color area. Can't think of a reason why you wouldn't want it to be visually hidden.

packages/@react-spectrum/color/src/ColorArea.tsx
```diff
@@ -59,13 +58,12 @@ function ColorArea(props: SpectrumColorAreaProps, ref: FocusableRef<HTMLDivEleme
       }
       ref={containerRef}
       style={{
-        ...colorAreaProps.style,
+        ...(isDisabled ? {} : colorAreaProps.style),
```
devongovett, line 61: Previously no gradient was returned by the hook when disabled, but I don't think we can assume that all color areas will work that way. This moves that logic into Spectrum instead.

packages/react-aria-components/src/ColorThumb.tsx
```diff
@@ -0,0 +1,107 @@
...
+  state: ColorState,
+  thumbProps: HTMLAttributes<HTMLElement>,
+  inputXRef: RefObject<HTMLInputElement>,
+  inputYRef?: RefObject<HTMLInputElement>,
+  xInputProps: InputHTMLAttributes<HTMLInputElement>,
+  yInputProps?: InputHTMLAttributes<HTMLInputElement>,
+  isDisabled?: boolean
+}
+
+export const InternalColorThumbContext = createContext<InternalColorThumbContextValue | null>(null);
```
devongovett, line 25: Ideally this context would be exported/not internal. However the API for the context is a bit abnormal in that it takes a lot of stuff that isn't just the same as the props, including props/refs for two hidden inputs. If we want to export it we would need to re-think this.

packages/react-aria-components/src/ColorWheel.tsx
```diff
@@ -0,0 +1,105 @@
...
+const _ColorWheel = forwardRef(ColorWheel);
+export {_ColorWheel as ColorWheel};
+
+export interface ColorWheelTrackRenderProps extends ColorWheelRenderProps {}
+export interface ColorWheelTrackProps extends StyleRenderProps<ColorWheelTrackRenderProps> {}
+interface ColorWheelTrackContextValue extends Omit<HTMLAttributes<HTMLElement>, 'children' | 'className' | 'style'>, ColorWheelTrackProps {}
+
+export const ColorWheelTrackContext = createContext<ContextValue<ColorWheelTrackContextValue, HTMLDivElement>>(null);
+
+function ColorWheelTrack(props: ColorWheelTrackProps, ref: ForwardedRef<HTMLDivElement>) {
```
devongovett, line 79: Ended up creating a separate component for the ColorWheelTrack. I wanted the track to just be rendered as the background of the ColorWheel itself, but the clip-path we use makes this complicated. The thumb shouldn't actually be clipped, so the only way to do that is to make a separate sibling element. Couldn't reuse SliderTrack here because that expects a SliderState and exposes a few other render props that are specific to sliders.

packages/react-aria-components/src/utils.tsx
```diff
@@ -132,11 +134,11 @@ export function useRenderProps<T>(props: RenderPropsHookOptions<T>) {
 
     return {
       className: computedClassName ?? defaultClassName,
-      style: computedStyle,
-      children: computedChildren,
+      style: (computedStyle || defaultStyle) ? {...defaultStyle, ...computedStyle} : undefined,
```
devongovett, line 137: note that defaultStyle works slightly differently to defaultClassName and defaultChildren: it is _merged_ with the computed style rather than being overridden. I think this makes sense otherwise whenever anyone used a `style` render prop they would need to remember to merge the defaults manually. Since the merge is shallow, it is still possible to override/remove a style property from the defaults by setting it to undefined.

packages/react-aria-components/src/utils.tsx
```diff
@@ -186,21 +188,31 @@ export function useSlottedContext<T>(context: Context<SlottedContextValue<T>>, s
...
+    if (typeof contextProps.style === 'function' || typeof props.style === 'function') {
+      // @ts-ignore
+      mergedProps.style = (renderProps) => {
+        let contextStyle = typeof contextProps.style === 'function' ? contextProps.style(renderProps) : contextProps.style;
+        let defaultStyle = {...renderProps.defaultStyle, ...contextStyle};
+        let style = typeof props.style === 'function' 
+          ? props.style({...renderProps, defaultStyle})
+          : props.style;
+        return {...defaultStyle, ...style};
+      };
```
devongovett, line 211: This merges the styles from context, defaultStyle, and local styles.

majornista:
Should React-Spectrum ColorArea have a border like the ColorSlider and ColorWheel with `forced-colors: active`?

packages/@react-aria/color/src/useColorField.ts
snowystinger: I think this has closed https://github.com/adobe/react-spectrum/issues/5484

packages/@react-aria/color/src/useColorAreaGradient.ts
snowystinger: Ran chromatic, it didn't appear to pick up on these https://www.chromatic.com/build?appId=5f0dd5ad2b5fc10022a2e320&number=737

packages/react-aria-components/src/ColorArea.tsx
```diff
@@ -0,0 +1,80 @@
...
+          [InternalColorThumbContext, {state, thumbProps, inputXRef, xInputProps, inputYRef, yInputProps, isDisabled: props.isDisabled}]
+        ]}>
+        {renderProps.children}
+      </Provider>
+    </div>
+  );
+}
+
+/**
+ * A color area allows users to adjust two channels of an RGB, HSL or HSB color value against a two-dimensional gradient background.
```
snowystinger, line 77: Are those just the color spaces we support? or is that actually all which use a square? I could've sworn LAB could use a square as well

packages/react-aria-components/docs/ColorSlider.mdx
```diff
@@ -0,0 +1,687 @@
...
+### RGBA
+
+This example shows how you could build an RGBA color picker using four color sliders bound to the same
+color value in state. The <TypeLink links={docs.links} type={docs.exports.parseColor} />
+function is used to parse the initial color from a hex value, stored in state. The `value` and `onChange` props
+of ColorSlider are used to make the sliders controlled, so that they all update when the color is modified.
+
+```tsx example
+function Example() {
+  let [color, setColor] = React.useState(parseColor('#ff00ff'));
```
snowystinger, line 282: this is hex but the title says rgba, not a big deal, though maybe we should wait to use the hex until we add support for the alpha channel? https://github.com/adobe/react-spectrum/pull/3302
instead we can use the `rgba(r, g, b, a)` syntax here

packages/react-aria-components/docs/ColorWheel.mdx
```diff
@@ -0,0 +1,436 @@
...
+  &[data-focus-visible] {
+    width: 24px;
+    height: 24px;
+  }
+}
+```
+
+</details>
+
+## Features
```
snowystinger, line 76: no example for `Creating a color picker`?
maybe we should link to that section?

snowystinger (commented):
Topic for further work, 2d color wheels and composing multiple areas

https://color.adobe.com/create/color-wheel

Maybe allow multiple color pickers to be stacked like this? we can't just layer them on top of each other with transparent backgrounds because the highest one will intercept the events and position only its thumb to wherever you clicked. Maybe we want a "headless" thumb mode down the road?

packages/react-aria-components/src/ColorArea.tsx
devongovett: yeah we only support those right now. I'd like to have some kind of plugin system for them so we can add more in the future though.

packages/react-aria-components/stories/ColorWheel.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+const RADIUS = 100;
+const TRACK_THICKNESS = 28;
+const THUMB_SIZE = 20;
+
+
+export const ColorWheelExample = (args) => (
+  <ColorWheel {...args} outerRadius={RADIUS} innerRadius={RADIUS - TRACK_THICKNESS}>
+    <ColorThumb
```
majornista, line 27:
```suggestion
    <ColorWheelTrack />
    <ColorThumb
```

majornista, line 27: Should ColorWheelTrack be included  with the ColorThumb as a child of ColorWheel?
```suggestion
  <ColorWheel {...args} outerRadius={RADIUS} innerRadius={RADIUS - TRACK_THICKNESS}>
        <ColorWheelTrack />
        <ColorThumb
```

majornista (changes requested):
Shouldn't ColorWhee

packages/react-aria-components/docs/ColorWheel.mdx
devongovett: Link where? That section doesn't exist in the RSP or hook docs either...

LFDanLu (dismissed):
LGTM, tested the stories and docs and verified the functionality + bugs from last testing session don't exist anymore. With regard to defaultClassName/defaultStyle/defaultChildren, I think it is fine to expose it on all components, albeit the lack of any actual defaultStyle/defaultChildren on most components might be a bit confusing to users, at the moment it doesn't really feel very obvious when a component actually has values provided for those.

packages/@adobe/spectrum-css-temp/components/colorarea/skin.css
```diff
@@ -17,10 +14,6 @@
       &:before {
         box-shadow: inset 0 0 0 var(--spectrum-colorarea-border-size) var(--spectrum-colorarea-border-color-disabled);
       }
-
-      .spectrum-ColorArea-gradient {
-        display: none;
-      }
     }
   }
```
majornista, line 19: Add `--spectrum-colorarea-border-color: ButtonBorder;` to `forced-colors: active` style declaration at: https://github.com/adobe/react-spectrum/blob/49ba0e7cec1f22f6da62b2a49872169594f661a1/packages/%40adobe/spectrum-css-temp/components/colorarea/skin.css#L21-L23

```css
 .spectrum-ColorArea { 
   --spectrum-colorarea-fill-color-disabled : GrayText; 
   --spectrum-colorarea-border-color: ButtonBorder;
 } 
```

majornista (changes requested):
Add `--spectrum-colorarea-border-color: ButtonBorder;` for consistency with `forced-colors: active`.

We may also want to update the custom swatches in the Storybook stories for ColorArea and the controlled examples of ColorField and ColorWheel to use `forced-color-adjust: none`.

devongovett:
> We may also want to update the custom swatches in the Storybook stories for ColorArea and the controlled examples of ColorField and ColorWheel to use forced-color-adjust: none.

I've replaced the custom swatches with real ones using `useColorSwatch` in #6272, which includes forced-color-adjust: none by default.

LFDanLu (approved):
LGTM, verified the new forced color border around the RSP ColorArea

dannify (approved):

packages/react-aria-components/docs/ColorSlider.mdx
snowystinger: yeah, but we're only sending in a color with three channels, I know it'll default to 1 for opacity, but maybe misleading name? or misleading value being input? all the other examples include it

packages/react-aria-components/docs/ColorWheel.mdx
snowystinger: ah, sorry, i meant link to the ColorArea section titled that

PR #6205, opened 2024-04-15
Switch from Babel to SWC in Jest

devongovett:
This switches Jest from using Babel to using SWC, which is faster and matches what Parcel uses. It also bumps Babel to support parsing import attributes when building docs for components with macros. The bump of Babel was causing issues with jest timing out/running out of memory, and switching to SWC solves this.


lib/jestResolver.js
```diff
@@ -15,15 +15,31 @@ const path = require('path');
...
 try {
-  fs.mkdirSync(cacheDir, {recursive: true});
+  fs.mkdirSync(intlCacheDir, {recursive: true});
+  fs.mkdirSync(globCacheDir, {recursive: true});
 } catch (err) {
   // ignore
 }
 
 module.exports = (request, options) => {
+  // Handle glob imports.
```
devongovett, line 30: This is in replacement of the previous babel plugin that handled globs.

packages/@react-spectrum/avatar/test/Avatar.test.js
```diff
@@ -31,26 +30,6 @@ describe('Avatar', () => {
...
-      const dimensionValueSpy = jest.spyOn(Utils, 'dimensionValue');
-      render(<Avatar src="http://localhost/some_image.png" size="size-100" />);
-      expect(dimensionValueSpy).toHaveBeenCalledWith('avatar-size-100');
-    });
-
-    it('defaults to default size when size is a string number', () => {
-      const dimensionValueSpy = jest.spyOn(Utils, 'dimensionValue');
-      render(<Avatar src="http://localhost/some_image.png" size="100" />);
-      expect(dimensionValueSpy).toHaveBeenCalledWith('avatar-size-100');
-    });
```
devongovett, line 53: These tests didn't seem particularly useful now that we have chromatic. Could be fixed by switching to module mocks like I did in focus tests but didn't seem worth it.

packages/@react-aria/focus/test/focusSafely.test.js
```diff
@@ -14,14 +14,20 @@
...
+import * as ReactAriaUtils from '@react-aria/utils';
 import {setInteractionModality} from '@react-aria/interactions';
 
+jest.mock('@react-aria/utils', () => {
+  let original = jest.requireActual('@react-aria/utils');
+  return {
+    ...original,
+    focusWithoutScrolling: jest.fn()
+  };
+});
```
devongovett, line 26: You can't spy on a module export with swc, which errors with "cannot redefine property". This uses jests module mocks to do that instead.

reidbarber (approved):

snowystinger (approved):

PR #6237, opened 2024-04-19
fix: DropZone button type

### LINKED ISSUE #6230: DropZone unexpectedly submits forms.

The hidden button rendered by DropZone for focus behaviors is of type `submit` (since the type is not explicitly specified): https://github.com/adobe/react-spectrum/blob/64ed13090ce77cc0e4cb4cd5602e75f655bff6bb/packages/react-aria-components/src/DropZone.tsx#L118

When using a DropZone inside of a form, activating the button with the keyboard (eg press enter or space) will submit the form. I don't think submitting the form is an expected behavior given that the button seems to be an accessibility implementation detail. 

### 🤔 Expected Behavior?

Probably nothing should happen when that button is activated. When a FileTrigger is present, maybe the click event should be passed to the FileTrigger? The HTML file input acts as both a drag and drop area and the file trigger and both enter/space open the file menu rather than submit the form.

### 😯 Current Behavior

The DropZone focus trap button can submit the form (usually by accident).

### 💁 Possible Solution

Add `type="button"` to the instance of `<button>` linked above.

### 🔦 Context

While trying to use DropZone within a form, we noticed unintentional early submits from keyboard users activating the DropZone button rather than the FileTrigger.

### 🖥️ Steps to Reproduce

https://codesandbox.io/p/sandbox/heuristic-joji-c8gv78

In the codesandbox, move the keyboard focus to the DropZone and press enter or space. Notice the submitCount increase each time.

### Version

React Aria Components 1.1.1

Firefox, Chrome, Safari

### If other, please specify.

_No response_

mac OS 14.3

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

adriantrunzo:
Closes #6230 

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/react-aria-components/src/DropZone.tsx
+++ b/packages/react-aria-components/src/DropZone.tsx
@@ -117,6 +117,7 @@ function DropZone(props: DropZoneProps, ref: ForwardedRef<HTMLDivElement>) {
           </div>
           <button
             {...mergeProps(buttonProps, focusProps, clipboardProps, labelProps)}
+            type="button"
             ref={buttonRef} />
         </VisuallyHidden>
         {renderProps.children}
--- a/packages/react-aria-components/test/DropZone.test.js
+++ b/packages/react-aria-components/test/DropZone.test.js
@@ -11,7 +11,7 @@
  */
 
 import {act, fireEvent, pointerMap, render} from '@react-spectrum/test-utils';
-import {Button, DropZone, DropZoneContext, FileTrigger, Link, Text} from '../';
+import {Button, DropZone, DropZoneContext, FileTrigger, Form, Link, Text} from '../';
 import {ClipboardEvent, DataTransfer, DataTransferItem, DragEvent} from '@react-aria/dnd/test/mocks';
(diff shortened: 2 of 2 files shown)


adriantrunzo:
It seems like `type="button"` should be coming from the call to `useButton`, but I haven't tracked down why the `type` attribute doesn't come through in the button props.

adriantrunzo:
Also strange here is the `type="button"` **_does_** apply to the button in the test environment even without these changes.

yihuiliao:
Actually thanks to your comment about how `type="button"` was getting applied in the tests I dug around a bit more and found that this already has been fixed by https://github.com/adobe/react-spectrum/pull/5958. However, this PR was made after the last release so you'll just have to wait until our next release to get this update. We're hoping to get one out by the end of next week but I can't guarantee anything. I think you can close out this PR but thank you for submitting it!

adriantrunzo:
Ah, thanks for the update. I realized soon after typing that comment about the tests that the issue had likely already been fixed. I wasn't sure where, though.

PR #6238, opened 2024-04-19
Give translators more flexibility with transparent color names

devongovett:
As requested by globalization team/translation vendor. Some languages need to be able to re-order the transparency with respect to the other color components, and add/remove punctuation. This creates a new string for partially transparent colors and swaps between them accordingly.


packages/@react-stately/color/intl/ar-AE.json
```diff
@@ -8,7 +8,7 @@
   "red": "أحمر",
   "saturation": "التشبع",
   "colorName": "{transparency} {lightness} {chroma} {hue}",
-  "transparency": "{percent} transparent",
+  "transparentColorName": "{lightness} {chroma} {hue}, {percentTransparent} transparent",
   "very dark": "very dark",
```
rgeraghty, line 12: @devongovett English is correct, but the localized still have **{transparency}** in the **colorName**.

Our automation will correct that when updating translations, just mentioning in case issue before then,

LFDanLu (approved):

dannify (approved):

PR #6244, opened 2024-04-22
fix: MenuTrigger children prop is required

PHILLIPS71:
The `MenuTrigger` component should require its children prop, matching the similar `DialogTrigger` component. This also aligns with the rendered `PressResponder` component which also requires the children prop to be passed.

## ✅ Pull Request Checklist:


snowystinger (approved):
Thanks so much for the PR, seems like the right thing to do.

reidbarber (approved):

PR #6254, opened 2024-04-23
useDrag: add support for isDisabled

reidbarber:
For parity with `useDrop` and `useClipboard`

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Check [new story](https://reactspectrum.blob.core.windows.net/reactspectrum/78be891d80b00fe7d7b910a8f54ff4dd5b7e8ce2/storybook/index.html?path=/story/drag-and-drop--draggable-disabled&providerSwitcher-express=false) and tests

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-aria/dnd/src/useDrag.ts
+++ b/packages/@react-aria/dnd/src/useDrag.ts
@@ -357,6 +357,7 @@ export function useDrag(options: DragOptions): DragResult {
       onDragEnd
     },
     dragButtonProps: {
+      ...descriptionProps,
       onPress
     },
     isDragging
--- a/packages/@react-aria/dnd/src/useDrag.ts
+++ b/packages/@react-aria/dnd/src/useDrag.ts
@@ -38,7 +38,11 @@ export interface DragOptions {
    * Whether the item has an explicit focusable drag affordance to initiate accessible drag and drop mode.
    * If true, the dragProps will omit these event handlers, and they will be applied to dragButtonProps instead.
    */
-  hasDragButton?: boolean
+  hasDragButton?: boolean,
+  /**
+   * Whether the drag operation is disabled. If true, the element will not be draggable.
+   */
+  isDisabled?: boolean
 }
 
 export interface DragResult {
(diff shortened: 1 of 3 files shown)


packages/@react-aria/dnd/src/useDrag.ts
```diff
@@ -341,7 +357,6 @@ export function useDrag(options: DragOptions): DragResult {
       onDragEnd
     },
     dragButtonProps: {
-      ...descriptionProps,
```
yihuiliao, line 344: did we mean to delete the descriptionProps when it isn't disabled?

reidbarber: That is the "Press Enter to start dragging" description, so I assume we don't want it while disabled.

yihuiliao: yes but we would want to announce it otherwise right? cause in the [docs](https://reactspectrum.blob.core.windows.net/reactspectrum/78be891d80b00fe7d7b910a8f54ff4dd5b7e8ce2/docs/react-aria/useDrag.html#drag-button) now, it doesn't announce "press enter to start dragging" even when it isn't disabled

reidbarber: Oh good catch, that was accidentally removed. Thanks!

packages/@react-aria/dnd/src/useDrag.ts
```diff
@@ -332,6 +336,18 @@ export function useDrag(options: DragOptions): DragResult {
     };
   }
 
+  if (isDisabled) {
+    return {
+      dragProps: {
+        draggable: 'false'
+      },
+      dragButtonProps: {
+        onPress() {}
```
LFDanLu, line 345: just curious, was this required? Any reason we shouldn't just provide an empty obj?

packages/@react-aria/dnd/src/useDrag.ts
LFDanLu, line 1: was thinking about what should happen if the drag target becomes disabled in the middle of a drag operation (e.g. keyboard dnd). I think the current behavior is fine (aka you can still complete the drop part of your current drag operation) but open question for the team in case there were any other opinions.

LFDanLu (approved):

packages/@react-aria/dnd/src/useDrag.ts
reidbarber: No reason, just updated that

ktabors (approved):

packages/@react-aria/dnd/src/useDrag.ts
reidbarber: I think we would have to stop the drag session, and that would be a bit weird. Open to implementing that though, if thats what we want to do.

Issue #4244, opened 2023-03-23
[Toast] exiting attribute should apply to the target on close


siriwatknp:
# 🐛 Bug Report 

The `exiting` does not apply to the target on close

## 🤔 Expected Behavior

The `exiting` always applies to the first element of the region

## 😯 Current Behavior

When remove any element, the first element always get the `exiting` attribute.

https://user-images.githubusercontent.com/18292247/227089618-07cb9cdb-bf88-4ea0-bc63-6b3b1f6e1b24.mov

https://codesandbox.io/s/snackbar-joy-ui-feat-react-aria-gme1rg?file=/JoySnackbar.tsx

## 🔦 Context

I'm integrating `useToast` from React Aria with [Joy UI Alert](https://mui.com/joy-ui/react-alert/). All of the code I got from the official React Aria website.

## 💻 Code Sample

Take a look at the CodeSandbox above.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| react-spectrum   |
| Browser          |
| Operating System |

marinalazarova:
I'm experiencing this as well. I believe it's because of the way the toasts are being sorted by priority here:
https://github.com/adobe/react-spectrum/blob/d677c31d7b4dd34641c1b8f809863ee1a735607c/packages/%40react-stately/toast/src/useToastState.ts#L172-L176

marinalazarova:
**UPDATE:**

I was able to "fix" this by extending the `ToastQueue` class and overriding the `add` method to always set the highest priority for newly added toasts. I'm posting the workaround here for anyone who needs this:

```typescript
class CustomToastQueue<T> extends ToastQueue<T> {
  add(content: T, options?: any) {
    return super.add(content, {
      ...options,
      // We need to access the private `queue` property to get the length
      // of the queue, which is used as the priority of the newly added toast.
      // We could use the `visibleToasts` property, but that would only give
      // us the length of the visible toasts, not the total length of the queue.
      // This is only a workaround until the following bug is fixed:
      // @see https://github.com/adobe/react-spectrum/issues/4244
      // @ts-ignore
      priority: (this.queue as QueuedToast<T>[]).length,
    });
  }
}
``` 

Then, you can instantiate your toast queue like this:
```typescript
const toastQueue = new CustomToastQueue<ToastContent>({
  maxVisibleToasts: 3,
  hasExitAnimation: true,
});
```

LFDanLu:
[...]

PR #6270, opened 2024-04-26
Refactor logic to prevent breaking changes in useToastState

LFDanLu:
Closes https://github.com/adobe/react-spectrum/issues/4244

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test toast stories. Behavior should still be that new toasts are inserted at the bottom of the stack. Toast exit animations should still work

## 🧢 Your Project:

RSP


snowystinger (dismissed):

ktabors:
Sorry I'm breaking it. I wrote some tests that found one bug that I fixed and a second bug I'll work on tomorrow.

The first bug: The `close()` code assumed all Toasts being closed are visible. When programmatically removing an item that isn't visible the `this.visibleToasts.find()` returned `undefined` and we can't set a prop on that.

Details about the second unfixed bug:
* There are two Toasts and it removes the first Toast so only the third Toast is left.
* `this.visibleItems` is empty while `this.queue` has one item, the third Toast.

packages/@react-stately/toast/src/useToastState.ts
```diff
@@ -159,30 +155,27 @@ export class ToastQueue<T> {
...
   }
 
   /** Removes a toast from the visible toasts after an exiting animation. */
   remove(key: string) {
-    let index = this.queue.findIndex(t => t.key === key);
-    this.visibleToasts = this.visibleToasts.filter(t => t.key !== key);
-    this.updateVisibleToasts(index);
+    this.queue = this.queue.filter(t => t.key !== key);
+    this.updateVisibleToasts('remove');
   }
```
LFDanLu, line 171: I'm unsure about whether or not to have `this.queue = this.queue.filter(t => t.key !== key);` in here or not. The code works without it assuming the user calls `close` and then `remove` and would allow us to remove the `key` input argument. However, if the user calls just `remove` then we need the aforementioned line to make sure the queue is properly updated and thus `visibleToasts` gets updated as well.

ktabors: Technically it does work, the remove() is just doing some cleanup to the dom by removing the object from the dom.

The close button not closing until you move the mouse was related to close() not remove(), right?

packages/@react-stately/toast/src/useToastState.ts
```diff
@@ -159,30 +155,27 @@ export class ToastQueue<T> {
     if (index >= 0) {
       this.queue[index].onClose?.();
       this.queue.splice(index, 1);
+      let closedToast = this.visibleToasts.find(t => t.key === key);
+      if (closedToast) {
+        closedToast.animation = 'exiting';
```
devongovett, line 160: wonder if there will be problems due to mutating the toast, which is passed into props of a component...

packages/@react-stately/toast/src/useToastState.ts
devongovett: Probably good to keep it to be safe.

packages/@react-stately/toast/src/useToastState.ts
```diff
@@ -159,30 +155,27 @@ export class ToastQueue<T> {
...
-          .map(t => ({...t, animation: 'exiting'}));
 
-      if (oldIndex !== -1) {
-        toasts.splice(oldIndex, 0, prevToasts?.[0]);
-      }
-
-      this.visibleToasts = toasts;
+    if (action === 'close' && this.hasExitAnimation) {
+      // Cause a rerender to happen for exit animation
+      this.visibleToasts = this.visibleToasts.map(t => t);
```
devongovett, line 178: One change in logic here is that previously, if the number of visible toasts exceeded the `maxVisibleToasts` setting, the toasts that got pushed off the end would animate out. Now they won't. Only toasts that are explicitly closed by clicking the X button will animate. Is that expected?

packages/@react-stately/toast/src/useToastState.ts
LFDanLu: yeah, the extra update to the queue in the `remove` call is to ensure the queue as our source of truth is still up to date in case the user only calls `remove`.  

> The close button not closing until you move the mouse was related to close() not remove(), right?

yeah it previous happened to use since we weren't always making an update to `visibleToast` that would cause a state change

packages/@react-stately/toast/src/useToastState.ts
LFDanLu: Oh didn't realize that was a case here. Will revisit how we apply the `.animation` then and see if we can move it back into here somehow

packages/@react-stately/toast/src/useToastState.ts
```diff
@@ -161,28 +157,33 @@ export class ToastQueue<T> {
...
-      if (oldIndex !== -1) {
-        toasts.splice(oldIndex, 0, prevToasts?.[0]);
-      }
-
-      this.visibleToasts = toasts;
+    if (action === 'add' && this.hasExitAnimation) {
+      let prevToasts: QueuedToast<T>[] = this.visibleToasts
+        .filter(t => !toasts.some(t2 => t.key === t2.key))
+        .map(t => ({...t, animation: 'exiting'}));
+      this.visibleToasts = prevToasts.concat(toasts).sort((a, b) => b.priority - a.priority);
```
LFDanLu, line 177: This should bring it back exiting animations theoretically, testing and seeing if I can simplify further

packages/@react-stately/toast/src/useToastState.ts
LFDanLu: I got rid of the extra update to queue in the `remove` call since this was prematurely removing toasts that get pushed out of visibility by higher priority toasts. In RSP the flow that happens when a toast is pushed out of visibility is as follows:
new toast `add` animates in -> toast that is no longer visible gets `close` called by its `onAnimationEnd`

LFDanLu:
You can test the behavior of the toasts with priority in the toast story here on this branch as well: https://github.com/adobe/react-spectrum/pull/6290

packages/@react-stately/toast/src/useToastState.ts
```diff
@@ -134,19 +134,15 @@ export class ToastQueue<T> {
       }
     }
 
-    if (toast.priority) {
-      this.queue.splice(low, 0, toast);
-    } else {
-      this.queue.unshift(toast);
-    }
+    this.queue.splice(low, 0, toast);
```
ktabors, line 137: Is this a breaking change or was `this.queue.unshift(toast);` a bug we hadn't found?

ktabors (approved):

packages/@react-stately/toast/src/useToastState.ts
LFDanLu: This reverts the logic to what is currently released

ktabors: Right, I just remembered that and was coming back to delete it, but you responded.

snowystinger (approved):

devongovett (approved):

snowystinger (approved):

devongovett (approved):
ok for this release but we are planning to refactor this more later

PR #6272, opened 2024-04-26
Add ColorPicker and ColorSwatchPicker components

devongovett:
Depends on #6204 and #6199.

This adds ColorPicker and ColorSwatchPicker components to both RSP and RAC. A ColorPicker synchronizes the state between multiple color components via context. In RSP, it also provides a swatch that triggers a popover, inside of which you can place a default `<ColorEditor>` or any other combination of color components.

ColorSwatchPicker is based on the RAC ListBox component, allowing users to select a preset swatch. It is intentionally limited, allowing only single selection of unique color values so that the value can be two-way synced with a ColorPicker.

ColorField also gains support for a `channel` prop (along with `colorSpace`), enabling it to edit an individual channel value as a number in addition to hex values. Internally, this is a wrapper around our existing number field components. There are new hooks in React Aria, and in RAC/RSP, we swap between these hooks automatically depending on whether the prop is present.

Finally, a `colorSpace` prop has been added to `ColorArea` and `ColorSlider`. This works together with the existing channel props. Previously, you would need to pass in a channel that actually exists in the `value` otherwise it would throw an error. Now that color components can receive values via context from a ColorPicker, this is less obvious. Declaring a `colorSpace` internally converts whatever `value` is passed in to that color space so that the channel always works no matter what value is provided by the ColorPicker.


packages/@react-spectrum/color/src/ColorArea.tsx
```diff
@@ -24,14 +25,15 @@ import {useProviderProps} from '@react-spectrum/provider';
 
 function ColorArea(props: SpectrumColorAreaProps, ref: FocusableRef<HTMLDivElement>) {
   props = useProviderProps(props);
+  let inputXRef = useRef(null);
+  let inputYRef = useRef(null);
+  let containerRef = useFocusableRef(ref, inputXRef);
+  [props, containerRef] = useContextProps(props, containerRef, ColorAreaContext);
```
devongovett, line 31: The existing RSP color components now consume from the RAC contexts even though I didn't rewrite them with RAC. That way they work inside a RAC ColorPicker.

packages/@react-spectrum/color/src/ColorPicker.tsx
```diff
@@ -0,0 +1,122 @@
...
+          UNSAFE_style={{
+            width: 'fit-content',
+            minWidth: 0,
+            margin: '0 auto' // Center within tray.
+          }}>
+          <Content 
+            UNSAFE_style={{
+              position: 'relative',
+              margin: 'calc(var(--spectrum-dialog-padding) * -1)',
+              padding: 'var(--spectrum-global-dimension-size-200)'
```
devongovett, line 108: This is a hack to reduce the padding of the popover. Ideally we'd add a density option to popover but we need to work out the API for it.

packages/@react-spectrum/color/src/ColorSwatchPicker.tsx
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * The size of the color swatches.
+   * @default "M"
+   */
+  size?: 'XS' | 'S' | 'M' | 'L',
+  /**
+   * The corner rounding of the color swatches.
+   * @default "none"
+   */
+  rounding?: 'none' | 'default' | 'full'
```
devongovett, line 38: I followed spectrum's options here, but it is sorta weird that density uses "regular" and rounding uses "default"...

packages/dev/parcel-transformer-mdx-docs/MDXTransformer.js
```diff
@@ -59,8 +59,10 @@ module.exports = new Transformer({
 
             // TODO: Parsing code with regex is bad. Replace with babel transform or something.
             let code = node.value;
-            code = code.replace(/import ((?:.|\n)*?) from (['"].*?['"]);?/g, (m, _, s) => {
-              if (s.slice(1, -1) !== 'your-component-library') {
+            code = code.replace(/import ((?:.|\n)*?) from (['"].*?['"]);?/g, (m, x, s) => {
+              if (meta === 'example' && s[1] === '.') {
+                exampleCode.push(`import ${x} from "extract:${s.slice(1, -1)}.mdx";`);
```
devongovett, line 64: Basically if you import a relative path (starting with "./") from an MDX file, it will import the Reusable Wrappers section of the referenced file. That allows components to be reused between pages.

majornista:
Within the ColorPicker, when the colorSpace changes with the Picker, should the colorSpace also change for the ColorArea and corresponding ColorSlider?

packages/@react-spectrum/color/src/ColorEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <Item key="hex">{formatter.format('hex')}</Item>
+          <Item key="rgb">{formatter.format('rgb')}</Item>
+          <Item key="hsl">{formatter.format('hsl')}</Item>
+          <Item key="hsb">{formatter.format('hsb')}</Item>
+        </Picker>
+        {format === 'hex'
+          ? <ColorField isQuiet width="size-1000" aria-label={formatter.format('hex')} />
+          : getColorChannels(format).map(channel => (
+            <ColorField key={channel} colorSpace={format === 'hex' ? 'rgb' : format} channel={channel} isQuiet width="size-400" flex UNSAFE_style={{'--spectrum-textfield-min-width': 0} as CSSProperties} />
+          ))}
```
majornista, line 45: You already filter out the special case where `format === 'hex'`:

```suggestion
        {format === 'hex'
          ? <ColorField isQuiet width="size-1000" aria-label={formatter.format('hex')} />
          : getColorChannels(format).map(channel => (
            <ColorField key={channel} colorSpace={format} channel={channel} isQuiet width="size-400" flex UNSAFE_style={{'--spectrum-textfield-min-width': 0} as CSSProperties} />
          ))}
```

packages/@react-spectrum/color/src/ColorEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {useDOMRef} from '@react-spectrum/utils';
+import {useLocalizedStringFormatter} from '@react-aria/i18n';
+
+export interface SpectrumColorEditorProps {
+  /** Whether to hide the alpha channel color slider and color field. */
+  hideAlphaChannel?: boolean
+}
+
+function ColorEditor(props: SpectrumColorEditorProps, ref: DOMRef<HTMLDivElement>) {
+  let [format, setFormat] = useState<ColorSpace | 'hex'>('hex');
```
majornista, line 21:
```suggestion
  let [format, setFormat] = useState<ColorSpace | 'hex'>('hex');
  let colorSpace = format === 'hex' ? 'hsb' : format;
  let channels = getColorChannels(colorSpace);
```

packages/@react-spectrum/color/src/ColorEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+  let domRef = useDOMRef(ref);
+  let formatter = useLocalizedStringFormatter(intlMessages, '@react-spectrum/color');
+
+  return (
+    <div className={style({display: 'flex', flexDirection: 'column', gap: 4})()} ref={domRef}>
+      <div className={style({display: 'flex', gap: 4})()}>
+        <ColorArea colorSpace="hsb" xChannel="saturation" yChannel="brightness" />
+        <ColorSlider colorSpace="hsb" channel="hue" orientation="vertical" />
```
majornista, line 29:
```suggestion
        <ColorArea colorSpace={colorSpace} xChannel={channels[1]} yChannel={channels[2]} />
        <ColorSlider colorSpace={colorSpace} channel={channels[0]} orientation="vertical" />
```

majornista (commented):
Suggestions to use appropriate ColorArea for colorSpace.

devongovett:
> Within the ColorPicker, when the colorSpace changes with the Picker, should the colorSpace also change for the ColorArea and corresponding ColorSlider? This would be similar to how the Photoshop desktop color picker works.

This behavior seems to vary widely among color pickers. In Photoshop/Illustrator web, the dropdown next to the color fields only affects those, and not the color wheel/color area. I believe the design I followed here is based on XD, which also works this way - the ColorArea/ColorSlider is always in HSB, and the dropdown next to the color fields only affects those, not the whole picker. Figma works similarly here.

Anyway this is just the default ColorEditor, you can build a custom one that does update similar to the example in the docs for color channel sliders. We'll see if we get feedback on the default and can adjust if needed.

ktabors (approved):

packages/@react-spectrum/color/src/ColorEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+        {!props.hideAlphaChannel &&
+          <ColorSlider channel="alpha" orientation="vertical" />
+        }
+      </div>
+      <div className={style({display: 'flex', gap: 4})()}>
+        <Picker aria-label="Color format" isQuiet width="size-700" menuWidth="size-1000" selectedKey={format} onSelectionChange={f => setFormat(f as typeof format)}>
```
majornista, line 35:
```suggestion
                <Picker
          aria-label="Color format"
          isQuiet
          width="size-700"
          menuWidth="size-1000"
          selectedKey={format}
          onSelectionChange={f => {
            if (f !== format) {
              setFormat(f as typeof format);
              setZChannelIndex(0);
            }
          }}>
```

majornista:
> > Within the ColorPicker, when the colorSpace changes with the Picker, should the colorSpace also change for the ColorArea and corresponding ColorSlider? This would be similar to how the Photoshop desktop color picker works.
> 
> This behavior seems to vary widely among color pickers. In Photoshop/Illustrator web, the dropdown next to the color fields only affects those, and not the color wheel/color area. I believe the design I followed here is based on XD, which also works this way - the ColorArea/ColorSlider is always in HSB, and the dropdown next to the color fields only affects those, not the whole picker. Figma works similarly here.
> 
> Anyway this is just the default ColorEditor, you can build a custom one that does update similar to the example in the docs for color channel sliders. We'll see if we get feedback on the default and can adjust if needed.

Fair enough.

Can we add  `--spectrum-colorarea-border-color: ButtonBorder;` for `forced-colors` mode here: https://github.com/adobe/react-spectrum/blob/7cc7549c0bf041f1e8fe01ced58d1e195b499056/packages/%40adobe/spectrum-css-temp/components/colorarea/skin.css#L28-L30

packages/@react-spectrum/color/src/ColorEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <Item key="hex">{formatter.format('hex')}</Item>
+          <Item key="rgb">{formatter.format('rgb')}</Item>
+          <Item key="hsl">{formatter.format('hsl')}</Item>
+          <Item key="hsb">{formatter.format('hsb')}</Item>
+        </Picker>
+        {format === 'hex'
+          ? <ColorField isQuiet width="size-1000" aria-label={formatter.format('hex')} />
+          : getColorChannels(format).map(channel => (
+            <ColorField key={channel} colorSpace={format === 'hex' ? 'rgb' : format} channel={channel} isQuiet width="size-400" flex UNSAFE_style={{'--spectrum-textfield-min-width': 0} as CSSProperties} />
```
majornista, line 44: You already filter out the special case where format === 'hex' at [Line 41](https://github.com/adobe/react-spectrum/pull/6272/files/6284aece9005065d4ac163c8394be36c26abf484#diff-f4419983fd216e15e3e9d79cf93d497159c1ffb2480774d499fcc43be6753c87R41):

```suggestion
            <ColorField key={channel} colorSpace={format} channel={channel} isQuiet width="size-400" flex UNSAFE_style={{'--spectrum-textfield-min-width': 0} as CSSProperties} />
```

devongovett:
> Can we add  --spectrum-colorarea-border-color: ButtonBorder; for forced-colors mode here:

I think I already did that over on #6199, and looks like it is in the current branch: https://github.com/adobe/react-spectrum/blob/color-picker/packages/%40adobe/spectrum-css-temp/components/colorarea/skin.css#L23

packages/@react-spectrum/color/src/ColorEditor.tsx
devongovett: not good enough for typescript unfortunately. it still complains unless I repeat the check :/

packages/@react-spectrum/color/intl/en-US.json
LFDanLu, line 1: will need to run the missingTranslation script for this one too

LFDanLu (dismissed):
Partial review since I'm still going through rest of the code changes and docs, but approving for testing next week

packages/@react-spectrum/color/intl/en-US.json
devongovett: that script only picks up individual missing strings from already existing languages, not entire missing language files. That is the case that errors. If an entire language is missing, we already fall back to English.

PR #6286, opened 2024-04-29
Fix crash in DateRangePicker when using SSR optimized localized strings

devongovett:
In our Next.js example app, when selecting a range in a DateRangePicker with reversed order, the page would crash with an error about a missing localized string. In that setup, the optimize-locales-plugin is used to remove all localized strings from the JS bundle, and they are injected into the page during SSR. However, this particular code used the dictionary imported from the JS bundle, and did not load the global dictionary from the window like other parts of the code.

### DIFF
--- a/packages/@react-stately/datepicker/src/utils.ts
+++ b/packages/@react-stately/datepicker/src/utils.ts
@@ -105,9 +105,10 @@ export function getRangeValidationResult(
 
   let result = mergeValidation(startValidation, endValidation);
   if (value.end != null && value.start != null && value.end.compare(value.start) < 0) {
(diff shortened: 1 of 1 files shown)


reidbarber (commented):
Ran locally and I'm getting this:

```
Unhandled Runtime Error
Error: Could not find intl message rangeReversed in en-US locale

Call Stack
$5b160d28a433310d$export$c17fa47878dc55b6.getStringForLocale
```

devongovett:
How did you run? I guess you have to either run verdaccio or do a build and then yarn link the packages into the example app

reidbarber:
Ahh yeah, I forgot the local wasn't already linked.

reidbarber (approved):

PR #6287, opened 2024-04-29
Fix styling of ColorPicker examples in RAC docs

devongovett:
Fixes 2 issues found in testing:

1. On iPhone, the ColorPicker had blue text at a small font size due to browser default styles.
2. The ColorPicker popover overflowed instead of scrolling. This was actually a problem with all popovers in the RAC docs (e.g. date picker with small browser viewport) as well. Please test all RAC components with popovers in the docs in addition to ColorPicker.

### DIFF
--- a/packages/react-aria-components/docs/ColorArea.mdx
+++ b/packages/react-aria-components/docs/ColorArea.mdx
@@ -64,6 +64,7 @@ import {ColorArea, ColorThumb} from 'react-aria-components';
   width: 192px;
   height: 192px;
   border-radius: 4px;
+  flex-shrink: 0;
 }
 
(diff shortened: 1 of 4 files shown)


majornista:
@devongovett When the ColorPicker popover overflows, the ColorThumb should scroll into view when the value changes using the keyboard.

snowystinger (approved):
Checked DatePickers, Popovers, Select, etc in docs, all of them are scrolling.

ktabors (approved):

majornista (approved):
Approving with useMove autoscrolling opened as a separate issue: https://github.com/adobe/react-spectrum/issues/6299

PR #6289, opened 2024-04-29
Add story with multiple rows in ColorSwatchPicker

devongovett:
Like the title says


ktabors (commented):
When I try to use the controls for this new story I get this error. Reload the page and it loads with this change fine. The default story doesn't have this problem.
image

devongovett:
hmm you don't like my random colors huh?

snowystinger (approved):

PR #6301, opened 2024-05-01
Workaround RSP tree shaking issue with RAC

devongovett:
This avoids all of RAC Checkbox and its dependencies being included in RSP checkbox when building an app with Parcel. The underlying issue is a bug in SWC, the minifier used by Parcel. In the meantime, this workaround moves the contexts we import from RSP when not using the RAC implementation to a separate file so that only the contexts are included and not the entire implementation. I verified this is working in a test app.


packages/react-aria-components/src/RSPContexts.ts
```diff
@@ -0,0 +1,33 @@
...
+// (they use hooks). When included in the same file as the implementation, Parcel includes both the context
+// and implementation even if only the context is imported. We moved these contexts to a separate file so
+// that the implementation does not get included in the bundle. This will be removed once the tree shaking
+// issue is fixed.
+export const CheckboxContext = createContext<ContextValue<CheckboxProps, HTMLLabelElement>>(null);
+export const ColorAreaContext = createContext<ContextValue<Partial<ColorAreaProps>, HTMLDivElement>>(null);
+export const ColorFieldContext = createContext<ContextValue<ColorFieldProps, HTMLDivElement>>(null);
+export const ColorSliderContext = createContext<ContextValue<Partial<ColorSliderProps>, HTMLDivElement>>(null);
+export const ColorWheelContext = createContext<ContextValue<Partial<ColorWheelProps>, HTMLDivElement>>(null);
+export const HeadingContext = createContext<ContextValue<HeadingProps, HTMLHeadingElement>>({});
```
devongovett, line 33: The most optimal thing to do would be to make each of these a separate file but that seemed a bit annoying. So if you import Checkbox you'll currently also get the other contexts, but they are tiny (one function call) so doesn't seem like a big deal.

snowystinger (approved):

LFDanLu (approved):

PR #6302, opened 2024-05-01
Make FormContext match our other contexts

devongovett:
Noticed in audit. FormContext should work like the other contexts in RAC. It should be a slotted value accepting all Form props/ref. Form should consume it, and then also provide its own props via children (not every component does that part right now). Also added some examples to the advanced customization section of the docs.

### DIFF
--- a/packages/react-aria-components/docs/Form.mdx
+++ b/packages/react-aria-components/docs/Form.mdx
@@ -162,59 +162,23 @@ See the [Forms](forms.html) guide to learn more about form validation in React A
 
 ### Validation behavior
 
-By default, native HTML form validation is used to display errors and block form submission.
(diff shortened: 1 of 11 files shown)


packages/react-aria-components/docs/Form.mdx
```diff
@@ -162,61 +162,25 @@ See the [Forms](forms.html) guide to learn more about form validation in React A
...
-```
-
-The `validationBehavior` for a Form can be accessed from within any component inside a given From by using the `FormContext`.
-
-```tsx
-import {FormContext} from 'react-aria-components';
-
-let {validationBehavior} = useContext(FormContext);
-```
-
```
devongovett, line 219: Removed default example since we already have one above showing native validation. Adjusted the aria validation example to show a simpler custom validation example. Not sure we should recommend people validate emails via regex. Moved the context example to advanced customization below.

reidbarber (approved):
LGTM

snowystinger (approved):

yihuiliao (approved):

PR #6312, opened 2024-05-01
Fix infinite resizing issue in overlay positioning

devongovett:
Closes https://github.com/adobe/react-spectrum/issues/6241


packages/@react-aria/overlays/test/calculatePosition.test.ts
```diff
@@ -199,123 +199,123 @@ describe('calculatePosition', function () {
   const testCases = [
     {
       placement: 'left',
-      noOffset: [50, 200, undefined, 196, 350],
-      offsetBefore: [-200, 50, undefined, 50, 500],
+      noOffset: [50, 200, undefined, 100, 350],
```
LFDanLu, line 202: For others who may be looking at the tests and wondering what these number mean:
- first value is either `position: left` of the overlay or a value used to calculate the corresponding `position: right`
- second value is either `position: top` of the overlay or a value used to calculate the corresponding `position: bottom`
- third value is arrowOffsetLeft, aka the number of px that the arrow is shifted horizontally with respect to the left edge of the overlay if any (only applies for top/bottom placements)
- fourth value is arrowOffsetTop, aka the number of px that the arrow is shifted with respect to the top edge of the overlay (only applies for left/right placements)
- the last value is the overlay's calculated max height, essentially the amount of space that the overlay can grow before hitting the boundary's edge 


Translation of these values for this test case (placement: left + noOffset):
```
        checkPosition(
          placement, getTargetDimension({left: 250, top: 250}), testCase.noOffset
        );
```
becomes:

- container is `600x600` with a padding of `50` so valid render area is a rect of `{left: 50, right: 550, top: 50, bottom: 550}`
- button is placed at `(250, 250)` with a height of `(100, 100)` => `{left: 250, right: 350, top: 250, bottom: 350}`
- overlay is `200x200` and has a calculated position of `{left: 150, right: 350, top: 200, bottom: 400}` and a maxHeight of `350` since it is `550 (boundary bottom) - 200 (top of overlay)` 
- arrow size is `8`, and an arrowOffsetTop of `100` aka it is placed 100px from the top edge of the overlay which is the center of overlay and matches the center of the button: `200 (overlay top) + 100 (arrow offset) = 300 (center of button)`
```

LFDanLu:
https://www.chromatic.com/build?appId=5f0dd5ad2b5fc10022a2e320&number=742

snowystinger (approved):
Checked chrome and FF, looking good

LFDanLu (approved):
LGTM, also stepped through most of the test logic and the numbers seem to check out

PR #6329, opened 2024-05-04
Internationalized/NumberParser: fix TS5.5 build errors

Renegade334:
This small change pre-emptively addresses build errors that will appear under TypeScript 5.5, caused by changes relating to Intl.NumberFormat in microsoft/TypeScript#56902.

Unit tests are unaffected.


snowystinger (approved):
Looks fine. We do the same `?? 0` up above and they should always be defined
Thanks for the PR!

LFDanLu (approved):

PR #6346, opened 2024-05-08
Fix synthetic links not using `useHref`

joshuajaco:
Closes https://github.com/adobe/react-spectrum/issues/6347

## ✅ Pull Request Checklist:


packages/@react-aria/utils/src/openLink.tsx
```diff
@@ -146,9 +146,10 @@ function openSyntheticLink(target: Element, modifiers: Modifiers) {
   getSyntheticLink(target, link => openLink(link, modifiers));
 }
 
-export function getSyntheticLinkProps(props: LinkDOMProps) {
+export function useSyntheticLinkProps(props: LinkDOMProps) {
+  let router = useRouter();
   return {
-    'data-href': props.href,
+    'data-href': props.href ? router.useHref(props.href) : undefined,
```
devongovett, line 152: This could cause issues due to a hook being called conditionally. If we're sure that there is always an href when this function is called, we could probably make href required in the type.

devongovett (commented):
Thanks, this makes sense.

packages/@react-aria/utils/src/openLink.tsx
joshuajaco: I agree that this can cause issues, however I directly mirrored the implementation of `useLinkProps`:

https://github.com/adobe/react-spectrum/blob/10bac541a2a5c666b443fe70f313455a405a5c77/packages/%40react-aria/utils/src/openLink.tsx#L160-L170

Also `useSyntheticLinkProps` is used in `useGridListItem`, where we do not always have an href.

If we want to prevent `useHref` from being called conditionally we would have to instead split up `useGridListItem` in to two hooks, one with href and one without. We'd also have to conditionally render between two components everywhere `useGridListItem` is used to swap between the new two hooks.

Alternatively we could call `useHref` with an empty string. After a quick search this also seems to be a valid href according to this detailed [stackoverflow comment](https://stackoverflow.com/a/43340108).

devongovett: ok if it was there before I guess it's probably fine for now.

devongovett (approved):

snowystinger (approved):

PR #6349, opened 2024-05-08
RAC: Support removing data-rac attribute

reidbarber:
For discussion:

This provides a way to remove the data-rac attribute from elements by passing in `null`.

## ✅ Pull Request Checklist:


reidbarber:
Will close until we see a good use case. This breaks the tailwind plugin so I would be wary of documenting it IMO.

Issue #5258, opened 2023-10-17
Support for `data-focus-within` in RAC


jaknas:
### Provide a general summary of the feature here

Originally opened in the discussion: https://github.com/adobe/react-spectrum/discussions/4849#discussion-5463826

For all grid components that support interactive children, I'd like to have possibility to style the `<Item/>` component based on whether the focus is contained in the descendant of this `<Item/>`. 

For example:
My solution requires to style a `<Row/>` component when it's being interacted with. Currently, this can be done with using `data-hovered` or `data-focus-visible` selectors. However, when my focus moves to one of the interactive children, there's no attribute on the `<Row/>` component I could use, that I could use to apply the same styles as e.g `data-hovered` (that'd signify that this row is being interacted with)

### 🤔 Expected Behavior?

Additional prop/data-attribute that would be enabled once the focus is on the element's descendants. Best candidate seems to be using focus-within selector, but I'm open to hear about drawbacks of this or some other ideas.			

### 😯 Current Behavior

No way of knowing that focus is contained in the element's descendants.

### 💁 Possible Solution

Possibly adding the `useFocusWithin` hook to the [TableRow](https://github.com/adobe/react-spectrum/blob/7120dd375ca6e1db4315a2940ad15c62791e9489/packages/react-aria-components/src/Table.tsx#L996) component.

### 🔦 Context

I'd like to style a grid `role="row"` element based on whether any of the descendants is currently focused.

LFDanLu:
Just to add to @reidbarber comment, we'd also want to add `focus-within` tracking to Cell as well

sookmax:
Hello guys, do we want to add a separate `data-focus-within` and leave the existing `data-focused` intact, or is it okay to just enhance `data-focused` to remain activated even when the focus shifts to one of its children?

I might be missing some details here, but if the latter is okay then I think this can be solved as simple as changing `useFocusRing()` in `TableRow` to `useFocusRing({ within: true })`, and in the case of `GridList`, it seems this is already handled by `useGridListItem()` and `useSelectableItem()` within it:
https://github.com/adobe/react-spectrum/blob/d2049a5831e57fcad7a2d489fa0eacbe155a6db9/packages/%40react-aria/gridlist/src/useGridListItem.ts#L62-L71

If what we want is the former.. then a little more guidance would be of great help 🙏

PR #6350, opened 2024-05-08
Add renderProp support for if a Table row has visible focus within it

LFDanLu:
Closes https://github.com/adobe/react-spectrum/issues/5258

useful for making a row style that remains applied when anything in the table row is focused, not just the row itself (e.g. make the whole row blue when a cell inside is keyboard focused). Needed for S2.

Other changes to come as I run into them while working on S2

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/react-aria-components/src/Table.tsx
LFDanLu, line 1: Unneeded, can instead just style the elements you pass to the Table cell directly for focusVisible etc. Will re-add if it turns out we need focusVisibleWithin for cell styling

LFDanLu:
@jaknas thanks for the issue link (may have slipped my mind to see if anyone filed a issue for this change when making the PR 😅). I'm still evaluating what other elements should also expose some version of isFocusVisibleWithin as I work on styling an adjacent component that uses RAC table but can definitely get this merged in for now so a nightly can built. Do you also need info for `isFocusWithin` or is `isFocusVisibleWithin` sufficient for your use case?

LFDanLu:
@jaknas gotcha, I'll see about getting this merged in early then so we can get a nightly out for testing.

ktabors (approved):

reidbarber (approved):

snowystinger (approved):

PR #6356, opened 2024-05-09
Translations Update, please integrate

rgeraghty:
--no-edit


snowystinger (approved):

LFDanLu (approved):

packages/@react-aria/tree/intl/de-DE.json
```diff
@@ -1,4 +1,4 @@
 {
-  "expand": "Expand",
-  "collapse": "Collapse"
+  "collapse": "Reduzieren",
+  "expand": "Erweitern"
```
levrik, line 3: These translations are technically correct but do not make a lot of sense in the context they are used. I doubt that especially vision impaired persons will understand what is meant here.

I would propose "Einklappen" for "collapse" and "Ausklappen"/"Aufklappen" for "expand". These literally translate to something like "fold up" and "fold out". Also see https://www.dict.cc/englisch-deutsch/to+expand.html and https://www.dict.cc/englisch-deutsch/to+collapse.html (see the translations for "[in a tree view]").

levrik: Opened a PR: #6455

PR #6410, opened 2024-05-17
Fix CommonJS build and add Jest setup to test app

devongovett:
Pulls in https://github.com/parcel-bundler/parcel/pull/9722 to fix the CommonJS build. Also adds a Jest test setup to one of our example apps that renders many components to verify they work in that environment. This will run in our verdaccio CI script. I verified that the test failed using the old build and should now pass.


devongovett:
Verdaccio deploy of Next.js app: https://reactspectrum.blob.core.windows.net/reactspectrum/b4d69f0b2a8b7a4222f600c24b85db415629e93d/verdaccio/next/index.html

The test also ran in this CircleCI run: https://app.circleci.com/pipelines/github/adobe/react-spectrum/20292/workflows/753efde2-72df-497d-9a91-c90523140ec7/jobs/301441 (search for "smoke")

LFDanLu (approved):
LGTM did a quick sweep of the verdaccio apps, lets get this tested some more on monday

yihuiliao (approved):
noticed some weird behavior in the RAC Tailwind Example app with the popover and combobox. when you select an item in the popover, it scrolls the page up and then you have to click on the page to scroll the page again. with the combobox, there's just some weird animation when you open the combobox. otherwise, didn't notice anything weird in the other test apps

devongovett:
@yihuiliao could you check if these also happen on main?

yihuiliao:
> @yihuiliao could you check if these also happen on main?

yep, just checked on main and seems like it also happens there so unrelated to the changes in this PR

PR #6494, opened 2024-06-04
Load more collection element support in RAC Table and Tree

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Try the RAC Table/Tree loading stories starting with https://reactspectrum.blob.core.windows.net/reactspectrum/830795d557f86549e355253e5a2a2cddd7cfc696/storybook/index.html?path=/story/react-aria-components--table-loading-body-wrapper-story&providerSwitcher-express=false. Toggle the isLoading button and verify that the extra loading row appears with the proper messaging and attributes. These rows should not interfere with keyboard navigation and should not be focusable. Also try the table loading RAC stories, both for virtualized and non virtualized

## 🧢 Your Project:

RSP


packages/react-aria-components/src/Table.tsx
```diff
@@ -102,7 +102,8 @@ class TableCollection<T> extends BaseCollection<T> implements ITableCollection<T
 
   getLastKey() {
     let rows = [...this.getChildren(this.body.key)];
-    return rows[rows.length - 1]?.key;
+    let lastRow = rows[rows.length - 1];
+    return (lastRow?.type === 'loader' ? lastRow.prevKey : lastRow?.key) as Key;
```
LFDanLu, line 106: not sure why typescript insists that this return is different from BaseCollection's (somehow BaseCollection's return type is Key and not Key | null)

packages/react-aria-components/src/Table.tsx
```diff
@@ -93,7 +93,7 @@ class TableCollection<T> extends BaseCollection<T> implements ITableCollection<T
   }
 
   get size() {
-    return [...this.getChildren(this.body.key)].length;
+    return [...this.getChildren(this.body.key)].filter(item => item.type !== 'loader').length;
```
devongovett, line 96: I think `size` is used to calculate the `aria-rowcount` right? Shouldn't the loader be included in that? It is presented as a row in the table so seems like maybe it should? What else is `size` used for that caused a problem here?

packages/react-aria-components/src/Table.tsx
```diff
@@ -813,6 +812,7 @@ const _ColumnResizer = forwardRef(ColumnResizer);
 export {_ColumnResizer as ColumnResizer};
 
 export interface TableBodyRenderProps {
+  // TODO: I wonder if this should also be true if the table only has the loading row. I figure we should keep loading and empty state as separate states
```
devongovett, line 815: Yeah I think separate.

packages/react-aria-components/src/Table.tsx
```diff
@@ -839,10 +839,11 @@ export const TableBody = /*#__PURE__*/ createBranchComponent('tablebody', <T ext
   let {dragAndDropHooks, dropState} = useContext(DragAndDropContext);
   let isDroppable = !!dragAndDropHooks?.useDroppableCollectionState && !dropState?.isDisabled;
   let isRootDropTarget = isDroppable && !!dropState && (dropState.isDropTarget({type: 'root'}) ?? false);
+  let isTableEmpty = collection.size === 0 && collection.rows.filter(row => row.type === 'loader').length === 0;
```
devongovett, line 842: If `collection.size` included loading rows I guess you wouldn't need this change?

packages/react-aria-components/src/Tree.tsx
```diff
@@ -469,6 +533,13 @@ function flattenTree<T>(collection: TreeCollection<T>, opts: TreeGridCollectionO
         if (hasContentNode) {
           clone.index = node?.index != null ? node?.index - 1 : 0;
         }
+
+        // For loader nodes that have a parent (aka non-root level loaders), these need their levels incremented by 1 for parity with their sibiling rows
+        // (Collection only increments the level if it is a "item" type node).
```
devongovett, line 538: Couldn't you render the loader as a child of the loading row instead of a sibling? That would give you more control over where it goes.

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -392,3 +393,134 @@ const MyCheckbox = ({children, ...props}: CheckboxProps) => {
     </Checkbox>
   );
 };
+
+const MyTableLoader = () => {
+  return (
+    <UNSTABLE_TableLoader>
+      {({isTableEmpty}) => (
+        <span>
+          {isTableEmpty ? 'Loading spinner' : 'Load more spinner'}
```
devongovett, line 402: If we recommend using `renderEmptyState` to show the initial loading spinner, do we need this `isTableEmpty` render prop?

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -392,3 +393,134 @@ const MyCheckbox = ({children, ...props}: CheckboxProps) => {
...
+  render: TableLoadingRowRenderWrapper,
+  args: {
+    isLoading: false
+  },
+  name: 'Table loading, row renderer wrapper and dep array'
+};
+
+
+// TODO: when writing the docs, prefer this pattern in the docs, establish renderEmptyState as something distinct from empty + loading
+// (aka users should use empty state for if their table is empty and not loading, but should render the TableLoader for empty + loading and for has items + loading)
```
devongovett, line 491: Oh. What was the reason for not using `renderEmptyState` for loading? Not sure which one is better I guess, but seems like there was a bit of extra complexity to handle the empty state with TableLoader.

packages/react-aria-components/src/Table.tsx
LFDanLu: Thats right, but I was hoping that we'd modify https://github.com/adobe/react-spectrum/blob/10bac541a2a5c666b443fe70f313455a405a5c77/packages/%40react-aria/table/src/useTable.ts#L71-L73 to essentially do `[...this.getChildren(this.body.key)].length` as part of its calculation. For the most part though, checks on `collection.size` like in https://github.com/adobe/react-spectrum/blob/10bac541a2a5c666b443fe70f313455a405a5c77/packages/%40react-aria/table/src/useTableColumnHeader.ts#L83 and https://github.com/adobe/react-spectrum/blob/10bac541a2a5c666b443fe70f313455a405a5c77/packages/%40react-aria/table/src/useTableSelectionCheckbox.ts#L67 would want to still resolve as if the collection were truly empty even when the loading spinner exists in an empty table.

> What else is size used for that caused a problem here?

The two lines linked above were problematic since the table would still support keyboard navigation when it was in a loading state, which was at odds from previous behavior since the table itself used to be the only tab stop if it was empty. There were some crashes on keyboard nav as well but those are fixed now by the changes in GridKeyboardDelegate. If we are ok with the keyboard nav experience being different between the loading state and the empty state, I can revert this change

packages/react-aria-components/src/Table.tsx
LFDanLu: yep thats right

packages/react-aria-components/src/Tree.tsx
LFDanLu: It actually already is rendered as the child of the loading row (it has the proper parent, nextKey, prevKey etc), it is just that this line only adds an extra +1 for item type nodes: https://github.com/adobe/react-spectrum/blob/10bac541a2a5c666b443fe70f313455a405a5c77/packages/react-aria-components/src/Collection.tsx#L302

I could also have it do so for "loader" type elements, assuming that we then accept "loaders" as a common type node for collections in general (which now that I think about it again makes sense, was a bit wary of adding that to the base Collection implementation previously)

packages/react-aria-components/stories/Table.stories.tsx
LFDanLu: we wouldn't need `isTableEmpty` any more, this loader would then be specifically for when we are in a "load more" case/for loading nested row loaders. However, it feels a bit confusing that the user would use `renderEmptyState` to render the initial loading spinner IMO since then they wouldn't use `UNSTABLE_TableLoader` in their `renderEmptyState` call, they would just pass it a div or something since https://github.com/adobe/react-spectrum/blob/10bac541a2a5c666b443fe70f313455a405a5c77/packages/react-aria-components/src/Table.tsx#L856-L864 wraps it in a row already. Also feels confusing that the end use would still need to render the  `UNSTABLE_TableLoader` for the `loadMore` case still as well

packages/react-aria-components/stories/Table.stories.tsx
LFDanLu: Mainly felt confusing that renderEmptyState would then have mixed responsibilities for rendering the true empty state and the initial loading state

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -392,3 +393,134 @@ const MyCheckbox = ({children, ...props}: CheckboxProps) => {
...
+      <TableHeader columns={columns}>
+        <Column><MyCheckbox slot="selection" /></Column>
+        <Collection items={columns}>
+          {(column) => (
+
+            <Column isRowHeader={column.isRowHeader}>{column.name}</Column>
+          )}
+        </Collection>
+      </TableHeader>
+      <TableBody renderEmptyState={() => renderEmptyLoader({isLoading})}>
```
LFDanLu, line 505: The idea now is that the user will handle the empty state and the "isLoading" state (aka no items but is loading) via renderEmptyState, no longer using UNSTABLE_TableLoader for this. Rendering the "loadMore" state (aka items exist but is loading) is the responsibility of the user where they should only include the UNSTABLE_TableLoader when they are actually performing load more and not when the table is still empty. 

This simplifies the collections changes we need to do since the way to compute collection size remains the same

packages/react-aria-components/src/Table.tsx
```diff
@@ -1219,3 +1223,34 @@ function RootDropIndicator() {
     </tr>
   );
 }
+
+// TOOD: no props for now, maybe get rid of this? Might be good to keep it just in case
+export interface TableLoaderProps {}
+
+export const UNSTABLE_TableLoader = createLeafComponent('loader', function TableLoader<T extends object>(props: TableLoaderProps, ref: ForwardedRef<HTMLTableRowElement>, item: Node<T>) {
```
LFDanLu, line 1230: Now that this loader is only for load more, open to suggestions for the naming

packages/@react-aria/loading/src/useLoadOnScroll.ts
LFDanLu, line 1: didn't really want to move this into a separate package but none of the existing aria packages seemed to be a good fit

packages/@react-aria/loading/src/useLoadOnScroll.ts
```diff
@@ -0,0 +1,88 @@
...
+  /** Whether data is currently being loaded. */
+  isLoading?: boolean,
+  /** Handler that is called when more items should be loaded, e.g. while scrolling near the bottom.  */
+  onLoadMore?: () => void,
+  // TODO: decide on default here
+  /**
+   * The amount of offset (in pixels) from the bottom of your scrollable region that should trigger load more.
+   * @default 25
+   */
+  scrollOffset?: number
```
LFDanLu, line 26: Open question, I could make this default to 0 I guess? Maybe a percentage of the scroll body height?

packages/@react-aria/loading/src/useLoadOnScroll.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// TODO: discuss if it would be ok to just attach event to ref...
+interface LoadOnScrollAria {
+  /** Props for the scrollable region. */
+  scrollViewProps: {
+    onScroll: () => void
+  }
```
LFDanLu, line 34: I shifted this over to be scrollViewProps so it could be used in useVirtualizer (though I guess I could have useVirtualizer just return a empty func for its onVisibleRectChange). I would've preferred attaching the scroll event directly to the ref via useEvent though...

packages/@react-aria/loading/src/useLoadOnScroll.ts
```diff
@@ -0,0 +1,88 @@
...
+    // not if a local state change occurred.
+    let wasLoading = isLoadingRef.current;
+    if (props !== prevProps.current) {
+      isLoadingRef.current = isLoading;
+      prevProps.current = props;
+    }
+
+    // TODO: this actually calls loadmore twice in succession on intial load because after the first load
+    // the scrollable element hasn't yet recieved its new height with the newly loaded items... Because of RAC collection needing two renders?
+    let shouldLoadMore = ref?.current
```
LFDanLu, line 67: This one I haven't quite figured out, but I believe the 2 renders RAC collections needs to populate the table body is what is causing two load more calls since the first load more will cause the items to be updated and isLoading to become false, but the height of the scroll body hasn't changed yet since the new items haven't been processed yet hence this will think we can still load more items since the scrollHeight hasn't increased yet and thus it thinks we still have room for more items.

packages/@react-aria/virtualizer/src/Virtualizer.tsx
```diff
@@ -98,48 +99,21 @@ interface VirtualizerOptions {
...
   let {isLoading, onLoadMore} = props;
-  let {setVisibleRect, virtualizer} = state;
+  let {setVisibleRect} = state;
+  let scrollOffset = ref.current?.clientHeight;
+
+  // TODO: The ref previously provided was the table's ref. That ref didn't actually do anything in the old implementation
+  // but now the ref provided to useVirtualizer MUST be the scrollable ref so its a breaking change kinda?
+  // TODO: Also, now that we use ref.currnet.clientHeight/scrollHeight/etc instead of relying on the rect to figure out if we should load more
+  // this will break peoples tests since they need to mock scrollHeight to simulate the virtualizer's total contents height...
+  let {scrollViewProps: {onScroll}} = useLoadOnScroll({isLoading, onLoadMore, scrollOffset}, ref);
```
LFDanLu, line 109: Maybe a breaking change to have useVirtualizer use useLoadOnScroll for the above reasons? Open to discussion if useVirtualizer should not use useLoadOnScroll

packages/@react-spectrum/listbox/test/ListBox.test.js
```diff
@@ -893,12 +900,7 @@ describe('ListBox', function () {
       let listbox = getByRole('listbox');
       let options = within(listbox).getAllByRole('option');
       expect(options.length).toBe(5);
-      let isReact19 = parseInt(React.version, 10) >= 19;
```
LFDanLu, line 896: These test are now consistent React 19 which is nice

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -75,6 +75,10 @@ export class TableLayout<T> extends ListLayout<T> {
     super.validate(invalidationContext);
   }
 
+  // TODO: in the RAC case, we don't explicity accept loadingState on the TableBody, but note that if the user
+  // does happen to set loadingState="loading" or loadingState="loadingMore" coincidentally, the isLoading
+  // part of this code will trigger and the layout will reserve more room for the loading spinner which we actually only use
+  // in RSP
   protected buildCollection(): LayoutNode[] {
```
LFDanLu, line 82: Perhaps will change if we split out all of the spectrum specific logic into a separate implementation

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -18,7 +18,7 @@ import {LayoutNode, ListLayout, ListLayoutOptions, ListLayoutProps} from './List
 import {TableCollection} from '@react-types/table';
 import {TableColumnLayout} from '@react-stately/table';
 
-export interface TableLayoutOptions<T> extends ListLayoutOptions<T> {
+export interface TableLayoutOptions<T> extends Omit<ListLayoutOptions<T>, 'loaderHeight'> {
```
LFDanLu, line 21: loaderHeight didn't actually do anything since buildCollection in TableLayout doesn't use it.

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -336,6 +341,12 @@ export class TableLayout<T> extends ListLayout<T> {
       }
     }
 
+    // TODO: perhaps make a separate buildLoader? Do we need to differentiate the layoutInfo information?
+    // I think the below is ok for now since we can just treat nested loaders/load more as rows
+    if (node.type === 'loader') {
+      height = this.rowHeight;
+    }
```
LFDanLu, line 348: this is an assumption, we may want to open up the loader height to be configurable by the user

packages/react-aria-components/src/Table.tsx
```diff
@@ -312,7 +313,18 @@ export interface TableProps extends Omit<SharedTableProps<any>, 'children'>, Sty
...
+  // Perhaps we should make isLoading be loadingKeys for the eventual section/nested rows loading case
+  // Or should it follow the loadingState api we have for our async hook
+  // TODO: should the below be loadMoreOptions? These are just passed to the hook and making a prop that takes in a object containing the
+  // options might be easier to adjust in the future
+  isLoading?: boolean,
+  onLoadMore?: () => void,
+  scrollOffset?: number,
+  // TODO: keep this for now since we might need it for the body, but eventually perhaps we could just do tableContainerContext?.scrollRef || table ref if it turns out we
+  // don't need the body to be scrollable
+  scrollRef?: RefObject<HTMLElement>
```
LFDanLu, line 327: Some api question for discussion

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -393,6 +394,265 @@ const MyCheckbox = ({children, ...props}: CheckboxProps) => {
...
+  render: OnLoadMoreTable,
+  name: 'onLoadMore table'
+};
+
+// TODO: this doens't quite work because the table body isn't initially rendered thus the ref is scrollRef is undef
+// This could be remedied perhaps by moving the onLoadMore call to TableBody internally and passing down the scrollRef and other load props via context from body.
+// That would guarentee that the ref for the table body is defined at the time the hook renders. Stashed this change locally, to be discussed.
+// TODO: Depenedent on whether or not we need to support the TableBody being scrollable, or can get away with just having the table resize container or the table itself
+// be scrollable. The vertical scrollbar extending to the column and the table body outline disappearing when horizontally scrolled are some of the problems with this
+const OnLoadMoreTableBodyScroll = () => {
```
LFDanLu, line 597: The table body itself being virtualized doesn't quite work/isn't supported yet. Will need to experiment in S2 table to see if we can get away with just having the resize wrapper or table be the only virtualized scroll element candidates.

packages/react-aria-components/src/Table.tsx
LFDanLu: When we finalize the api/useLoadOnScroll, I'll go ahead and add this to all the other collection components

packages/@react-aria/loading/src/useLoadOnScroll.ts
LFDanLu: Also this and many of the other props might need to be prefixed with UNSTABLE

packages/react-aria-components/src/Table.tsx
```diff
@@ -312,7 +313,18 @@ export interface TableProps extends Omit<SharedTableProps<any>, 'children'>, Sty
...
   /** The drag and drop hooks returned by `useDragAndDrop` used to enable drag and drop behavior for the Table. */
-  dragAndDropHooks?: DragAndDropHooks
+  dragAndDropHooks?: DragAndDropHooks,
+  // TODO: finalize these
+  // Perhaps we should make isLoading be loadingKeys for the eventual section/nested rows loading case
+  // Or should it follow the loadingState api we have for our async hook
+  // TODO: should the below be loadMoreOptions? These are just passed to the hook and making a prop that takes in a object containing the
+  // options might be easier to adjust in the future
+  isLoading?: boolean,
+  onLoadMore?: () => void,
```
LFDanLu, line 323: Also should these props be on TableBody instead to match RSP's api (and thus have S2 also match?)

packages/@react-aria/loading/src/useLoadOnScroll.ts
devongovett: I wonder if this should also go in the new `@react-aria/collections` package I am working on. Would it fit there? I guess it's not even really specific to collections though... 🤔

packages/@react-aria/loading/src/useLoadOnScroll.ts
devongovett: In virtualizer this defaulted to 2 page heights from the bottom I believe. It does seem like incorporating the the height would be useful as a trigger for loading more. 25px is quite small if loading is slow.

packages/@react-aria/loading/src/useLoadOnScroll.ts
LFDanLu: Yeah, feels a bit weird to place it there because as you said its not specific to collections really since it could really just be applied to any scrollable body that doesn't use our collections stuff. However, I wouldn't be too upset if we wanted to portray it as being more tightly coupled with collections and chuck it in there still.

packages/@react-aria/loading/src/useLoadOnScroll.ts
LFDanLu: should it just be based off using the ref and querying the client height then? Just worried about the case where the ref isn't defined on first render and thus you get a default that doesn't exist if that makes sense

packages/react-aria-components/src/Table.tsx
devongovett: I also wondered if we wanted to go with one that's specific to the collection it's in or not. We are inconsistent at the moment. Items are specific because they accept different props per collection, but Section and DropIndicator are not (they pass a render function down via context). Which way do we want to go here?

packages/@react-stately/layout/src/TableLayout.ts
snowystinger: I think someone was just asking for this somewhere, but I can't find where. Seems like something we should support

packages/react-aria-components/src/Collection.tsx
```diff
@@ -927,10 +927,22 @@ export function Collection<T extends object>(props: CollectionProps<T>): JSX.Ele
 
   let doc = useContext(CollectionDocumentContext);
   if (doc) {
-    return <CollectionRoot>{children}</CollectionRoot>;
+    // TODO: operates off the assumption that dependencies set on a higher level Collection should propagate down
+    // Makes it simpler for someone to setup nested Tree loaders (only need to set dependency at top level Collection instead of remembering to propagate it to every instance of Collection)
+    // TODO: With the refactor to Collections, I can either apply the CollectionContext.Provider here or in Tree
+    // itself in the content passed to CollectionBuilder. Feels like here is more appropriate so we don't need to remember to do this for each component.
+    return (
```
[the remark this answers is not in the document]
snowystinger: What does this mean for a menu inside a tree?

packages/react-aria-components/src/Table.tsx
```diff
@@ -312,7 +313,18 @@ export interface TableProps extends Omit<SharedTableProps<any>, 'children'>, Sty
   /** Handler that is called when a user performs an action on the row. */
   onRowAction?: (key: Key) => void,
   /** The drag and drop hooks returned by `useDragAndDrop` used to enable drag and drop behavior for the Table. */
-  dragAndDropHooks?: DragAndDropHooks
+  dragAndDropHooks?: DragAndDropHooks,
+  // TODO: finalize these
+  // Perhaps we should make isLoading be loadingKeys for the eventual section/nested rows loading case
+  // Or should it follow the loadingState api we have for our async hook
+  // TODO: should the below be loadMoreOptions? These are just passed to the hook and making a prop that takes in a object containing the
```
snowystinger, line 320: seems reasonable, are any of them dependent on each other? as in, if one is provided, another must be?

packages/react-aria-components/src/Table.tsx
snowystinger: Does it accept different props? or do we think it will eventually?
does it matter if the loader is between items vs at the end? or in a different "level" even though we've flattened the collection?

packages/react-aria-components/src/Tree.tsx
snowystinger: resolved?

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -393,6 +407,265 @@ const MyCheckbox = ({children, ...props}: CheckboxProps) => {
...
+};
+
+export const OnLoadMoreTableStory  = {
+  render: OnLoadMoreTable,
+  name: 'onLoadMore table'
+};
+
+// TODO: this doens't quite work because the table body isn't initially rendered thus the ref is scrollRef is undef
+// This could be remedied perhaps by moving the onLoadMore call to TableBody internally and passing down the scrollRef and other load props via context from body.
+// That would guarentee that the ref for the table body is defined at the time the hook renders. Stashed this change locally, to be discussed.
```
snowystinger, line 607: i think earlier you said it'd match another API to move it down?

packages/@react-aria/virtualizer/src/Virtualizer.tsx
devongovett: Should virtualizer even be responsible for this at all? Maybe we should just call `useLoadOnScroll` directly from within collection components that need it?

packages/@react-aria/loading/src/useLoadOnScroll.ts
```diff
@@ -0,0 +1,88 @@
...
+
+// TODO: discuss if it would be ok to just attach event to ref...
+interface LoadOnScrollAria {
+  /** Props for the scrollable region. */
+  scrollViewProps: {
+    onScroll: () => void
+  }
+}
+
+export function useLoadOnScroll(props: LoadOnScrollProps, ref: RefObject<HTMLElement | null>): LoadOnScrollAria {
```
devongovett, line 37: `useLoadMore`?

packages/react-aria-components/src/Collection.tsx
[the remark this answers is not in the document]
devongovett: I think I also added this in the move to the new package in #6640 fyi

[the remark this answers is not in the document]
devongovett: > What does this mean for a menu inside a tree?

It might still need to invalidate. For example, say you have something like this:

```tsx
<Tree dependencies={[something]}>
  {item => (
    <TreeItem>
      <TreeItemContent>
        {item.name}
        <MenuTrigger>
            <Button>...</Button>
            <Menu items={menuItems}>
              {menuItem => <MenuItem>{menuItem.name} • {something}</MenuItem>}
            </Menu>
        </MenuTrigger>
      </TreeItemContent>
    </TreeItem>
  )}
</Tree>
```

In this case, when `something` changes in the outer scope, we need to invalidate everything that accesses it. That means we need to re-render the Menu item. We could put `dependencies` on the `Menu`, but that wouldn't work because the `TreeItem` containing it still won't re-render. So we need to move it up to the `Tree`. Currently you'd have to put it on both Tree and Menu. I think Daniel was trying to make it a bit easier so you'd only need to set dependencies at the top-level.

packages/@react-stately/layout/src/TableLayout.ts
devongovett: would need to handle when row heights are estimated too

packages/react-aria-components/src/Table.tsx
devongovett: Oh I guess it does: TreeLoader has a `level` render prop. https://github.com/adobe/react-spectrum/pull/6494/files#diff-f6a41f61624976962cf916378a6e89a361a9c2067767dda15fbe6441890b996bR445

packages/react-aria-components/src/Table.tsx
LFDanLu: Devon and I discussed this some more, for now we will be pulling these out entirely and having the user call the useLoadMore hook outside of Table. This emulates how things like useDragAndDrop are called and we can always add these back in the future if need be

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -393,6 +407,265 @@ const MyCheckbox = ({children, ...props}: CheckboxProps) => {
...
+};
+
+interface Character {
+  name: string,
+  height: number,
+  mass: number,
+  birth_year: number
+}
+
+const OnLoadMoreTable = () => {
```
LFDanLu, line 550: Curiously, I just noticed that React complains about "React has detected a change in the order of Hooks called by Component" when the items are loaded... Will have to dig to see why that is

packages/react-aria-components/stories/Table.stories.tsx
LFDanLu: The only that comes to mind is that moving loading to TableBody would match RSP more closely, but anyway for now the user will call this hook outside of table so this is inapplicable for now I guess (bug will still exist since the body ref isn't defined upon the first render but we can move this back in if it turns out we need the body to scroll I guess)

packages/@react-spectrum/listbox/test/ListBox.test.js
LFDanLu: The load more scroll logic now uses the scrollable body's client height vs the scroll height rather to check if it should load more rather than being dependent on us mocking the height of the items and having ScrollView's calculations rect updates happening at the right times. This made it more straight forward to mock.

However, there is still some weird behavior when I just tried to get rid of useVirtualizer and rely on attaching the scroll handler on the ref since some of these tests relied on onScroll being called even with a scroll event doesn't actually trigger via the test (gets called by ScrollView). I think the core issue there is that these tests aren't very representative of a real flow (isLoading doesn't get set/updated by onLoadMore and the scrollHeight isn't updated when onLoadMore is finished being called so it isn't actually simulating additional items being loaded) hence I saw extra load more calls happening since a "load" would finish but then the height wouldn't be changed and thus it would try to load again. Will try to resolve after break, was some additional weirdness that I couldn't quite puzzle out.

packages/@react-aria/virtualizer/src/Virtualizer.tsx
LFDanLu: For posterity: since we are bumping virtualizer package, will try and get rid of useVirtualizer entirely and have RSP components call useLoadOnScroll directly. See https://github.com/adobe/react-spectrum/pull/6494#discussion_r1659422726 for some issues I ran into when trying to do this, will hopefully resolve after break

packages/@react-aria/loading/src/useLoadOnScroll.ts
LFDanLu: See updated comment in this part of the code for some additional issues with trying to use ref.current.clientElemnetcount  as an additional check to try and fix this issue

packages/@react-aria/loading/src/useLoadOnScroll.ts
LFDanLu: updated to be a percentage value based of the clientHeight of your ref

packages/@react-aria/loading/src/useLoadOnScroll.ts
LFDanLu: for posterity: for now we are moving it to the `react-aria/utils` package because RSP will need to call it and we'd rather not import the new collections package (aka `react-aria/collections`) since RSP doesn't use new collections. Can reevaluate later

packages/react-aria-components/src/Table.tsx
LFDanLu: ah thats right, forgot about that one. For now we will go with unique loader collection elements per collection component to match the pattern we have with items. DropIndicator makes sense to be shared since that is more of a useDnD hook think

packages/@react-aria/utils/src/useLoadMore.ts
```diff
@@ -0,0 +1,78 @@
...
+    // not if a local state change occurred.
+    if (props !== prevProps.current) {
+      isLoadingRef.current = isLoading;
+      prevProps.current = props;
+    }
+
+    // TODO: this actually calls loadmore twice in succession on intial load because after the first load
+    // the scrollable element hasn't yet recieved its new height with the newly loaded items... Because of RAC collection needing two renders?
+    // Using ref.current.clientElementCount doesn't work either because the scrollable body may be the Table resize container (which contains the table, header and body: 3 children always)
+    // or the Table (which contains the header and body so 2 children) or the virtualizer in RSP (just contains the rows so has a variable count)
```
LFDanLu, line 61: Not quite sure the best way to go about handling this since the direct children of the scrollable element may not be the actual rows...

packages/@react-aria/utils/src/useLoadMore.ts
```diff
@@ -0,0 +1,78 @@
...
+    let shouldLoadMore = ref?.current
+      && !isLoadingRef.current
+      && onLoadMore
+      && ref.current.clientHeight === ref.current.scrollHeight;
```
LFDanLu, line 65: Note that I got rid of the equivalent logic here from https://github.com/adobe/react-spectrum/blob/b46d23b9919eaec8ab1f621b52beced82e88b6ca/packages/%40react-aria/virtualizer/src/Virtualizer.tsx#L129-L131. Previously, [onVisibleRect](https://github.com/adobe/react-spectrum/blob/b46d23b9919eaec8ab1f621b52beced82e88b6ca/packages/%40react-aria/virtualizer/src/Virtualizer.tsx#L103-L113) change was actually responsible for calling load more for cases in ComboBox when the user opened the dropdown before/during/after the initial set of items were loaded and there was still room available in the dropdown for more items. Now that `onVisibleRect`'s equivalent is `onScroll` which will ONLY trigger on scroll, I've made this layoutEffect less restrictive on when it should call onLoadMore (aka it will call onLoadMore whenever the scrollable body isn't actually scrollable and loading isn't currently happening)

packages/@react-aria/utils/src/useLoadMore.ts
```diff
@@ -0,0 +1,78 @@
...
+      isLoadingRef.current = true;
+      onLoadMore?.();
+    }
+
+  }, [isLoading, onLoadMore, props, ref]);
+
+  // TODO: maybe this should still just return scroll props?
+  // Test against case where the ref isn't defined when this is called
+  // Think this was a problem when trying to attach to the scrollable body of the table in OnLoadMoreTableBodyScroll
+  useEvent(ref, 'scroll', onScroll);
```
LFDanLu, line 77: Bit surprised that this is an issue, double checking if useEvent just doesn't quite work if the ref provided to it isn't immediately attached to an actual node

ktabors:
I'm testing the RSP Table async stories. They load, you scroll to the end of the table (bottom or last row), async loading starts, and you're scrolled back to the first row with no table scroll bar until the async load finishes.
https://reactspectrum.blob.core.windows.net/reactspectrum/6fd0167e7dbea19511d1cb10d28b5de29d5958c5/storybook/index.html?path=/story/tableview--async-loading-server-filtering&providerSwitcher-express=false
https://reactspectrum.blob.core.windows.net/reactspectrum/6fd0167e7dbea19511d1cb10d28b5de29d5958c5/storybook/index.html?path=/story/tableview--async-loading&providerSwitcher-express=false

packages/@react-aria/utils/src/useLoadMore.ts
ktabors: Isn't `isLoadingRef` supposed to indicate if there is already a `onLoadMore()` call, which should prevent the second call? Did the removal of `wasLoading` introduce this? You confirmed this wasn't pre-exixisting?

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -393,6 +408,271 @@ const MyCheckbox = ({children, ...props}: CheckboxProps) => {
...
+  let isLoading = list.loadingState === 'loading' || list.loadingState === 'loadingMore';
+  let scrollRef = useRef<HTMLDivElement>(null);
+  let memoedLoadMoreProps = useMemo(() => ({
+    isLoading: isLoading,
+    onLoadMore: list.loadMore
+  }), [isLoading, list.loadMore]);
+  useLoadMore(memoedLoadMoreProps, scrollRef);
+
+  return (
+    // TODO: but why typescript?
```
ktabors, line 578: I don't understand this TODO.

LFDanLu: ah thanks for the reminder, I can remove this

packages/@react-aria/utils/src/useLoadMore.ts
LFDanLu: The core reason behind this is the two renders required for RAC collections to process new items: https://github.com/adobe/react-spectrum/pull/6494#discussion_r1670904973

High level flow that happens from what I can tell:
Table renders -> first time load is triggered (at this time the scrollable body's scrollHeight === clientHeight aka it isn't scrollable because no content) -> load finishes and returns first page of items, isLoading changes from true to false -> useLoadMore useLayoutEffect runs and collection begins processing the new items. scrollHeight of scrollable body is still the same as clientHeight since the collection hasn't done its second render yet and thus hasn't finished processing the collection so it hasn't rendered the new rows yet -> 2nd loadmore is triggered because of that

packages/@react-aria/utils/src/useLoadMore.ts
LFDanLu: At the point this useEvent is called, Table hasn't rendered anything within it
image
so we hit https://github.com/adobe/react-spectrum/blob/2964e9aafc281888cd65da5bd77512e7f999bd84/packages/%40react-aria/utils/src/useEvent.ts#L26-L28. When the Table's internals are rendered after the collection is processed, nothing in the useEffect def array has changed to cause it to trigger again, hence addEventListener is never called on the table body's ref

packages/@react-spectrum/table/test/Table.test.js
```diff
@@ -4208,19 +4209,25 @@ export let tableTests = () => {
...
     });
 
     it('should automatically fire onLoadMore if there aren\'t enough items to fill the Table', function () {
+      let scrollHeightMock = jest.spyOn(window.HTMLElement.prototype, 'scrollHeight', 'get').mockImplementation(() => 1000);
       let items = [{id: 1, foo: 'Foo 1', bar: 'Bar 1'}];
       let onLoadMoreSpy = jest.fn();
+      let onLoadMore = () => {
+        onLoadMoreSpy();
+        scrollHeightMock = jest.spyOn(window.HTMLElement.prototype, 'scrollHeight', 'get').mockImplementation(() => 2000);
+      };
```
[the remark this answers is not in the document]
LFDanLu: ah thanks, lemme update this

packages/@react-aria/utils/src/useLoadMore.ts
```diff
@@ -0,0 +1,78 @@
...
+  useLayoutEffect(() => {
+    // Only update isLoadingRef if props object actually changed,
+    // not if a local state change occurred.
+    if (props !== prevProps.current) {
+      isLoadingRef.current = isLoading;
+      prevProps.current = props;
+    }
+
+    // TODO: this actually calls loadmore twice in succession on intial load because after the first load
+    // the scrollable element hasn't yet recieved its new height with the newly loaded items... Because of RAC collection needing two renders?
```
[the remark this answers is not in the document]
snowystinger: could you check if the collection changed? I feel like if nothing has changed, then the loadmore call should not be fired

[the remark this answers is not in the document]
LFDanLu: the hook itself isn't really tied to the collections api though, but we could make accept the collection as a prop if need be I guess but it would reduce its flexibility for people who just want the scroll loading behavior without being tied to our collections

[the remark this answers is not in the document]
snowystinger: I'm caught up to your thinking now, thanks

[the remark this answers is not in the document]
LFDanLu: how would said prop work? Would it ignore subsequent triggers of triggers of the useLayoutEffect?

packages/@react-aria/utils/src/useLoadMore.ts
ktabors: These seems fine unless you think there is a performance issue to solve.

packages/@react-aria/utils/src/useLoadMore.ts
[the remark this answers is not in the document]
LFDanLu: For future reference: for now I've included a items options to which a user may pass their currently loaded items so that we can make sure only to try to load more items ONLY if after the previous set of newly loaded items was added, the scrollable body still isn't scrollable. Note that this doesn't actually solve the double loadMore call for RAC table since the new set of items will loaded but the scrollable element is still waiting for the second render for the collection to process these new items and thus add these items to the body.

packages/@react-aria/utils/src/useLoadMore.ts
LFDanLu: For future reference, we may be moving useLoadMore back into the RAC components in the future so that it can access the component's collection for https://github.com/adobe/react-spectrum/pull/6494#discussion_r1669614731. When this happens, we can always place it in TableBody itself and walk up to find the first scrollable element and set that as the scroll ref, thus allowing for the body/table/ or resizable table container to be the scrollable body

packages/react-aria-components/src/Table.tsx
```diff
@@ -916,12 +916,23 @@ export const TableBody = /*#__PURE__*/ createBranchComponent('tablebody', <T ext
...
-    let TR = useElementType('tr');
-    let TD = useElementType('td');
+    let rowProps = {};
+    let rowHeaderProps = {};
+    let style = {};
+    if (isVirtualized) {
+      rowProps['aria-rowindex'] = state.collection.headerRows.length + 1;
+      rowHeaderProps['aria-colspan'] = state.collection.columnCount;
+      style = {display: 'contents'};
+    } else {
```
LFDanLu, line 929: Another interesting this is that making the loader sticky doesn't quite seem to work anymore in the virtualized case since the table body wrapper has "overflow: hidden". In RSP v3 we got around this via https://github.com/adobe/react-spectrum/blob/caff84082a16b3ceec0f409a3f4d9bb982001d7c/packages/%40react-spectrum/table/src/TableViewBase.tsx#L715-L722 which changed all the `overflow: hidden` that VirtualizerItem would set to visible. I can get around this perhaps by extending/subclassing the tablelayout and make the table body layoutinfo have `allowOverflow: true`, wonder if that should be done by default though in the RAC table layout

devongovett (approved):

packages/@react-aria/virtualizer/src/Virtualizer.tsx
```diff
@@ -69,11 +69,14 @@ function Virtualizer<T extends object, V extends ReactNode, O>(props: Virtualize
     layoutOptions
   });
 
-  let {virtualizerProps, scrollViewProps} = useVirtualizer(props, state, ref);
+  useLoadMore({isLoading, onLoadMore, scrollOffset: 1}, ref);
```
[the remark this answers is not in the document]
LFDanLu: I believe the two calculations are equivalent no (aka we always made sure to have one page worth of content loaded)? Maybe my math isn't mathing:

previous code in useVirtualizer:
```
let scrollOffset = virtualizer.contentSize.height - rect.height * 2;
if (rect.y > scrollOffset) {

becomes:

rect.y >  virtualizer.contentSize.height - rect.height * 2
->
scrollTop > scrollHeight - clientHeight * 2
->
scrollTop > scrollHeight - clientHeight - clientHeight *  1 (aka 1 = scrollOffset multiplier value in useLoadMore)
```

useLoadMore code:
```
let shouldLoadMore = ref.current.scrollHeight - ref.current.scrollTop - ref.current.clientHeight < ref.current.clientHeight * scrollOffset;

becomes:
 ref.current.clientHeight * scrollOffset > ref.current.scrollHeight - ref.current.scrollTop - ref.current.clientHeight
->
clientHeight * scrollOffset > scrollHeight - scrollTop - clientHight
->
scrollTop > scrollHeight - clientHeight - clientHeight * scrollOffset (aka scrollOffset = 1 which is the same scrollOffset multiplier as in useVirtualizer)
```

packages/react-aria-components/src/Table.tsx
```diff
@@ -712,7 +711,7 @@ export const Column = /*#__PURE__*/ createLeafComponent('column', (props: Column
 
   let style = renderProps.style;
   if (layoutState) {
-    style = {...style, width: layoutState.getColumnWidth(column.key)};
+    style = {...style, width: isVirtualized ? undefined : layoutState.getColumnWidth(column.key)};
```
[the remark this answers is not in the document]
LFDanLu: The thing is that the virtualizer wrapper applied around the column element already has the same width applied to it. In S2 I then applied the padding to the column to position the arbitrary column text properly which wouldn't work if the column also had the virtualizer calculated width applied directly on it.
image
  
I could revert this and add another wrapper around the text that adds this padding instead but this just emulated v3 table

packages/react-aria-components/src/Table.tsx
LFDanLu: Ah sorry, if you try https://reactspectrum.blob.core.windows.net/reactspectrum/0085279a1631f18a2e48e9c2d2baf108f17cf875/storybook/index.html?path=/story/react-aria-components--virtualized-table-with-empty-state-story&providerSwitcher-express=false and expand the columns + scroll it horizontally you'll notice that the loading text isn't sticky. 

This was an attempt to replicate v3 table's sticky loading indication here (which I just noticed seems to have broken on main, not sure why there is extra vertical scroll space): https://reactspectrum.blob.core.windows.net/reactspectrum/0085279a1631f18a2e48e9c2d2baf108f17cf875/storybook/index.html?path=/story/tableview--is-loading&providerSwitcher-express=false

packages/react-aria-components/src/Table.tsx
```diff
@@ -1288,3 +1303,47 @@ function RootDropIndicator() {
...
 }
+
+export interface TableLoadingIndicatorProps extends StyleProps {
+  children?: ReactNode
+}
+
+export const UNSTABLE_TableLoadingIndicator = createLeafComponent('loader', function TableLoadingIndicator<T extends object>(props: TableLoadingIndicatorProps, ref: ForwardedRef<HTMLTableRowElement>, item: Node<T>) {
+  let state = useContext(TableStateContext)!;
+  let {isVirtualized} = useContext(CollectionRendererContext);
+  let numColumns = state.collection.columns.length;
```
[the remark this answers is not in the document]
LFDanLu: ah lol, I had copy pastaed the if statement from a section higher up, I'll fix this

packages/react-aria-components/src/Table.tsx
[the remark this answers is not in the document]
devongovett: oh I forgot about the wrapper, ok

packages/react-aria-components/src/Table.tsx
devongovett: ah right. hmm, well I don't think we'd want _all_ of the elements to have overflow: visible. I guess only the body. I didn't realize position: sticky required overflow visible...

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -506,11 +796,163 @@ export function VirtualizedTableWithEmptyState() {
...
+  let scrollRef = useRef<HTMLDivElement>(null);
+  let memoedLoadMoreProps = useMemo(() => ({
+    isLoading: isLoading,
+    onLoadMore: list.loadMore,
+    items: list.items
+  }), [isLoading, list.loadMore, list.items]);
+  useLoadMore(memoedLoadMoreProps, scrollRef);
+
+  return (
+    <ResizableTableContainer ref={scrollRef} style={{height: 150, width: 400, overflow: 'auto'}}>
```
snowystinger, line 927: columns of this story don't appear to be resizable?

packages/react-aria-components/stories/Tree.stories.tsx
```diff
@@ -295,25 +316,161 @@ WithLinks.story = {
...
+    </UNSTABLE_Tree>
+  );
+}
+
+export const LoadingStoryDepOnCollectionStory = {
+  render: LoadingStoryDepOnCollection,
+  args: {
+    isLoading: false
+  },
+  name: 'Loading, static root loader and dynamic rows'
```
snowystinger, line 367: maybe add a description to this story and the ones that follow it? I'm not sure what I'm supposed to see or do with it

akyllus (approved):

packages/react-aria-components/test/Table.test.js
```diff
@@ -1257,6 +1266,471 @@ describe('Table', () => {
...
+                  return <Cell>filler</Cell>;
+                }}
+              </Row>
+            </TableBody>
+          </Table>
+        );
+      }
+
+      expect(() => render(<StaticRowDynamicCell />)).toThrow();
+      expect(consoleWarnSpy).toHaveBeenCalledWith('No id detected for the Row element. The Row element requires a id to be provided to it when the cells are rendered dynamically.');
```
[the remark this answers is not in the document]
LFDanLu: updated the test name

packages/react-aria-components/stories/Table.stories.tsx
LFDanLu: they are not, more of a test that the scroll ref could be attached to this wrapper and still works with load more.

ktabors (approved):

packages/react-aria-components/src/Table.tsx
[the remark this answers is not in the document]
devongovett: we may want to revisit this at some point. this does sort of assume the DOM structure of our virtualizer, but might not always be true if someone built a different virtualized collection renderer. Would the padding work if you set `box-sizing: border-box`?

devongovett (approved):

packages/react-aria-components/src/Table.tsx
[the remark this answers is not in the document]
LFDanLu: yep that works as well, I'll add a PR to revert this change

PR #6518, opened 2024-06-08
Add Virtualizer to React Aria Components

### LINKED ISSUE #5067: RAC: Multi select with Shift is not working in Virtualized GridList

When testing GridList with virtualization library like `@tanstack/react-virtual` kinda looks like it works 🎉  

One issue noticed, when using selectionBehavior `replace` and selecting multi rows with Shift. If we select item from current render items then it works, issue is when selecting one and scroll down a bit to new rendered items, then it breaks, item is focused but no selection.

https://github.com/adobe/react-spectrum/assets/82964/6472cbf9-6cc3-4b11-bd00-c19339c19d5f

Also when scrolling focus is changing, is there some autofocus enabled?

### 🤔 Expected Behavior?

Items should be selected.

### 😯 Current Behavior

Items are not selected and row is focused.

### 💁 Possible Solution

_No response_

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

https://codesandbox.io/s/friendly-tamas-3zpl6q?file=/src/App.js

### Version

1.0.0-beta.0

Chrome

### If other, please specify.

_No response_

MacOS

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

devongovett:
Closes #5067

This PR implements a `Virtualizer` in React Aria Components, using the collection renderer system added in #5912.

There are 3 parts which you can review commit by commit:

1. Add a new `LayoutDelegate` interface, which enables us to share a common KeyboardDelegate implementation between virtualized and non-virtualized collections. Previously, ListLayout implemented the KeyboardDelegate interface itself, which was mostly duplicated with the DOM-based ListKeyboardDelegate. It also resulted in us needing to sync data into the layout such as `disabledKeys`. Now we can use ListKeyboardDelegate in both cases. It accepts a `LayoutDelegate`, which provides the rectangles of item elements and the collection itself. There is a default `DOMLayoutDelegate` implementation, and in addition, all `Layout` classes automatically implement this interface as well.
2. Unify the Spectrum and Aria column resizing implementations. We now use the hook in Spectrum and pass the resulting `columnWidths` into the virtualizer layout via `layoutOptions`, rather than storing state in the layout. I've also managed to simplify the algorithm for resizing a column. This enables RAC to implement resizing the same way as RSP.
3. Add `Virtualizer` to React Aria Components. This is implemented as a CollectionRenderer, which has been refactored a bit. It now provides two components: `CollectionRoot` renders the root items of a collection, and `CollectionBranch` renders the children of a branch item. Virtualizer attaches scroll events to the provided `scrollRef` at the root, determines which views need to be displayed, and renders those.

The end result is that you can wrap any collection component in a `Virtualizer`, pass in a `layout` as a prop, and the component will have virtualized scrolling. The rest of the API is the same as it is today.

```jsx
let layout = useMemo(() => {
  return new ListLayout({
    rowHeight: 25
  });
}, []);

return (
  <Virtualizer layout={layout}>
    <GridList style={{width: 200, height: 400}} items={items}>
      {item => <GridListItem>{item.name}</GridListItem>}
    </GridList>
  </Virtualizer>
);
```

This means the layouts are also customizable. For example, you could pass a `GridLayout` here and the items would be arranged in a grid. `Virtualizer` delegates to the layout for keyboard navigation and drag and drop automatically, and persists the focused key in the DOM. Individual components may also add additional ARIA attributes when virtualized, e.g. `aria-setsize` and `aria-posinset`. Otherwise, they are unaware of virtualization, which means you only pay the bundle size cost when needed, and you could even implement a custom CollectionRenderer to use a different virtualization library if you wanted.


packages/@react-aria/table/src/useTable.ts
```diff
@@ -15,18 +15,17 @@ import {GridAria, GridProps, useGrid} from '@react-aria/grid';
...
 import {TableKeyboardDelegate} from './TableKeyboardDelegate';
 import {tableNestedRows} from '@react-stately/flags';
 import {TableState, TreeGridState} from '@react-stately/table';
 import {useCollator, useLocale, useLocalizedStringFormatter} from '@react-aria/i18n';
 
-export interface AriaTableProps<T> extends GridProps {
+export interface AriaTableProps extends GridProps {
   /** The layout object for the table. Computes what content is visible and how to position and style them. */
-  layout?: Layout<Node<T>>
+  layoutDelegate?: LayoutDelegate
```
devongovett, line 28: Technically a breaking change but I'm hoping we can get away with it since virtualizer wasn't previously public?

packages/@react-stately/layout/src/ListLayout.ts
```diff
@@ -253,7 +249,7 @@ export class ListLayout<T> extends Layout<Node<T>, ListLayoutProps> implements K
       y = loader.rect.maxY;
     }
 
-    if (nodes.length === 0) {
+    if (nodes.length === 0 && this.enableEmptyState) {
```
devongovett, line 252: In RAC, the loading and empty states are handled in the component rather than in the layout. Ideally this would also the the case in RSP, but I haven't found a good way to implement that yet.

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -10,114 +10,72 @@
...
-import {LayoutNode, ListLayout, ListLayoutOptions} from './ListLayout';
+import {LayoutNode, ListLayout, ListLayoutOptions, ListLayoutProps} from './ListLayout';
+import {TableCollection} from '@react-types/table';
 import {TableColumnLayout} from '@react-stately/table';
 
-type TableLayoutOptions<T> = ListLayoutOptions<T> & {
-  columnLayout: TableColumnLayout<T>,
-  initialCollection: TableCollection<T>
+export interface TableLayoutOptions<T> extends ListLayoutOptions<T> {
+  scrollContainer?: 'table' | 'body'
```
devongovett, line 22: This is to handle a difference between RSP and RAC. In RSP, only the TableBody scrolls, whereas (so far) in RAC, the whole table scrolls and the column headers have `position: sticky`. This affects the positioning of the body rows, which are relative to the body in RSP, and relative to the whole table in RAC.

packages/@react-stately/layout/src/ListLayout.ts
```diff
@@ -278,15 +274,15 @@ export class ListLayout<T> extends Layout<Node<T>, ListLayoutProps> implements K
     );
   }
 
-  buildChild(node: Node<T>, x: number, y: number): LayoutNode {
+  protected buildChild(node: Node<T>, x: number, y: number, parentKey: Key | null): LayoutNode {
```
devongovett, line 277: This now takes a parentKey as a parameter rather than using the node's parentKey in order to properly handled TreeView's flattened collection. We want the parent LayoutInfo here, not the parent Node.

packages/@react-stately/table/src/TableCollection.ts
```diff
@@ -317,6 +317,14 @@ export class TableCollection<T> extends GridCollection<T> implements ITableColle
     return this.getItem(keys[idx]);
   }
 
+  getChildren(key: Key): Iterable<GridNode<T>> {
```
devongovett, line 320: Interoperability with RAC TableCollection.

packages/@react-stately/table/src/TableColumnLayout.ts
```diff
@@ -80,78 +78,25 @@ export class TableColumnLayout<T> {
...
+      } else if (freeze) {
+        // freeze columns to the left to their previous pixel value
+        newWidths.set(column.key, prevColumnWidths.get(column.key));
+      } else {
+        newWidths.set(column.key, column.props.width ?? uncontrolledWidths.get(column.key));
       }
-      newWidths.set(key, width);
     });
+
     return newWidths;
```
devongovett, line 100: Perhaps I am missing something here, but it seemed like we could simplify this algorithm a lot. We just need to lock in all of the columns to the left of the resizing one to their previous pixel values, and set the width of the resizing column. I don't think we need to run the sizing algorithm here at all, we can do that after the next render when the new widths are set. All the tests passed and it seemed to behave ok.

packages/react-aria-components/src/Collection.tsx
```diff
@@ -765,6 +767,14 @@ export function useCollectionDocument<T extends object, C extends BaseCollection
...
   }, [document]);
   let collection = useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot);
+  useLayoutEffect(() => {
+    document.isMounted = true;
+    return () => {
+      // Mark unmounted so we can skip all of the collection updates caused by 
+      // React calling removeChild on every item in the collection.
+      document.isMounted = false;
+    };
+  }, [document]);
```
devongovett, line 777: Improves performance a lot when unmounting a collection component with many items (e.g. 10,000 rows). Noticed when switching stories.

packages/react-aria-components/stories/Table.stories.tsx
```diff
@@ -392,3 +392,75 @@ const MyCheckbox = ({children, ...props}: CheckboxProps) => {
...
+  let layout = useMemo(() => {
+    return new TableLayout({
+      rowHeight: 25,
+      headingHeight: 25
+    });
+  }, []);
+
+  return (
+    <Virtualizer layout={layout}>
+      <Table aria-label="virtualized table" style={{height: 400, width: 400, overflow: 'auto'}}>
```
majornista, line 411:
```suggestion
      <Table aria-label="virtualized table" style={{height: 400, width: 400, overflow: 'auto', scrollPaddingTop: 25}}>
```

majornista (dismissed):

packages/@react-stately/table/src/TableColumnLayout.ts
snowystinger: at one point i think we needed to run the layout here so that the internal values for the columns was up to date
I think the changes you made that no longer needed since it'll actually update in render

packages/@react-stately/table/src/useTableColumnResizeState.ts
```diff
@@ -133,9 +133,11 @@ export function useTableColumnResizeState<T>(props: TableColumnResizeStateProps<
       columnLayout.getColumnMinWidth(key),
     getColumnMaxWidth: (key: Key) =>
       columnLayout.getColumnMaxWidth(key),
-    tableState: state
+    tableState: state,
+    columnWidths
   }), [
     columnLayout,
+    columnWidths,
```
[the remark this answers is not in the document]
devongovett: Just adding them so seems ok? What did you mean?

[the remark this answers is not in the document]
snowystinger: I was worried that we were now depending on it, but I guess all packages will get a version bump and the range will keep them correct.

reidbarber (dismissed):

packages/react-aria-components/src/TableLayout.ts
```diff
@@ -0,0 +1,28 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {TableLayout as BaseTableLayout, TableLayoutProps} from '@react-stately/layout';
+import {LayoutOptionsDelegate} from './Virtualizer';
+import {TableColumnResizeStateContext} from './Table';
+import {useContext, useMemo} from 'react';
+
+export class TableLayout<T> extends BaseTableLayout<T> implements LayoutOptionsDelegate<TableLayoutProps> {
```
LFDanLu, line 18: Since this is using the BaseTableLayout, I found that it doesn't handle the "loader" type node that the load more spinner collection element due to https://github.com/adobe/react-spectrum/blob/535e1596d2f00c3e30cab2a64106435fdb696f01/packages/%40react-stately/layout/src/TableLayout.ts#L290-L302 not understanding what that node type is. Bit torn on whether or not the TableLayout should be updated to handle "loader" type nodes or if the leaf component for the loader should be changed to "item" (this would pose some difficulties in differentiating the loaders from normal rows but can probably be handled collection side some how)

Or would the expectation be that the TableLayout be replaced/extended to handle new node types? Thinking from an angle where perhaps a section header/loader would need its calculated row height differentiated from a typical row

devongovett: If they were all items, how would we differentiate them elsewhere? Do we need to?

Otherwise I think it's fine to add more item types to the layout, would just work the same as item. I don't expect we'd be adding them that often.

packages/react-aria-components/src/Virtualizer.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+export function Virtualizer(props: VirtualizerProps) {
+  let {children, layout} = props;
+  let renderer: CollectionRenderer = useMemo(() => ({
+    isVirtualized: true,
+    layoutDelegate: layout,
+    dropTargetDelegate: layout.getDropTargetFromPoint ? layout as DropTargetDelegate : undefined,
+    CollectionRoot({collection, focusedKey, scrollRef}) {
+      let layoutOptions = layout.useLayoutOptions?.();
+      let state = useVirtualizerState({
```
LFDanLu, line 38: I'm still digging to see where best to fix this, but I noticed that a virtualized table with renderEmptyState won't actually display the rendered empty state element because the wrapping div has a height of 0. It seems like that height of 0 is because that visible view (aka the body) doesn't have any items within it and thus it gets a layoutInfo calculated for it with a height of 0px.

Passing `enableEmptyState: true` to the TableLayout alleviates this somewhat (actually allocates a height), but the row element positioning still isn't quite right

devongovett: hmm yeah I didn't want the table to be responsible for rendering the empty state. I guess I can force the table body to always take up the entire height of the table so the empty state isn't clipped.

LFDanLu: that makes sense, that way the user can still control the styling. I was initially thinking on making `enableEmptyState` be true and have `CollectionBranch` take children but that was all kinds of janky

packages/react-aria-components/src/Table.tsx
```diff
@@ -209,31 +213,50 @@ export interface ResizableTableContainerProps extends DOMProps, ScrollableProps<
 }
 
 function ResizableTableContainer(props: ResizableTableContainerProps, ref: ForwardedRef<HTMLDivElement>) {
-  let objectRef = useObjectRef(ref);
+  let containerRef = useObjectRef(ref);
+  let tableRef = useRef<HTMLTableElement>(null);
```
snowystinger, line 217:
```suggestion
  let tableRef = useRef<HTMLTableElement | null>(null);
```

snowystinger:
https://reactspectrum.blob.core.windows.net/reactspectrum/a7fb8036a032800b1592b30e8b52885beeb374b2/storybook/index.html?path=/story/react-aria-components--virtualized-select&providerSwitcher-express=false&providerSwitcher-theme=light isn't opening at all for me

snowystinger:
Heh, I can see this bug https://github.com/adobe/react-spectrum/issues/6549 in the Empty Table Virtualized example
Don't worry, pre-existing, you didn't break it. Hopefully one of us can look into it in the next week or two.

devongovett:
@snowystinger looks like the select was broken by https://github.com/adobe/react-spectrum/pull/6497...

packages/react-aria-components/src/TableLayout.ts
LFDanLu: For now I've made the loader node trigger `this.buildRow` as well. I do still wonder if there should be additional customization allowed for setting the height of the loader row independent of the row height but that can be handled in the future if requested.

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -305,27 +260,31 @@ export class TableLayout<T> extends ListLayout<T> {
       // Add some margin around the loader to ensure that scrollbars don't flicker in and out.
       let rect = new Rect(40,  Math.max(y, 40), (width || this.virtualizer.visibleRect.width) - 80, children.length === 0 ? this.virtualizer.visibleRect.height - 80 : 60);
       let loader = new LayoutInfo('loader', 'loader', rect);
```
LFDanLu, line 262: just wanted to point out that this `isLoading` block doesn't trigger for RAC tables since RAC TableBody doesn't take in `loadingState` (which is good and expected) but if the user coincidentally passes `loadingState='loading'` then this does trigger and messes up the layout by creating the loader layout info that we only actually use in RSP TableView.

I don't think this is really something to worry about probably but just wanted to call it out since this happens since we access the props set on the collection nodes in several places

packages/react-aria-components/src/TableLayout.ts
LFDanLu, line 18: Another thing I just realized is that this exposes options like `loaderHeight` that actually don't do anything due to how TableLayout implements its own `buildCollection`. I've gotten rid of it from stately's TableLayout but I'll double check if there are any other options that need to be removed/aren't applicable to RAC TableLayout

packages/@react-stately/layout/src/TableLayout.ts
```diff
@@ -378,7 +337,7 @@ export class TableLayout<T> extends ListLayout<T> {
 
     this.setChildHeights(children, height);
 
-    rect.width = this.layoutInfos.get('header').rect.width;
+    rect.width = this.layoutInfos.get(this.collection.head?.key ?? 'header').rect.width;
     rect.height = height + 1; // +1 for bottom border
```
LFDanLu, line 341: I noticed that our layouts have some extra calculations built in like the above that adds 1 to the row height for the border. Will this be a bit strange for a RAC user who provides something like `rowHeight: 25` to their TableLayout but then gets an actual row height of 26px?

LFDanLu (approved):
Verified that drag and drop works now and that behavior for RAC/RSP virtualized components work as expected. Only a couple of small questions but not blockers IMO

packages/@react-stately/layout/src/TableLayout.ts
devongovett: hmm yeah. I am wondering if we split out a spectrum-specific layout from the generic one...

packages/@react-spectrum/card/src/BaseLayout.tsx
```diff
@@ -55,8 +55,8 @@ export class BaseLayout<T> extends Layout<Node<T>, CardViewLayoutOptions> implem
 
   validate(invalidationContext: InvalidationContext<CardViewLayoutOptions>) {
     this.collection = this.virtualizer.collection as GridCollection<T>;
-    this.isLoading = invalidationContext.layoutOptions.isLoading;
-    this.direction = invalidationContext.layoutOptions.direction;
+    this.isLoading = invalidationContext.layoutOptions?.isLoading || false;
+    this.direction = invalidationContext.layoutOptions?.direction || 'rtl';
```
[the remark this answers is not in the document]
devongovett: typo, good catch

snowystinger (approved):
At this point, happy to get in and we can follow up on the other ideas we had. Would be good if we can split any of that work amongst us

rostero1:
I'm testing out this branch and if I have dynamic heights for each element and scroll to the bottom it does not show all the elements and there is a large white gap at the bottom resulting in a blank viewport.  I apologies if dynamic height isn't supported.

This is for the VirtualizedListBox story, where I moved sections out of the component.

```
function generateRandomString(minLength: number, maxLength: number): string {
  const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
  const length = Math.floor(Math.random() * (maxLength - minLength + 1)) + minLength;

  let result = '';
  for (let i = 0; i < length; i++) {
    result += characters.charAt(Math.floor(Math.random() * characters.length));
  }

  return result;
}

let sections: {id: string, name: string, children: {id: string, name: string}[]}[] = [];
for (let s = 0; s < 10; s++) {
  let items: {id: string, name: string}[] = [];
  for (let i = 0; i < 100; i++) {
    const l = (s * 5) + i + 10;
    items.push({id: `item_${s}_${i}`, name: `Item ${i} ${generateRandomString(l, l)}`});
  }
  sections.push({id: `section_${s}`, name: `Section ${s}`, children: items});
}
```

Also, will these virtualizer handle the case where an element could resize at any point?  Such, as a row getting more data to render?

devongovett:
It should work if you specify an `estimatedRowHeight` rather than `rowHeight` on the ListLayout.

rostero1:
> It should work if you specify an `estimatedRowHeight` rather than `rowHeight` on the ListLayout.

I forgot to mention that I also changed `rowHeight` to `estimatedRowHeight`

devongovett:
Hmm your example is working for me locally... Any other steps to reproduce what you're seeing?

rostero1:
Thanks for checking.  The only other difference is I have `style={{ wordBreak: 'break-word'}}` added to `<MyListBoxItem `.

Sorry for not creating a codebandbox for this.

LFDanLu (dismissed):
Happy for this to go in as is, we can discuss breaking out the RSP specific layout logic separately

packages/@react-stately/layout/src/TableLayout.ts
LFDanLu: IMO it would ideal to split them. I'll add to the canvas to discuss this further

devongovett:
@rostero1 thanks, I see what you mean. I will look at it in a followup PR.

snowystinger (approved):

PR #6526, opened 2024-06-11
Selection checkbox should always toggle regardless of selectionBehavior

devongovett:
Opening for feedback on behavior. When rendering a GridList or Table with `selectionBehavior="replace"` that also has checkboxes, clicking the checkboxes currently replaces the selection with that row (the same as clicking anywhere on the row). This PR changes the behavior so the checkbox always behaves like a checkbox, and toggles the selection, while clicking on the row still replaces the selection. This is kind of a hybrid between replace and toggle, and actually matches how Windows Explorer behaves as well.

[Storybook link for testing](https://reactspectrum.blob.core.windows.net/reactspectrum/c477443b87aa3e21bb63925a2a5a142fec81753f/storybook/index.html?path=/story/react-aria-components--grid-list-example&args=selectionMode:multiple;selectionBehavior:replace&providerSwitcher-express=false)

### DIFF
--- a/packages/@react-aria/grid/src/useGridSelectionCheckbox.ts
+++ b/packages/@react-aria/grid/src/useGridSelectionCheckbox.ts
@@ -31,7 +31,8 @@ export function useGridSelectionCheckbox<T, C extends GridCollection<T>>(props: 
   let isDisabled = !state.selectionManager.canSelectItem(key);
   let isSelected = state.selectionManager.isSelected(key);
 
-  let onChange = () => manager.select(key);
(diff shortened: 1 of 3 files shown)


snowystinger (commented):
I think this is a good idea. It feels a bit more intuitive since I'm expressly choosing to click on the checkbox. And it doesn't take two hands for me to add to my selection anymore (holding CMD + click)

snowystinger (approved):

LFDanLu (approved):
LGTM, will just leave https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=67750260 here for additional context for potential followup

PR #6535, opened 2024-06-12
Always render ListBox section headers in RSP for dividers

devongovett:
Closes #6490

Alternative approach to #6490. Adds an option to `ListLayout` instead of modifying the collection.


snowystinger (commented):
I don't recall how to test this (or why we don't need to set this for other `new ListLayout`s, can you add directions to the description?

LFDanLu:
@snowystinger you can test the "listbox w/sections and no title" story for this, just need to confirm that the divider shows up. For now apply this in the ListBoxBase's ListLayout should be good enough since I believe ListBox is the only collection component with Sections and components like Picker/ComboBox use it directly.

LFDanLu (approved):

snowystinger (approved):

Issue #6183, opened 2024-04-11
"AM/PM" segment for time fields with 12-hour clock format does not work on mobile


paulodetarsofm:
### Provide a general summary of the issue here

In the time field component, the "AM/PM" segment does not have an `inputmode` defined, which in turn displays the standard alphanumeric keyboard when focused:

![image](https://github.com/adobe/react-spectrum/assets/4804498/25a106e3-c805-4bf1-95cb-4a526d8e6243)

But no matter what I type, the value does not change. Furthermore, there is a small bug that when pressing "p", the value of the segment was not changed as expected, but rather concatenated with what it was previously, becoming "AMPM":

![image](https://github.com/adobe/react-spectrum/assets/4804498/f944495e-5ab4-492d-a29e-1b0d02c019ab)

With English, I somehow managed to insert some value as shown in the image above, but in other languages, this did not happen. I was able to test using the Korean keyboard on my Android, even though I don't understand what the symbols displayed on the keyboard mean, I pressed all of them and nothing changed anyway:

![image](https://github.com/adobe/react-spectrum/assets/4804498/e605a8f6-81a2-4f57-9b8b-108d42b4745f)

When I was testing Korean, I changed the segment's `inputmode` manually using Chrome DevTools, which is why the screenshot above shows the numeric keyboard, but the keyboard displayed is normally alphanumeric.

A video demonstrating how the "AM/PM" segment was displayed as "AMPM":

https://github.com/adobe/react-spectrum/assets/4804498/1d40e24a-6844-4a9b-abe4-ce97beeb3867

### 🤔 Expected Behavior?

We should be able to change the value of "AM/PM" somehow and the value displayed in the UI should match the new selected value.

### 😯 Current Behavior

In English, the segment was rendered as "AMPM" when trying to change the segment value, and in Korean, the value is not changed.

### 💁 Possible Solution

As the segment only has two possible values, I don't know if displaying a keyboard for it is the best solution, as we practically can't type anything. However, due to the size of the field/segment and because it is a mobile device, I don't know if we would have enough space in the UI to offer an alternative, such as a dropdown, for example.

I was thinking: what if we used something like a hidden checkbox that would change the value to "AM" or "PM"? Behind the scenes, this is what we would store there, 0 or 12, which would be used to calculate the correct value of the time if it is AM or PM - if we have "1" set as the hour and we have "AM", we basically do 1 + 0 = 1 (01:00), or if it is "PM", 1 + 12 = 13 (13:00).

[...]

PR #6562, opened 2024-06-18
Removing date segment placeholder default width

ktabors:
Closes 

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/datepicker/src/DatePickerSegment.tsx
+++ b/packages/@react-spectrum/datepicker/src/DatePickerSegment.tsx
@@ -60,13 +60,9 @@ function EditableSegment({segment, state}: DatePickerSegmentProps) {
         'is-placeholder': segment.isPlaceholder,
         'is-read-only': !segment.isEditable
       })}
-      style={{
-        ...segmentProps.style,
-        minWidth: segment.maxValue != null ? String(segment.maxValue).length + 'ch' : null
-      }}
+      style={segmentProps.style}
       data-testid={segment.type}>
-      <span aria-hidden="true" className={classNames(styles, 'react-spectrum-DatePicker-placeholder')}>{segment.placeholder}</span>
-      {segment.isPlaceholder ? '' : segment.text}
+      {segment.isPlaceholder ? segment.placeholder : segment.text}kyle
     </div>
   );
 }
--- a/packages/@react-spectrum/datepicker/src/styles.css
+++ b/packages/@react-spectrum/datepicker/src/styles.css
@@ -111,22 +111,13 @@
   }
 }
 
-.react-spectrum-DatePicker-placeholder {
-  display: block;
-  width: 100%;
-  text-align: center;
-  font-style: italic;
-  visibility: hidden;
-  height: 0;
-  pointer-events: none;
-}
-
 .react-spectrum-DatePicker-cell.is-placeholder {
   color: var(--spectrum-gray-600);
 
   .react-spectrum-DatePicker-placeholder {
-    visibility: visible;
-    height: auto;
+    text-align: center;
+    font-style: italic;
+    pointer-events: none;
   }
 }
 


packages/@react-spectrum/datepicker/src/DatePickerSegment.tsx
```diff
@@ -60,13 +60,9 @@ function EditableSegment({segment, state}: DatePickerSegmentProps) {
...
       })}
-      style={{
-        ...segmentProps.style,
-        minWidth: segment.maxValue != null ? String(segment.maxValue).length + 'ch' : null
-      }}
+      style={segmentProps.style}
       data-testid={segment.type}>
-      <span aria-hidden="true" className={classNames(styles, 'react-spectrum-DatePicker-placeholder')}>{segment.placeholder}</span>
-      {segment.isPlaceholder ? '' : segment.text}
+      {segment.isPlaceholder ? <span aria-hidden="true" className={classNames(styles, 'react-spectrum-DatePicker-placeholder')}>{segment.placeholder}</span> : segment.text}
```
devongovett, line 65: I'm not seeing this span in the DOM at all, not sure why. It looks like the placeholder text (e.g. mm) is directly inside the div. 🤔

packages/@react-spectrum/datepicker/src/styles.css
```diff
@@ -111,22 +111,13 @@
...
-}
-
 .react-spectrum-DatePicker-cell.is-placeholder {
   color: var(--spectrum-gray-600);
 
   .react-spectrum-DatePicker-placeholder {
-    visibility: visible;
-    height: auto;
+    text-align: center;
+    font-style: italic;
```
devongovett, line 119: This element isn't working at all right now so I'm not seeing the italics. But I guess we should get rid of them anyway?

devongovett:
One other problem appears to be that when you type into the field, the entire field gets wider or shorter...

packages/@react-spectrum/datepicker/src/DatePickerSegment.tsx
ktabors: Which build are you testing? When I look at bc8e7a2a8d6b89da4f821cefe50488b494433f35 I see the following.
image

devongovett: hmm weird, it is working now... not sure what happened before.

Should we get rid of the italics though while we're at it? @dannify seemed to think so

packages/@react-spectrum/datepicker/src/styles.css
ktabors: Ditto the above for not seeing italics.
We are purging italics, but do we need that visual indicator for placeholder?

packages/@react-stately/datepicker/src/useDatePickerState.ts
```diff
@@ -62,7 +62,9 @@ export interface DatePickerState extends OverlayTriggerState, FormValidationStat
   /** Whether the date picker is invalid, based on the `isInvalid`, `minValue`, and `maxValue` props. */
   isInvalid: boolean,
   /** Formats the selected value using the given options. */
-  formatValue(locale: string, fieldOptions: FieldOptions): string
+  formatValue(locale: string, fieldOptions: FieldOptions): string,
+  /** Gets a formatter based on state's props. */
+  getDateFormatter(ocale: string, formatOptions: FormatterOptions): DateFormatter
```
snowystinger, line 67:
```suggestion
  getDateFormatter(locale: string, formatOptions: FormatterOptions): DateFormatter
```

packages/@react-spectrum/datepicker/src/utils.ts
```diff
@@ -77,3 +77,16 @@ export function useFocusManagerRef(ref: FocusableRef<HTMLElement>) {
...
+export function useDateCharacterWidth(state) {
+  let locale = useLocale()?.locale;
+  let currentDate = new Date();
+  let formatedDate = state.getDateFormatter(locale, {shouldForceLeadingZeros: true}).format(currentDate, locale);
+  let totalCharacters =  formatedDate.length;
+
+  // The max of two is for times with only hours.
+  // As the length of a date grows we need to proportionally increase the width.
+  // We use the characts with 'ch' and months and time dashes are wider and this
+  // accomates for that.
```
snowystinger, line 90:
```suggestion
  // accommodates for that.
```

packages/@react-spectrum/datepicker/src/utils.ts
```diff
@@ -77,3 +77,16 @@ export function useFocusManagerRef(ref: FocusableRef<HTMLElement>) {
...
+
+export function useDateCharacterWidth(state) {
+  let locale = useLocale()?.locale;
+  let currentDate = new Date();
+  let formatedDate = state.getDateFormatter(locale, {shouldForceLeadingZeros: true}).format(currentDate, locale);
+  let totalCharacters =  formatedDate.length;
+
+  // The max of two is for times with only hours.
+  // As the length of a date grows we need to proportionally increase the width.
+  // We use the characts with 'ch' and months and time dashes are wider and this
```
snowystinger, line 89:
```suggestion
  // We use the characters width 'ch' and months and time dashes are wider and this
```
?? this sentence didn't make much sense

I assume this is trying to explain why we do a divide by 5? i don't really follow why that is. Maybe some examples would help?

packages/@react-spectrum/datepicker/src/TimeField.tsx
```diff
@@ -52,6 +52,8 @@ function TimeField<T extends TimeValue>(props: SpectrumTimeFieldProps<T>, ref: F
 
   let validationState = state.validationState || (isInvalid ? 'invalid' : null);
 
+  let characterCount = useDateCharacterWidth(state) + 'ch';
```
snowystinger, line 55: this isn't actually a character count though is it? nor a character width?

packages/@react-spectrum/datepicker/src/utils.ts
```diff
@@ -77,3 +77,16 @@ export function useFocusManagerRef(ref: FocusableRef<HTMLElement>) {
   }));
   return domRef;
 }
+
+export function useDateCharacterWidth(state) {
```
snowystinger, line 81: is it actually a character width? or is it the width of the characters + some fudging number? is this actually `useDateSegmentWidth`?

packages/@react-spectrum/datepicker/src/DateRangePicker.tsx
```diff
@@ -112,6 +112,9 @@ function DateRangePicker<T extends DateValue>(props: SpectrumDateRangePickerProp
   let visibleMonths = useVisibleMonths(maxVisibleMonths);
   let validationState = state.validationState || (isInvalid ? 'invalid' : null);
 
+  // Multiplying by two for the two dates, adding one character for the dash, and then the dash left and right padding
```
snowystinger, line 115:
```suggestion
  // Multiplying by two for the two dates, adding one character for the dash, and then the padding around the dash
```

packages/@react-spectrum/datepicker/src/utils.ts
LFDanLu: I don't quite follow the calculation either, I think I've formatted as wide of a date as possible but the min width still gives some extra padding here
image

The character count is 24 and the calculated character width with the additional accommodation factor sets the width to 28ch which gives it that extra purple padding. Using 24ch makes it pretty flush with the inputted date
image
which seems better?

LFDanLu (commented):
image
Feels like a lot of whitespace on the end but I don't think there is much we can do about that without making the field width shift which feels worse haha.

I also found that Android now crashes when trying to change AM/PM in the time field. Steps to reproduce:
1. Go to https://reactspectrum.blob.core.windows.net/reactspectrum/1a948f002e8a7279c153110cbe0b9d4f4e8f5820/storybook/index.html?path=/story/date-and-time-timefield--default&providerSwitcher-express=false
2. Fill out the field with any time
3. Focus the AM/PM field and hit backspace to clear it. Focus should move on to the previous segment
4. tap the AM/PM segment and try to change it to PM by pressing P. Note it doesn't change. 
5. Press P again and note it crashes

packages/@react-spectrum/datepicker/src/utils.ts
devongovett: I think the problem is when you change it to a placeholder. "mm" is much wider than two numbers, so it would push beyond the min-width causing the field to grow.

packages/@react-spectrum/datepicker/src/TimeField.tsx
ktabors: Both... It's the character count being turned into a width. And on Picker it's a `calc()` which makes it more width.

packages/@react-spectrum/datepicker/src/utils.ts
LFDanLu: still seems to have a bit of extra space by ~1 or 2ch:
image
I'm fine with it being a bit off/not exactly precise, just wanna understand the underlying calculation here to see if there is any way we can make it more precise, the white space after the field is already pretty large compared to before (of which people already complained about haha)

packages/@react-spectrum/datepicker/src/utils.ts
ktabors: It's more like `useApproximateFormattedDateWidth`, but that is too long. I'll update it along with `characterCount` to something better.

packages/@react-spectrum/datepicker/src/utils.ts
ktabors: TLDR: This works remarkably well as a guess/approximation and is because of the width of month and dash characters in placeholder values.

It does widen a few cases. I tried a bunch of things with ch and mm units, including a granularity-based switch statement. This ratio works well, and it is remarkably close. Since we're guessing and using today's date, not the placeholder or the actual date value, I'm not sure if getting more precise would be better.

When Devon and I were looking at it and talking about it yesterday, it ended up being that we need approximately two characters for any data value less than 10 characters. For example, we have a problem with TimeField width shifting for granularity hour if we don't give it four characters of space. Like Devon stated above, for dates it's the months and for time's it's the dashes that create the need for extra characters. As we grow past ten characters, we're adding time which is dashes so we need more characters. 

Also, this seems to work with locales, but we don't have a collection of rigorous tests with all the possible font, language, and calendar combinations. Getting more precise for English might cause problems elsewhere, while being more generous has worked so for with the locale testing that has been done.

Yes, it looks "bad" when you have a DateRange with granularity seconds with two values, where it only fills about half the field, but when you switch it to placeholder it fills it back up. We're shifting from all the white space being in the segments to the white space being after the date values. The field widths didn't change much. Looking at chromatic most of the field widths didn't grow in width and it was only by a little for those that did. It was about equivalent width grown to the purple space in that image Daniel shared.

ktabors:
> I also found that Android now crashes when trying to change AM/PM in the time field. Steps to reproduce:

I was able to reproduce it on an another PR build. Follow the same steps, but when it doesn't crash press backspace and then it crashes. We should enter a bug for this.

packages/@react-spectrum/datepicker/src/utils.ts
snowystinger: I'm fine with the hook name being long, it's private and i'd prefer the name be accurate

packages/@react-spectrum/datepicker/src/utils.ts
snowystinger: Instead of guesstimating, could we make use of useLayoutEffect to render the placeholder, measure, then replace the placeholder with the real value before paint actually happens? Then we should be exact.

We could also render it position absolute and visibility hidden behind the real one, that way we don't need to make a change to state and it'd pick up all the correct CSS inheritance

snowystinger (commented):
I assume we're ok with the minWidth on all these fields changing in response to the time? Especially for ones with shorter durations, like fields containing seconds. Or even just moving from AM -> PM I think was noted?

Sounds like it might cause problems for chromatic unless we mock the time for those?

akyllus (approved):

packages/@react-spectrum/datepicker/src/utils.ts
ktabors: That was my first approach. I recall there being edge cases, it broke a lot of tests even with aria-hidden, and we still have the DateRangePicker problem of needing to set the value higher up than we have the value, placeholder, and segments.

LFDanLu (approved):
Approving for testing, filed https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=70985842 for the Android crash since it is pre-existing.

yihuiliao:
> filed https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=70985842 for the Android crash

Seems like this bug is basically https://github.com/adobe/react-spectrum/issues/6183 except that instead of crashing, it concatenates AM and PM (so that it looks like AMPM)

yihuiliao (approved):
Approving for testing

PR #6577, opened 2024-06-19
Fix tooltip positioning when containerPadding is set

LFDanLu:
Caught by Chromatic

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

See https://www.chromatic.com/build?appId=5f0dd5ad2b5fc10022a2e320&number=750. You can also try building the chromatic storybook locally and testing the story there but I noticed it doesn't flip like it does in chromatic (seems to be a separate issue though)

## 🧢 Your Project:

RSP


packages/@react-aria/overlays/src/useOverlayPosition.ts
```diff
@@ -171,6 +171,9 @@ export function useOverlayPosition(props: AriaPositionProps): PositionAria {
     // This is so we don't have to delay autoFocus scrolling or delay applying preventScroll for popovers
     overlay.style.top = '';
     overlay.style.bottom = '';
+    overlay.style.left = '';
+    overlay.style.right = '';
+
```
LFDanLu, line 176: When the tooltip is flipped from its previously calculate position, there is a scenario where both left and right are applied on the overlay at the same time where the left/right set respectively are the values to position the tooltip to the right/left of the button respectively, resulting in the tooltip squishing itself in an attempt to fufill both constraints. Clearing it out here means this won't happen

snowystinger (approved):

ktabors (approved):
I was trying to reproduce this in Firefox and accidently found some interesting behavior. It's pre-existing. The following is an .mov file.
https://github.com/adobe/react-spectrum/assets/309540/fd4e5a2f-f24a-4281-aff0-ab136136e2be

LFDanLu:
@ktabors ah, super weird, looks like it is constantly flipping back and forth I guess? Def file a bug for it so we have it tracked if its preexisting

PR #6588, opened 2024-06-20
Fix incremental layout with variable heights

devongovett:
Reported by @rostero1 in https://github.com/adobe/react-spectrum/pull/6518#issuecomment-2178438003

In ListLayout with variable row heights, sometimes a large empty space could appear at the end of the list. This was due to the `validRect` not being invalidated properly when the height of an item changed. When this occurs, all items after that item need to be repositioned.

For clarity, I've also renamed two properties:

* `lastValidRect` is now called `validRect` - this is the rectangle that is currently valid. We can reuse previously laid out items within this rectangle.
* `validRect` is now called `requestedRect` - this is the rectangle that has been requested for layout so far. After a layout pass, `validRect` will equal `requestedRect`.

### DIFF
--- a/packages/@react-stately/layout/src/ListLayout.ts
+++ b/packages/@react-stately/layout/src/ListLayout.ts
@@ -74,8 +74,10 @@ export class ListLayout<T> extends Layout<Node<T>, ListLayoutProps> implements D
   protected loaderHeight: number;
   protected placeholderHeight: number;
   protected enableEmptyState: boolean;
-  protected lastValidRect: Rect;
+  /** The rectangle containing currently valid layout infos. */
   protected validRect: Rect;
+  /** The rectangle of requested layout infos so far. */
+  protected requestedRect: Rect;
 
   /**
    * Creates a new ListLayout with options. See the list of properties below for a description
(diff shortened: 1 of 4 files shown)


reidbarber (approved):
The rendering of the items looks good. I'm seeing interesting behavior when scrolling quickly, towards the end of the list, with variable height enabled. Although once all the items get rendered once, it doesn't reproduce. This is MacOS Chrome.

https://github.com/adobe/react-spectrum/assets/8961049/1f3ef6b6-9960-4b13-9cc6-0c007eae2851

https://github.com/adobe/react-spectrum/assets/8961049/d86c981f-a497-482e-b052-bba9e7050efb

LFDanLu (approved):
LGTM. The behavior that @reidbarber saw is interesting, I suppose that is because we don't actually know the heights of the rows that haven't been actually rendered and thus we use are using the estimated row height to calculate a theoretical scroll height for the scrollable region? Then as more rows render as you scroll down, the actual total scroll height grows, meaning your mouse's drag position relative to the scrollbar is incrementally no longer in sync with the actual scroll position? Looks like we actually have the same problem in RSP TableView if `overflowMode: wrap` is applied with highly variable row heights.

devongovett:
Yeah exactly. That has always been the case.

PR #6631, opened 2024-06-27
Render drop indicators with collection renderers and add GridLayout

### LINKED ISSUE #6617: RAC Virtualizer content element should have `role="presentation"`
The generic `div` container that wraps the virtualized elements and provides the content height for native scrolling should have `role="presentation"` so that it does not break the accessibility tree structure for the containing grid, menu or listbox.

https://github.com/adobe/react-spectrum/blob/2cd8f0b509cd9a1d93322dd75fc1306e5f0e217f/packages/react-aria-components/src/Virtualizer.tsx#L65

The problem can be seen in the Virtualized RAC table at: https://reactspectrum.blob.core.windows.net/reactspectrum/ff19a7c1ca7fccfd3e46eeef1c6b71ef23f36437/storybook/index.html?path=/story/react-aria-components--virtualized-table&providerSwitcher-express=false

We may wish to return `role="presentation"` at: https://github.com/adobe/react-spectrum/blob/2cd8f0b509cd9a1d93322dd75fc1306e5f0e217f/packages/%40react-aria/virtualizer/src/ScrollView.tsx#L245-L247

devongovett:
Also fixes #6617.

This refactor makes collection renderers responsible for rendering the drop indicators between each item in a collection. That enables Virtualizer layouts to handle positioning them correctly, rather than being inside each item. This is implemented by passing a `renderDropIndicator` function to the `CollectionRoot` and `CollectionBranch` components from the collection component (e.g. ListBox/GridList). Each layout has a `getDropTargetLayoutInfo` method to retrieve the size and position of a drop indicator on demand. It is not included in the collection or returned as part of the visible layout infos.

There is also a refactor to Virtualizer to remove the `renderWrapper` function, which has been moved into the `Virtualizer` component rather than in stately. In RAC we handle this differently since children are rendered on demand (by `CollectionBranch` rather than eagerly.

Finally, I've added a new GridLayout that is a simplified version of the one from CardView but with fewer Spectrum-specific details. It also implements the new drag and drop methods to position drop indicators, including flipping from vertical to horizontal if only one column is shown in a narrow view.


packages/react-aria-components/src/DragAndDrop.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export interface DragAndDropContextValue {
+  dragAndDropHooks?: DragAndDropHooks,
+  dragState?: DraggableCollectionState,
+  dropState?: DroppableCollectionState
+}
+
+export const DragAndDropContext = createContext<DragAndDropContextValue>({});
+export const DropIndicatorContext = createContext<DropIndicatorContextValue | null>(null);
+
```
devongovett, line 26: Broke these out of useDragAndDrop.tsx into a separate file to improve tree shaking. Components depend on these contexts all the time but we shouldn't pull in the actual dnd hooks unless they are used.

packages/@react-spectrum/table/src/TableViewBase.tsx
```diff
@@ -690,6 +691,17 @@ function TableVirtualizer<T>(props: TableVirtualizerProps<T>) {
   );
 }
 
+function renderChildren<T extends object>(parent: View | null, views: View[], renderWrapper: TableVirtualizerProps<T>['renderWrapper']) {
```
ktabors, line 694: Is there a way to export the version of this (renderChildren()) in Virtualizer and use it here? The method signature seemed to be the only change. Or is there a reason we don't do that?

packages/@react-stately/layout/src/GridLayout.ts
```diff
@@ -0,0 +1,215 @@
...
+  constructor(options: GridLayoutOptions) {
+    super();
+    this.minItemSize = options.minItemSize || new Size(200, 200);
+    this.maxItemSize = options.maxItemSize || new Size(Infinity, Infinity);
+    this.minSpace = options.minSpace || new Size(18, 18);
+    this.maxColumns = options.maxColumns || Infinity;
+    this.dropIndicatorThickness = options.dropIndicatorThickness || 2;
+  }
+
+  validate(): void {
```
ktabors, line 64: This seems more like a "calculate" function.

ktabors (commented):
This is great!

BUG - Virtualized List Box Grid Story - I started DnD using keyboard with item 21. I moved the drop location indicator up, trying to get to item 1. It moved the scroll view once, then the second time it moves the scroll view my drop indicator is lost. Story crashes and I have to reload page. There isn't a console error. I tried with a different item and it crashed the first time I triggered a scroll.

Looking at some of the changes, do we need a Virtualized ListBox with Sections and a Virtualized Table with Nested columns stories to test?

snowystinger (commented):
VO on iOS I can't drag an item past what the virtualizer initially loaded. For example, Virtualized Table initiate a drag on item 1, then try to drop it past ~item 20, I can't move past that point.

Pre-existing, happens in RSP Main on May 16th

snowystinger (commented):
ListBox, when dragging via mouse the reused row appears as hovered even if it's a different item due to virtualization

GridList, cannot mouse drag to a position outside the initial viewport, it won't scroll the collection

snowystinger (approved):

ktabors (approved):

packages/@react-stately/layout/src/GridLayout.ts
devongovett: yeah this was existing in the Layout base class. But I agree, we should rename it before we release the major

packages/@react-spectrum/table/src/TableViewBase.tsx
devongovett: We could but I was trying to minimize the amount of "deprecated" things we export. The way RAC does it is the preferred way, but we need the old way in RSP for now. Hopefully in the future we can update RSP accordingly.

devongovett:
Merging to unblock merge conflicts in upcoming PR. Will fix bugs in subsequent PR.

Issue #6646, opened 2024-06-29
[internationalized/date] Add eachMonthOfInterval query


yajihum:
### Provide a general summary of the feature here

We would like to make the `eachMonthOfInterval` query available in `@internationalized/date`, similar to what is available in [date-fns](https://date-fns.org/v3.6.0/docs/eachMonthOfInterval). 
Without this feature, users have to either define the function themselves or install date-fns, which is inconvenient.

I don't think it's necessary to cover everything available in date-fns, but it would be helpful to add convenient features (usable with the DateValue type) for using RAC.

### 🤔 Expected Behavior?

It would be nice to be able to obtain each month spanning the specified start and end dates.

```tsx
let start = new CalendarDateTime(2023, 11, 15, 9, 20);
let end = new CalendarDateTime(2024, 2, 10, 9, 0);

eachMonthOfInterval(start, end); 
// => [ 2023-12-01T00:00, 2024-01-01T00:00, 2024-02-01T00:00 ]
```

### 😯 Current Behavior

Since this query does not exist, we need to either define it from scratch or install `date-fns`.

### 💁 Possible Solution

I think it can be implemented as follows. I plan to submit a PR.

```tsx
/** Returns an array of the start dates of each month within the specified date interval. */
export function eachMonthOfInterval(start: DateValue, end: DateValue): DateValue[] {
  const startDate = startOfMonth(start).set({hour: 0, minute: 0});
  const endDate = endOfMonth(end).set({hour: 0, minute: 0});

  if (startDate.compare(endDate) > 0) {
    throw new Error('The start date must be before the end date.');
  }

  function collectDates(currentDate: DateValue, endDate: DateValue): DateValue[] {
    if (currentDate.compare(endDate) > 0) {
      return [];
    }
    return [currentDate, ...collectDates(startOfMonth(currentDate.add({months: 1})), endDate)];
  }

  return collectDates(startDate.add({months: 1}), endDate);
}
```

### 🔦 Context

When implementing calendar display using RAC, I wanted to obtain `eachMonthOfInterval`, but it was not available in `@internationalized/date`, so I had to install date-fns separately. 
Since handling dates in RAC requires the DateValue type instead of the standard JavaScript Date type, converting to DateValue is also cumbersome. 
Therefore, it would be great if we could obtain `eachMonthOfInterval `directly in the DateValue type from the start.

yihuiliao:
[...]

PR #6647, opened 2024-06-29
feat: add eachMonthOfInterval query

yajihum:
Closes https://github.com/adobe/react-spectrum/issues/6646

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Adding test files

### DIFF
--- a/packages/@internationalized/date/src/queries.ts
+++ b/packages/@internationalized/date/src/queries.ts
@@ -180,6 +180,25 @@ export function getMinimumDayInMonth(date: AnyCalendarDate) {
   return 1;
 }
 
+/** Returns an array of the start dates of each month within the specified date interval. */
+export function eachMonthOfInterval(start: DateValue, end: DateValue): DateValue[] {
+  const startDate = startOfMonth(start).set({hour: 0, minute: 0});
+  const endDate = endOfMonth(end).set({hour: 0, minute: 0});
+
+  if (startDate.compare(endDate) > 0) {
+    throw new Error('The start date must be before the end date.');
+  }
+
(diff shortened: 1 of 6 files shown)


yajihum:
[ask]
In date-fns, it is possible to set [the step option](https://github.com/date-fns/date-fns/blob/main/src/eachMonthOfInterval/index.ts#L41). Should I add this as well?

packages/@internationalized/date/src/queries.ts
```diff
@@ -180,6 +180,25 @@ export function getMinimumDayInMonth(date: AnyCalendarDate) {
...
+  }
+
+  function collectDates(currentDate: DateValue, endDate: DateValue): DateValue[] {
+    if (currentDate.compare(endDate) > 0) {
+      return [];
+    }
+    return [currentDate, ...collectDates(startOfMonth(currentDate.add({months: 1})), endDate)];
+  }
+
+  return collectDates(startDate.add({months: 1}), endDate);
```
yihuiliao, line 199: Why did you choose to exclude the start month in the interval?

I was looking at the [date-fns docs](https://date-fns.org/v3.6.0/docs/eachMonthOfInterval#examples) and it seems like the interval is inclusive of the start month so I was just curious about this choice. Not sure if I'm maybe just missing something.

yajihum: @yihuiliao 
Thank you for your comment!

I forgot that the JavaScript Date type starts counting months from 0, which led to an incorrect implementation.
I have fixed it to include the start month in the result, so I would appreciate it if you could check it again.

[fix: include the start month in the result](https://github.com/adobe/react-spectrum/pull/6647/commits/a12e1b349a138ff6ece3727447535d9bac3de551)

PR #6657, opened 2024-07-03
fix: remove autoCapitalize to prevent mismatch warning

ryo-manba:
Closes https://github.com/adobe/react-spectrum/issues/5599

This bug occurs because `autoCapitalize: 'off'` is set for each segment of DateInput, but Firefox defaults to `none`. This mismatch between server-side and client-side rendering causes the warning to be displayed; Safari and Chrome leave it at `off`, so no warning is displayed. DateInput has `input-mode=numeric` and already displays a numeric keyboard, so setting `autoCapitalize` is not necessary.

[A similar error was also occurring](https://github.com/nextui-org/nextui/pull/3376) in NextUI, which is based on react-aria.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Open the following URL in Firefox:
https://stackblitz.com/~/github.com/ryo-manba/auto-capitalize-error

1. Click on the link "Page with autoCapitalize mismatch error".
2. Observe the warning in the console.
3. Go back to the home page.
4. Click on the link "Page without autoCapitalize mismatch error" (autoCapitalize is deleted).
5. Observe that no warning is displayed in the console.


packages/@react-aria/datepicker/src/useDateSegment.ts
```diff
@@ -23,7 +23,7 @@ import {useSpinButton} from '@react-aria/spinbutton';
 
 export interface DateSegmentAria {
   /** Props for the segment element. */
-  segmentProps: DOMAttributes
+  segmentProps: React.HTMLAttributes<HTMLDivElement>;
```
ryo-manba, line 26: Fixed the type definitions to prevent type errors caused by props such as `autoCapitalize` and `spellCheck`.

packages/@react-aria/datepicker/src/useDateSegment.ts
```diff
@@ -23,7 +22,7 @@ import {useSpinButton} from '@react-aria/spinbutton';
 
 export interface DateSegmentAria {
   /** Props for the segment element. */
-  segmentProps: DOMAttributes
+  segmentProps: React.HTMLAttributes<HTMLDivElement>
```
snowystinger, line 25: why change this?

ryo-manba: There was a warning with the original type.

https://github.com/adobe/react-spectrum/blob/2964e9aafc281888cd65da5bd77512e7f999bd84/packages/%40react-aria/datepicker/src/useDateSegment.ts#L368-L397
image

snowystinger: huh odd it's not coming up for me, I wonder if when we turn on strict mode that this will show up

snowystinger (approved):
I think this is fine, in all segments I can think of we either control the capitalization or it's numeric.
Pending next release before we merge.

yihuiliao (approved):

packages/@react-aria/datepicker/src/useDateSegment.ts
ryo-manba: Hmmm, maybe it is.

PR #6700, opened 2024-07-12
Update TanStack Router example in docs

### LINKED ISSUE #6413: The instructions for client side routing for Tantsack Router doesn't work.
### 🙋 Documentation Request

Currently, if you implement it as the documentation has explained, you get this error:
image

To get rid of the error, you can change it to this:
```ts
declare module 'react-aria-components' {
  interface RouterConfig {
    href: ToPathOption<RegisteredRouter>;
    routerOptions: Omit<NavigateOptions, keyof ToOptions>;
  }
}
```

But this makes the `href` becomes just `string`, and you get no auto complete.

The way to get autocomplete is doing this:
```ts
declare module 'react-aria-components' {
	interface RouterConfig {
		href: ToPathOption<RegisteredRouter, '/', '/'> | ({} & string);
		routerOptions: Omit<NavigateOptions, 'to' | 'from'>;
	}
}
```

And change the router provider to this:
```ts
function RootComponent() {
	const router = useRouter();

	return (
		<RouterProvider
			navigate={(to, options) =>
				router.navigate({
					...options,
					to: to as ToPathOption<RegisteredRouter, '/', '/'>,
				})
			}
			useHref={(to) => {
				return router.buildLocation({ to }).href;
			}}
		>
			<Outlet />
		</RouterProvider>
	)
}
```
Now you will get auto complete in the links, and you can use any string as well:
image

This would also enable using the `params` and `search` from `tanstack/react-router` to be typesafe:
image

But right now since there is no way to constrict them based on the given `href` they show all the options for all the routes.

I think the way to fix this would be to give `RouterConfig` a generic string parameter and pass it to `ToPathOption` but I'm not sure.

```ts
declare module 'react-aria-components' {
	interface RouterConfig<Path extends string> {
		href: ToPathOption<RegisteredRouter, '/', Path> | ({} & string);
		routerOptions: Omit<NavigateOptions<RegisteredRouter, '/', Path>, 'to' | 'from'>;
	}
}
```

I'm not too familiar with how the types are set up for React Aria, but what I'm hoping for is this will link the `href` value to `ToPathOption` and `NavigateOptions` and show the correct values for `search` and `params` options.

### 🧢 Your Company/Team

_No response_

devongovett:
Fixes #6413

The types exposed by TanStack Router have changed since the docs were originally written.

This keeps `href` as a string. We could also consider typing it as an object as shown here but not sure if everyone would want that or not? https://github.com/adobe/react-spectrum/issues/6587#issuecomment-2224235268

### DIFF
--- a/packages/dev/docs/pages/react-spectrum/routing.mdx
+++ b/packages/dev/docs/pages/react-spectrum/routing.mdx
@@ -278,7 +278,7 @@ To use [TanStack Router](https://tanstack.com/router) with React Spectrum, rende
 import {useRouter, type NavigateOptions, type ToOptions} from '@tanstack/react-router';
 import {Provider, defaultTheme} from '@adobe/react-spectrum';
 
-declare module 'react-aria-components' {
+declare module '@adobe/react-spectrum' {
   interface RouterConfig {
     href: ToOptions['to'];
     routerOptions: Omit<NavigateOptions, keyof ToOptions>;
--- a/packages/dev/docs/pages/react-spectrum/routing.mdx
+++ b/packages/dev/docs/pages/react-spectrum/routing.mdx
@@ -275,13 +275,13 @@ export default function App() {
 To use [TanStack Router](https://tanstack.com/router) with React Spectrum, render React Spectrum's `Provider` inside your root route. Use `router.navigate` in the `navigate` prop, and `router.buildLocation` in the `useHref` prop. You can also configure TypeScript to get autocomplete for the `href` prop by declaring the `RouterConfig` type using the types provided by TanStack Router.
 
 ```tsx
-import {useRouter, type RegisteredRouter, type NavigateOptions, type ToOptions} from '@tanstack/react-router';
+import {useRouter, type NavigateOptions, type ToOptions} from '@tanstack/react-router';
 import {Provider, defaultTheme} from '@adobe/react-spectrum';
 
-declare module '@adobe/react-spectrum' {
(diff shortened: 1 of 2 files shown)


packages/dev/docs/pages/react-spectrum/routing.mdx
```diff
@@ -275,13 +275,13 @@ export default function App() {
 To use [TanStack Router](https://tanstack.com/router) with React Spectrum, render React Spectrum's `Provider` inside your root route. Use `router.navigate` in the `navigate` prop, and `router.buildLocation` in the `useHref` prop. You can also configure TypeScript to get autocomplete for the `href` prop by declaring the `RouterConfig` type using the types provided by TanStack Router.
 
 ```tsx
-import {useRouter, type RegisteredRouter, type NavigateOptions, type ToOptions} from '@tanstack/react-router';
+import {useRouter, type NavigateOptions, type ToOptions} from '@tanstack/react-router';
 import {Provider, defaultTheme} from '@adobe/react-spectrum';
 
-declare module '@adobe/react-spectrum' {
+declare module 'react-aria-components' {
```
LFDanLu, line 281: Shouldn't this be `'@adobe/react-spectrum'` still?

devongovett: ah yeah good catch

```suggestion
declare module '@adobe/react-spectrum' {
```

LFDanLu (approved):

PR #6717, opened 2024-07-16
Fix crash in TimeField on Android

devongovett:
Closes https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=70985842

This could occur when tapping on an already focused date segment. This moved the cursor between letters. Then when entering a character, Chrome would fire a composition event rather than a regular text entry event. Composition events are not cancelable, so chrome would mutate the DOM. Later, React would try to update the DOM but it was no longer in the expected state and this resulted in a crash.

Fix is to always collapse the selection to the start of the segment. We already did this on initial focus, but tapping on the segment again could change the selection. Now we listen for selection change events and collapse them if they are inside the segment.

### DIFF
--- a/packages/@react-aria/datepicker/src/useDateSegment.ts
+++ b/packages/@react-aria/datepicker/src/useDateSegment.ts
@@ -269,6 +269,17 @@ export function useDateSegment(segment: DateSegment, state: DateFieldState, ref:
     selection.collapse(ref.current);
   };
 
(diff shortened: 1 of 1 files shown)


ktabors (approved):

LFDanLu (approved):
LGTM, verified the crash no longer happens on Android Chrome and that behavior is still the same in other browser/devices

PR #6723, opened 2024-07-17
Fix virtualizer bug when an item changes parents

devongovett:
We saw this issue in doc search due to section keys changing. Virtualizer saw the child keys as already visible even though their parents changed and didn't create new views for them. Now we treat parent key changes the same way we treat deletes. The view will then be re-created under its new parent. Added a test for this case when a listbox item changes sections.

In doc search we also make this more efficient by using only the title as the key rather than including the index, so that when sections are re-ordered the keys remain consistent and views can be reused.


snowystinger (approved):

LFDanLu (approved):
Approving since the bug in docs on mobile is unrelated, behavior looks good to me

PR #6724, opened 2024-07-18
Get rid of sizeToFit to fix cases where listbox rect wasn't updating properly

LFDanLu:
The listbox in the async case and the Docs search on mobile were rendering with very small scroll rects, causing only a couple of items to be visible at a time even though there was actually more vertical real estate available. This was because of the `sizeToFit` logic [here](https://github.com/adobe/react-spectrum/blob/9b48ec5d528adeba1a96592945459726def7ba56/packages/%40react-aria/virtualizer/src/ScrollView.tsx#L166) being called by the ResizeObserver that would then set the height of the scroll rect to the content size at the time (placeholder height). Before the changes made in https://github.com/adobe/react-spectrum/pull/6508, `updateSize` would be called much more frequently when `contentSize` changed hence why this wasn't a problem before. We think that we actually don't need `sizeToFit` anymore since the browser should handle updating the height of the scrollview when more content loads async so cases where a dropdown goes from having a single placeholder row to having loaded items should still work

To be tested further, if it breaks stuff we can fallback to calling `updateSize` if `contentSize` changes in scrollView

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test anything with a ListBox (Listbox, Combobox, Picker, tray versions of those, docs search)

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-aria/dnd/stories/VirtualizedListBox.tsx
+++ b/packages/@react-aria/dnd/stories/VirtualizedListBox.tsx
@@ -168,7 +168,6 @@ export const VirtualizedListBox = React.forwardRef(function (props: any, ref) {
         {...mergeProps(collectionProps, listBoxProps)}
         ref={domRef}
         className={classNames(dndStyles, 'droppable-collection', 'is-virtualized', {'is-drop-target': isDropTarget})}
-        sizeToFit="height"
         scrollDirection="vertical"
         layout={layout}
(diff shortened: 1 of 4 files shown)


devongovett (approved):

snowystinger (approved):
please run chromatic on main after merging this, it looks like our baseline hasn't been update in too long

PR #6778, opened 2024-07-26
(WIP) S2 table

devongovett:


akyllus (approved):

packages/@react-aria/virtualizer/src/VirtualizerItem.tsx
```diff
@@ -57,8 +57,12 @@ export function layoutInfoToStyle(layoutInfo: LayoutInfo, dir: Direction, parent
...
 
   let rectStyles = {
-    top: layoutInfo.rect.y - (parent ? parent.rect.y : 0),
-    [xProperty]: layoutInfo.rect.x - (parent ? parent.rect.x : 0),
+    // TODO: For layoutInfos that are sticky that have parents with overflow visible, their "top" will be relative to the to the nearest scrolling container
+    // which WON'T be the parent since the parent has overflow visible. This means we shouldn't offset the height by the parent's position
+    // Not 100% about this change here since it is quite ambigious what the scrolling container maybe and how its top is positioned with respect to the
+    // calculated layoutInfo.y here
+    top: layoutInfo.rect.y - (parent && !(parent.allowOverflow && layoutInfo.isSticky) ? parent.rect.y : 0),
+    [xProperty]: layoutInfo.rect.x - (parent && !(parent.allowOverflow && layoutInfo.isSticky) ? parent.rect.x : 0),
```
LFDanLu, line 65: Checked that this doesn't break things in RSP TableView but still might be a bit of a tricky change due to how sticky positioned elements calculate their position

packages/@react-spectrum/s2/intl/ar-AE.json
LFDanLu, line 1: all intl changes are from the script

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+  /**
+   * Sets the overflow behavior for the cell contents.
+   * @default 'truncate'
+   */
+  overflowMode?: 'wrap' | 'truncate',
+  /** How selection should be displayed. */
+  selectionStyle?: 'checkbox' | 'highlight',
+  // TODO: will we contine with onAction or rename to onRowAction like it is in RAC?
+  /** Handler that is called when a user performs an action on a row. */
+  onAction?: (key: Key) => void,
```
LFDanLu, line 83: up for discussion, depends on if we might want to bring onCellAction over

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+          aria-label={stringFormatter.format('table.loadingMore')} />
+      </div>
+    </UNSTABLE_TableLoadingIndicator>
+  );
+
+  // If the user is rendering their rows in dynamic fashion, wrap their render function in Collection so we can inject
+  // the loader. Otherwise it is a static renderer and thus we can simply add the table loader after
+  // TODO: this assumes that the user isn't providing their children in some wrapper though and/or isn't doing a map of children
+  // (though I guess they wouldn't provide items then so the check for this is still valid in the latter case)...
+  if (typeof children === 'function' && items) {
```
LFDanLu, line 356: will need to think a bit more about this and how best to tell if someone is providing a render function or just mapping over items directly

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
-          {props.children}
-          {allowsSorting && (
-            <span aria-hidden="true" className="sort-indicator">
-              {sortDirection === 'ascending' ? '▲' : '▼'}
-            </span>
-          )}
+          {/* Note this is mainly for column's without a dropdown menu. If there is a dropdown menu, the button is styled to have a focus ring for simplicity
+          (no need to juggle showing this focus ring if focus is on the menu button and not if it is on the resizer) */}
+          {/* Separate absolutely positioned element because appyling the ring on the column directly via outline means the ring's required borderRadius will cause the bottom gray border to curve as well */}
+          <CellFocusRing isFocusVisible={isFocusVisible} />
```
LFDanLu, line 486: I know we were trying to cut down on all the extra elements but relying on the outline styles for the column directly for the focus ring ran into the above issues

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+
+  let buttonAlignment = 'start';
+  let menuAlign = 'start' as 'start' | 'end';
+  // TODO: align center is quite strange, copied this from S1 but really there isn't a good place to put the menu when the column text is centered
+  if (align === 'center') {
+    buttonAlignment = 'center';
+  } else if (align === 'end') {
+    buttonAlignment = 'end';
+    menuAlign = 'end';
+  }
```
LFDanLu, line 723: image
wonder if there should be an align center option for menu

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+            of having a separate white rectangle div base below a div with the row background color set above it as a mask
+            is that it doesn't come out as the same color as the other cells because the base below the sticky cell when other selected cells are scrolled below it will be the blue of the
+            other cells, not the same white base. If I could convert informative-900/10 (and the rest of the rowBackgroundColors) to an equivalent without any opacity
+            then I could do away with this styling. To reproduce this, comment out the stickyCell gray-25, get rid of the below div and apply backgroundColor: '--rowBackgroundColor' to checkboxCellStyle.
+            Having the CellFocusRing here instead of applying a outline on the cell directly also makes it NOT overlap with the border (can be remedied with a -3px outline offset) and applying a border radius to get the curved outline focus ring messes
+            with the divider rendered on the cell since those are also borders
+          */}
+          <div role="presentation" className={cellBackground({isSticky})} />
```
LFDanLu, line 982: Once I get token values for all table row states with full opacity I can do away with this extra element

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+  '--rowFocusIndicatorColor': {
+    type: 'backgroundColor',
+    value: {
+      default: 'focus-ring',
+      forcedColors: 'Highlight'
+    }
+  },
+  // TODO: outline here is to emulate v3 forcedColors experience but runs into the same problem where the sticky column covers the outline
+  // This doesn't quite work because it gets cut off by the checkbox cell background masking element, figure out another way. Could shrink the checkbox cell's content even more
+  // and offset it by margin top but that messes up the checkbox centering a bit
```
LFDanLu, line 1040: for context, this is what it looks like in HCM for windows for S1 tableview:
image

packages/@react-spectrum/s2/style/spectrum-theme.ts
```diff
@@ -87,8 +87,23 @@ export function baseColor(base: keyof typeof color) {
   };
 }
 
-export function lightDark(light: keyof typeof color, dark: keyof typeof color): `[${string}]` {
-  return `[light-dark(${color[light]}, ${color[dark]})]`;
+export function lightDark(light: Color<keyof typeof color>, dark:Color<keyof typeof color>): `[${string}]` {
```
LFDanLu, line 90: support syntax like `lightDark('informative-900/10', 'informative-700/10')`

packages/@react-spectrum/s2/style/spectrum-theme.ts
```diff
@@ -513,7 +528,11 @@ export const style = createTheme({
       base: colorToken('background-base-color'),
       'layer-1': colorToken('background-layer-1-color'),
       'layer-2': weirdColorToken('background-layer-2-color'),
-      pasteboard: weirdColorToken('background-pasteboard-color')
+      pasteboard: weirdColorToken('background-pasteboard-color'),
+      'focus-ring': {
+        default: colorToken('focus-indicator-color'),
+        forcedColors: 'Highlight'
+      }
```
LFDanLu, line 535: wanted the focus ring color that only exists in outline styles as the background color for the nubbin and resizer handle

packages/@react-spectrum/s2/style/spectrum-theme.ts
```diff
@@ -87,8 +87,23 @@ export function baseColor(base: keyof typeof color) {
   };
 }
 
-export function lightDark(light: keyof typeof color, dark: keyof typeof color): `[${string}]` {
-  return `[light-dark(${color[light]}, ${color[dark]})]`;
+export function lightDark(light: Color<keyof typeof color>, dark:Color<keyof typeof color>): `[${string}]` {
+  let [lightColor, lightOpacity] = light.split('/');
+  let [darkColor, darkOpacity] = dark.split('/');
+  // @ts-ignore
+  lightColor = color[lightColor];
```
snowystinger, line 94: could just make a new variable no?

packages/@react-spectrum/s2/style/spectrum-theme.ts
snowystinger: I don't love this being part of the theme, could you use a variable instead to get the focus-ring color and then set the background color to that variable?

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+}
+
+let InternalTableContext = createContext<TableProps & {scale?: Scale, layout?: S2TableLayout<unknown>, setIsInResizeMode?:(val: boolean) => void, isInResizeMode?: boolean}>({});
+
+const tableWrapper = style({
+  width: 'full'
+});
+
+const table = style<TableRenderProps & S2TableProps>({
+  height: 'full',
```
snowystinger, line 116: does this still work if the parent is a `flex child` instead of explicitly setting a height?

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+    forcedColors: 'ButtonBorder'
+  },
+  outlineWidth: {
+    default: 1,
+    isQuiet: 0,
+    isFocusVisible: 2
+  },
+  outlineStyle: 'solid',
+  borderRadius: {
+    default: size(6),
```
snowystinger, line 140: the border radius for the table really doesn't fall into their scale? seems like a mistake in spectrum

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+type Scale = 'large' | 'medium';
+
+export class S2TableLayout<T> extends UNSTABLE_TableLayout<T> {
+  protected isStickyColumn(node: GridNode<T>): boolean {
+    if (node.props.isSticky) {
+      return true;
+    }
+
+    return false;
+  }
```
snowystinger, line 175:
```suggestion
  protected isStickyColumn(node: GridNode<T>): boolean {
    return node.props.isSticky
  }
```
Cast to `Boolean` if it can be undefined

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+    loadingState,
+    onLoadMore,
+    onResize: propsOnResize,
+    onResizeStart: propsOnResizeStart,
+    onResizeEnd: propsOnResizeEnd,
+    onAction,
+    ...otherProps
+  } = props;
+
+  // TODO: perhaps just make a useScale
```
snowystinger, line 255: 👍🏻

packages/@react-spectrum/s2/stories/Table.stories.tsx
```diff
@@ -10,49 +10,556 @@
...
+              {column.name}
+            </Column>
+          )}
+        </TableHeader>
+        <TableBody items={list.items}>
+          {item =>
+            (<Row id={item.data.id} columns={columns}>
+              {(column) => {
+                return column.id === 'title'
+                  ? <Cell textValue={item.data.title}><Link isQuiet><a href={item.data.url} target="_blank">{item.data.title}</a></Link></Cell>
```
reidbarber, line 545: It also looks like there is an extra focus stop in the Link cells?

packages/@react-spectrum/s2/stories/Table.stories.tsx
```diff
@@ -10,49 +10,556 @@
...
+
+export const ResizingUncontrolledSortableColumns = {
+  render: (args) => <AsyncLoadingExample {...args} />,
+  args: {
+    ...Example.args,
+    onResize: action('onResize'),
+    onResizeStart: action('onResizeStart'),
+    onResizeEnd: action('onResizeEnd')
+  },
+  name: 'resizable, sortable, reddit example'
```
reidbarber, line 564: Can you add a story with many items (~10k or similar)

packages/@react-spectrum/s2/style/spectrum-theme.ts
LFDanLu: sure, mainly did this for any other cases where the background color of an element needed to match the same blue as the focus ring/outline but I'll move it to a variable

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: So if the parent has display flex and there are other flex children alongside the table that the height of the container gets distributed amongst?

packages/@react-spectrum/s2/stories/Table.stories.tsx
LFDanLu: good catch, forgot to update the Link implementation to match what we have in S2 now so had the extra `<a>`

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: good call out, I'll left a question in the token figma

packages/@react-spectrum/s2/src/Table.tsx
snowystinger: yeah, or a flex laid out horizontally and relying on align-items stretch

packages/@react-spectrum/s2/style/spectrum-theme.ts
snowystinger: I don't think that's going to be common outside of components, but if it is, then we should probably get it added to spectrum itself

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1110 @@
...
+    default: 8,
+    scale: {
+      large: 14
+    }
+  }
+});
+
+const selectAllCheckboxColumn = style({
+  padding: 0,
+  height: '[calc(100% - 1px)]',
```
snowystinger, line 803:
```suggestion
  height: '[calc(100% - self(borderBottomWidth))]',
```
is this calculation actually this?

Note: this may have a bug if you try to do it, see my PR internally for more information. I'll open it here in the near future

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1110 @@
...
+  let {selectionBehavior, selectionMode} = useTableOptions();
+  let [isHeaderRowHovered, setHeaderRowHovered] = useState(false);
+
+  return (
+    <InternalTableHeaderContext.Provider value={{isHeaderRowHovered}}>
+      <RACTableHeader onHoverChange={setHeaderRowHovered} className={tableHeader}>
+        {/* Add extra columns for selection. */}
+        {selectionBehavior === 'toggle' && (
+          // Also isSticky prop is applied just for the layout, will decide what the RAC api should be later
+          // @ts-ignore
```
snowystinger, line 828: this is just to get it on the node properties?

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1110 @@
...
+  ...stickyCell,
+  paddingStart: 16,
+  paddingEnd: {
+    default: 8,
+    scale: {
+      large: 14
+    }
+  },
+  alignContent: 'center',
+  height: '[calc(100% - 1px)]',
```
snowystinger, line 918: same with this one

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1110 @@
...
+    default: 'transparent',
+    isSticky: '--rowBackgroundColor'
+  }
+});
+
+export interface CellProps extends RACCellProps, Pick<ColumnProps, 'align' | 'showDivider'> {
+  /** @private */
+  isSticky?: boolean,
+  /** The content to render as the cell children. */
+  children?: ReactNode
```
snowystinger, line 971: can this actually be optional?

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1110 @@
...
+  let {children, isSticky, showDivider, align, textValue, ...otherProps} = props;
+  let tableVisualOptions = useContext(InternalTableContext);
+  textValue ||= typeof children === 'string' ? children : undefined;
+
+  return (
+    <RACCell
+      // Also isSticky prop is applied just for the layout, will decide what the RAC api should be later
+      // @ts-ignore
+      isSticky={isSticky}
+      // This is a inline style because it needs to set properties ONLY for the end border (don't want to set a color for the bottom border)
```
snowystinger, line 984: i assume this isn't possible with the theme, but can it be?

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1110 @@
...
+        </>
       )}
-      {selectionBehavior === 'toggle' && (
-        <Cell>
-          <Checkbox slot="selection" />
+    </RACCell>
+  );
+}
+
+// TODO: reminder, these dark theme colors were grabbed from the spectrum tokens but they feel quite dark IMO. Will comment in design file
```
snowystinger, line 1012: resolve these?

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: Ah actually now that we are in the monorepo now, I can just pull from `@react-spectrum/utils` so I'll opt for that

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -139,6 +137,21 @@ const table = style<TableRenderProps & S2TableProps>({
   borderRadius: {
     default: size(6),
     isQuiet: 'none'
+  },
+  // These scroll padding values are kinda odd (expected them to match the header height and checkbox column width) but they are the best fit from testing
```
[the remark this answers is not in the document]
LFDanLu: yeah, I'm not quite sure why. Even for our RAC [minimally styled example](https://reactspectrum.blob.core.windows.net/reactspectrum/dcf312bc21ac08e9202e110f10572d4df47784b1/storybook/index.html?path=/story/react-aria-components--virtualized-table&providerSwitcher-express=false) the scrollPaddingTop of 25px that matches the header's height causes the row to be scrolled too far but the story below it works as expected. Need to dig into it some more

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1110 @@
...
+  cursor: {
+    default: 'none',
+    resizableDirection: {
+      'left': 'e-resize',
+      'right': 'w-resize',
+      'both': 'ew-resize'
+    }
+  },
+  // So that the user can still hover + drag the resizer even though it's hit area is partially in the adjacent column's space
+  zIndex: 1000,
```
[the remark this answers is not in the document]
LFDanLu: ah thanks, doesn't look like I need this anymore, looks to be a remanent from when we didn't have virtualizer yet and I was still using native table elements

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: which PR specifically? Mind DMing it to me?

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: yep that is correct

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: ah good catch

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: yeah, can be potentially added to the theme but I didn't add it since it would mean top/bottom/start/end border colors+widths+style would need to be added but I didn't know how many cases would actually use it

packages/@react-spectrum/s2/src/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {useMediaQuery} from '@react-spectrum/utils';
+
+export function useIsMobileDevice(): boolean {
+  return useMediaQuery('(max-width: 640px)');
+}
```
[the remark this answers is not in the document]
LFDanLu: removed in favor of importing it directly from `@react-spectrum/utils`

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: I did some testing around this and it seems fine in the case where the flex direction is row but flex direction column was problematic since the table wouldn't respect the container's height (probably what you were alluding to here). Think I fixed it now, lemme know if the new stories reflect what you were getting at

packages/@react-spectrum/s2/src/Table.tsx
snowystinger: I think it's fine to add all of those, but @devongovett may have an opinion as well

LFDanLu: kk, talked about this in stand up, I'll see about adding something for this in the theme

packages/@react-spectrum/s2/src/Table.tsx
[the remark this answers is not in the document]
LFDanLu: I did some more digging and its bizarre, a 1px change in the scroll padding value corresponds to a 1.5px change in the actual scroll position when keyboard navigating... The same scaling factor seems to also occurs in this [story](https://reactspectrum.blob.core.windows.net/reactspectrum/dcf312bc21ac08e9202e110f10572d4df47784b1/storybook/index.html?path=/story/react-aria-components--virtualized-table&providerSwitcher-express=false) but this [story](https://reactspectrum.blob.core.windows.net/reactspectrum/dcf312bc21ac08e9202e110f10572d4df47784b1/storybook/index.html?path=/story/react-aria-components--virtualized-table-with-resizing&providerSwitcher-express=false) has a 1:1 scaling. Meanwhile, this rough base [example](https://codepen.io/lfdanlu/pen/XWLBzmP) has a 1px scrollPaddingTop to .5 actual scroll position scaling positioning change when you tab bring to the input field into position. Not quite sure what to make of this tbh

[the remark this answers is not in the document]
snowystinger: Are the differences consistent across browsers?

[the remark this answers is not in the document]
LFDanLu: ah good call, looks to be consistent in Chrome and Firefox, Safari actually works as I would've expected with the Table (aka using the column header height for the scroll padding top works perfectly). Safari behaves the same as the other browsers though with my codepen if you scroll to the bottom and tab to the input which is even more strange...

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1103 @@
...
+  outlineStyle: 'solid',
+  borderRadius: {
+    default: size(6),
+    isQuiet: 'none'
+  },
+  // These scroll padding values are kinda odd (expected them to match the header height and checkbox column width) but they are the best fit from testing
+  scrollPaddingTop: {
+    scale: {
+      medium: '[21px]',
+      large: '[27px]'
```
devongovett, line 149: Wouldn't these need to be rems in case the user has a non-default font size set? Use `size` macro? But still weird that they don't match. Is it the border or something?

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1103 @@
...
+    loadingState,
+    onLoadMore,
+    onResize: propsOnResize,
+    onResizeStart: propsOnResizeStart,
+    onResizeEnd: propsOnResizeEnd,
+    onAction,
+    ...otherProps
+  } = props;
+
+  let scale = (useIsMobileDevice() ? 'large' : 'medium') as Scale;
```
devongovett, line 267: Does useIsMobileDevice match the media query we use for scale in S2? I think we changed it from v3...

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1103 @@
...
+  let {items, renderEmptyState, children} = props;
+  let tableVisualOptions = useContext(InternalTableContext);
+  let {loadingState} = tableVisualOptions;
+  let emptyRender;
+  let renderer = children;
+  let stringFormatter = useLocalizedStringFormatter(intlMessages, '@react-spectrum/s2');
+  let loadMoreSpinner = (
+    <UNSTABLE_TableLoadingIndicator className={style({height: 'full', width: 'full'})}>
+      <div className={centeredWrapper}>
+        <ProgressCircle
```
devongovett, line 359: Maybe make the load more spinner a bit smaller? It seems big relative to the row height at the moment IMO

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1103 @@
...
+} as const;
+
+const cell = style<CellRenderProps & S2TableProps>({
+  ...commonCellStyles,
+  color: {
+    default: 'gray-800', // neutral-content-color-default
+    isHovered: 'gray-900', // neutral-content-color-hover
+    isPressed: 'gray-900', // neutral-content-color-down
+    isFocusVisible: 'gray-900' // neutral-content-color-key-focus
+  },
```
devongovett, line 868: Can't you use `color: "neutral"` here? That should set all of these states for you

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1103 @@
...
+      compact: size(6) // table-row-top-to-text-medium-compact
+    }
+  },
+  paddingBottom: {
+    default: 12, // table-row-bottom-to-text-medium-spacious
+    density: {
+      spacious: size(15), // table-row-bottom-to-text-medium-spacious
+      compact: 8 // table-row-bottom-to-text-medium-compact
+    }
+  },
```
devongovett, line 883: We shouldn't use hard coded paddings for centering anymore with the new font. You should be able to calculate these similar to what we do in Button

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1103 @@
...
+  backgroundColor: 'gray-25'
+} as const;
+
+const checkboxCellStyle = style({
+  ...commonCellStyles,
+  ...stickyCell,
+  paddingStart: 16,
+  paddingEnd: {
+    default: 8,
+    scale: {
```
devongovett, line 909: Shouldn't our sizes already scale up automatically?

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1103 @@
...
+      large: 14
+    }
+  },
+  alignContent: 'center',
+  height: '[calc(100% - 1px)]',
```
devongovett, line 914: What's the minus 1 for?

packages/@react-aria/virtualizer/src/VirtualizerItem.tsx
devongovett: Is this assuming that an element with allowsOverflow and isSticky also won't have position relative or absolute? Not sure we can assume that

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: I'm not quite sure why these values work to be honest, I did some digging here: https://github.com/adobe/react-spectrum/pull/6778#discussion_r1735343357 and got some varied results

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: oh whoops I forgot that was the case, I'll bring back the S2 specific useIsMobileDevice that I had added before

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: ah good call thanks, I'll do the same for the cell styles too.

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: It needs to accommodate for the the row's border since it is sticky. If it was full height it would cover it like so:
image

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: it only scales up by a factor of 1.25 right? I double checked the token designs and the spacing from checkbox to text goes from 24px to 30px but the text to cell spacing remains the same regardless of desktop vs mobile (aka 16px. This means this needs to be 24 - 16 = 8px on desktop but 30 -16 = 14px on mobile hence the need for scale here. I just noticed that the large scale values aren't applying for some reason though, digging

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: Hmm, I used `size="S"` instead but that seems too small IMO. Custom sizes aren't supported for ProgressCircle either 
image
FWIW the progress circle sizing is the same as it was in S1

devongovett: with compact, it is right up against the edges. I think the smaller one looks better 
Screenshot 2024-08-29 at 5 44 13 PM

packages/@react-aria/virtualizer/src/VirtualizerItem.tsx
LFDanLu: well, `isSticky` would mean the position is set to `sticky` rather than `relative` or `absolute` no?

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: ah I see what you mean, I'll go ahead and change it to small then

LFDanLu: actually what do you think about this:
image

I've made the loader height row larger (60px like in S1)

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1082 @@
...
+    layout
+  }), [isQuiet, density, overflowMode, loadingState, scale, layout, isInResizeMode, setIsInResizeMode]);
+
+  let isLoading = loadingState === 'loading' || loadingState === 'loadingMore';
+  let scrollRef = useRef(null);
+  let memoedLoadMoreProps = useMemo(() => ({
+    isLoading: isLoading,
+    onLoadMore
+  }), [isLoading, onLoadMore]);
+  useLoadMore(memoedLoadMoreProps, scrollRef);
```
LFDanLu, line 313: This may change when we revisit `useLoadMore` to possibly accept collections, but one thing to note is that the `items` aren't available here since they are provided directly to the TableBody instead of the Table. This causes the issue described in https://github.com/orgs/adobe/projects/19/views/32?sliceBy%5Bvalue%5D=S2&pane=issue&itemId=79185605 where a new load more isn't triggered if items are removed from view manually. We could perhaps add the `items` prop to the Table as well but I'm not sure what form `useLoadMore` to support `collections` so I'm punting for now

packages/@react-spectrum/s2/src/Table.tsx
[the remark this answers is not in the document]
LFDanLu: Filed https://issues.chromium.org/issues/365913982 against chrome and updated to the values that we expect to work. Core issue is that https://github.com/adobe/react-spectrum/blob/086ad3115c5f99115f623ddcd91a724599712cfd/packages/%40react-aria/utils/src/scrollIntoView.ts#L98 causes the scroll container to scroll even though the container is already in view

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: Filed https://issues.chromium.org/issues/365913982 against chrome and updated to the values that we expect to work. Core issue is that https://github.com/adobe/react-spectrum/blob/086ad3115c5f99115f623ddcd91a724599712cfd/packages/%40react-aria/utils/src/scrollIntoView.ts#L98 causes the scroll container to scroll even though the container is already in view

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: Got a reply, the 6 is correct here. It was a middle ground decision where 4px felt too much like S1 and 8px felt too rounded.

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: Added via https://github.com/adobe/react-spectrum/pull/6778/commits/ae0e6f1eccd86a1220eaded37cf34479aa7188f3

packages/@react-spectrum/s2/style/__tests__/style-macro.test.js
LFDanLu, line 1: Note the changes to the style macro test here. I think these are expected due to the changes to the spectrum-theme (specifically the addition of the discrete border styles) and the resulting changes are just to the prefixes and not to the overall generated structure, but its kinda hard to tell tbh...

packages/@react-spectrum/s2/src/Table.tsx
snowystinger: sigh

packages/@react-spectrum/table/stories/Table.stories.tsx
```diff
@@ -2053,6 +2053,29 @@ export const ColumnHeaderFocusRingTable = {
   }
 };
 
+export const Blah = {
```
snowystinger, line 2056: ? nice name ;)

LFDanLu: oh whoops lol, test story that I forgot to remove, thanks

packages/@react-spectrum/s2/src/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {useMediaQuery} from '@react-spectrum/utils';
+
+export function useIsMobileDevice(): boolean {
+  return useMediaQuery('(max-width: 640px)');
```
devongovett, line 16: Since this is being used to determine the scale, shouldn't it look for `@media not ((hover: hover) and (pointer: fine))`, which is what the style macro uses for that?

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1079 @@
...
+   */
+  onResizeEnd?: (widths: Map<Key, ColumnSize>) => void,
+  /** The current loading state of the table. */
+  loadingState?: LoadingState,
+  /** Handler that is called when more items should be loaded, e.g. while scrolling near the bottom. */
+  onLoadMore?: () => any
+}
+
+// TODO: Note that loadMore and loadingState are now on the Table instead of on the TableBody
+export interface TableProps extends Omit<RACTableProps, 'style' | 'disabledBehavior' | 'className' | 'onRowAction' | 'selectionBehavior' | 'onScroll' | 'onCellAction' | 'dragAndDropHooks'>, StyleProps, S2TableProps {
```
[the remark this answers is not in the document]
LFDanLu: good call, I'm not sure why it didn't complain in story where I make use of both height and width via styles hmm

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1079 @@
...
+      onResize={propsOnResize}
+      onResizeEnd={onResizeEnd}
+      onResizeStart={onResizeStart}
+      className={(UNSAFE_className || '') + mergeStyles(tableWrapper, styles)}
+      style={UNSAFE_style}>
+      <UNSTABLE_Virtualizer layout={layout}>
+        <InternalTableContext.Provider value={context}>
+          <RACTable
+            ref={scrollRef}
+            style={{WebkitTransform: 'translateZ(0)'}}
```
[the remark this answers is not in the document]
LFDanLu: this fixes the issue of checkboxes appearing above the scrollbar, I'll add a comment
image

packages/@react-spectrum/s2/src/Table.tsx
LFDanLu: Sure, I'll post in the room

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1079 @@
...
+  // the loader. Otherwise it is a static renderer and thus we can simply add the table loader after
+  // TODO: this assumes that the user isn't providing their children in some wrapper though and/or isn't doing a map of children
+  // (though I guess they wouldn't provide items then so the check for this is still valid in the latter case)...
+  if (typeof children === 'function' && items) {
+    renderer = (
+      <>
+        <Collection items={items}>
+          {children}
+        </Collection>
+        {loadingState === 'loadingMore' && [...items].length > 0 && loadMoreSpinner}
```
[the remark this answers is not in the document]
LFDanLu: the length check here is to prevent a situation where somehow the `loadingState = "loadingMore"' but no items are actually loaded aka the below:
image

For `loadingState = "loading"` we don't need these checks because the rendering of that circle/empty state is provided to `renderEmptyState` instead so RAC table handles rendering it only when the collection size is 0.

We could consider the "loadingMore" case a user error and get rid of the length check perhaps, what do you think? Otherwise I need some way of knowing the collection is empty...

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1079 @@
...
+  },
+  outlineOffset: -2,
+  outlineWidth: 2,
+  outlineColor: 'focus-ring',
+  borderRadius: size(6)
+} as const;
+
+function CellFocusRing(props: {isFocusVisible: boolean}) {
+  let {isFocusVisible} = props;
+  return <div role="presentation" style={{position: 'absolute', top: 0, left: 0, bottom: 0, right: 0}} className={style({...cellFocus})({isFocusVisible})} />;
```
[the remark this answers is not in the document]
LFDanLu: ah it previously was but the cellFocus gets reused in other places so I put this inline to ensure it didn't propagate elsewhere unexpectedly. I dunno why I didn't put it in the style macro in the className, whoops will do thanks!

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1096 @@
...
+    isHovered: 'block'
+  },
+  width: 12,
+  height: 'full',
+  position: 'absolute',
+  top: 0,
+  insetEnd: size(-6),
+  cursor: {
+    default: 'none',
+    resizableDirection: {
```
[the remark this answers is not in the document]
LFDanLu: oh thats right, completely slipped my mind, missed it since I didn't tag it with TODO lol. I'll add it to the followup ticket and see if we need to submit something to the icons team

packages/@react-spectrum/s2/src/Table.tsx
```diff
@@ -11,85 +11,1089 @@
...
+        break;
+      case 'sort-desc':
+        sort('descending');
+        break;
+      case 'resize':
+        setIsInResizeMode?.(true);
+        // TODO: Need to wait for the menu to fully transition out, otherwise mouse movements may hover the menu items and steal focus from the resizer,
+        // causing resizing to end prematurely due to blur. Open to ideas for how to handle this since it affects RAC. Ideally we'd freeze
+        // focus from being able to move if we are in resizing mode except if the user clicks away or moves focus with something other than hover
+        setTimeout(() => startResize(), 200);
```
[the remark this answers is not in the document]
LFDanLu: oh interesting, I can give that a shot

packages/@react-spectrum/s2/src/utils.ts
LFDanLu: I considered doing that but I figured we'd want to follow a similar logic style that we had in S1 in preparation for Trays aka we would check screen size. If we use the touch media query that the style macro uses, that would match large screen touch devices right?

packages/@react-spectrum/s2/src/Table.tsx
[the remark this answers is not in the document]
LFDanLu: Looks like adding that to Popover worked, pushed a update

packages/@react-spectrum/s2/src/Table.tsx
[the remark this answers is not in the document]
devongovett: Hmm yeah I'd consider that programmer error I think

packages/@react-spectrum/s2/src/utils.ts
devongovett: But this is being used to determine pixel sizes not whether to show a tray right? The style macro uses the media query I posted to do the same thing, ie scale up dimensions by 25%. Whether to show a tray is indeed based on screen size but that's different from touch scale, which might also occur on large screen devices like iPads or laptops.

LFDanLu: ahh right, makes sense, guess I conflated the two. I'll make a separate `useScale` that uses the media query and use it for table

packages/@react-spectrum/s2/src/Table.tsx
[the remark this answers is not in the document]
LFDanLu: filed an airtable ticket for those cursors

packages/react-aria-components/test/DatePicker.test.js
```diff
@@ -50,6 +50,10 @@ describe('DatePicker', () => {
     user = userEvent.setup({delay: null, pointerMap});
   });
   it('provides slots', async () => {
+    // Mock console.error for React Canary "Received the string `true` for the boolean attribute `inert`." warning
```
snowystinger, line 53: So these errors are going to appear in downstream projects tests as well

packages/react-aria-components/src/Popover.tsx
```diff
@@ -173,6 +173,8 @@ function PopoverInner({state, isExiting, UNSTABLE_portalContainer, ...props}: Po
         ref={ref}
         slot={props.slot || undefined}
         style={style}
+        // @ts-ignore - React < 19 compat
+        inert={(isEntering || isExiting) ? 'true' : undefined}
```
snowystinger, line 177: what is actually the issue?

devongovett: I think this is causing focus not to move into the popover when it opens now. e.g. try opening the column header menu with a keyboard - focus stays on the header instead of going to the menu item.

devongovett: Reverted the inert change and used pointer-events: none instead so it only applies to mouse during the exit animation.

packages/@react-aria/virtualizer/src/VirtualizerItem.tsx
devongovett: This isn't quite right because it is relative to the nearest scrolling or `overflow: hidden` container, which might not be the root. So really `parent` here should be either the immediate parent in case of `position: absolute`, or the nearest ancestor without `allowOverflow` if `position: sticky`. In the table case that doesn't happen because all of the ancestors have `allowOverflow` set, but it could in theory.

Alternatively, if we used `overflow: clip` instead of `overflow: hidden` the sticky elements would always be positioned relative to the root like you are assuming here. Looks like I tried that in https://github.com/adobe/react-spectrum/pull/6613 but reverted it due to some other side effects. We may be able to revisit that at some point.

For now I'd like to ship this as is since we've already tested it, and follow up later with a more correct solution.

packages/@react-spectrum/s2/src/Table.tsx
devongovett: Same problem in cardview. We will have to move it into RAC to handle this.

packages/@react-spectrum/s2/src/Table.tsx
devongovett: probably could only mount this element when focus is actually visible?
```suggestion
          {/* Note this is mainly for column's without a dropdown menu. If there is a dropdown menu, the button is styled to have a focus ring for simplicity
          (no need to juggle showing this focus ring if focus is on the menu button and not if it is on the resizer) */}
          {/* Separate absolutely positioned element because appyling the ring on the column directly via outline means the ring's required borderRadius will cause the bottom gray border to curve as well */}
          {isFocusVisible && <CellFocusRing />}
```

packages/@react-spectrum/s2/style/spectrum-theme.ts
```diff
@@ -603,7 +618,11 @@ export const style = createTheme({
     borderEndWidth: createRenamedProperty('borderInlineEndWidth', borderWidth),
     borderTopWidth: borderWidth,
     borderBottomWidth: borderWidth,
-    borderStyle: ['solid', 'dashed', 'dotted', 'double', 'hidden', 'none'] as const,
+    borderStartStyle: createRenamedProperty('borderInlineStartStyle', borderStyle),
+    borderEndStyle: createRenamedProperty('borderInlineEndStyle', borderStyle),
+    borderTopStyle: borderStyle,
+    borderBottomStyle: borderStyle,
+    borderStyle: borderStyle,
```
devongovett, line 625: This will break when both the shorthand and longhand properties are set. We would need to expand `borderStyle` to all 4 properties in all cases. But I don't think we need these individual properties since we already have individual width properties. I don't think we'll need borders with different styles on each side.

packages/@react-spectrum/s2/package.json
```diff
@@ -128,12 +128,16 @@
     "@react-aria/i18n": "^3.12.2",
     "@react-aria/interactions": "^3.22.2",
     "@react-aria/utils": "^3.25.2",
+    "@react-aria/visually-hidden": "^3.8.15",
```
devongovett, line 131: we should try to use the react-aria monopackage

reidbarber (approved):

dannify (approved):

PR #6779, opened 2024-07-26
S2 tabs

devongovett:


packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,289 @@
...
+      setStyle(styleObj);
+    }
+  }, [direction, setStyle, selectedTab, orientation]);
+
+  useLayoutEffect(() => {
+    onResize();
+  }, [onResize, state?.selectedKey, state, orientation]);
+  
+  return (
+    <div style={{...style}} className={selectedIndicator({isDisabled, orientation}) + ' ' + raw('transition: transform 130ms ease-in-out 0s, width 130ms ease-in-out 0s')} />
```
yihuiliao, line 263: There's no tokens for the animations when the selection indicator moves between the different tabs so I mostly guessed for these. Not sure if there's a way to add transitions without using raw?

yihuiliao:
There are no designs for vertical tabs so everything is just a guess of how it might look like based on V3. We can choose to pull this out and not support it for right now until they release some designs for it.

packages/@react-spectrum/s2/src/Tabs.tsx
snowystinger: shouldn't need raw
see https://github.com/adobe/react-spectrum/blob/1dcc8705115364a2c2ead2ececae8883dd6e9d07/packages/%40react-spectrum/s2/src/ProgressBar.tsx#L105

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,289 @@
...
+import {createContext, forwardRef, ReactNode, useCallback, useContext, useEffect, useRef, useState} from 'react';
+import {focusRing, getAllowedOverrides, StyleProps} from './style-utils' with {type: 'macro'};
+import {IconContext} from './Icon';
+import {raw} from '../style/style-macro' with {type: 'macro'};
+import {size, style} from '../style/spectrum-theme' with {type: 'macro'};
+import {useDOMRef} from '@react-spectrum/utils';
+import {useLayoutEffect} from '@react-aria/utils';
+import {useLocale} from '@react-aria/i18n';
 
+interface TabsProps extends Omit<AriaTabsProps, 'className' | 'style' | 'children'>, StyleProps { // is height a style prop we want to allow for tabs?
```
snowystinger, line 24: I think height for the entire thing makes sense, not for TabList or Tab though
Width should also be allowed I think

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,289 @@
...
+}
+
+const tab = style({
+  ...focusRing(),
+  display: 'inline-flex',
+  color: {
+    default: 'neutral-subdued',
+    isSelected: 'neutral',
+    isHovered: 'neutral-subdued',
+    isDisabled: 'disabled'
```
snowystinger, line 54: don't forget HCM

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,289 @@
...
+    orientation: {
+      vertical: 20
+    }
+  },
+  paddingStart: {
+    orientation: {
+      vertical: 12
+    }
+  },
+  borderColor: 'gray-400'
```
snowystinger, line 130: HCM

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,289 @@
...
+    onResize();
+  }, [onResize, state?.selectedKey, state, orientation]);
+  
+  return (
+    <div style={{...style}} className={selectedIndicator({isDisabled, orientation}) + ' ' + raw('transition: transform 130ms ease-in-out 0s, width 130ms ease-in-out 0s')} />
+  );
+}
+
+const tabs = style({
+  display: 'flex',
```
snowystinger, line 268: I think the default for flexShrink is 1, we should be explicit and set it to 0

packages/@react-spectrum/s2/stories/Tabs.stories.tsx
```diff
@@ -42,3 +42,41 @@ export const Example = (args: any) => (
...
+    <TabPanel id="Emp">
+      Alea jacta est.
+    </TabPanel>
+  </Tabs>
+);
+
+export const Icons = (args: any) => (
+  <Tabs {...args} styles={style({width: 320})}>
+    <TabList aria-label="History of Ancient Rome">
+      <Tab id="FoR"><Edit /></Tab>
```
snowystinger, line 68: haha, lets use a couple different icons

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,289 @@
...
+  isDisabled: boolean | undefined,
+  selectedTab: HTMLElement | undefined,
+  orientation?: Orientation
+}
+
+const selectedIndicator = style({
+  position: 'absolute',
+  color: {
+    default: 'neutral',
+    isDisabled: 'disabled'
```
snowystinger, line 198: HCM

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,289 @@
...
+    }
+    selectedKey = collection.getFirstKey();
+    
+    let index = 0;
+    while (index < collection.size) {
+      if (selectedKey && !disabledKeys.has(selectedKey)) {
+        return false;
+      }
+
+      // @ts-ignore
```
snowystinger, line 178: why the ts-ignore?

packages/@react-spectrum/s2/src/Tabs.tsx
yihuiliao: ah right, i always forget. thanks for reminding me

packages/@react-spectrum/s2/src/Tabs.tsx
yihuiliao: originally it was because i thought it would be solved with fixing ts-strict but now that i'm looking at it more i think i need to update the logic

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+      compact: 32,
+      regular: 48
+    }
+  },
+  alignItems: 'center',
+  position: 'relative'
+});
+
+const icon = style({
+  paddingEnd: 4,
```
snowystinger, line 76: `text-to-visual`?

any reason for extra padding here? can it be the `columnGap` on `tab`? like Button?

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
 
+interface TabsProps extends Omit<AriaTabsProps, 'className' | 'style' | 'children'>, UnsafeStyles {
+  /** Spectrum-defined styles, returned by the `style()` macro. */
+  styles?: StylesPropWithHeight,
+  /** The content to display in the tabs. */
+  children?: ReactNode,
+  density?: 'compact' | 'regular'
+}
+
+interface TabProps extends Omit<AriaTabProps, 'children' | 'style' | 'className'> {
```
snowystinger, line 32: missing StyleProps on Tab, TabList, and TabPanel
See https://react-spectrum.adobe.com/react-spectrum/Tabs.html#props for the kinds of style props we've previously allowed

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+import {useDOMRef} from '@react-spectrum/utils';
+import {useLayoutEffect} from '@react-aria/utils';
+import {useLocale} from '@react-aria/i18n';
 
+interface TabsProps extends Omit<AriaTabsProps, 'className' | 'style' | 'children'>, UnsafeStyles {
+  /** Spectrum-defined styles, returned by the `style()` macro. */
+  styles?: StylesPropWithHeight,
+  /** The content to display in the tabs. */
+  children?: ReactNode,
+  density?: 'compact' | 'regular'
```
snowystinger, line 29:
```suggestion
  /** The amount of space between the tabs. */
  density?: 'compact' | 'regular'
```

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+
+export function TabPanel(props: TabPanelProps) {
+  return (
+    <AriaTabPanel {...props} className={style({marginTop: 4, color: 'neutral'})} />
+  );
+}
+
+const tab = style({
+  ...focusRing(),
+  display: 'inline-flex',
```
snowystinger, line 52: why inline? in our older implementation they were block elements which were typically flex children

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+export function Tab(props: TabProps) {
+  let {density} = useContext(TabsInternalContext);
+
+  return (
+    <RACTab 
+      {...props}
+      className={renderProps => tab({...renderProps, density})}>
+      <Provider
+        values={[
+          [IconContext, {styles: icon}]
```
snowystinger, line 92: missing render centerBaseline

probably model all of this after button, for instance, ensure ordering of the icons and text as well
and you'll want flexShrink: 0 on the icon

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+        ]}>
+        {props.children}
+      </Provider>
+    </RACTab>
+  );
+}
+
+const tablist = style({
+  display: 'flex',
+  height: {
```
snowystinger, line 102: should be able to leave the height up to the content?

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+    orientation: {
+      horizontal: {
+        density: {
+          compact: 32,
+          regular: 48
+        }
+      }
+    }
+  },
+  gap: {
```
snowystinger, line 112: this seems like a very large gap between tabs?

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+  },
+});
+
+export function TabList<T extends object>(props: TabListProps<T>) {
+  let {density, isDisabled, disabledKeys, orientation = 'horizontal'} = useContext(TabsInternalContext);
+  let state = useContext(TabListStateContext);
+  let [selectedTab, setSelectedTab] = useState<HTMLElement | undefined>(undefined);
+  let tablistRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
```
snowystinger, line 145: I think this should be added to `useTagGroup, I see previously we handled it in RSP Tabs.
Looks like it's dependent on rendering, however, we could probably make use of the collection/selectionManager instead of querying the dom. I have no idea why we were doing that.

I'm ok for it to be followup as well

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+        ref={tablistRef}
+        className={renderProps => tablist({...renderProps, density})} />
+      {orientation === 'horizontal' && 
+        <TabLine disabledKeys={disabledKeys} isDisabled={isDisabled} selectedTab={selectedTab} orientation={orientation} />}
+    </div>
+  );
+}
+
+function isAllTabsDisabled<T>(collection: Collection<Node<T>> | null, disabledKeys: Set<Key>) {
+  let selectedKey: Key | null = null;
```
snowystinger, line 171: I got confused and thought this was about selection
```suggestion
  let testKey: Key | null = null;
```

it'd be good to memo this based on the collection and disabledKeys

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+  if (collection) {
```
snowystinger, line 172: can do a quick and easy optimization if the disabledKeys size is 0
or if its size is smaller than the collection size

might also want to check collection size > 0 instead of just truthiness of collection

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+      setStyle(styleObj);
+    }
+  }, [direction, setStyle, selectedTab, orientation]);
+
+  useLayoutEffect(() => {
+    onResize();
+  }, [onResize, state?.selectedKey, state, orientation]);
+  
+  return (
+    <div style={{...style}} className={selectedIndicator({isDisabled, orientation}) + ' ' + selectionAnimation} />
```
snowystinger, line 275: use mergeStyles?

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+      } else {
+        styleObj.height = `${selectedTab.offsetHeight}px`;
+      }
+      setStyle(styleObj);
+    }
+  }, [direction, setStyle, selectedTab, orientation]);
+
+  useLayoutEffect(() => {
+    onResize();
+  }, [onResize, state?.selectedKey, state, orientation]);
```
snowystinger, line 272: we'll need direction in this list
and I don't think we need the state?

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+        values={[ 
+          [TabsInternalContext, {density: props.density || 'regular', isDisabled: props.isDisabled, disabledKeys: props.disabledKeys, orientation: props.orientation}]
+        ]}>
+        {props.children}
+      </Provider>
+    </RACTabs>
+  );
 }
+
+const _Tabs = forwardRef(Tabs);
```
snowystinger, line 310:
```suggestion
/**
 * Tabs organize content into multiple sections and allow users to navigate between them. The content under the set of tabs should be related and form a coherent unit.
 */
const _Tabs = forwardRef(Tabs);
```

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+  position: 'absolute',
+  color: {
+    default: 'neutral',
+    isDisabled: 'disabled',
+    forcedColors: {
+      default: 'Highlight',
+      isDisabled: 'GrayText'
+    }
+  },
+  height: size(2),
```
snowystinger, line 207: Shouldn't change with scale or anything
```suggestion
  height: '[2px]',
```

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+  bottom: {
+    orientation: {
+      horizontal: 0
+    }
+  },
+  borderStyle: 'solid',
+  borderRadius: 'lg',
```
snowystinger, line 214:
```suggestion
  borderRadius: 'full',
```

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,302 @@
...
+      isDisabled: 'GrayText'
+    }
+  },
+  height: size(2),
+  bottom: {
+    orientation: {
+      horizontal: 0
+    }
+  },
+  borderStyle: 'solid',
```
snowystinger, line 213: instead of a border, use no border and a background
you should be able to model it after `Divider`

otherwise at zoom you can get funny looking things like this:
Screenshot 2024-08-06 at 12 52 39 PM

snowystinger (commented):
Don't know if you saw this issue at all? https://github.com/adobe/react-spectrum/blob/main/packages/%40adobe/spectrum-css-temp/components/tabs/index.css#L83

Set cursor on tabs so we don't get the text insertion indicator

I think `focusRing` has a bug, I think its `outlineWidth` and `outlineOffset` should both be `[2px]`

packages/@react-spectrum/s2/src/Tabs.tsx
yihuiliao: i thought about that but it's possible that disabledKeys has a size of 0 but `isDisabled` is set on each individual Tab in which case we wouldn't want to return early. or alternatively, there's a mix of disabledKeys and `isDisabled` in which case checking if `disabledKeys.size` < `collection.size` wouldn't work either. i know it's a weird case so maybe it's not something we care to check for

packages/@react-spectrum/s2/src/Tabs.tsx
yihuiliao: maybe something to bring to design? the spacing between tabs is supposed to be 32px for the regular density and 24px for the compact density.

packages/@react-spectrum/s2/src/Tabs.tsx
yihuiliao: i ended up removing the height from here since i actually set the height on the tabs themselves. the tokens have a set height on the tabs depending on the density so not sure if we just want to leave it up to the content? also, setting a height is necessary for the correct positioning of the selection indicator.

packages/@react-spectrum/s2/src/Tabs.tsx
yihuiliao: to be honest, i cannot remember why i added padding here...

packages/@react-spectrum/s2/src/Tabs.tsx
snowystinger: ah, i forgot it could be `isDisabled` directly on the item. Womp womp. Thanks for clarifying

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -154,32 +156,34 @@ export function TabList<T extends object>(props: TabListProps<T>) {
   }, [tablistRef, state?.selectedItem]);
 
   return (
-    <div className={style({position: 'relative'})}>
+    <div
+      style={props.UNSAFE_style}
+      className={(props.UNSAFE_className || '') + mergeStyles(style({position: 'relative'}), props.styles)}>
```
snowystinger, line 161: probably need to call getAllowedOverrides for the style as well
```suggestion
      className={(props.UNSAFE_className || '') + style({position: 'relative'}, props.styles)}>
```

snowystinger: should do this with the other components as well

snowystinger (commented):
I think we should provide a width/height to the story, otherwise it shifts around a bunch in orientation vertical

yihuiliao:
> Don't know if you saw this issue at all? https://github.com/adobe/react-spectrum/blob/main/packages/%40adobe/spectrum-css-temp/components/tabs/index.css#L83 I don't see it, but might need a bit of chasing down to remember what it was. Or see if removing it in v3 still causes it.

so i tried reproducing on safari and it doesn't seem to happen anymore? didn't notice any difference between having it and removing it so maybe it's been fixed in newer version of safari since this line was added 3 years ago. 

> I think `focusRing` has a bug, I think its `outlineWidth` and `outlineOffset` should both be `[2px]`

oh are you saying that rather than focus ring being `outlineWidth: 2` is should be `outlineWidth: '[2px]'`

packages/@react-spectrum/s2/stories/Tabs.stories.tsx
```diff
@@ -25,9 +27,9 @@ const meta: Meta<typeof Tabs> = {
 export default meta;
 
 export const Example = (args: any) => (
-  <Tabs {...args}>
+  <Tabs {...args} styles={style({maxWidth: '[450px]', height: 144})}>
```
snowystinger, line 30:
```suggestion
  <Tabs {...args} styles={style({width: '[450px]', height: 144})}>
```
This way changing tabs won't change the size overall

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,320 @@
...
+export function TabPanel(props: TabPanelProps) {
+  return (
+    <AriaTabPanel 
+      {...props}
+      style={props.UNSAFE_style}
+      className={(props.UNSAFE_className || '') + style({marginTop: 4, color: 'gray-800'}, getAllowedOverrides())(null, props.styles)} />
+  );
+}
+
+const tab = style({
```
snowystinger, line 56: when a tab is icon-only, it appears there should be some padding in the X direction, I noticed that the selection indicator sticks out over the sides. I don't see a token for it, so you may need to ask for that

snowystinger: Screenshot 2024-08-07 at 12 06 27 PM

yihuiliao: interesting, yeah i definitely looks like there should be padding. the tokens seem to say otherwise cause there's a section under `Spacing (edge to edge)` (not sure if i want to post a pic of the tokens here) where it's defined as none but that might only be for tabs that aren't icon only so i've reached out to the spectrum team for clarification

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,320 @@
...
+    isSelected: 'neutral',
+    isHovered: 'neutral-subdued',
+    isDisabled: 'disabled',
+    forcedColors: {
+      isSelected: 'Highlight',
+      isDisabled: 'GrayText'
+    }
+  },
+  borderRadius: 'sm',
+  gap: size(6),
```
snowystinger, line 70: text-to-visual?

snowystinger (dismissed):

snowystinger:
Happy to have anything else be followup, I think we can get started testing this

snowystinger (dismissed):

akyllus (approved):

akyllus (approved):

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,342 @@
...
+  flexBasis: '[0%]'
+});
+
+export function TabList<T extends object>(props: TabListProps<T>) {
+  let {density, isDisabled, disabledKeys, orientation} = useContext(TabsInternalContext);
+  let state = useContext(TabListStateContext);
+  let [selectedTab, setSelectedTab] = useState<HTMLElement | undefined>(undefined);
+  let tablistRef = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
```
yihuiliao, line 164: known issue: when you set the orientation to vertical and toggle between the two densities, the selection indicator won't resize properly. 

i took a look at it with rob and we couldn't get to a solution and we also found that it's an issue in v3. it just was never noticeable because we didn't have controls on the stories. 

we tried a couple of things. mainly moving the resize code out of `<TabLine/>` and into `<TabList/>`, updating some the dependencies, using refs instead of a state, removing one of the `<TabLine/>` so that there was only one, etc but none of these fixed the issue.

snowystinger (approved):

packages/@react-spectrum/s2/src/Tabs.tsx
yihuiliao: link to the ticket: https://github.com/orgs/adobe/projects/19/views/32?sliceBy%5Bvalue%5D=%F0%9F%8C%88&pane=issue&itemId=74700408

devongovett: Maybe eventually we could implement the [FLIP animation technique](https://css-tricks.com/animating-layouts-with-the-flip-technique/). This would render the line inside the selected tab so it automatically has the right layout for the current tab and we don't need to do any manual syncing. When switching tabs, the tab line inside the newly selected tab gets offset such that it appears visually like the line that was previously inside the old selected tab, and then transitions to its final position. [This video](https://www.youtube.com/watch?v=kep_Iaxuzy0) shows the overall technique. This is easier to implement with a library like Framer Motion but we could probably make it work.

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,342 @@
...
+      setStyle(styleObj);
+    }
+  }, [direction, setStyle, selectedTab, orientation]);
+
+  useLayoutEffect(() => {
+    onResize();
+  }, [onResize, state?.selectedItem?.key, direction, orientation, density]);
+
+  return (
+    <div style={{...style}} className={mergeStyles(selectedIndicator({isDisabled, orientation}), selectionAnimation)} />
```
devongovett, line 304: Could the `selectionAnimation` style be merged into the definition of `selectedIndicator`? Seems like that would avoid needing to merge these at runtime

devongovett (commented):
I can't approve because I opened the PR but looks good. Couple tiny things

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,342 @@
...
+  height: {
+    density: {
+      compact: 32,
+      regular: 48
+    }
+  },
+  alignItems: 'center',
+  position: 'relative',
+  cursor: 'default',
+  flexShrink: 0
```
devongovett, line 94: Add `transition: 'default'` here as well for the color change on hover.

packages/@react-spectrum/s2/src/Tabs.tsx
```diff
@@ -10,9 +10,340 @@
...
+      setStyle(styleObj);
+    }
+  }, [direction, setStyle, selectedTab, orientation]);
+
+  useLayoutEffect(() => {
+    onResize();
+  }, [onResize, state?.selectedItem?.key, direction, orientation, density]);
+
+  return (
+    <div style={{...style}} className={mergeStyles(selectedIndicator({isDisabled, orientation}))} />
```
devongovett, line 302: probably don't need `mergeStyles` at all now right?

devongovett:
✅ (since I cannot approve)

snowystinger (approved):

reidbarber (approved):

PR #6781, opened 2024-07-26
S2 AvatarGroup

devongovett:


packages/@react-spectrum/s2/src/Avatar.tsx
```diff
@@ -22,20 +22,33 @@ export interface AvatarProps extends StyleProps, DOMProps {
   /** Text description of the avatar. */
   alt?: string,
   /** The image URL for the avatar. */
-  src?: string
+  src?: string,
+  /** The stroke color for the avatar. */
+  strokeColor?: 'white' | 'black'
```
snowystinger, line 27: is this really staticColor?
Though seems like there's actually a whole bunch of options for the strokeColor. Possibly every "layer" color is a choice?
Screenshot 2024-08-06 at 10 45 21 AM

packages/@react-spectrum/s2/src/Avatar.tsx
```diff
@@ -60,7 +74,7 @@ function Avatar(props: AvatarProps, ref: DOMRef<HTMLImageElement>) {
       ref={domRef}
       alt={alt}
       style={UNSAFE_style}
-      className={UNSAFE_className + imageStyles(null, props.styles)}
+      className={UNSAFE_className + imageStyles({strokeColor}, props.styles)}
```
snowystinger, line 77:
```suggestion
      className={(UNSAFE_className ?? '') + imageStyles({strokeColor}, props.styles)}
```

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+});
+
+const container = style({display: 'flex', alignItems: 'center'}, getAllowedOverrides());
+
+export function AvatarGroup(props: AvatarGroupProps) {
+  let {children, label, size = 'M'} = props;
+
+  return (
+    <div className={container(null, props.styles)}>
```
snowystinger, line 82: need to ass UNSAFE classname and style along with a spread of whatever other props are allowed. (labelling?, id, data-*)

snowystinger: role group?
focusable?

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,93 @@
...
+      XL: 'ui-xl',
+      XXL: 'ui-2xl',
+      XXXL: 'ui-3xl'
+    }
+  }
+});
+
+const container = style({display: 'flex', alignItems: 'center'}, getAllowedOverrides());
+
+export function AvatarGroup(props: AvatarGroupProps) {
```
snowystinger, line 78: forward ref?

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+function AvatarGroup(props: AvatarGroupProps) {
+  let {children, label, size = 'M', strokeColor = 'base', styles, UNSAFE_style, UNSAFE_className, ...otherProps} = props;
+  let labelId = useId();
+
+  return (
+    <ul
+      aria-labelledby={labelId}
+      {...filterDOMProps(otherProps)}
+      tabIndex={0}
```
jluyau, line 130: was getting a lint error for this because there's no handler on the list. is that ok?

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+      role="group"
```
snowystinger, line 131: why are we using an implicit list and converting it to a group? should we just be using divs? or should their be a wrapper element that is the group, and inside is a list and a span label thing?

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+export {_AvatarGroup as AvatarGroup};
+
+export function AvatarGroupItem(props: AvatarGroupItemProps) {
+  let {
+    alt = '',
+    src
+  } = props;
+  let {styles, strokeColor} = useContext(AvatarGroupContext);
+  return (
+    <li className={style({display: 'flex'})}>
```
snowystinger, line 156: why display flex?

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+  size: {
+    size: {
+      S: 20,
+      M: 24,
+      L: 28,
+      XL: 32,
+      XXL: 36,
+      XXXL: 40
+    }
+  }
```
snowystinger, line 74: should probably set
`flexShrink: 0` and `flexGrow: 0` on avatars

snowystinger (commented):
should the '145 members' trailing label truncate?

packages/@react-spectrum/s2/src/AvatarGroup.tsx
jluyau: without flex, the list item element looks like this:
Screenshot 2024-08-09 at 2 10 05 PM

definitely felt weird using it to fix the styling, so definitely open to another solution

packages/@react-spectrum/s2/src/AvatarGroup.tsx
jluyau: your previous comment mentioned that it should probably be focusable, so i added the tab index for that. was there something else i should've done?

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+      className={(UNSAFE_className ?? '') + container(null, styles)}
+      style={UNSAFE_style}>
+      {
+        Children.map(children, (child, index) => (
+          <AvatarGroupContext.Provider value={{styles: avatar({size, isFirst: index === 0}), strokeColor}}>
+            {child}
+          </AvatarGroupContext.Provider>
+        ))
+      }
+      {label && <span id={labelId} className={text({size})}>{label}</span>}
```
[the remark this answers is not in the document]
jluyau: i think devon had brought it up in a past Spectrum sync, but i haven't heard if they gave any feedback on how the label should relate to the displayed avatars/whole component

packages/@react-spectrum/s2/src/AvatarGroup.tsx
jluyau: i had an issue with the dimensions of the avatars in the `<li>` element, but since i've removed it, it works fine without the flex

packages/@react-spectrum/s2/src/AvatarGroup.tsx
snowystinger: ah, interesting https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role
This is i think pretty specific to menu implementations

I don't know if it has further implications, would be best to consult with @majornista about this

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -77,6 +76,9 @@ const avatar = style({
 const text = style({
   marginStart: 8,
   fontFamily: 'sans',
+  textOverflow: 'ellipsis',
```
snowystinger, line 79: you should just be able to use `truncate` to handle this

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -114,32 +116,28 @@ const AvatarGroupContext = createContext<AvatarGroupContextProps>({});
...
-          </AvatarGroupContext.Provider>
-        ))
-      }
-      {label && <span id={labelId} className={text({size})}>{label}</span>}
-    </ul>
+    <AvatarGroupContext.Provider value={{styles: avatar({size}), strokeColor}}>
+      <div
+        id={groupId}
+        aria-labelledby={`${groupId} ${labelId}`}
+        aria-label="online members"
```
snowystinger, line 132: this should be a prop, we don't know if it'll always be online members

jluyau: ah whoops, used this to test and accidentally committed it. will remove

majornista: If the containing group with `id={groupID}` does not have an `aria-label`, or some default label, like `Avatar Group`, we should not include `groupId` in the `aria-labelledby`, otherwise the group will include the accessibility name for all the descendants, including the label in the accessibility name, which may lead to redundancy and verbosity.

majornista (changes requested):

packages/@react-spectrum/s2/src/AvatarGroup.tsx
snowystinger: gah, i see how that would be confusing. Sorry about that. Spectrum has been conflicting on this and it's now coming through me. Either way is probably ok for the moment, we'll get it sorted out.

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+const AvatarGroupContext = createContext<AvatarGroupContextProps>({});
+
+const container = style({
+  display: 'flex',
+  alignItems: 'center',
+  margin: 0
+}, getAllowedOverrides());
+
+function AvatarGroup(props: AvatarGroupProps) {
```
snowystinger, line 120: we're not passing through the ref, something like so
```suggestion
function AvatarGroup(props: AvatarGroupProps, ref: ForwardedRef<HTMLElement>) {
```

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+import {createContext, forwardRef, ReactNode, useContext} from 'react';
+import {DOMProps} from '@react-types/shared';
+import {filterDOMProps, useId} from '@react-aria/utils';
+import {getAllowedOverrides, StyleProps} from './style-utils' with {type: 'macro'};
+import {size, style} from '../style/spectrum-theme' with {type: 'macro'};
+import type {StyleString} from '../style/types';
+
+
+interface AvatarGroupProps extends StyleProps, DOMProps {
```
snowystinger, line 21:
```suggestion
interface AvatarGroupProps extends StyleProps, DOMProps, AriaLabelingProps {
```

packages/@react-spectrum/s2/src/AvatarGroup.tsx
jluyau: whoops did the first forward but forgot to add/connect the ref 😢

devongovett:
Pushed 3 changes:

1. Moved to using `Avatar` directly instead of `AvatarGroupItem`. Also moved the stroke into Avatar itself (via an `isOverBackground` prop), since this will also be needed in other places such as Cards.
2. Moved to pixel-based sizing rather than t-shirt sizing (since there are so many size options). Also added back the `size` prop to `Avatar` instead of using macros for sizing, which enables us to adjust the stroke width depending on the size. Updated codemod for this change as well.
3. Made the stroke color automatic. This is done by making all container components with a background color set the `--s2-container-bg` variable. Then the stroke can automatically match the nearest container.

snowystinger (approved):

reidbarber (approved):
LGTM

reidbarber (dismissed):

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -0,0 +1,98 @@
...
+import {ContextValue} from 'react-aria-components';
+import {createContext, CSSProperties, forwardRef, ReactNode} from 'react';
+import {DOMProps, DOMRef, DOMRefValue} from '@react-types/shared';
+import {filterDOMProps, useId} from '@react-aria/utils';
+import {getAllowedOverrides, StylesPropWithoutWidth, UnsafeStyles} from './style-utils' with {type: 'macro'};
+import {style} from '../style/spectrum-theme' with {type: 'macro'};
+import {useDOMRef} from '@react-spectrum/utils';
+import {useSpectrumContextProps} from './useSpectrumContextProps';
+
+export interface AvatarGroupProps extends UnsafeStyles, DOMProps {
```
LFDanLu, line 23: Noticed that the aria labeling props were added then removed, any particular reason why?

devongovett: hmm looks like they should, they do get used. I can add them back

snowystinger (approved):

dannify (approved):

PR #6792, opened 2024-07-29
Yarn plugin to detect multiple versions of packages

devongovett:
This is a plugin for modern Yarn versions that detects when there are multiple versions of React Spectrum/React Aria packages installed, which often causes issues due to context and other singletons. When multiple versions of packages are found, the Yarn install will fail and we print out a list of packages with duplicates. Running `yarn dedupe` often fixes this, but sometimes it can be more complicated if there are pinned versions.

The plugin can be installed into projects by running:

```
yarn plugin import https://raw.githubusercontent.com/adobe/react-spectrum/main/lib/yarn-plugin-rsp-duplicates.js
```

See https://yarnpkg.com/cli/plugin/import


snowystinger (approved):

LFDanLu (approved):
Verified that it caught a dupe package that I forced in the test app:
```
➤ YN0000: └ Completed in 0s 255ms
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed in 0s 233ms
➤ YN0001: · Found multiple installed versions of React Spectrum packages: 

    @react-spectrum/combobox: 3.12.4, 3.13.1

 You may be able to fix this by running `yarn dedupe`.
```

PR #6812, opened 2024-07-31
Add quiet styling to S2 Picker

yihuiliao:
Closes 

## ✅ Pull Request Checklist:


packages/@react-spectrum/s2/src/style-utils.ts
```diff
@@ -113,7 +113,8 @@ export const fieldInput = () => ({
   contain: {
     // Only apply size containment if contain-intrinsic-width is supported.
     // In older browsers, this will fall back to the default browser intrinsic width.
-    '@supports (contain-intrinsic-width: 1px)': 'inline-size'
+    '@supports (contain-intrinsic-width: 1px)': 'inline-size',
+    isQuiet: 'none'
```
yihuiliao, line 117: honestly not too sure about this but without it, there's too much spacing between the value text and the dropdown chevron which doesn't match the figma designs

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -195,7 +225,8 @@ function Picker<T extends object>(props: PickerProps<T>, ref: FocusableRef<HTMLB
     necessityIndicator,
     UNSAFE_className = '',
     UNSAFE_style,
-    placeholder = 'Select an option...',
+    placeholder = 'Select...',
```
yihuiliao, line 228: based on https://github.com/adobe/react-spectrum/pull/6796

can change it back if we want "Select an option" for S2

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -67,7 +67,12 @@ export interface PickerStyleProps {
    *
    * @default 'M'
    */
-  size?: 'S' | 'M' | 'L' | 'XL'
+  size?: 'S' | 'M' | 'L' | 'XL',
+  /**
+   * Whether the picker should be displayed with a quiet style.
+   *
```
snowystinger, line 73:
```suggestion
```

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -101,6 +106,11 @@ interface PickerButtonProps extends PickerStyleProps, ButtonRenderProps {}
 const inputButton = style<PickerButtonProps | AriaSelectRenderProps>({
   ...focusRing(),
   ...fieldInput(),
+  outlineStyle: {
+    default: 'none',
+    isFocusVisible: 'solid',
```
snowystinger, line 111: did we just forget this earlier?

packages/@react-spectrum/s2/src/Picker.tsx
snowystinger: Should be the short version, thanks

packages/@react-spectrum/s2/src/Picker.tsx
yihuiliao: honestly im not too sure why this was taken out before. there was a comment before that said "The focus ring shows up after menu close if default and isFocusVisible from focusRing() aren't added to this definition" but i haven't been able to reproduce unless im not understanding it correctly.

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -111,19 +118,39 @@ const inputButton = style<PickerButtonProps | AriaSelectRenderProps>({
...
   color: {
     default: 'neutral',
     isDisabled: 'disabled'
   }
 });
 
+const quietFocusLine = style({
+  width: 'full',
+  // Use pixels since we are emulating a border.
+  height: '[2px]',
```
LFDanLu, line 144: should this really not be `size(2)`? Bit surprised that this focus line is an exception unlike our other focus rings but perhaps I'm misunderstanding when `size` should be used

LFDanLu (commented):
Not sure if it is expected or not, but I noticed that the label for a quiet Picker sizes itself with respect to the length of the text content in the picker, whereas the label in v3 seems to prefer having the label as a single line regardless of the picker's selected text: https://reactspectrum.blob.core.windows.net/reactspectrum/1dcc8705115364a2c2ead2ececae8883dd6e9d07/storybook/index.html?path=/story/picker--quiet&providerSwitcher-express=false. Not sure if this is important if we are only adding quiet for Tabs (if that is even the case, lemme know if I'm misremembering)

reidbarber (dismissed):
One minor nit: there might be slightly too much space between the invalid icon and the value in quiet

snowystinger:
Screenshot 2024-08-02 at 9 40 08 AM

custom width menu isn't quite wide enough, not sure what's going on there

snowystinger:
Screenshot 2024-08-02 at 9 44 36 AM

popover alignment seems off when setting `align="end"` it's too far left I think, might need that negative margin adjustment? or maybe just need to get rid of the current one?

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -67,7 +67,11 @@ export interface PickerStyleProps {
    *
    * @default 'M'
    */
-  size?: 'S' | 'M' | 'L' | 'XL'
+  size?: 'S' | 'M' | 'L' | 'XL',
+  /**
+   * Whether the picker should be displayed with a quiet style.
+   */
+  isQuiet?: boolean
```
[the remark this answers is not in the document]
yihuiliao: yeah it's just for Tabs so i'd be okay with making it private

packages/@react-spectrum/s2/src/Picker.tsx
yihuiliao: oh hmm im not sure...i just copied it over from another pr and im not sure if we used size() then. i'll take a closer look

packages/@react-spectrum/s2/src/Field.tsx
```diff
@@ -80,9 +81,10 @@ function FieldLabel(props: FieldLabelProps, ref: DOMRef<HTMLLabelElement>) {
         contain: {
           labelPosition: {
             top: 'inline-size'
-          }
+          },
```
yihuiliao, line 84: adding this allows the label be on a single line regardless of the picker's selected text

packages/@react-spectrum/s2/src/Field.tsx
```diff
@@ -80,9 +81,10 @@ function FieldLabel(props: FieldLabelProps, ref: DOMRef<HTMLLabelElement>) {
...
+          isQuiet: 'none'
```
yihuiliao, line 85: adding this allows the label be on a single line regardless of the picker's selected text

yihuiliao:
> One minor nit: there might be slightly too much space between the invalid icon and the value in quiet

hmm yeah, it seems like it's because we add `marginStart` to the `FieldErrorIcon`. honestly, i'm not sure why we added this but after some digging it's been here since we created the component
https://github.com/adobe/react-spectrum/blob/617ad2ed3c3ee46715ca505a315cbc2b945b1f4e/packages/%40react-spectrum/s2/src/Field.tsx#L329-L330

yihuiliao:
> Screenshot 2024-08-02 at 9 40 08 AM
> 
> custom width menu isn't quite wide enough, not sure what's going on there

actually, what is the expected behavior for the width of a popover when a custom width is applied?  

after i pushed the latest code, the width of the popover is not the same as the applied custom width because i had to add "fit-content" to the width of the button for quiet pickers to prevent the focus ring from extending past the dropdown chevron. as a result, the popover will always be the width of the trigger/button (unless the width of the trigger is less than the min width) 

i tried it out in v3 and seems like if you apply any custom width to a quiet picker, the width of the popover will always be whatever the width of the trigger is so S2 now matches the behavior of V3 but is this what we want?

yihuiliao:
> popover alignment seems off when setting `align="end"` it's too far left I think, might need that negative margin adjustment? or maybe just need to get rid of the current one?

it looks fine to me? or maybe im misunderstanding something but it looks to me like the popover is aligning to the end of the trigger. maybe my eyes just aren't picking it up...

LFDanLu:
> hmm yeah, it seems like it's because we add marginStart to the FieldErrorIcon. honestly, i'm not sure why we added this but after some digging it's been here since we created the component. maybe someone else knows?

Isn't that from the designs for spacing between text value to icon? 

> i tried it out in v3 and seems like if you apply any custom width to a quiet picker, the width of the popover will always be whatever the width of the trigger is, so S2 now matches the behavior of V3 but is this what we want?

Thats probably fine for now IMO, especially since it will just be for tabs for now and thus just always the min popover width

snowystinger:
Screenshot 2024-08-05 at 2 59 44 PM

I mean the align start actually doesn't align with the start of the input, instead it's pushed further left. But align end is exactly lined up, it isn't pushed further right.

packages/@react-spectrum/s2/src/Field.tsx
snowystinger: I guess we'll have to figure out if we want this, might interfere with form layouts

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -111,19 +119,42 @@ const inputButton = style<PickerButtonProps | AriaSelectRenderProps>({
...
   transition: 'default',
-  columnGap: 'text-to-control',
-  paddingX: 'edge-to-text',
+  columnGap: {
+    default: 'text-to-control',
+    isQuiet: 'text-to-visual'
+  },
+  paddingX: {
+    default: 'edge-to-text',
+    isQuiet: centerPadding()
```
snowystinger, line 128: isn't this for vertical centering/spacing?

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -111,19 +119,42 @@ const inputButton = style<PickerButtonProps | AriaSelectRenderProps>({
...
+  },
+  maxWidth: {
+    isQuiet: 'max'
   }
 });
 
+const quietFocusLine = style({
+  width: 'full',
+  // Use pixels since we are emulating a border.
+  height: size(2),
```
snowystinger, line 148: using size will adjust to font size and will get bigger for touch, if this was copied from somewhere else, it probably needs to be adjusted there as well
```suggestion
  height: `[2px]`,
```

yihuiliao: ohhh lol it was [2px] before and then i changed it size(2) but i see okay i'll change it back

LFDanLu: oh thats my bad, I thought the focus rings had adaptive sizing via the outline values but I see now that they just convert the `2` into the token value which is static

packages/@react-spectrum/s2/src/Field.tsx
yihuiliao: yeah im open to either. for tabs, there isn't going to be a label anyway so this won't matter too much. i added this to mainly to match the v3 behavior. i don't really think isQuiet should be used in forms given that the other form components don't have isQuiet styles but if you're curious, this is what it looks like having an isQuiet Picker in a form
Screenshot 2024-08-06 at 11 03 17 AM
Screenshot 2024-08-06 at 11 03 25 AM

for the second image, i had to apply a width to the form, otherwise, the text would overflow

packages/@react-spectrum/s2/src/Picker.tsx
yihuiliao: hm yeah that was copied over from when we removed isQuiet so not sure why we chose `centerPadding`. i'll just remove it and set the padding to 0 since we don't need any

snowystinger: apologies, that might've been me
thank you!

snowystinger:
> I mean the align start actually doesn't align with the start of the input, instead it's pushed further left. But align end is exactly lined up, it isn't pushed further right.

Ah I see this was actually the behavior in v3 as well. I just checked the storybook for it. I think it's ok to leave it how you had it because I think we want that slight offset at the start. We can handle the align end case later. Would you mind opening a discussion with Spectrum about it?

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -299,14 +337,25 @@ function Picker<T extends object>(props: PickerProps<T>, ref: FocusableRef<HTMLB
...
-                width: '[var(--trigger-width)]'
-              })}>
+                width: {
+                  default: '[var(--trigger-width)]',
+                  isQuiet: '[calc(var(--trigger-width) + var(--quietOffset))]'
+                },
+                '--quietOffset': {
+                  type: 'width',
+                  value: 24
+                }
```
snowystinger, line 357: can maybe simplify with?
```suggestion
                  isQuiet: '[calc(var(--trigger-width) + (-1 * self(marginStart))]'
                }
```

reidbarber (approved):

snowystinger (approved):

akyllus (approved):

PR #6831, opened 2024-08-06
Update calculatePosition to handle overlays that have margins on them

LFDanLu:
Found in testing

This is the case for s2 tooltip where a margin surrounds the tooltip making the dom node larger than the actual visible tooltip. the margins werent accomodated for and thus the size of the overlay and the coordinate transform were off

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

s2 tooltips should be properly centered on the trigger element by default. Setting a large offset should never detach the tooltip's arrow from the tooltip itself

## 🧢 Your Project:
RSP


packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -417,15 +417,20 @@ export function calculatePositionInternal(
...
+  // childOffset[crossAxis] + .5 * childOffset[crossSize] = absolute position with respect to the trigger's coordinate system that would place the arrow in the center of the trigger
+  // position[crossAxis] - margins[AXIS[crossAxis]] = value use to transform the position to a value with respect to the overlay's coordinate system. A child element's (aka arrow) position absolute's "0"
+  // is positioned after the margin of its parent (aka overlay) so we need to subtract it to get the proper coordinate transform
+  let preferredArrowPosition = childOffset[crossAxis] + .5 * childOffset[crossSize] - position[crossAxis] - margins[AXIS[crossAxis]];
 
   // Min/Max position limits for the arrow with respect to the overlay
   const arrowMinPosition = arrowSize / 2 + arrowBoundaryOffset;
-  const arrowMaxPosition = overlaySize[crossSize] - (arrowSize / 2) - arrowBoundaryOffset;
+  // overlaySize[crossSize] - margins['left'] - margins['right'] = true size of the overlay
+  const arrowMaxPosition = (overlaySize[crossSize] - margins['left'] - margins['right']) - (arrowSize / 2) - arrowBoundaryOffset ;
```
LFDanLu, line 428: Bit wary about this since it assumes the margins shouldn't be considered part of the physical overlay (aka like S2 tooltip). Not sure about complications from negative margins too...

snowystinger: should we be using a cross offset instead?

LFDanLu: ah yep thanks, forgot to convert these as well

LFDanLu:
https://www.chromatic.com/build?appId=5f0dd5ad2b5fc10022a2e320&number=769 updated chromatic, note the failures are unrelated to this change and are fixed in https://github.com/adobe/react-spectrum/pull/6829 instead

packages/@react-aria/overlays/src/calculatePosition.ts
```diff
@@ -417,15 +417,21 @@ export function calculatePositionInternal(
...
+  // position[crossAxis] - margins[AXIS[crossAxis]] = value use to transform the position to a value with respect to the overlay's coordinate system. A child element's (aka arrow) position absolute's "0"
+  // is positioned after the margin of its parent (aka overlay) so we need to subtract it to get the proper coordinate transform
+  let preferredArrowPosition = childOffset[crossAxis] + .5 * childOffset[crossSize] - position[crossAxis] - margins[AXIS[crossAxis]];
 
   // Min/Max position limits for the arrow with respect to the overlay
   const arrowMinPosition = arrowSize / 2 + arrowBoundaryOffset;
-  const arrowMaxPosition = overlaySize[crossSize] - (arrowSize / 2) - arrowBoundaryOffset;
+  // overlaySize[crossSize] - margins = true size of the overlay
+  const overlayMargin = AXIS[crossAxis] === 'left' ? margins.left + margins.right : margins.top + margins.bottom;
+  const arrowMaxPosition = overlaySize[crossSize] - overlayMargin - (arrowSize / 2) - arrowBoundaryOffset ;
```
reidbarber, line 429:
```suggestion
  const arrowMaxPosition = overlaySize[crossSize] - overlayMargin - (arrowSize / 2) - arrowBoundaryOffset;
```
super-nit: extra whitespace

reidbarber (dismissed):
LGTM

snowystinger (dismissed):
Noticed that we aren't making use of the arrowBoundaryOffset
Screenshot 2024-08-07 at 6 09 08 AM

Not sure how to use it with macros though.

Want to add here or should we handle it separately?

LFDanLu:
@snowystinger added a fix for the `arrowBoundaryOffset`, here is the new chromatic: https://www.chromatic.com/build?appId=65ef96081d99868c05594590&number=32

snowystinger (commented):
unsure what happened to the docs
Screenshot 2024-08-07 at 8 55 19 AM

LFDanLu:
weird I don't see that here: https://reactspectrum.blob.core.windows.net/reactspectrum/4b079fc16176d8a59ec943274363ab3dc9480bac/storybook-s2/index.html?path=/docs/tooltip--docs, what reproduction steps?
image

snowystinger:
Wut? That's the same URL I go to. Even if I click the link you just gave me, it shows up as how my screenshot shows. It happens in both Chrome and Safari for me, Firefox will shift it to the correct placement after 5seconds or so. Doesn't matter if I'm on my high dpi or regular dpi monitor.
If I resize the screen, it'll go to the right place as well. But it always starts off wrong.

LFDanLu:
what version of chrome/safari? I'm on latest Chrome (127.0.6533.100) and Safari 17.4, but can't get it to happen regardless of monitor :/.

https://github.com/user-attachments/assets/aa488b15-879d-4f1c-bc52-a1f9627ddd2d

snowystinger:
Chrome 126.0.6478.183 & 127.0.6533.100 (I updated after checking 126)
Safari 17.5

Let's check if anyone else on the team can reproduce. Can you do that during standup tomorrow?

yihuiliao:
yeah im able to reproduce on safari and chrome with the link daniel shared

chrome 126.0.6478.128
safari 16.6

LFDanLu:
Odd, does it happen only on load and only on the docs page? What if you toggle the controlled isOpen off then on?

yihuiliao:
for me, it only happens on load and on the docs page. if i try to inspect it with dev tools, then it fixes itself, and if i toggle isOpen, it also fixes itself and then i'm not able to reproduce it again until i reload the page.

LFDanLu:
@yihuiliao thanks for the additional info, I wonder if the layout/extra stuff the docs page loads has something do with it. I'm still unable to reproduce it on any of my browsers :/

LFDanLu:
@snowystinger @yihuiliao I was able to reproduce it on my windows machine (I should've tried it on there earlier lol). I confirmed that the issue still occurs even with the version of positioning code on main so I'm inclined to handle that elsewhere given that it seems unrelated to the changes made in the PR.

I'll see if I can figure what is going on and if I can't come up with a specific fix I'll see about modifying the docs to not be controlled open as a stop gap.

packages/@react-spectrum/s2/stories/Tooltip.stories.tsx
```diff
@@ -78,6 +80,22 @@ export const Example = (args: any) => {
   );
 };
 
+Example.play = async ({canvasElement}) => {
+  await userEvent.tab();
+  let body = canvasElement.ownerDocument.body;
+  await within(body).findByRole('tooltip');
+};
```
LFDanLu, line 87: I removed the default `isOpen` on the tooltip stories because visible jank happens for the positioning of tooltips rendered out of view and if the container story div rendered a scroll bar and removed it during calculation of the overlay position. Ideally, the autodocs would not have `isOpen: true` so that I could still preserve the chromatic tests that snap shot the tooltips as open but I believe this isn't possible short of creating a completely custom docs page. 

Instead used storybook interactions test as a substitute since they don't affect the docs page

packages/@react-aria/overlays/src/calculatePosition.ts
LFDanLu, line 1: Note that the issue mentioned in https://github.com/adobe/react-spectrum/pull/6831#pullrequestreview-2222362100 is due to storybook's docs story container rendering with a smaller height than the story contents -> causes a scroll bar to appear shifting the trigger's position as we calculate the tooltip's position -> scrollbar disappears after position is calculated causing the popover to be misplaced.

Unless we recalculate the overlay's position on every shift of the trigger's position this will be a problem. Opening a separate issue since it is out of scope and a larger change that will require discussion

LFDanLu:
https://www.chromatic.com/build?appId=65ef96081d99868c05594590&number=34

snowystinger (approved):

reidbarber (approved):

devongovett (approved):

PR #6832, opened 2024-08-06
Add prop disabledBehavior to RSP TableView api

ktabors:
Closes [issue](https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=37183649)

- [ ] Do we need an expandable rows with `disabledKeys` story?
- [ ] Do we need a story with `disabledKeys` and content like buttons in the cells?
- [ ] Documentation wasn't mentioned in the scoping, but the docs would be updated to match ListView, correct?
- [ ] Add or update a chromatic story?

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Added control "disabledBehavior" to TableView stories.
Easiest to test is you set `selectionMode` to "multiple" because the checkboxes will be disabled in both modes.

Test with "dynamic with disabled keys" story.

Test DnD with "Drag between tables" story.

Didn't see a story to test expandable row with.

* Confirm "selection" is the default for TableView. (Should be opposite ListView with is "all".)
* Confirm "all" shows the cell content text as disabled.
* Confirm "all" prevents DnD, focus, hover, actions, etc.

## 🧢 Your Project:
RSP

### DIFF
--- a/packages/@adobe/spectrum-css-temp/components/table/skin.css
+++ b/packages/@adobe/spectrum-css-temp/components/table/skin.css
@@ -180,6 +180,10 @@ tbody.spectrum-Table-body {
     background-color: var(--spectrum-table-row-background-color-down);
   }
 
+  &.is-disabled .spectrum-Table-cell.is-disabled {
+    color: var(--spectrum-alias-text-color-disabled);
+  }
+
   /* Alternative to border on rows. Using box shadow since they don't take room unlike border which would cause wiggles
    * in the highlight case and displace the sticky indicator. Also allows for a nicer bottom curved border to match the container,
    * the bottom border curved corners were cut off when using borders.
--- a/packages/@react-spectrum/table/src/TableViewWrapper.tsx
+++ b/packages/@react-spectrum/table/src/TableViewWrapper.tsx
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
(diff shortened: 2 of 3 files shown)


snowystinger (commented):
Noticed an interesting side effect of this change.
https://reactspectrum.blob.core.windows.net/reactspectrum/51dfb35aff8dea1e82bd72c76a19c5e5415f0d7e/storybook/index.html?path=/story/tableview--dynamic-with-disabled-keys&args=disabledBehavior:selection;selectionMode:multiple&providerSwitcher-express=false

If I enter selection mode by selecting a row. Then clicking another enabled row will select it and will NOT call the action for that row.
However, while in selection mode with a row selected, if I click a disabled row, then the action IS called.

I think while a row is selected actions should not be called regardless of the disabledBehavior.

ktabors:
> However, while in selection mode with a row selected, if I click a disabled row, then the action IS called.

I checked [ListView](https://reactspectrum.blob.core.windows.net/reactspectrum/51dfb35aff8dea1e82bd72c76a19c5e5415f0d7e/storybook/index.html?path=/story/listview-selection--on-action-disable-files-row-actions&providerSwitcher-express=false) and it's the same. You can select an enabled item (folder) and clicking on a disabled item will trigger the action.

snowystinger (approved):
pre-existing then is what you're saying? can we add an item to the backlog?

ktabors:
> pre-existing then is what you're saying? can we add an item to the backlog?

Added https://github.com/orgs/adobe/projects/19/views/4?pane=issue&itemId=73752687

reidbarber (approved):

PR #6833, opened 2024-08-06
Conditionally render field label only if label is provided

jluyau:
Closes 

Currently, RSP components without labels will render with an empty element with padding, causing flex alignment issues. Any component using `<FieldLabel>` will be affected as passing an empty string will not render the containers with padding (causing the misalignment).

Screenshot 2024-08-14 at 4 06 50 PM
Screenshot 2024-08-14 at 4 07 12 PM

## ✅ Pull Request Checklist:


snowystinger (commented):
I see non-conditional FieldLabels in a lot of other components. Are they not issues?
Screenshot 2024-08-07 at 9 54 06 AM

jluyau:
> I see non-conditional FieldLabels in a lot of other components. Are they not issues? Screenshot 2024-08-07 at 9 54 06 AM

the ones i fixed were because i encountered them though testing in my use cases. the other ones i'll need to test out in a storybook

jluyau:
unsure if i need to update slider, progress bar, color slider, and meter. the number value on the right prevents the dictates the min height of the component, but would it be better to remove the empty field label div?

snowystinger:
qq, should FieldLabel just return null if no children?

reidbarber (dismissed):

snowystinger:
How should I be testing this? We don't have any stories that omit labels in favor of aria-labels. I'd like to see it in a Form as well. This is affecting a lot of components and from past experience with fields, this is high risk. We really need to be able to easily test this.

jluyau:
@snowystinger let me know if that story i added is sufficient for testing or if i need to make any adjustments

snowystinger:
Ah, sorry, I meant in a mixed form, some items in the form have a label, some don't. I used your example to try it out, looks like top label and side label both work as expected.

I did find some unexpected behavior though, which is, ContextualHelp no longer renders. This might be desireable though since there is no longer a visible label to kind of anchor it into the UI. However, maybe we should be more explicit about this and throw or warn if you try to use contextual help with a non-visible label?

> unsure if i need to update slider, progress bar, color slider, and meter. the number value on the right prevents the dictates the min height of the component, but would it be better to remove the empty field label div?

I do not know what to do about this either. Should that extra area also not render? much like the contextual help? that seems less desirable since it has important information.

snowystinger (dismissed):
Sorry, I meant still inside a Form without additional wrapping divs around things. To make sure that Form layout still worked well.

I checked a few locally, looks like it still lays out forms well. Up to you if you want to revise it in this PR. I'm going to approve it because it's tagged for release and it's working well. Just hard to test is all.

jluyau:
> Sorry, I meant still inside a Form without additional wrapping divs around things. To make sure that Form layout still worked well.
> 
> I checked a few locally, looks like it still lays out forms well. Up to you if you want to revise it in this PR. I'm going to approve it because it's tagged for release and it's working well. Just hard to test is all.

ah sorry, i didnt realize that Form now has an underlying grid to do alignment. will add a mixed form in there

devongovett:
Added hiding of value label for Meter and ProgressBar when no `label` prop is provided (to match v3), but important to note that if you do this it is not accessible unless you also display the percentage externally yourself (the track doesn't pass contrast). In the eventual docs we will add a warning about this.

We will likely do something for sliders as well but it requires a bit more work so punting for now. Tracking here https://github.com/orgs/adobe/projects/19/views/32?sliceBy%5Bvalue%5D=🌈&pane=issue&itemId=74955775

Updated the example to fix the warnings about missing aria-labels, and create aria groups for each section.

devongovett (approved):

LFDanLu (approved):

PR #1424, opened 2020-12-23
fix 1423 - add isReadOnly and isDisabled in useRadioGroupState hook

intergalacticspacehighway:
Closes #1423 

## ✅ Pull Request Checklist:

- [x] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues/1423).
- [x] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [x] Filled out test instructions.
- [x] Updated documentation (if it already exists for this component).
- [x] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/TR/wai-aria-practices-1.1/)

## 📝 Test Instructions:
yarn test radio

PR #6864, opened 2024-08-13
Segmented control

yihuiliao:
Closes 

## ✅ Pull Request Checklist:


snowystinger (commented):
One thing you could try in order to set a default would be to search through the dom for the first non-disabled radio, fire a change event on it.

Another would be to use a registration pattern on the context. Each non-disabled item would call a callback with it's `value` from the context, the first time this callback is called, the default state is set. All subsequent calls to this callback would do nothing.

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+
+interface SegmentedControlProps extends ValueBase<string|null, string>, InputDOMProps, FocusEvents, StyleProps, AriaLabelingProps {
+  children?: ReactNode,
+  /**
+   * The style of the track.
+   * 
+   * @default 'none'
+   */
+  trackStyle?: 'none' | 'track',
```
devongovett, line 31: Probably should have mentioned this before (sorry), but we probably should not support the non-track version for now since it isn't accessible.

yihuiliao: oh lol it's okay. should we keep the prop still and define it like `trackStyle?: 'track'` so that there isn't as big of a API change later?

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,225 @@
...
+          isDisabled: 'GrayText',
+          isSelected: {
+            default: 'Highlight',
+            isDisabled: 'GrayText'
+          }
+        }
+      }
+    }
+  },
+  borderRadius: 'lg',
```
snowystinger, line 136: looks like maybe it should be 'm'? or more probably, 'control'

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,225 @@
...
+    <RadioGroup 
+      {...props}
+      ref={domRef}
+      className={segmentedControl({trackStyle: props.trackStyle}, props.styles)}
+      aria-label={props['aria-label'] || 'Segmented Control'}>
+      <Provider
+        values={[
+          [SegmentedControlInternalContext, {trackStyle: props.trackStyle || 'none'}]
+        ]}>
+        <div className={controlWrapper({trackStyle: props.trackStyle})}>
```
snowystinger, line 183: this div seems unneeded, i'd move any relevant styles up to the segmetedControl

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: i can switch it to 'control' but it looks the same as 'lg' to me

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,225 @@
...
+        forcedColors: {
+          isSelected: 'Highlight',
+          isDisabled: {
+            isSelected: 'GrayText'
+          },
+        }
+      }
+    },
+  },
+  // the padding should be a little less for segmented controls that only contain icons but not sure of a good way to do that
```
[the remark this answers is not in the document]
yihuiliao: im not sure if this will work in this scenario because there's a hidden input so it looks like
```
<div>
  <span with the visually hidden input>
  <icon>
</div>
```

so it's not actually an only-child like the way it is in button for example

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,204 @@
...
+  } = props
+  let domRef = useDOMRef(ref);
+
+  return (
+    <RadioGroup 
+      {...props}
+      ref={domRef}
+      style={props.UNSAFE_style}
+      className={(props.UNSAFE_className || '') + segmentedControl({size: 'M'}, props.styles)}
+      aria-label={props['aria-label'] || 'Segmented Control'}>
```
yihuiliao, line 134: how do we wanna add translations for this?

snowystinger: You can add it to the intl files, just add the english string across all of them so it doesn't crash. it'll get picked up in next translation PR after this PR is merged

packages/@react-spectrum/s2/src/SegmentedControl.tsx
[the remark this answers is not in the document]
snowystinger: tch... annoying
well, you could use callback refs to know if you have both an icon and text

we should be using the same requirements, either just a string, in which case you wrap it in a Text. Or users must pass an Icon and/or/not Text

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -97,7 +106,13 @@ const controlItem = style({
 }, getAllowedOverrides())
 
 interface SegmentedControlInternalContextProps extends SegmentedControlProps {
-  register?: (value) => void;
+  register?: (value: string) => void;
```
snowystinger, line 109: I think we generally follow the pattern of prefix the internal context with Internal https://github.com/adobe/react-spectrum/blob/main/packages/%40react-spectrum/s2/src/Picker.tsx#L219
It makes it a little easier to skim for

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -125,11 +140,11 @@ function SegmentedControl(props: SegmentedControlProps, ref: DOMRef<HTMLDivEleme
   )
 }
 
-function DefaultSelectionTracker(props) {
+function DefaultSelectionTracker(props: DefaultSelectionTrackProps) {
   let state = useContext(RadioGroupStateContext);
   let isRegistered = useRef(!(props.defaultValue == null && props.value == null));
 
-  let register = useCallback((value) => {
+  let register = useCallback((value: string) => {
```
snowystinger, line 146: string or number i believe

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -97,7 +106,13 @@ const controlItem = style({
 }, getAllowedOverrides())
 
 interface SegmentedControlInternalContextProps extends SegmentedControlProps {
-  register?: (value) => void;
+  register?: (value: string) => void;
+}
+
+interface DefaultSelectionTrackProps {
+  defaultValue?: string,
```
snowystinger, line 113: these are string or number i think

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,204 @@
...
+import {centerBaseline} from './CenterBaseline';
+import {ContextValue, Radio, RadioGroup, RadioProps, Provider, RadioGroupStateContext, useSlottedContext} from "react-aria-components"
+import {createContext, forwardRef, ReactNode, useContext, useRef, useCallback, useEffect} from 'react';
+import {focusRing, StyleProps, getAllowedOverrides} from './style-utils' with {type: 'macro'};
+import {IconContext} from './Icon';
+import {style} from '../style/spectrum-theme' with {type: 'macro'};
+import {useDOMRef, useFocusableRef} from '@react-spectrum/utils';
+import {useSpectrumContextProps} from './useSpectrumContextProps';
+
+export interface SegmentedControlProps extends ValueBase<string|null, string>, InputDOMProps, FocusEvents, StyleProps, AriaLabelingProps {
```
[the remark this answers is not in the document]
reidbarber: I think we'll need `size` also

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+const segmentedControl = style<{size: string}>({
+  font: 'control',
+  display: 'flex',
+  backgroundColor: 'gray-100',
+  borderRadius: 'lg',
+  width: 'full'
+}, getAllowedOverrides())
+
+const controlItem = style({
```
reidbarber, line 50: Should these have `transition: 'default'`?

reidbarber (commented):
Looks great!

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: in our docs it looks like defaultValue and value are defined as `string | null`

packages/@react-spectrum/s2/src/SegmentedControl.tsx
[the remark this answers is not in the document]
yihuiliao: so we removed the non-track variant since it wasn't accessible. originally it defined a prop like `trackStyle: 'none' | 'track'` but we've removed that in favor of just defaulting to the track style for now until design figures out the accessibility. 

and right now i don't see any t-shirt sizing in the figma designs. it's still under review so they might add it later

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: actually, i added it but not sure how i feel about it. you can compare the previous builds with the first build after this comment and see the difference

packages/@react-spectrum/s2/src/SegmentedControl.tsx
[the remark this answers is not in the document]
yihuiliao: the only thing is that it throws an error if you try to do `{typeof props.children === 'string' ? <Text>{props.children}</Text> : props.children}` in Radio because in RAC we expect the slot "description" or "error message" with `<Text>`. 

i could just use a span instead of `<Text>`

packages/@react-spectrum/s2/src/SegmentedControl.tsx
reidbarber: It looks better IMO, I think we should keep it.

packages/@react-spectrum/s2/intl/en-US.json
```diff
@@ -12,6 +12,7 @@
   "label.(optional)": "(optional)",
   "menu.moreActions": "More actions",
   "picker.placeholder": "Select…",
+  "segmentedcontrol.aria-label": "Segmented control",
```
reidbarber, line 15: Not blocking: I think we should require an `aria-label` for this component, or choose a default that is more understandable.

reidbarber (approved):

packages/@react-spectrum/s2/src/SegmentedControl.tsx
snowystinger: Ah, right, hasn't gone in yet [fix: add number type in value](https://github.com/adobe/react-spectrum/pull/6636#top)

packages/@react-spectrum/s2/intl/en-US.json
snowystinger: Agreed, I have no idea what a 'Segmented control' is, despite having been in many meetings about them. So we'll want to ask Spectrum for a default or require it.

Also, I think we should name the string a little differently `segmentedcontrol.defaultlabel` or `segmentedcontrol.defaultname` or something.

Finally, I don't think I have logic to handle strings only existing in s2 in my migrate intl script. If you have some cycles, maybe see about adding that. Otherwise, I can handle it soon.

devongovett: yeah I don't think we need a default label, we should just require one like we did before in ActionGroup

yihuiliao: sounds good, will update it to be required

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: talked with spectrum, we're going with a sliding animation instead

snowystinger (commented):
Noticed some bugs with the animation if I use my mouse. You may find it easier to follow https://css-tricks.com/everything-you-need-to-know-about-flip-animations-in-react/ if you introduce a new div between `RACRadioGroup` and the `Provider with InternalSegmentedControlContext` that you use instead of `domRef` for searching

In addition, you will want to wait to run either the effect that gets the initial position until the default selected item is found. Something like

```
  let isRegistered = useRef(!(props.defaultValue == null && props.value == null));
  let [isReady, setIsReady] = useState(!(props.defaultValue == null && props.value == null));

  const rects = useRef(new Map()).current;

  // default select the first available item
  let register = useCallback((value: string) => {
    if (state && !isRegistered.current) {
      isRegistered.current = true;

      state.setSelectedValue(value);
      // wait for selection to be ready, only then do we want to get the initial state for the animations
      setIsReady(true);
    }
  }, []);

  // First: get the current bounds of the element
  useEffect(() => {
    if (domRef.current && isReady) {
      let item = domRef.current.querySelector('label[data-selected=true]');

      rects.set('slider', item.getBoundingClientRect());
    }
  }, [isReady]);
```

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -156,16 +189,97 @@ function DefaultSelectionTracker(props: DefaultSelectionTrackProps) {
...
+
+  // style={{transform: move ? `translateX(100px)` : 'translateX(0px)'}}
+  // style={{transform: 'translateX(120px)'}}
   return (
     <Provider
       values={[
         [InternalSegmentedControlContext, {register: register}]
     ]}>
       {props.children}
+      <span style={style} id="animate" className={test}/>
```
reidbarber: Can we skip this behavior for `prefers-reduced-motion`?

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -156,16 +189,97 @@ function DefaultSelectionTracker(props: DefaultSelectionTrackProps) {
...
+        // styleObj.height = finalH;
+        // setStyle(styleObj);
+        
+        // Play: animate the final element from its first bounds to its last bound
+        slide.animate(
+          [
+            {transform: `translateX(${finalX}px)` },
+          ],
+          {
+            duration: 150,
```
reidbarber, line 231: We may want to consider having the duration to be dynamic based on the deltaX so that it always goes the same speed.

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -156,16 +189,97 @@ function DefaultSelectionTracker(props: DefaultSelectionTrackProps) {
...
+            easing: 'ease-in',
```
reidbarber, line 232: We may want to test the options, but I think `ease-out` is better for this use case.

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,309 @@
...
+      inputRef={inputRef}
+      style={props.UNSAFE_style}
+      className={renderProps => (props.UNSAFE_className || '') + controlItem({...renderProps}, props.styles)} >
+      <Provider 
+        values={[
+          [IconContext, {
+            render: centerBaseline({slot: 'icon', styles: style({order: 0, flexShrink: 0})})
+          }]
+        ]}>
+        {props.children}
```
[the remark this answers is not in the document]
yihuiliao: hmm not sure if i can do that without throwing a TS error since it excepts a slot prop because of the RAC implementation. it expects either "description" or "error message" which i don't think we want to add

packages/@react-spectrum/s2/stories/SegmentedControl.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <ControlItem value="week">Week</ControlItem>
+    <ControlItem value="month">Month</ControlItem>
+    <ControlItem value="year">Year</ControlItem>
+  </SegmentedControl>
+);
+
+
+export const WithIcons = (args: any) => (
+  <SegmentedControl {...args}>
+    <ControlItem value="unordered"><ListBulleted />Unordered</ControlItem>
```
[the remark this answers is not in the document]
yihuiliao: same issue as before. i have to add a slot to Text...maybe we should add a default one through context?

packages/@react-spectrum/s2/src/SegmentedControl.tsx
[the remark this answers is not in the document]
snowystinger: ah, that's an interesting case to bring up
Can you override that slot inside the RadioGroup to null it out?

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: oo yes thanks for bringing that up

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: might bring up to design and ask them what they prefer

packages/@react-stately/radio/src/useRadioGroupState.ts
```diff
@@ -74,7 +74,7 @@ export function useRadioGroupState(props: RadioGroupProps): RadioGroupState  {
   });
 
   let setSelectedValue = (value) => {
-    if (!props.isReadOnly && !props.isDisabled) {
+    if (!props.isReadOnly) {
```
yihuiliao, line 77: the reason for this change is because we need to default select the first available item even if the radio group is disabled. without this change, the selected value would never be committed. 

not sure if there are other repercussions though. tests ran fine, behavior seemed the same but i could be missing something.

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+      state.setSelectedValue(value);
+    } else if (isDisabled && disabledIsRegistered.current == null) {
+      disabledIsRegistered.current = value;
+    }
+  }, []);
+
+  // if the registration fails, then we will default select the first item
+  useLayoutEffect(() => {
+    if (state && isRegistered.current === false) {
+      console.log(disabledIsRegistered.current);
```
snowystinger, line 175:
```suggestion
```

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+        {props.children}
+      </DefaultSelectionTracker>
+    </RadioGroup>
+  );
+}
+
+function DefaultSelectionTracker(props: DefaultSelectionTrackProps) {
+  let state = useContext(RadioGroupStateContext);
+  let isRegistered = useRef(!(props.defaultValue == null && props.value == null));
+  let disabledIsRegistered = useRef<string | null>(null);
```
snowystinger, line 159:
```suggestion
  let disabledIsRegistered = useRef<string | null>(props.defaultValue ?? props.value ?? null);
```
should probably also set this

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+  let isSelected = props.value === state?.selectedValue;
+  // do not apply animation if a user has the prefers-reduced-motion setting
+  const isReduced = window.matchMedia('(prefers-reduced-motion: reduce)').matches;
+
+  useLayoutEffect(() => {
+    if (!props.isDisabled) {
+      register?.(props.value, props.isDisabled);
+    } else if (props.isDisabled) {
+      register?.(props.value, props.isDisabled);
+    }
```
snowystinger, line 208: what's the difference here?

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+          <Provider 
+            values={[
+              [IconContext, {
+                render: centerBaseline({slot: 'icon', styles: style({order: 0, flexShrink: 0})})
+              }],
+              [RACTextContext, {slots: {[DEFAULT_SLOT]: {}}}],
+              [TextContext, {styles: style({order: 1, truncate: true})}
+              ]
+            ]}>
+            {/* still need this zIndex unfortunately. without it, the slider renders on top of the text */}
```
snowystinger, line 252: yep, that's normal and expected, we can remove this comment

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+function ControlItem(props: ControlItemProps, ref: FocusableRef<HTMLLabelElement>) {
+  let inputRef = useRef<HTMLInputElement>(null);
+  let domRef = useFocusableRef(ref, inputRef);
+  let divRef = useRef<HTMLDivElement>(null);
+  let {register, prevRef, currentSelectedRef} = useContext(InternalSegmentedControlContext);
+  let state = useContext(RadioGroupStateContext);
+  let isSelected = props.value === state?.selectedValue;
+  // do not apply animation if a user has the prefers-reduced-motion setting
+  const isReduced = window.matchMedia('(prefers-reduced-motion: reduce)').matches;
```
devongovett, line 201: We might need to put this behind an if statement that checks if window?.matchMedia exists otherwise it might not work in environments like server rendering where there is no window

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+            values={[
+              [IconContext, {
+                render: centerBaseline({slot: 'icon', styles: style({order: 0, flexShrink: 0})})
+              }],
+              [RACTextContext, {slots: {[DEFAULT_SLOT]: {}}}],
+              [TextContext, {styles: style({order: 1, truncate: true})}
+              ]
+            ]}>
+            {/* still need this zIndex unfortunately. without it, the slider renders on top of the text */}
+            <div ref={divRef} style={pressScale(divRef)({isPressed})} className={style({zIndex: 1, display: 'flex', gap: 'text-to-visual'})}>
```
devongovett, line 253: Add transition: default here so press scale animates

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+    <Provider
+      values={[
+        [InternalSegmentedControlContext, {register: register, prevRef: props.prevRef, currentSelectedRef: props.currentSelectedRef}]
+      ]}> 
+      {props.children}
+    </Provider>
+  );
+}
+
+function ControlItem(props: ControlItemProps, ref: FocusableRef<HTMLLabelElement>) {
```
devongovett, line 193: Should it be named SegmentedControlItem for consistency?

packages/@react-stately/radio/src/useRadioGroupState.ts
devongovett: If we're gonna get rid of it for disabled it seems weird to leave it for read only? Did you look at the blame to see when these were added?

yihuiliao: oh good point. look's like since radio was implemented that it checked for read only. we added the check for disabled later on in this pr https://github.com/adobe/react-spectrum/pull/1424

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+import {focusRing, getAllowedOverrides, StyleProps} from './style-utils' with {type: 'macro'};
+import {IconContext} from './Icon';
+import {pressScale} from './pressScale';
+import {style} from '../style/spectrum-theme' with {type: 'macro'};
+import {Text, TextContext} from './Content';
+import {useDOMRef, useFocusableRef} from '@react-spectrum/utils';
+import {useLayoutEffect} from '@react-aria/utils';
+import {useSpectrumContextProps} from './useSpectrumContextProps';
+
+export interface SegmentedControlProps extends Omit<RadioGroupProps, 'isReadOnly' | 'name' | 'isRequired' | 'isInvalid' | 'validate' | 'validationBehavior' | 'children' | 'className' | 'style' | 'aria-label'>, StyleProps{
```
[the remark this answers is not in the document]
yihuiliao: oh yep good point

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: yeah im fine with that

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,274 @@
...
+    }
+  }, []);
+
+  // if the registration fails, then we will default select the first item
+  useLayoutEffect(() => {
+    if (state && isRegistered.current === false) {
+      console.log(disabledIsRegistered.current);
+      state.setSelectedValue(disabledIsRegistered.current);
+    } else {
+      new Error('Could not determine a default selected item');
```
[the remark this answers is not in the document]
yihuiliao: doesn't have to be. i guess the only scenario i imagine this failing is if they didn't provide any children in the first place but we've also made that required. worse case it just doesn't default select anything

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: we don't actually care if whether the radio group is disabled but whether the individual radios are disabled. the first condition is to handle radios that are not individually disabled, the second one is to handle when the radios are individually disabled. 

if all the radio's are disabled and the registration fails, that's when we'll select the first radio.

i can probably simplify this statement a bit

packages/@react-spectrum/s2/src/SegmentedControl.tsx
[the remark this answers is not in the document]
reidbarber: I think it's just missing `throw`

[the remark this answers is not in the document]
yihuiliao: ohhh yeah my b

packages/@react-spectrum/s2/src/SegmentedControl.tsx
reidbarber: Thanks for providing both. For those testing, it's easiest to see the difference when switching between the left-most and rightmost controls. I like the second *better*, but it does still seem a little slow. I'm curious what design's feedback is. Maybe a middle ground (?)

reidbarber (dismissed):

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: Seems like design (or more like Lindsey since she was the one I shared this with) liked the first one better. Her reasoning was that the second one felt like it was taking a long time between when you clicked on end to the other end. She also worried that it would feel even more slow if there were more items between the two.  

That said, if it feels too slow, we can adjust the speed to be quicker. However, that would also mean for shorter distances, the animation would be pretty fast.

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -238,10 +252,10 @@ function SegmentedControlItem(props: ControlItemProps, ref: FocusableRef<HTMLLab
       ref={domRef} 
       inputRef={inputRef}
       style={props.UNSAFE_style}
-      className={renderProps => (props.UNSAFE_className || '') + controlItem({...renderProps}, props.styles)} >
+      className={renderProps => (props.UNSAFE_className || '') + controlItem({...renderProps, isIconOnly: isIconOnly?.current}, props.styles)} >
```
snowystinger, line 255: can't read from ref during render, you'll need to make this state and trigger a re-render
you have it in a layouteffect, so this will update fast enough that there won't be a flash of wrong spacing

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+      state.setSelectedValue(value);
+    } else if (isDisabled && disabledIsRegistered.current == null) {
+      disabledIsRegistered.current = value;
+    }
+  }, []);
+
+  // if the registration fails, then we will default select the first item
+  useLayoutEffect(() => {
+    if (state && isRegistered.current === false) {
```
snowystinger, line 178:
```suggestion
    if (state && !isRegistered.current) {
```

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,289 @@
...
+      state.setSelectedValue(disabledIsRegistered.current);
+    } else if (isRegistered.current === false && disabledIsRegistered.current == null) {
```
snowystinger, line 180:
```suggestion
    } else if (!isRegistered.current&& disabledIsRegistered.current == null) {
```

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,289 @@
...
+      }
+    }
+  }, []);
+
+  useLayoutEffect(() => {
+    if (!props.isDisabled) {
+      register?.(props.value, false);
+    } else if (props.isDisabled) {
+      register?.(props.value, true);
+    }
```
snowystinger, line 224:
```suggestion
  register?.(props.value, !!props.isDisabled);
```

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,289 @@
...
+  let {register, prevRef, currentSelectedRef} = useContext(InternalSegmentedControlContext);
+  let state = useContext(RadioGroupStateContext);
+  let isSelected = props.value === state?.selectedValue;
+  // do not apply animation if a user has the prefers-reduced-motion setting
+  let isReduced = false;
+  if (window?.matchMedia) {
+    isReduced = window.matchMedia('(prefers-reduced-motion: reduce)').matches;
+  }
+  
+  let isIconOnly = useRef<boolean>(false);
```
devongovett, line 209: I think you could do this in CSS only with no JS if you use `:has()`. I believe we do that in button.

yihuiliao: yeah i think button does something like this `:has([slot=icon]:only-child)': 'control'`

but i don't think it would work in this scenario because there's a hidden input so it looks like
```
<div>
  <span with the visually hidden input>
  <icon>
</div>

```
so it's not really an only-child

reidbarber: You could try something like this: `':has([slot=icon]:nth-child(2):nth-last-child(1))'`

packages/@react-stately/radio/src/useRadioGroupState.ts
snowystinger: i think useRadio needs updates to handle this, you could move the whole if to surround them instead of handling it here
https://github.com/adobe/react-spectrum/blob/26b9d5db8a0dea35190f032514143e3b18e67439/packages/%40react-aria/radio/src/useRadio.ts#L74
https://github.com/adobe/react-spectrum/blob/26b9d5db8a0dea35190f032514143e3b18e67439/packages/%40react-aria/radio/src/useRadio.ts#L63

snowystinger (commented):
Not sure what's going on, but there's a wiggle in the animation. Only in LTR languages, I've used the Chrome animation tools to slow down the animation to see what's happening. You can see the animation moves the element farther away before starting to animate in.

I only see it happen in the built versions of storybook as well. I logged out the
```
console.log('should move to', currentItem.left, 'from', prevRef?.current.left);
```
in the useLayoutEffect, but it seems to be correctly measuring/moving.

https://github.com/user-attachments/assets/39f68dde-f590-49c7-ac84-a86504ed1154

Happy to help look into it.

reidbarber:
> Not sure what's going on, but there's a wiggle in the animation. 

Maybe try `transform-origin: left` / `transform-origin: right` ?

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,260 @@
...
+    default: 'gray-700',
+    isHovered: 'neutral-subdued',
+    isSelected: 'neutral',
+    isDisabled: 'disabled',
+    forcedColors: {
+      default: 'ButtonText',
+      isDisabled: 'GrayText',
+      isSelected: 'HighlightText'
+    }
+  },
```
LFDanLu, line 71: looks like the text in light mode doesn't show up in HCM, at least when using chrome's emulation
image

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,260 @@
...
+    default: 'edge-to-text',
+    ':has([slot=icon]:only-child)': size(6)
+  },
+  height: 32,
+  alignItems: 'center',
+  flexBasis: 0,
+  flexGrow: 1,
+  flexShrink: 0,
+  justifyContent: 'center',
+  whiteSpace: 'nowrap',
```
LFDanLu, line 83: should this have `truncate`? Tried using a real long label and got the following
image

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,260 @@
...
+}, getAllowedOverrides());
+
+const controlItem = style({
+  position: 'relative',
+  display: 'flex',
+  forcedColorAdjust: 'none',
+  color: {
+    default: 'gray-700',
+    isHovered: 'neutral-subdued',
+    isSelected: 'neutral',
```
LFDanLu, line 64: It feels really hard to distinguish the hover style in light theme IMO, maybe something to bring up with Spectrum

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,260 @@
...
+      {({isSelected, isFocusVisible, isPressed, isDisabled}) => (
+        <>
+          {isSelected && <div className={slider({isFocusVisible, isDisabled})} ref={currentSelectedRef} />}
+          <Provider 
+            values={[
+              [IconContext, {
+                render: centerBaseline({slot: 'icon', styles: style({order: 0, flexShrink: 0})})
+              }],
+              [RACTextContext, {slots: {[DEFAULT_SLOT]: {}}}],
+              [TextContext, {styles: style({order: 1, truncate: true})}]
```
LFDanLu, line 237: oh hm, I guess this was supposed to apply the truncation, not sure why it isn't taking

packages/@react-spectrum/s2/src/SegmentedControl.tsx
ktabors: And when disabled as well.

packages/@react-spectrum/s2/src/SegmentedControl.tsx
ktabors: I wish we were following `color: baseColor('gray-700')` or `neutral` for defining the default, hovered, focused, selected, and pressed?

devongovett (approved):

LFDanLu (approved):

PR #6882, opened 2024-08-14
Add docs for S2 illustrations

devongovett:
Adds a page to storybook similar to the list of icons, but for gradient and linear illustrations.


packages/dev/parcel-transformer-icon/IconTransformer.js
```diff
@@ -33,6 +37,12 @@ module.exports = new Transformer({
                   inlineStyles: {
                     onlyMatchedOnce: false,
                     removeMatchedSelectors: true
+                  },
+                  cleanupIDs: {
+                    prefix
+                  },
+                  convertPathData: {
+                    makeArcs: false
```
devongovett, line 45: Adds a hash prefix to all id attributes so they don't conflict when multiple illustrations are on the page at the same time. Also disables `makeArcs` because it was breaking one of the linear illustrations.

reidbarber (approved):

LFDanLu (approved):

PR #6894, opened 2024-08-16
Remove isCollapsed from S2 NumberField and add hideStepper

devongovett:
The isCollapsed option was removed from the Figma designs since it is not accessible (hit targets too small). Added hideStepper option from v3 as well.

### DIFF
--- a/packages/@react-spectrum/s2/src/NumberField.tsx
+++ b/packages/@react-spectrum/s2/src/NumberField.tsx
@@ -21,8 +21,7 @@ import {
   Text,
   useContextProps
 } from 'react-aria-components';
-import {baseColor, style} from '../style/spectrum-theme' with {type: 'macro'};
-import ChevronIcon from '../ui-icons/Chevron';
+import {baseColor, size, style} from '../style/spectrum-theme' with {type: 'macro'};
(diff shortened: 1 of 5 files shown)


packages/@react-spectrum/s2/src/NumberField.tsx
```diff
@@ -38,23 +37,21 @@ import {useSpectrumContextProps} from './useSpectrumContextProps';
...
   HelpTextProps {
   /**
-   * Whether the NumberField step buttons should be collapsed into a single column.
-   *
-   * @default "false"
+   * Whether to hide the increment and decrement buttons.
+   * @default false
    */
-    isCollapsed?: boolean,
+  hideStepper?: boolean,
```
reidbarber, line 48: There's a codemod for commenting this out that can be removed now.

reidbarber (commented):
There's a codemod for commenting this out that can be removed now.

ktabors (dismissed):
Works and looks good!

reidbarber (commented):
One more thing, Migrating.jsx

LFDanLu (approved):

yihuiliao (approved):

PR #6896, opened 2024-08-16
Tabs codemods

reidbarber:
## ✅ Pull Request Checklist:


packages/dev/codemods/src/s1-to-s2/src/codemods/transforms.ts
```diff
@@ -526,33 +526,144 @@ function commentIfParentCollectionNotDetected(
...
+        }
+      }
+    });
+
+    if (tabListNode) {
+      path.node.children = [tabListNode, ...tabPanelsNodes];
+    }
+  }
+
+  function transformItem(
```
devongovett, line 575: Is this different from `updateComponentWithinCollection`?

packages/dev/codemods/src/s1-to-s2/__tests__/__snapshots__/tabs.test.ts.snap
```diff
@@ -1,8 +1,105 @@
...
+"import { Tab, TabPanel, Tabs, TabList } from "@react-spectrum/s2";
+import { TabPanels, Item } from '@adobe/react-spectrum';
+let isEmphasized = true;
+let props = {isEmphasized: true};
 <div>
+  <Tabs aria-label="History of Ancient Rome"><TabList>
+      <Tab id="FoR">Founding of Rome</Tab>
+      <Tab id="MaR">Monarchy and Republic</Tab>
+      <Tab id="Emp">Empire</Tab>
+    </TabList><TabPanel id="FoR">Arma virumque cano, Troiae qui primus ab oris.</TabPanel><TabPanel id="MaR">Senatus Populusque Romanus.</TabPanel><TabPanel id="Emp">Alea jacta est.</TabPanel></Tabs>
```
devongovett, line 36: would be nice if the formatting was nicer but I guess they can run prettier or something

reidbarber: Since the node gets moved, not sure there's much we can do about this at the AST-level. Hopefully people are running their formatter+linter afterwards anyway

packages/dev/codemods/src/s1-to-s2/__tests__/__snapshots__/tabs.test.ts.snap
```diff
@@ -1,8 +1,110 @@
...
+  {name: 'Tab 6', children: 'Tab Body 6'}
+];
+
+<Tabs aria-label="Dynamic tabs"><TabList items={tabs}>
+    {(item) => (
+      <Item key={item.name}>
+        {item.name}
+      </Item>
+    )}
+  </TabList><Collection>
```
devongovett, line 22: I think `items` needs to passed into `Collection` as well.

devongovett (dismissed):

devongovett (approved):

LFDanLu (commented):
I verified that the codemod properly replaced transformed some v3 Tabs that I configured in our test apps but I had to change 
https://github.com/adobe/react-spectrum/blob/ac6116d1ae30666493d0d28ad676857e92b945bc/packages/dev/codemods/src/s1-to-s2/src/getComponents.ts#L10
to
```
const indexPath = path.join(path.dirname(packagePath), '../src/index.ts');
``` 
in order to get it to compile a list of available component in S2 since the former points to the dist folder of the installed S2 package which doesn't contain the `index.ts` file

ktabors (commented):
Not related to this PR, but I'm surprised we don't have to have an aria-label on icon only tabs. It passed axe.

Also not related to this PR, we don't have `autodocs` turned on for tabs. We should exclude the icon story because we don't have an equivalent v3, react-aria, or RAC example.

devongovett:
I will update these issues in the release notes PR.

devongovett:
I guess the problem is that inside the repo `require.resolve('@react-spectrum/s2')` resolves to `packages/@react-spectrum/s2/index.ts`, but outside it will resolve to `dist/index.cjs`...

packages/dev/codemods/src/s1-to-s2/__tests__/__snapshots__/tabs.test.ts.snap
```diff
@@ -1,8 +1,110 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Renames variants 1`] = `
-"// import {Button} from '@adobe/react-spectrum';
+exports[`Move items from Tabs to TabList 1`] = `
+"import { Collection, TabPanel, Tabs, TabList } from "@react-spectrum/s2";
+import { TabPanels, Item } from '@adobe/react-spectrum';
```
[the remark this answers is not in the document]
reidbarber: I think due to how the node was reused for . I had to manually do this

ktabors (approved):

devongovett (approved):

PR #6902, opened 2024-08-19
Release fixes

devongovett:
* Codemod needs a hashbang to specify interpreter when it is run via npx
* We need to resolve from `src/index.ts` rather than `.js`
* Convert Section to PickerSection and ComboBoxSection - this still doesn't handle dynamic collections. For that we would need to wrap the function child in `<Collection>`
* Small fix for cursor in NumberField stepper buttons
* Fix build of illustrations so that it includes our wrapper (actually processed by our Icon transformer).

### DIFF
--- a/.parcelrc-build
+++ b/.parcelrc-build
@@ -7,6 +7,10 @@
     "packages/**/intl/*.json": ["parcel-transformer-intl"],
     "bundle-text:*.svg": ["@parcel/transformer-svg", "@parcel/transformer-inline-string"],
     "illustration:*.svg": ["@react-spectrum/parcel-transformer-s2-icon"],
+    "illustrations-module:*svg": ["@react-spectrum/parcel-transformer-s2-icon"],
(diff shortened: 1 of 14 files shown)


devongovett:
I fixed the lint errors but it looks like the lint script does not exit with a non-zero code anymore which means CI does not fail when the linter has errors. We might need to bring back `concurrently` @snowystinger.

LFDanLu (commented):
Verified that Combobox/Picker/Menu all had their v3 -> S2 sections transformed properly. I'm not exactly sure how to get the Illustrations built though, tried running `generateS2Icon` from the scripts folder but it didn't seem to generate anything. Also noticed that running the codemod caused `"undefined": "^0.1.0",` to be added to the package.json, digging to see why

reidbarber (approved):

devongovett:
>  I'm not exactly sure how to get the Illustrations built though

It's part of `make build`

LFDanLu (approved):
thanks, verified the illustration wrapper

dannify (approved):

PR #6907, opened 2024-08-19
Codemod: handle legacy Link API

reidbarber:
We previously allowed anchor links or custom router links to be explicitly used as a child of `Link` (a deprecated API), so this handles those cases.

## ✅ Pull Request Checklist:


packages/dev/codemods/src/s1-to-s2/src/codemods/transforms.ts
```diff
@@ -918,6 +918,24 @@ function updateAvatarSize(
...
 }
 
+/**
+ * Handles the legacy `Link` API where an `a` tag could be used within a `Link` component.
+ * Removes the `a` tag and moves its attributes to the `Link` component.
+ */
+function updateLegacyLink(
+  path: NodePath<t.JSXElement>
+) {
+  let anchorElement = path.node.children.find((child) => t.isJSXElement(child) && t.isJSXIdentifier(child.openingElement.name) && getName(path, child.openingElement.name) === 'a') as t.JSXElement;
```
devongovett, line 928: This should handle more than just `<a>`, for example react-router links. According to the code, if there is no `href` prop directly on the `<Link>`, and the children is a React element (and not a string), we clone the child. Not quite sure what we should do with the props on components other than `<a>` inside Link though, perhaps leave a comment for manual migration.

snowystinger (approved):

yihuiliao (approved):

PR #6912, opened 2024-08-20
Fix import path in icon and illustration .d.ts files

devongovett:
`./dist/types` was not correct. The icons are in a subdirectory, and the illustrations can have multiple paths in between. We need to generate the import dynamically.


reidbarber (approved):

LFDanLu (approved):
Verified in the dist for icons and illustrations that the  `.d.ts` file all have a proper relative path pointing to IconProps

PR #6914, opened 2024-08-20
fix v3 storybook logo

reidbarber:
I think storybook has a cache-related bug where the logo from the S2 config was getting used in the v3 storybook, despite the S2 config not getting used. By setting a logo in the v3 config, we can prevent this.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Check the v3 and S2 storybook builds for the correct logo.


snowystinger (approved):
heh, when i first load the page, it has the opposite icon. then settles to the correct one. how strange

LFDanLu (approved):
LGTM, saw the same thing as @snowystinger but it doesn't seem to happen in incognito so might just be a cache thing

PR #6916, opened 2024-08-20
Fix buttons in S2 dialogs stuck in hover state

devongovett:
(review without whitespace)

Simple fix for now is to reset `OverlayTriggerStateContext` in dialog contents so it doesn't pick up the `isOpen` state from the parent DialogTrigger, which is meant to only affect the trigger button. However this means that users can't use `OverlayTriggerStateContext` to control the dialog (e.g. close it). Maybe fine for now in S2, but this is documented in RAC so I did not make the change there.

Perhaps in the future we could add an `isExpanded` render prop to Button, Link, and other components that could trigger dialogs/menus/etc. That could be used to style this state. Right now we unfortunately conflate `isPressed` with `isExpanded` in PressResponder, and changing that would be breaking... This requires more thought.

### DIFF
--- a/packages/@react-spectrum/s2/src/Dialog.tsx
+++ b/packages/@react-spectrum/s2/src/Dialog.tsx
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


LFDanLu (approved):
LGTM,  tested the S2 dialogs and verified that the hover states for the buttons work properly.

yihuiliao (approved):

snowystinger (approved):

PR #6917, opened 2024-08-20
types: fix S2 types/defaults

reidbarber:
## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/s2/src/Meter.tsx
+++ b/packages/@react-spectrum/s2/src/Meter.tsx
@@ -25,8 +25,10 @@ import {useDOMRef} from '@react-spectrum/utils';
 import {useSpectrumContextProps} from './useSpectrumContextProps';
 
 interface MeterStyleProps {
-  /** The [visual style](https://spectrum.adobe.com/page/meter/#-Options) of the Meter. */
-  variant: 'informative' | 'positive' | 'notice' | 'negative',
+  /** The [visual style](https://spectrum.adobe.com/page/meter/#-Options) of the Meter.
+   * @default 'informative'
+   */
+  variant?: 'informative' | 'positive' | 'notice' | 'negative',
   /**
    * The size of the Meter.
    *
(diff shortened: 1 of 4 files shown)


snowystinger (approved):

LFDanLu (approved):
LGTM, thanks for catching these

packages/@react-spectrum/s2/src/Meter.tsx
```diff
@@ -25,8 +25,10 @@ import {useDOMRef} from '@react-spectrum/utils';
 import {useSpectrumContextProps} from './useSpectrumContextProps';
 
 interface MeterStyleProps {
-  /** The [visual style](https://spectrum.adobe.com/page/meter/#-Options) of the Meter. */
-  variant: 'informative' | 'positive' | 'notice' | 'negative',
+  /** The [visual style](https://spectrum.adobe.com/page/meter/#-Options) of the Meter.
+   * @default 'informative'
```
devongovett, line 29: hmm I thought we had discussed making this one required but I can't remember. anyone?

LFDanLu: I just remember that `variant` used to be required in S1 but that was changed via https://github.com/adobe/react-spectrum/pull/6282.

devongovett: ohh maybe that was it

devongovett (approved):

packages/@react-spectrum/s2/src/Meter.tsx
LFDanLu: I have a faint recollection that we may have been worried about new variants of meter coming from Spectrum, but not sure if that really connects with this prop becoming required unless the default might change in the future? In which case making this required might guard against that being a breaking change since users would need to be explicit about what variant they are rendering?

devongovett: ehh I think it's ok to match S1 if we made the change there. I just forgot about it

PR #6931, opened 2024-08-22
Accordion

reidbarber:
# Accordion

This implements the [Accordion](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) pattern, with an emphasis on individual accordion items as opposed to groups.

## Features

- Implements [hidden=until-found](https://github.com/WICG/display-locking/blob/main/explainers/hidden-content-explainer.md)
- Supports additional interactive elements adjacent to Heading (i.e. Menu)
- Not a collection component

### React Aria/Stately hooks: 
- useDisclosure
- useDisclosureState

### React Aria components: 
- AccordionItem
  - Heading
    - Button
  - AccordionPanel

### RSP v3 components:

- AccordionGroup
    - AccordionItem
        - AccordionHeader
        - AccordionPanel

### Spectrum 2 components:

- AccordionGroup
    - AccordionItem
        - AccordionHeader
        - AccordionPanel

### RAC

```tsx
<AccordionItem>
  {({isExpanded}) => (
    <>
      <Header>
        <Heading level={3}>
          <Button slot="trigger">{isExpanded ? '⬇️' : '➡️'} This is an accordion header</Button>
        </Heading>
      </Header>
      <AccordionPanel>
        <p>This is the content of the accordion panel.</p>
      </AccordionPanel>
    </>
  )}
</AccordionItem>
```

### Spectrum 2

```tsx
<AccordionGroup isQuiet styles={style({width: 384})}>
  <AccordionItem>
    <AccordionHeader>
      Files
    </AccordionHeader>
    <AccordionPanel>
      Files content
    </AccordionPanel>
  </AccordionItem>
  <AccordionItem>
    <AccordionHeader>
      People
    </AccordionHeader>
    <AccordionPanel>
      <TextField label="Name" />
    </AccordionPanel>
  </AccordionItem>
</AccordionGroup>
```

## Additional comments

With regards to an Accordion (or AccordionGroup) wrapper around AccordionItem elements in React Aria, there is no ARIA semantics for that element, and keyboard arrow navigation is listed as optional, so we're going to prioritize the individual AccordionItem element for now but may revisit this in a follow-up. The main benefit would be controlling the expanded items' state at the group level and using Up/Down or Home/End keyboard navigation between item headers.

## ✅ Pull Request Checklist:


nwidynski:
Hey @reidbarber thanks getting to work on this! 

I've got a couple of questions regarding the spec the team chose here and would love to provide some input from our own learnings as we implemented both `<Disclosure />` and `<Accordion />` pattern using react-aria.

1. Is there a reason not to follow the `<...Trigger/>` pattern instead of `<AccordionItem>`?

```tsx
<DisclosureTrigger isOpen>
  <Heading level={3}>
    <Button />
  </Heading>
  <Disclosure>
    <Text>I am controlled.</Text>
  </Disclosure>
</DisclosureTrigger>
```

2. Since the W3C patterns do not collide in their ARIA properties or interactivity, would there be reasoning against re-cycling `<Disclosure />` inside `<Accordion />`, similar to how `<Checkbox/>` works in both grouped and ungrouped scenario?

```tsx
const { buttonProps, regionProps } = isNil(ctxState)
  ? // eslint-disable-next-line react-hooks/rules-of-hooks
    useDisclosureTrigger(state)
  : // eslint-disable-next-line react-hooks/rules-of-hooks
    useAccordionItem(props, ctxState, objectRef);
```

3. Have you guys considered adding something like `useDisclosureAnimation()`? 

```tsx
export const useDisclosureAnimation = (
  ...option: AtomicDisclosureAnimationOptions
): AtomicDisclosureAnimation => {
  const [ref, isOpen] = option;

  const previous = useRef(isOpen);
  const [, rerender] = useState({});

  const isAnimating = previous.current !== isOpen;
  const onEnd = useCallback(() => rerender({}), []);
  
  if (isAnimating) previous.current = isOpen;
  
  useAnimation(ref, isAnimating, onEnd);

  return {
    isOpening: isOpen && isAnimating,
    isClosing: !isOpen && isAnimating,
  };
};
```

4. Does the team already have an idea how to work in arrow key navigation inside `<Accordion />`? I suppose it would work similar to the `<Tabs />` pattern but i would be very interested in any guidance.

reidbarber:
@nwidynski  Thanks for the input! We are definitely still considering different API options, so I appreciate the comments.

> Is there a reason not to follow the `<...Trigger/>` pattern instead of `<AccordionItem>`?

I think the main reason is that we want to support additional interactive elements adjacent to the accordion header, and that pattern usually means the entire first child is the trigger.

> would there be reasoning against re-cycling `<Disclosure />` inside `<Accordion />`

We haven't given much thought into including a disclosure component, but as it seems like a simple subset of accordion features, I think we could consider having accordion build on top it, maybe at the hook level.

> Have you guys considered adding something like `useDisclosureAnimation()`

This looks great! I think we definitely want to include these states.

> Does the team already have an idea how to work in arrow key navigation inside `<Accordion />`?

Arrow key navigation between items is optional in ARIA, so we're still considering options for this.

packages/@react-aria/accordion/src/useAccordionItem.ts
```diff
@@ -0,0 +1,87 @@
...
+          state.open();
+        };
+      }
+    }
+  }, [isControlled, panelRef, props.isOpen, state, supportsBeforeMatch]);
+
+  return {
+    buttonProps: {
+      'aria-expanded': state.isOpen,
+      'aria-controls': state.isOpen ? regionId : undefined,
```
LFDanLu, line 75: I know we only set `aria-controls` on our triggers when they are open but I figured that was mainly because the overlay isn't actually rendered at that point. However, the accordion panel is available so maybe it should always have the `aria-controls` set here? Not sure what difference there is though

packages/@react-spectrum/accordion/src/Accordion.tsx
```diff
@@ -10,91 +10,106 @@
...
+
+export interface SpectrumAccordionHeaderProps extends DOMProps, AriaLabelingProps {
+  children: React.ReactNode
+}
+
+function AccordionHeader(props) {
+  let {direction} = useLocale();
+  return (
+    <Header {...props}>
+      <Heading className={classNames(styles, 'spectrum-Accordion-itemHeading')}>
```
LFDanLu, line 82: Should a user be able to specify the heading level?

packages/@react-stately/accordion/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@react-stately/accordion",
+  "version": "3.0.0",
```
LFDanLu, line 3: just a reminder that this will need to be alpha

LFDanLu (commented):
Just some initial comments, but I think overall the api makes sense. Might be a bit odd to have a AccordionItem only for RAC without a container Accordion element, but the wrapping container doesn't seem to have much of a purpose so it might be ok. If we feel like that might be confusing, we could change it from "Item" so that it doesn't sound like it is a collection element

packages/@react-spectrum/accordion/src/Accordion.tsx
```diff
@@ -10,91 +10,106 @@
...
+import {filterDOMProps} from '@react-aria/utils';
+import React, {forwardRef} from 'react';
 import styles from '@adobe/spectrum-css-temp/components/accordion/vars.css';
-import {TreeState, useTreeState} from '@react-stately/tree';
-import {useAccordion, useAccordionItem} from '@react-aria/accordion';
-import {useHover} from '@react-aria/interactions';
 import {useLocale} from '@react-aria/i18n';
 import {useProviderProps} from '@react-spectrum/provider';
 
+// TODO: handle types
```
snowystinger, line 25: ? want another set of eyes here? or this done?

packages/@react-spectrum/accordion/src/Accordion.tsx
```diff
@@ -10,91 +10,106 @@
...
+export interface SpectrumAccordionGroupProps extends StyleProps, DOMProps, AriaLabelingProps {
+  children: React.ReactNode
+}
 
-function Accordion<T extends object>(props: SpectrumAccordionProps<T>, ref: DOMRef<HTMLDivElement>) {
+function AccordionGroup(props: SpectrumAccordionGroupProps, ref: DOMRef<HTMLDivElement>) {
```
snowystinger, line 30: why are we keeping the AccordionGroup if it doesn't appear to *do* anything? It doesn't even have a role

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,128 @@
+/*
```
snowystinger, line 1: Just checking, but recommendation for building a mutually exclusive AccordionGroup with arrow key navigation would be something like this?
```
function AccordionGroup() {
  let [openIndex, setOpenIndex] = useState(0);
  let focusManager = useFocusManager({accept: (node) => {
    return node.getAttribute('data-accordion-trigger-g1') === true;
  }});
  return (
    <div onKeyDown={(e) => {
      if (e.key === 'ArrowDown') {
        focusManager.focusNext();
        e.preventDefault();
      } else if (e.key === 'ArrowUp') {
        focusManager.focusPrevious();
        e.preventDefault();
      }
    }}>
      <AccordionItem isOpen={openIndex === 0} onOpenChange={val => val && setOpenIndex(0)}>
        {({isOpen}) => (
          <>
            <Header>
              <Heading level={3}>
                <Button slot="trigger" data-accordion-trigger-g1>{isOpen ? '⬇️' : '➡️'} Item 1</Button>
              </Heading>
            </Header>
            <AccordionPanel>
              <p>Content for panel 1.</p>
            </AccordionPanel>
          </>
        )}
      </AccordionItem>
      <AccordionItem isOpen={openIndex === 1} onOpenChange={val => val && setOpenIndex(1)}>
        {({isOpen}) => (
          <>
            <Header>
              <Heading level={3}>
                <Button slot="trigger" data-accordion-trigger-g1>{isOpen ? '⬇️' : '➡️'} Item 2</Button>
              </Heading>
            </Header>
            <AccordionPanel>
              <p>Panel 2's content.</p>
            </AccordionPanel>
          </>
        )}
      </AccordionItem>
    </div>
  );
}

export const AccordionGroupExample = {
  render: () => <FocusScope><AccordionGroup /></FocusScope>
};
```

snowystinger (commented):
Please add react-stately/accordion to the ts strict config

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Whether the accordion item is open (controlled).
+   * @selector [data-open]
+   */
+  isOpen?: boolean,
+  /**
+   * Whether the accordion item is focused, either via a mouse or keyboard.
+   * @selector [data-focused]
+   */
+  // isFocused: boolean,
```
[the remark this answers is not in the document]
reidbarber: Mostly wanted to figure out if we want these, and if so, what do they reflect. If focus is in the entire item (on trigger or in panel)?

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,128 @@
...
+  children: ReactNode
+}
+
+function AccordionPanel(props: AccordionPanelProps, ref: ForwardedRef<HTMLElement>) {
+  let {regionProps, panelRef} = useContext(InternalAccordionContext)!;
+  let renderProps = useRenderProps({
+    ...props,
+    defaultClassName: 'react-aria-AccordionPanel',
+    values: {
+      // TODO: add more values?
```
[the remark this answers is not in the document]
reidbarber: I think that makes sense

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,368 @@
...
+  isQuiet?: boolean,
+  /** Whether the accordion should be disabled. */
+  isDisabled?: boolean
+}
+
+const groupStyles = style({
+  display: 'flex',
+  flexDirection: 'column',
+  minWidth: sizeValue(200)
+});
```
snowystinger, line 53: getAllowedOverrides

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * A accordion group is a container for multiple accordion items.
+ */
+let _AccordionGroup = /*#__PURE__*/ (forwardRef as forwardRefType)(AccordionGroup);
+export {_AccordionGroup as AccordionGroup};
+
+const itemStyles = style({
+  color: 'heading',
+  padding: '[2px]'
```
snowystinger, line 96: this never changes? won't spacing be static, so `2` should be fine?
I think devon recently added support for 2 in a PR, don't recall if it was in cards though, might have a check there

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,368 @@
...
+    UNSAFE_style,
+    UNSAFE_className = '',
+    isDisabled,
+    ...otherProps
+  } = props;
+  const domProps = filterDOMProps(otherProps);
+  let {isQuiet, isDisabled: isGroupDisabled} = useContext(InternalAccordionGroupContext);
+  return (
+    <>
+      {!isQuiet && <Divider size="S" />}
```
snowystinger, line 125: I don't think these are divider, i think they are just borders. In the Figma files they don't have rounded caps.
I think we don't want them as dividers in the accessibility tree either

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,368 @@
...
+      <RACAccordionItem
+        {...domProps}
+        isDisabled={isDisabled || isGroupDisabled}
+        ref={domRef}
+        style={UNSAFE_style}
+        className={(UNSAFE_className ?? '') + itemStyles}>
+        {props.children}
+      </RACAccordionItem>
+      {!isQuiet && <Divider UNSAFE_className={lastDividerStyles} size="S" />}
```
snowystinger, line 134: will remove need for unsafe classname to make them just borders as well

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,357 @@
...
+export interface AccordionHeaderProps extends UnsafeStyles, DOMProps {
+  children: React.ReactNode
+}
+
+const headingStyle = style({
+  margin: 0
+});
+
+const buttonStyles = style({
+  // Custom focus ring so it doesn't overlap divider
```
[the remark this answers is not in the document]
snowystinger: outline inset? or a negative offset?

Should be able to still use the `focusRing` macro and just override the relevant properties

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,368 @@
...
+  },
+  gap: {
+    size: {
+      S: sizeValue(7),
+      M: sizeValue(11),
+      L: sizeValue(14),
+      XL: sizeValue(17)
+    }
+  },
+  height: '[inherit]',
```
snowystinger, line 220: full?

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,368 @@
...
+      XL: {
+        isFocusVisible: 'lg',
+        isQuiet: {
+          isHovered: 'lg',
+          isFocusVisible: 'lg'
+        }
+      }
+    }
+  },
+  textAlign: 'start'
```
snowystinger, line 263: should be the default

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,368 @@
...
+let _AccordionHeader = /*#__PURE__*/ (forwardRef as forwardRefType)(AccordionHeader);
+export {_AccordionHeader as AccordionHeader};
+
+export interface AccordionPanelProps extends RACAccordionPanelProps, UnsafeStyles, DOMProps, AriaLabelingProps {
+  children: React.ReactNode
+}
+
+const panelStyles = style({
+  font: 'body',
+  height: 'auto',
```
snowystinger, line 323: should be the default and is unneeded

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,368 @@
...
+  paddingTop: {
+    isExpanded: sizeValue(8)
```
snowystinger, line 325: fits on the grid, can just use `8` right?

snowystinger: same with some other values below

packages/react-aria-components/stories/Accordion.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+export const AccordionExample = (args: any) => (
+  <AccordionItem {...args}>
+    {({isExpanded}) => (
+      <>
+        <Header>
+          <Heading level={3}>
+            <Button slot="trigger">{isExpanded ? '⬇️' : '➡️'} This is an accordion header</Button>
+          </Heading>
+        </Header>
+        <AccordionPanel>
```
snowystinger, line 32: are we worried about people conditionally rendering the panel based on `isExpanded`
@devongovett thoughts?

snowystinger (commented):
v3 chevrons don't rotate when open?

s2 HCM isn't quite done

packages/@react-aria/accordion/src/useDisclosure.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }, [isControlled, contentRef, props.isExpanded, state, supportsBeforeMatch]);
+  
+
+  return {
+    triggerProps: {
+      id: triggerId,
+      'aria-expanded': state.isExpanded,
+      'aria-controls': contentId,
+      onPress: state.toggle,
```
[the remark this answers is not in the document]
reidbarber: I think so. Looks like the APG examples for accordion/disclosure do this as well.

packages/@react-aria/accordion/src/useAccordionItem.ts
```diff
@@ -0,0 +1,87 @@
...
+import {useId} from '@react-aria/utils';
+
+export interface AccordionItemAriaProps {
+  /** Whether the accordion item is disabled. */
+  isDisabled?: boolean,
+  /**
+   * The accessibility role for the accordion item panel.
+   * @default 'region'
+   */
+  role?: string,
```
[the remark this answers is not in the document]
reidbarber: After a closer look, I wonder if this type should be `'region' | undefined` instead?

> Optionally, each element that serves as a container for panel content has role [region](https://w3c.github.io/aria/#region) and [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) with a value that refers to the button that controls display of the panel.

> Avoid using the region role in circumstances that create landmark region proliferation, e.g., in an accordion that contains more than approximately 6 panels that can be expanded at the same time.

> Role region is especially helpful to the perception of structure by screen reader users when panels contain heading elements or a nested accordion.

https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#wai-ariaroles,states,andproperties:

packages/@react-spectrum/s2/src/Accordion.tsx
reidbarber: This is to override the user agent stylesheet's default button styles of `text-align: center`

packages/@react-spectrum/s2/src/Accordion.tsx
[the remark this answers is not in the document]
reidbarber: Negative offset worked here, thanks!

packages/@react-spectrum/s2/src/Accordion.tsx
reidbarber: Removed this after adding negative outline offset

packages/@react-spectrum/s2/src/Accordion.tsx
reidbarber: It was overflowing originally, but looks like its fixed now.

packages/@react-spectrum/accordion/src/Accordion.tsx
reidbarber: I think there is value in having it for propagating `size`, `isQuiet`, `density`, and `isDisabled` down to child items.

packages/react-aria-components/src/Accordion.tsx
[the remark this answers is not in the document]
reidbarber: Just used `isFocusVisibleWithin` instead

packages/@react-aria/accordion/src/useDisclosure.ts
```diff
@@ -0,0 +1,80 @@
+/*
```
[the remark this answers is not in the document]
reidbarber: I reverted the accordion hook changes for now. I think marking it as deprecated is the best route

packages/@react-aria/disclosure/src/useDisclosure.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export interface DisclosureAria {
+  /** Props for the disclosure button. */
+  triggerProps: AriaButtonProps,
+  /** Props for the content element. */
+  contentProps: HTMLAttributes<HTMLElement>
+}
+
+export function useDisclosure(props: AriaDisclosureProps, state: DisclosureState): DisclosureAria {
```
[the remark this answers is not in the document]
reidbarber: That ref wouldn't get used anywhere right now, but I guess we could require it to avoid a breaking change if we ever needed it?

packages/@react-aria/accordion/src/useAccordionItem.ts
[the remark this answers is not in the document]
reidbarber: Discussed with accessibility team, and we will default to `'group'` and allow `'region'` for now.

packages/@react-spectrum/accordion/test/Accordion.test.js
```diff
@@ -10,136 +10,141 @@
...
+//       </Accordion>
+//     </Provider>
+//   );
+// }
 
 describe('Accordion', function () {
-  let user;
-  beforeAll(() => {
-    user = userEvent.setup({delay: null, pointerMap});
+  it('TOOD: remove', function () {
```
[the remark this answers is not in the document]
reidbarber: Added these back

packages/@react-aria/disclosure/src/useDisclosure.ts
```diff
@@ -0,0 +1,89 @@
...
+  let triggerId = useId();
+  let contentId = useId();
+  let isControlled = props.isExpanded !== undefined;
+  let supportsBeforeMatch = 'onbeforematch' in document.body;
+
+  useEffect(() => {
+    // Until React supports hidden="until-found": https://github.com/facebook/react/pull/24741
+    if (supportsBeforeMatch && contentRef.current && !isControlled && !isDisabled) {
+      if (state.isExpanded) {
+        // @ts-ignore
```
snowystinger, line 52: wrap the `if (state.isExpanded) { } else {}` in `if (contentRef.current) {`? then hopefully you don't need the ts-ignores

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,429 @@
...
+  size?: 'S' | 'M' | 'L' | 'XL',
+  /**
+   * The amount of space between the accordion items.
+   * @default "regular"
+   */
+  density?: 'compact' | 'regular' | 'spacious',
+  /** Whether the accordion item should be displayed with a quiet style. */
+  isQuiet?: boolean,
+  /** The contents of the accordion item, consisting of an AccordionHeader and AccordionPanel. */
+  children: [ReactElement<AccordionHeaderProps>, ReactElement<AccordionPanelProps>],
```
snowystinger, line 104: React 16/17/18 don't support this, not sure about 19, but I didn't think we can support enforcing this at the TS level. Which probably means we can't enforce this at all, other than visually, which isn't that helpful. Maybe it doesn't matter the order

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,429 @@
...
+    isFocusVisible: 'transparent-black-100',
+    isHovered: 'transparent-black-100'
+  },
+  borderWidth: 0,
+  borderRadius: {
+    // Only rounded for keyboard focus and quiet hover.
+    default: 'none',
+    size: {
+      S: {
+        isFocusVisible: '[7px]',
```
snowystinger, line 298: these odd values seem wrong, have you brought up with Spectrum?

packages/@react-aria/disclosure/src/useDisclosure.ts
reidbarber: I think this one isn't needed, but the others are

reidbarber: These are needed for the `hidden` attribute boolean type.

packages/@react-spectrum/s2/src/Accordion.tsx
reidbarber: I haven't. I think they just wanted the smallest difference possible between sizes.

Screenshot 2024-09-16 at 7 04 59 PM

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,397 @@
...
+  paddingBottom: {
+    isExpanded: sizeValue(16)
+  },
+  paddingX: {
+    isExpanded: {
+      size: {
+        S: sizeValue(8),
+        M: sizeValue(9),
+        L: sizeValue(12),
+        XL: sizeValue(15)
```
devongovett, line 365: hmm wonder where these sizes came from. can they be derived from something? Probably to align with the chevron icon?

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,397 @@
...
+  transitionDuration: '100ms',
+  transitionProperty: 'rotate',
+  '--iconPrimary': {
+    type: 'fill',
+    value: 'currentColor'
+  },
+  flexShrink: 0
+});
+
+function AccordionHeader(props: AccordionHeaderProps, ref: DOMRef<HTMLDivElement>) {
```
devongovett, line 313: Wondering if we will need to split out a separate component for the header vs the heading/button so you could add adjacent actions or if that's possible to achieve with the current api?

packages/@react-spectrum/s2/style/spectrum-theme.ts
```diff
@@ -356,6 +356,8 @@ const fontSize = {
   'ui-xl': fontSizeToken('font-size-300'),
   'ui-2xl': fontSizeToken('font-size-400'),
   'ui-3xl': fontSizeToken('font-size-500'),
+  'ui-4xl': fontSizeToken('font-size-600'),
+  'ui-5xl': fontSizeToken('font-size-700'),
```
devongovett, line 360: Probably should just be using the heading tokens instead of ui tokens here.

packages/@react-stately/disclosure/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "src"
+  ],
+  "sideEffects": false,
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/adobe/react-spectrum"
+  },
+  "dependencies": {
+    "@react-stately/utils": "^3.10.3",
+    "@react-types/accordion": "3.0.0-alpha.23",
```
devongovett, line 26: Unused?

packages/@react-stately/disclosure/src/useDisclosureState.ts
```diff
@@ -0,0 +1,65 @@
...
+  collapse(): void,
+  /** Toggles the disclosure's visibility. */
+  toggle(): void
+}
+
+/**
+ * Manages state for a disclosure widget. Tracks whether the disclosure is expanded, and provides
+ * methods to toggle this state.
+ */
+export function useDisclosureState(props: DisclosureProps): DisclosureState  {
```
devongovett, line 43: hmm, is this actually the same as `useToggleState` or `useOverlayTriggerState`? should we reuse one of those or create a new one? Looks like just the naming is different...

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+import React, {createContext, DOMAttributes, ForwardedRef, forwardRef, ReactNode, useContext, useRef} from 'react';
+
+export interface AccordionItemProps extends Omit<AriaDisclosureProps, 'children' | 'contentRef'>, HoverEvents, RenderProps<AccordionItemRenderProps>, SlotProps {}
+
+export interface AccordionItemRenderProps {
+  /**
+   * Whether the accordion item is expanded (controlled).
+   * @selector [data-expanded]
+   */
+  isExpanded?: boolean,
```
devongovett, line 29: Should not be optional.

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+import {forwardRefType} from '@react-types/shared';
+import {HoverEvents, useFocusRing} from 'react-aria';
+import {mergeProps, mergeRefs, useObjectRef} from '@react-aria/utils';
+import React, {createContext, DOMAttributes, ForwardedRef, forwardRef, ReactNode, useContext, useRef} from 'react';
+
+export interface AccordionItemProps extends Omit<AriaDisclosureProps, 'children' | 'contentRef'>, HoverEvents, RenderProps<AccordionItemRenderProps>, SlotProps {}
+
+export interface AccordionItemRenderProps {
+  /**
+   * Whether the accordion item is expanded (controlled).
```
devongovett, line 26:
```suggestion
   * Whether the accordion item is expanded.
```

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+export const DisclosureStateContext = createContext<DisclosureState | null>(null);
+
+interface InternalAccordionContextValue {
+  contentProps: DOMAttributes<HTMLElement>,
+  contentRef: React.RefObject<HTMLElement | null>
+}
+
+const InternalAccordionContext = createContext<InternalAccordionContextValue | null>(null);
+
+function AccordionItem(props: AccordionItemProps, ref: ForwardedRef<HTMLDivElement>) {
```
devongovett, line 56: Should this be called Disclosure to match the hook and to indicate it can be used standalone?

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+      state
+    }
+  });
+
+  return (
+    <Provider
+      values={[
+        [ButtonContext, {
+          slots: {
+            trigger: triggerProps
```
devongovett, line 82: Should we also add a `DEFAULT_SLOT` here so you can add additional buttons without errors?

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+export interface AccordionPanelProps extends RenderProps<{}> {
+  /**
+   * The accessibility role for the accordion item's panel.
+   * @default 'group'
+   */
+  role?: 'group' | 'region',
+  children: ReactNode
+}
+
+function AccordionPanel(props: AccordionPanelProps, ref: ForwardedRef<HTMLElement>) {
```
devongovett, line 110: DisclosurePanel?

packages/react-aria-components/src/Accordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+        values={[
+          [ButtonContext, null]
+        ]}>
+        {props.children}
+      </Provider>
+    </div>
+  );
+}
+
+/**
```
devongovett, line 143: We should probably have an `Accordion` component in RAC at some point as well to support mutually exclusive accordions.

packages/react-aria-components/stories/Accordion.stories.tsx
devongovett: Should be ok aside from not being findable during ctrl+f. But if we don't document it that way people probably won't use it that way.

devongovett (commented):
Looks great! A few API suggestions. Feel free to address some of them in followup if you want.

packages/@react-aria/disclosure/src/useDisclosure.ts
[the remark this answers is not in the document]
reidbarber: May be confusing about whether it goes on the content or the trigger? I guess we can include that in the JSDoc

packages/@react-stately/disclosure/src/useDisclosureState.ts
snowystinger: I'd be ok with reusing toggle state, it's toggling
not really related to overlays

packages/react-aria-components/src/Accordion.tsx
snowystinger: shouldn't it?

packages/@react-stately/disclosure/src/useDisclosureState.ts
reidbarber: If we want to stick with the `expanded` naming, we'll need to keep the new hook. I'm open to changing this, but I like the expanded name

packages/@react-spectrum/s2/src/Accordion.tsx
```diff
@@ -0,0 +1,397 @@
...
+  alignItems: 'center',
+  paddingX: {
+    size: {
+      S: sizeValue(7),
+      M: sizeValue(11),
+      L: sizeValue(14),
+      XL: sizeValue(17)
+    }
+  },
+  paddingY: {
```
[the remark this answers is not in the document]
reidbarber: Not sure how this will work since it varies by density

packages/react-aria-components/stories/Accordion.stories.tsx
```diff
@@ -0,0 +1,54 @@
+/*
```
devongovett, line 1: name this file Disclosure

packages/@react-aria/accordion/src/useAccordion.ts
```diff
@@ -23,18 +23,21 @@ export interface AccordionAria {
...
-  /** Props for the accordion item content element. */
+  /** Props for the disclosure content element. */
   regionProps: DOMAttributes
 }
 
-export function useAccordionItem<T>(props: AccordionItemAriaProps<T>, state: TreeState<T>, ref: RefObject<HTMLButtonElement | null>): AccordionItemAria {
+/**
+ * @deprecated Use useDisclosure from `@react-aria/disclosure` instead.
+ */
+export function useDisclosure<T>(props: DisclosureAriaProps<T>, state: TreeState<T>, ref: RefObject<HTMLButtonElement | null>): DisclosureAria {
```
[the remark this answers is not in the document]
reidbarber: Good catch

devongovett (dismissed):

packages/react-aria-components/package.json
```diff
@@ -39,8 +39,10 @@
   "dependencies": {
     "@internationalized/date": "^3.5.5",
     "@internationalized/string": "^3.2.3",
+    "@react-aria/accordion": "3.0.0-alpha.33",
```
LFDanLu, line 42: can we remove this dep now? Noticed it in `@react-spectrum/accordion`'s package json too

LFDanLu (approved):
behavior looks good, just one other dep that I think we can get rid of

PR #6941, opened 2024-08-23
feat: codemods for Badge and Well

reidbarber:
Thanks to @ktabors for noticing these were missing.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Check tests and new story for style macro well-equivalent.

### DIFF
--- a/packages/dev/codemods/src/s1-to-s2/src/codemods/changes.ts
+++ b/packages/dev/codemods/src/s1-to-s2/src/codemods/changes.ts
@@ -132,6 +132,23 @@ export const changes: ChangesJSON = {
       }
     ]
   },
+  Badge: {
+    changes: [
+      {
+        description: "Change variant='info' to variant='informative'",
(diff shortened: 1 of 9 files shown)


packages/dev/codemods/src/s1-to-s2/src/codemods/changes.ts
```diff
@@ -426,19 +443,6 @@ export const changes: ChangesJSON = {
       }
     ]
   },
-  Flex: {
```
reidbarber, line 429: realized these transforms aren't needed since this is already handled in `styleProps.ts`

snowystinger (commented):
Where'd the definition of Well come from? could we have a story that shows a div with those styles applied? I just want to check it out in dark/light mode for comparison

snowystinger (approved):

ktabors (approved):

PR #6950, opened 2024-08-26
Fix TreeView so that the row height always matches the chevron height

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test the TreeView docs and make sure the checkbox is properly centered in the row

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-spectrum/tree/src/TreeView.tsx
+++ b/packages/@react-spectrum/tree/src/TreeView.tsx
@@ -140,8 +140,7 @@ const treeCellGrid = style({
   display: 'grid',
   width: 'full',
   alignItems: 'center',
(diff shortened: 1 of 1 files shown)


packages/@react-spectrum/tree/src/TreeView.tsx
LFDanLu, line 1: Since font size on the docs page is 14px and the height of the row is set to `10` (aka 2.5rems) the TreeView's row height is actually 35 instead of 40. Not sure if we intend the font size to be 16px or if the row should always just be 40px.

snowystinger (approved):

reidbarber (approved):

PR #6978, opened 2024-08-30
S2 CardView and ghost loading

devongovett:
This adds S2 Cards, which come in the following configurations:

* `Card` - a generic card container, into which you can place a `<CardPreview>`, `<Content>` with `title`, `description` and action menu slots, and a `<Footer>`. Cards come in 4 sizes, 3 densities, and 4 variants. A `<CollectionCardPreview>` can also be used to display a grid of multiple images in a preview.
* `AssetCard` - a pre-configured `Card` for displaying an asset, with an image preview and metadata.
* `UserCard` - a pre-configured `Card` for displaying a user, with an optional preview, avatar, and metadata.
* `ProductCard` - a pre-configured `Card` for displaying a product, with an optional preview, thumbnail, metadata, and footer call to action.

`CardView` displays a grid or waterfall layout of cards, with selection, keyboard navigation, item actions/links, infinite loading, etc.


packages/@react-spectrum/s2/src/Content.tsx
```diff
@@ -102,17 +103,26 @@ export const TextContext = createContext<ContextValue<ContentProps, DOMRefValue>
...
-  let {UNSAFE_className = '', UNSAFE_style, styles, isHidden, slot, ...otherProps} = props;
+  let {UNSAFE_className = '', UNSAFE_style, styles, isHidden, slot, children, ...otherProps} = props;
+  let racContext = useContext(RACTextContext);
+  let isSkeleton = useIsSkeleton();
+  [children, UNSAFE_style] = useSkeletonText(children, UNSAFE_style);
   if (isHidden) {
     return null;
   }
+
+  slot = slot && racContext && 'slots' in racContext && !racContext.slots?.[slot] ? undefined : slot;
```
devongovett, line 114: Not sure if we want this or not. The idea is to not forward Spectrum-specific slots that RAC doesn't expect

packages/@react-spectrum/s2/src/Form.tsx
```diff
@@ -36,11 +37,29 @@ export interface FormProps extends FormStyleProps, Omit<RACFormProps, 'className
...
+      }
+    }
+
+    // Skeleton always wins over local props.
+    if (isSkeleton) {
+      result.isDisabled = true;
+    }
+
+    return result;
+  }, [ctx, props, isSkeleton]);
```
devongovett, line 62: Is it a bad idea to do this here? It was an easy way to mark all form components within a skeleton as disabled.

packages/@react-stately/virtualizer/src/LayoutInfo.ts
```diff
@@ -36,6 +36,11 @@ export class LayoutInfo {
    */
   parentKey: Key | null;
 
+  /** 
+   * Content for this view if it was generated by the layout rather than coming from the Collection.
+   */
+  content: any | null;
```
devongovett, line 42: Ideally we'd use a generic here but that would require a lot of changes to propagate everywhere...

packages/@react-spectrum/s2/src/CardView.tsx
```diff
@@ -0,0 +1,511 @@
...
+  layout?: 'grid' | 'waterfall',
+  size?: 'XS' | 'S' | 'M' | 'L' | 'XL',
+  density?: 'compact' | 'regular' | 'spacious',
+  variant?: 'primary' | 'secondary' | 'tertiary' | 'quiet',
+  selectionStyle?: 'checkbox' | 'highlight',
+  loadingState?: LoadingState,
+  onLoadMore?: () => void
+}
+
+class FlexibleGridLayout<T extends object, O> extends Layout<Node<T>, O> {
```
devongovett, line 38: This is different from the GridLayout in RAC because it supports variable row heights - the height of a row is equal to the maximum height of the items in that row. The RAC GridLayout just scales the sizes of its items proportionally, which won't work for wrapping text. I could see potentially adding this as an option in RAC at some point.

packages/@react-stately/virtualizer/src/LayoutInfo.ts
snowystinger: lot of changes, but would they be breaking?

packages/@react-spectrum/s2/src/ImageCoordinator.tsx
```diff
@@ -0,0 +1,150 @@
...
+  for (let isLoaded of loaded.values()) {
+    if (!isLoaded) {
+      return false;
+    }
+  }
+  return true;
+}
+
+export function ImageCoordinator(props: ImageCoordinatorProps) {
+  // If we are already inside another ImageList, just pass
```
snowystinger, line 111:
```suggestion
  // If we are already inside another ImageCoordinator, just pass
```
Is this what it's saying? Or can there be multiple lists in one coordinator?

packages/@react-spectrum/s2/src/Link.tsx
```diff
@@ -83,15 +85,27 @@ const link = style<LinkRenderProps & LinkStyleProps>({
...
-  let {variant = 'primary', staticColor, isQuiet, isStandalone, UNSAFE_style, UNSAFE_className = '', styles} = props;
+  let {variant = 'primary', staticColor, isQuiet, isStandalone, UNSAFE_style, UNSAFE_className = '', styles, children} = props;
 
   let domRef = useFocusableRef(ref);
+  let isSkeleton = useContext(SkeletonContext) || false;
+  [children, UNSAFE_style] = useSkeletonText(children, UNSAFE_style);
+
+  useLayoutEffect(() => {
+    if (domRef.current) {
+      // TODO: should RAC Link pass through inert?
```
snowystinger, line 96: I think so, seems like an ok thing to pass through in general and might make it easier for people to make their website/form builders instead of having to wrap all their components

packages/@react-spectrum/s2/src/SkeletonCollection.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export interface SkeletonCollectionProps {
+  children: () => ReactNode
+}
+
+let cache = new WeakMap();
+export const SkeletonCollection = createLeafComponent('skeleton', (props: SkeletonCollectionProps, ref, node) => {
+  // Cache rendering based on node object identity. This allows the children function to randomize
+  // its content (e.g. heights) and preserve on re-renders.
+  // TODO: do we need a `dependencies` prop here?
```
snowystinger, line 25: Can you expand on why we might need dependencies here? what kinds of things can change where we'd want to update?

packages/@react-stately/virtualizer/src/Virtualizer.ts
```diff
@@ -114,8 +114,8 @@ export class Virtualizer<T extends object, V> {
   }
 
   private _renderView(reusableView: ReusableView<T, V>) {
-    let {type, key} = reusableView.layoutInfo;
-    reusableView.content = this.collection.getItem(key);
+    let {type, key, content} = reusableView.layoutInfo;
```
snowystinger, line 117: I don't quite follow the implications of this change

packages/dev/parcel-transformer-s2-icon/IconTransformer.js
```diff
@@ -57,9 +57,10 @@ module.exports = new Transformer({
           'var(--spectrum-global-color-gray-800, #292929)': `var(--iconPrimary, var(--lightningcss-light, ${tokens['gray-800'].sets.light.value}) var(--lightningcss-dark, ${tokens['gray-800'].sets.dark.value}))`
         },
         typescript: true,
+        ref: true,
         plugins: ['@svgr/plugin-svgo', '@svgr/plugin-jsx']
       })
-    ).replace('export default SvgComponent;', '');
+    ).replace('export default ForwardRef;', '');
```
snowystinger, line 63: can delete comment below?

packages/@react-aria/utils/src/useLoadMore.ts
```diff
@@ -29,7 +29,7 @@ export interface LoadMoreProps {
    */
   scrollOffset?: number,
   /** The data currently loaded. */
-  items?: any[]
+  items?: any
```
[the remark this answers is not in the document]
devongovett: Iterable in this case but we don't actually use the value, it's just for invalidating an effect.

packages/@react-spectrum/s2/src/ActionButton.tsx
```diff
@@ -175,6 +177,7 @@ export const ActionButtonContext = createContext<ContextValue<ActionButtonProps,
 
 function ActionButton(props: ActionButtonProps, ref: FocusableRef<HTMLButtonElement>) {
   [props, ref] = useSpectrumContextProps(props, ref, ActionButtonContext);
+  props = useFormProps(props as any);
```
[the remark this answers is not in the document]
devongovett: Disabled

packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -0,0 +1,669 @@
...
+export interface CardPreviewProps extends UnsafeStyles {
+  children: ReactNode
+}
+
+export const CardPreview = forwardRef(function CardPreview(props: CardPreviewProps, ref: DOMRef<HTMLDivElement>) {
+  let {size, isQuiet, isHovered, isFocusVisible, isSelected} = useContext(InternalCardContext);
+  let {UNSAFE_className, UNSAFE_style} = props;
+  let domRef = useDOMRef(ref);
+  return (
+    <div
```
snowystinger, line 437: should we add filterDOMProps to all of these, people always want to pass their data-*s

packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -0,0 +1,669 @@
...
+});
+
+const collectionImage = style({
+  width: 'full',
+  aspectRatio: {
+    default: 'square',
+    ':nth-last-child(4):first-child': '[3/2]'
+  },
+  gridColumnEnd: {
+    ':nth-last-child(4):first-child': 'span 3'
```
snowystinger, line 469: hah, nice selector, good way to only do it for first image in a set of 4

packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -0,0 +1,669 @@
...
+              aspectRatio: '[3/1]',
+              objectFit: 'cover',
+              pointerEvents: 'none',
+              userSelect: 'none'
+            })
+          }],
+          [AvatarContext, {
+            size: avatarSize[size],
+            UNSAFE_style: {
+              '--size': avatarSize[size] + 'px'
```
snowystinger, line 569: we've got to start an audit of all these variables we introduce

or hash them for internal module use or something. it'd be nice if we could export/import them internally as well, so we know where they are defined vs used

don't worry about in this PR

packages/@react-spectrum/s2/src/CardView.tsx
```diff
@@ -0,0 +1,513 @@
...
+      newLayoutInfo.rect.height = size.height;
+      newLayoutInfo.estimatedSize = false;
+      this.layoutInfos.set(key, newLayoutInfo);
+      return true;
+    }
+
+    return false;
+  }
+
+  // Override keyboard navigation to work spacially.
```
snowystinger, line 315: Technically these overrides work for all grid layouts if you include keyAbove/keyBelow the same way. It's just that in a grid there's only one overlap candidate ever possible.

Any reason not to change how the base version works to this and then not worry about overriding?

packages/@react-spectrum/s2/src/CardView.tsx
```diff
@@ -0,0 +1,513 @@
...
+    }
+  }
+};
+
+const cardViewStyles = style({
+  overflowY: {
+    default: 'auto',
+    isLoading: 'hidden'
+  },
+  scrollPadding: '[18px]',
```
snowystinger, line 466: i know @LFDanLu was having some issues with inconsistency across browsers, did you resolve that?

packages/@react-spectrum/s2/src/CardView.tsx
```diff
@@ -0,0 +1,513 @@
...
+  alignItems: 'center',
+  justifyContent: 'center',
+  outlineStyle: 'none'
+}, getAllowedOverrides({height: true}));
+
+export function CardView<T extends object>(props: CardViewProps<T>) {
+  let {children, layout: layoutName = 'grid', size = 'M', density = 'regular', variant = 'primary', selectionStyle = 'checkbox', UNSAFE_className = '', UNSAFE_style, styles, ...otherProps} = props;
+  let options = layoutOptions[size][density];
+  let layout = useMemo(() => {
+    variant; // needed to invalidate useMemo
```
snowystinger, line 480: i think this is preferable over eslint-ignoring the entire dep array, should update this in other places. outside scope of pr

packages/@react-spectrum/s2/src/CardView.tsx
```diff
@@ -0,0 +1,513 @@
...
+    return layoutName === 'waterfall' ? new WaterfallLayout(options) : new FlexibleGridLayout(options);
+  }, [options, variant, layoutName]);
+
+  let ref = useRef(null);
+  useLoadMore({
+    isLoading: props.loadingState !== 'idle' && props.loadingState !== 'error',
+    items: props.items, // TODO: ideally this would be the collection. items won't exist for static collections, or those using <Collection>
```
snowystinger, line 487: this is fine though because a static collection wouldn't ever be in a loadMore/loading state?

snowystinger (commented):
Noticed a bit of a bug with virtualizer reuse. Selection can appear to move to other un-selected items when moving around via keyboard. I know we have this in TableView and ListView, but it feels more noticeable because of the more obvious selected border.

https://github.com/user-attachments/assets/69aea1f0-84de-477b-9a9c-dec8f423eb32

The docs page for Cards has a warning about missing labels.

packages/@react-spectrum/s2/src/Skeleton.tsx
```diff
@@ -0,0 +1,142 @@
...
+        ],
+        {
+          duration: 2000,
+          iterations: Infinity,
+          easing: 'ease-in-out'
+        }
+      );
+      animationRef.current.startTime = 0;
+    } else if (!isAnimating && animationRef.current) {
+      animationRef.current.cancel();
```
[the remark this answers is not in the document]
devongovett: Animations don't continue after unmounting

packages/@react-spectrum/s2/src/Skeleton.tsx
```diff
@@ -0,0 +1,142 @@
...
+  );
+}
+
+export const loadingStyle = raw(`
+  background-image: linear-gradient(to right, ${colorToken('gray-100')} 33%, light-dark(${colorToken('gray-25')}, ${colorToken('gray-300')}), ${colorToken('gray-100')} 66%);
+  background-size: 300%;
+  * {
+    visibility: hidden;
+  }
+`, 'UNSAFE_overrides');
```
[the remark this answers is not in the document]
devongovett: Maybe but I don't want to refactor the entire style system just to support cards haha

packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -0,0 +1,669 @@
...
+  }
+});
+
+let footer = style({
+  display: 'flex',
+  flexDirection: 'row',
+  alignItems: 'end',
+  justifyContent: 'space-between',
+  gap: 8,
+  paddingTop: '[calc(var(--card-spacing) * 1.5 / 2)]'
```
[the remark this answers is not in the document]
devongovett: It's spectrum design math from a document Lindsay made

packages/@react-spectrum/s2/src/CardView.tsx
devongovett: The spacial navigation algorithm isn't that simple to implement performantly with just the DOM. There's not a good way to query for elements inside a rectangle that I know of

packages/@react-spectrum/s2/src/CardView.tsx
devongovett: If you used `<Collection>` or another component that produced multiple children it could.

packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -0,0 +1,669 @@
...
+  M: 'S',
+  L: 'M',
+  XL: 'L'
+} as const;
+
+export const Card = forwardRef(function Card(props: CardProps, ref: DOMRef<HTMLDivElement>) {
+  [props] = useSpectrumContextProps(props, ref, CardContext);
+  let domRef = useDOMRef(ref);
+  let {density = 'regular', size = 'M', variant = 'primary', UNSAFE_className = '', UNSAFE_style, styles} = props;
+  let isQuiet = variant === 'quiet';
```
reidbarber, line 330: Screenshot 2024-09-03 at 2 28 56 PM

Should the background be different for the quiet variant here, or are we using the wrong background in storybook?

packages/@react-spectrum/s2/stories/CardView.stories.tsx
```diff
@@ -0,0 +1,231 @@
...
+  links: {html: string},
+  preview_photos: {id: string, urls: {small: string}}[]
+}
+
+function TopicCard({topic}: {topic: Topic}) {
+  return (
+    <Card href={topic.links.html} target="_blank" textValue={topic.title}>
+      <CollectionCardPreview>
+        {topic.preview_photos.slice(0, 4).map(photo => (
+          <Image key={photo.id} alt="" src={photo.urls.small} />
```
reidbarber, line 165: Would we support wrapping these in links?

reidbarber (commented):
The styles and behavior look great! Just a few initial comments

packages/@react-spectrum/s2/src/CardView.tsx
```diff
@@ -0,0 +1,513 @@
...
+  display: {
+    isEmpty: 'flex'
+  },
+  flexDirection: 'column',
+  alignItems: 'center',
+  justifyContent: 'center',
+  outlineStyle: 'none'
+}, getAllowedOverrides({height: true}));
+
+export function CardView<T extends object>(props: CardViewProps<T>) {
```
LFDanLu, line 476: I remember for the old CardView implementation that we offered the ability to customize the layout via maxColumns, inter row/column padding, etc. I assume these were scrapped by Spectrum but just wanna make sure.

packages/@react-spectrum/s2/src/CardView.tsx
```diff
@@ -0,0 +1,513 @@
...
+            <AriaGridList
+              ref={ref}
+              {...otherProps}
+              layout="grid"
+              selectionBehavior={selectionStyle === 'highlight' ? 'replace' : 'toggle'}
+              style={{
+                ...UNSAFE_style,
+                scrollPadding: options.minSpace.height
+              }}
+              className={renderProps => UNSAFE_className + cardViewStyles({...renderProps, isLoading: props.loadingState === 'loading'}, styles)}>
```
LFDanLu, line 505: Will need a focus ring when the CardView is empty and tabbed to.

LFDanLu (commented):
Still going through the Skeleton and layout logic, some initial comments from testing. Overall, performs nicely, tested a variety of keyboard operations/prefers reduced motion/etc

packages/@react-spectrum/s2/src/Skeleton.tsx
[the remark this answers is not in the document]
snowystinger: sure, we can follow it up if it presents an issue

packages/@react-spectrum/s2/src/Card.tsx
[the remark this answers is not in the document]
snowystinger: haha, ok, then i guess keep as is so it matches design

packages/@react-spectrum/s2/src/CardView.tsx
snowystinger: ahhh ok, now i understand the comment better.
so the issue here is that we haven't built the collection yet, that happens below.

packages/@react-spectrum/s2/src/Card.tsx
snowystinger: one of the non-supported combinations?

packages/@react-spectrum/s2/src/Content.tsx
snowystinger: Seems related: https://github.com/adobe/react-spectrum/issues/6992

packages/@react-spectrum/s2/src/Form.tsx
devongovett: So, the alternative is to manually do this in every individual component rather than here in one place. Or I guess we could rename `useFormProps` to something else...

packages/@react-spectrum/s2/src/Skeleton.tsx
```diff
@@ -0,0 +1,142 @@
...
+  let isLoading = useContext(SkeletonContext);
+  let animation = useLoadingAnimation(isLoading || false);
+  if (isLoading == null) {
+    return children;
+  }
+
+  let childRef = 'ref' in children ? children.ref as any : children.props.ref;
+  return (
+    <SkeletonContext.Provider value={null}>
+      {isLoading ? cloneElement(children, {
```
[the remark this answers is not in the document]
devongovett: No, `SkeletonWrapper` is not exported publicly.

packages/@react-spectrum/s2/src/ImageCoordinator.tsx
devongovett: Nested ImageCoordinators don't do anything, they just delegate to the parent.

packages/@react-spectrum/s2/src/SkeletonCollection.tsx
devongovett: just like `<Collection>`, this component caches the results. So if the child function depends on values from outside, they'd need to be dependencies so that the cache invalidates.

packages/@react-spectrum/s2/src/CardView.tsx
devongovett: yeah

packages/@react-stately/virtualizer/src/Virtualizer.ts
devongovett: yep. this enables the layout to produce items that aren't in the original collection.

packages/@react-spectrum/s2/src/ImageCoordinator.tsx
```diff
@@ -0,0 +1,150 @@
...
+    loaded: new Map()
+  });
+  
+  let register = useCallback((url: string) => dispatch({type: 'register', url}), []);
+  let unregister = useCallback((url: string) => dispatch({type: 'unregister', url}), []);
+  let load = useCallback((url: string) => dispatch({type: 'load', url}), []);
+
+  useEffect(() => {
+    if (!loadedAll) {
+      let timeoutId = setTimeout(() => {
```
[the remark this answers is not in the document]
devongovett: how could there be multiple at the same time? The effect cleanup will run when `timeout` changes and will clear the previous one.

packages/@react-spectrum/s2/src/Image.tsx
```diff
@@ -0,0 +1,207 @@
...
+  let animation = useLoadingAnimation(isAnimating);
+  useLayoutEffect(() => {
+    if (hidden) {
+      return;
+    }
+
+    // If the image is already loaded, update state immediately instead of waiting for onLoad.
+    if (state === 'loading' && imgRef.current?.complete) {
+      // Queue a microtask so we don't hit React's update limit.
+      // TODO: is this necessary?
```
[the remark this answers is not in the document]
devongovett: It's sort of hard to tell if I'm seeing placebo or how much difference it actually makes, but I noticed that the image's `complete` property could be true before the `onLoad` event fires (async). So the theory here is that we could show the image slightly earlier, which is important while scrolling. Just not sure how much difference it makes.

packages/@react-spectrum/s2/stories/CardView.stories.tsx
```diff
@@ -0,0 +1,231 @@
...
+  return (
+    <Card id={item.id} textValue={item.description || item.alt_description}>
+      <CardPreview>
+        <Image 
+          src={item.urls.regular}
+          styles={style({
+            width: 'full',
+            pointerEvents: 'none'
+          })}
+          UNSAFE_style={{
```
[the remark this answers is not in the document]
devongovett: not the theme. These values are dynamic: `${item.width} / ${item.height}`. I did wonder if we should somehow support "safe" inline styles (limited overridable properties) for cases like this. Probably a larger discussion than cards though.

packages/@react-spectrum/s2/stories/CardView.stories.tsx
```diff
@@ -0,0 +1,231 @@
...
+  onAction: null,
+  selectionMode: 'multiple'
+};
+
+export const Empty = (args: CardViewProps<any>, {viewMode}) => {
+  return (
+    <CardView 
+      aria-label="Assets"
+      {...args}
+      UNSAFE_style={{height: viewMode === 'docs' ? 600 : '100vh', width: viewMode === 'docs' ? '100%' : '100vw'}}
```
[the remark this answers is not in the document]
devongovett: well.... those values aren't in the theme. Sizes only go up to 384px, and we don't have `100vw`.

packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -0,0 +1,669 @@
...
+  backgroundColor: '--s2-container-bg',
+  boxShadow: {
+    default: 'emphasized',
+    isHovered: 'elevated',
+    isFocusVisible: 'elevated',
+    isSelected: 'elevated',
+    forcedColors: '[0 0 0 1px ButtonBorder]',
+    variant: {
+      tertiary: {
+        // Render border with box-shadow to avoid affecting layout.
```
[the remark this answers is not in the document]
devongovett: idk... I don't like this at all, and I don't want to encourage using box-shadow as a border. It doesn't show in high contrast mode by default and feels like a hack. Ideally `outline` would be used, but here we already used outline for the focus ring.

packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -0,0 +1,669 @@
...
+  borderColor: 'gray-1000',
+  transition: 'default',
+  opacity: {
+    default: 0,
+    isSelected: 1
+  }
+  // outlineColor: 'white',
+  // outlineOffset: -4,
+  // outlineStyle: 'solid',
+  // outlineWidth: 2
```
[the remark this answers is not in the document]
devongovett: waiting for design

packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -0,0 +1,669 @@
...
+          {children}
+        </InternalCardContext.Provider>
+      </div>
+    );
+  }
+
+  let press = pressScale(domRef, UNSAFE_style);
+  return (
+    <ElementType
+      {...props}
```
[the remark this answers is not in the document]
devongovett: that happens inside GridListItem but probably needed in the div case above.

packages/@react-spectrum/s2/src/CardView.tsx
devongovett: actually this one isn't used. it gets overridden below by the inline style. removing this.

packages/@react-spectrum/s2/src/Card.tsx
devongovett: hmm, yes, the background is wrong in storybook docs. I originally had `transparent` here but then realized I shouldn't override `--s2-container-bg` in that case (it is used for the border around avatars etc.). Guess I should add it back.

packages/@react-spectrum/s2/stories/CardView.stories.tsx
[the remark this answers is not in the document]
devongovett: No, density at the CardView level only affects the spacing between the cards, not within the cards. Density at the card level affects the spacing within the cards (but there is no story control for that).

packages/@react-spectrum/s2/stories/CardView.stories.tsx
```diff
@@ -0,0 +1,231 @@
...
+        <div className={style({display: 'flex', alignItems: 'center', gap: 8, gridColumnEnd: 'span 2'})}>
+          <Avatar src={item.user.profile_image.small} />
+          <Text slot="description">{item.user.name}</Text>
+        </div>
+      </Content>
+    </Card>
+  );
+}
+
+export const Example = (args: CardViewProps<any>, {viewMode}) => {
```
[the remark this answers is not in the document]
devongovett: hmm I haven't been able to reproduce this. any idea how you got this?

packages/@react-spectrum/s2/src/ImageCoordinator.tsx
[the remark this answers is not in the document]
snowystinger: ah yep, over thought it or missed that

packages/@react-spectrum/s2/stories/CardView.stories.tsx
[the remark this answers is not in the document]
snowystinger: yeah, I think some safe ones should be provided for places where we know it to be so

snowystinger (dismissed):
Testing was pretty good, happy to get in and address anything else in new PRs

packages/@react-aria/virtualizer/src/ScrollView.tsx
```diff
@@ -177,27 +182,38 @@ export function useScrollView(props: ScrollViewProps, ref: RefObject<HTMLElement
...
+
+      if (lastContentSize.current && prevShowX === curShowX && prevShowY === curShowY) {
+        return;
+      }
+
+      // React doesn't allow flushSync inside effects, so queue a microtask.
+      // We also need to wait until all refs are set (e.g. when passing a ref down from a parent).
+      // If we are in an `act` environment, update immediately without a microtask so you don't need
+      // to mock timers in tests. In this case, the update is synchronous already.
+      if (globalThis.IS_REACT_ACT_ENVIRONMENT) {
```
snowystinger: possibly not going to work, as some Issues out there recommend setting this to false in the tests... 
https://github.com/testing-library/react-testing-library/issues/1108
I can't imagine this is common, but I imagine this will be fun for people to debug, hahaimgonnagocrynow

snowystinger (approved):
I still see the horizontal scrollbar on initial render. I'm not fussed about it. It goes away if I resize the page.

devongovett:
> I still see the horizontal scrollbar on initial render

That's embarrassing. I had it working and then optimized it and apparently it broke and I didn't notice? Reverted to the previous version I had, which updates more frequently (whenever the content size changes at all) rather than trying to detect only when the scrollbars show and hide.

snowystinger (dismissed):

packages/@react-spectrum/s2/stories/CardView.stories.tsx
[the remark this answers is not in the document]
reidbarber: Not able to reproduce now

reidbarber (dismissed):

devongovett:
Updated with the following changes based on my meeting with design today:

* Checkbox wrapper changed from 6px to 4px
* An extra inner white/black stroke was added to quiet cards without a checkbox in order to distinguish the selection outline from the preview image. All other card variants do not have this stroke.
* The whole card now has press scaling in all cases rather than only the checkbox. For quiet cards, only the preview scales down.
* Checkbox now changes size depending on the card t-shirt size, and the space from the edge of the card to the checkbox also changes based on size/density (aligning with the metadata).
* Updated some spacing/sizing values to match designs.
* Fixed grid definition when there is no action menu.

snowystinger (approved):

devongovett:
> Only thing is that we might want to hide DnD from the props since isn't ready yet right?

I can do this in a followup

PR #7027, opened 2024-09-12
add all S2 components to example apps

reidbarber:
Adds all S2 components to Parcel and Webpack example apps for testing purposes. 

We'll need to investigate TagGroup crashing in a follow-up.

## ✅ Pull Request Checklist:

- [ ] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- [ ] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [ ] Filled out test instructions.
- [ ] Updated documentation (if it already exists for this component).
- [ ] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/WAI/ARIA/apg/)

## 📝 Test Instructions:

Check [Parcel example](https://reactspectrum.blob.core.windows.net/reactspectrum/ae9dd3f270a7d50ed2c21152b27cd664b4c99c08/verdaccio/s2-parcel-example/index.html) and [Webpack example](https://reactspectrum.blob.core.windows.net/reactspectrum/ae9dd3f270a7d50ed2c21152b27cd664b4c99c08/verdaccio/s2-webpack-5-example/index.html)

PR #7035, opened 2024-09-13
Fix verdaccio deploy step

reidbarber:
A wildcard in the middle of the path isn't support by AzCopy, so this fixes the verdaccio deploy step.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Check [this commit ](https://github.com/adobe/react-spectrum/commit/f8bd2de41195c77160c5c4e9368e190143afd991#comments) for working links

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -512,7 +512,9 @@ jobs:
           name: deploy
           command: |
             if [ $AZURE_STORAGE_SAS_TOKEN ]; then
-              azcopy copy "/tmp/verdaccio_dist/*/verdaccio/*" "https://reactspectrum.blob.core.windows.net/reactspectrum/$CIRCLE_SHA1/verdaccio${AZURE_STORAGE_SAS_TOKEN}" --recursive
+              for dir in /tmp/verdaccio_dist/*/verdaccio; do
+                azcopy copy "$dir/*" "https://reactspectrum.blob.core.windows.net/reactspectrum/$CIRCLE_SHA1/verdaccio${AZURE_STORAGE_SAS_TOKEN}" --recursive
+              done
             fi
 
   deploy-production:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -645,9 +645,9 @@ workflows:
           requires:
             - install
       - docs-verdaccio:
-          # filters:
-          #   branches:
-          #     only: main
+          filters:
+            branches:
+              only: main
           requires:
             - install
       - deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -512,7 +512,7 @@ jobs:
           name: deploy
           command: |
             if [ $AZURE_STORAGE_SAS_TOKEN ]; then
-              azcopy copy "/tmp/verdaccio_dist/*/verdaccio/*" "https://reactspectrum.blob.core.windows.net/reactspectrum/$CIRCLE_SHA1/verdaccio${AZURE_STORAGE_SAS_TOKEN}" --recursive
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -512,7 +512,7 @@ jobs:
           name: deploy
           command: |
             if [ $AZURE_STORAGE_SAS_TOKEN ]; then
-              azcopy copy "/tmp/verdaccio_dist/*/verdaccio/*" "https://reactspectrum.blob.core.windows.net/reactspectrum/$CIRCLE_SHA1/verdaccio${AZURE_STORAGE_SAS_TOKEN}" --recursive
+              azcopy copy "/tmp/verdaccio_dist/*/verdaccio/*" "https://reactspectrum.blob.core.windows.net/reactspectrum/${CIRCLE_SHA1}/verdaccio${AZURE_STORAGE_SAS_TOKEN}" --recursive
```
LFDanLu, line 515: Should `${AZURE_STORAGE_SAS_TOKEN}` be appended to the end of the url here? Wouldn't this mean the generated url would have said (secret?) token appended to the end?

reidbarber: Pretty sure this is a standard API for moving files into blob storage. The URL won't get exposed publicly (unless we logged it out somewhere explicitly). I'm pretty sure the azure-cli was doing this under the hood, you just need to do it yourself with AzCopy's API:  https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-blobs-upload?toc=%2Fazure%2Fstorage%2Fblobs%2Ftoc.json&bc=%2Fazure%2Fstorage%2Fblobs%2Fbreadcrumb%2Ftoc.json#get-started

LFDanLu: gotcha thanks for the link! Not sure why the comment didn't seem to get made on the commit though, do you have a link for the generated test apps? I tried replacing a working build link commit's hash but no luck

reidbarber: Not sure why comments aren't showing up, I'm having the same issue over at https://github.com/adobe/react-spectrum/pull/7027

So I've been replacing the hash in an old link like you mentioned. The deploy step logs show the files were successfully moved, so not sure where they are ending up.

LFDanLu: Ah, the comment on the commit isn't showing up because https://github.com/adobe/react-spectrum/blob/73414999fcc6afab5c4bc8ec4740fe3dc1eb8502/.circleci/comment.js#L14 needs to be true as well as https://github.com/adobe/react-spectrum/blob/73414999fcc6afab5c4bc8ec4740fe3dc1eb8502/.circleci/comment.js#L44. That being said, that will just comment the links with the commit hash which is what we've been trying to access manual with no luck :/.

reidbarber: I think the issue may be with the wildcard in the path: https://stackoverflow.com/questions/44395658/azcopy-wildcards-in-middle-of-pattern

.circleci/config.yml
```diff
@@ -512,7 +512,9 @@ jobs:
           name: deploy
           command: |
             if [ $AZURE_STORAGE_SAS_TOKEN ]; then
-              azcopy copy "/tmp/verdaccio_dist/*/verdaccio/*" "https://reactspectrum.blob.core.windows.net/reactspectrum/$CIRCLE_SHA1/verdaccio${AZURE_STORAGE_SAS_TOKEN}" --recursive
```
devongovett, line 515: huh, wildcards definitely show up in their docs. wonder what's different here: https://learn.microsoft.com/en-us/azure/storage/common/storage-ref-azcopy-copy

LFDanLu (approved):
Good catch with the wildcard, verified the PR build comment links

PR #7046, opened 2024-09-17
Fix CardView quiet heights in Safari

devongovett:
Fixes a bug found during testing where in Safari the heights of the cards would become massive, only in the quiet variant, and this could result in a "maximum update depth exceeded" crash. That resulted from the cards getting larger and larger each render, triggering more renders, etc.

I don't quite understand the reason this happened, only in Safari, but the difference between quiet and non-quiet cards that seemed to cause it was `overflow: visible` vs `overflow: clip`. We need `visible` so the shadow on the preview shows. Adding `contain: layout` seems to fix this, but again, still not sure why exactly. I also noticed that the issue is fixed in Safari Tech Preview, even without this change. 🤷

### DIFF
--- a/packages/@react-spectrum/s2/src/CardView.tsx
+++ b/packages/@react-spectrum/s2/src/CardView.tsx
@@ -551,11 +551,13 @@ function CardView<T extends object>(props: CardViewProps<T>, ref: DOMRef<HTMLDiv
     items: props.items, // TODO: ideally this would be the collection. items won't exist for static collections, or those using <Collection>
     onLoadMore: props.onLoadMore
   }, domRef);
+
+  let ctx = useMemo(() => ({size, variant}), [size, variant]);
   
   return (
     <UNSTABLE_Virtualizer layout={layout} layoutOptions={options}>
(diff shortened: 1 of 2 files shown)


packages/@react-spectrum/s2/src/CardView.tsx
```diff
@@ -551,11 +551,13 @@ function CardView<T extends object>(props: CardViewProps<T>, ref: DOMRef<HTMLDiv
     items: props.items, // TODO: ideally this would be the collection. items won't exist for static collections, or those using <Collection>
     onLoadMore: props.onLoadMore
   }, domRef);
+
+  let ctx = useMemo(() => ({size, variant}), [size, variant]);
```
devongovett, line 555: Unrelated to the original bug, just something I did while debugging. Avoids re-rendering every card if CardView re-renders but these props don't change.

snowystinger (approved):
Verified it's fixed in Safari 18. Continues to work in Chrome and FF.

reidbarber (approved):

PR #7053, opened 2024-09-18
S2 Card updates

devongovett:
Three changes from conversation with design:

1. Make the white inner stroke for selected quiet cards with highlight selection more subtle, by making it partially transparent. I believe this still passes contrast, and it also matches the overlay around the checkbox when that appears.
2. Prevent the gap between the title and description from expanding when an action button is present. The action button is taller than a single line of text so it makes the grid row taller, making it appear like there's more space. Now the button overlaps the gap and doesn't change the height of the row.
3. When resizing a CardView with waterfall layout, the cards seemed to jump around since we always chose the shortest column to append to during layout and this could change due to small changes in text wrapping, etc. Now we store the current column index for each card and try to preserve that so cards don't jump around unless the number of columns changes.

### DIFF
--- a/packages/@react-spectrum/s2/src/Card.tsx
+++ b/packages/@react-spectrum/s2/src/Card.tsx
@@ -331,6 +331,13 @@ let content = style({
   }
 });
 
+let actionMenu = style({
+  gridArea: 'menu',
+  // Don't cause the row to expand, preserve gap between title and description text.
+  // Would use -100% here but it doesn't work in Firefox.
+  marginY: '[calc(-1 * self(height))]'
+});
+
 let footer = style({
   display: 'flex',
   flexDirection: 'row',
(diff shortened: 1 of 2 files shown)


LFDanLu (approved):
LGTM, verified the outline color, spacing between the title and description before and after an action button is present, and the waterfall layout resizing behavior. Personally I feel like the new outline color for quiet + highlight + waterfall is too subtle now but its fine haha

ktabors (approved):
The code changes look good. 

I did notice some behavior toggling card size `s` to `xs` back and forth, while being on the last row of CardView in `waterfall`, where on a second or third toggle the card positions don't match their initial state, like they'd start with a gap and after toggling from `xs` to `s` to `xs` there was no longer a gap in the cards. Might be related to my next observation.
A few times I saw a gap in the last row that looked wrong. I drew a line along the bottom of that row of cards and it looked like the parent card above the gap, without a child, was higher than its siblings with child cards, so I'd expect it to have a card.
I don't see these as blocking issues. This was using the "Collection Card" stories.

PR #7078, opened 2024-09-24
Export pressScale and focusRing utilities from S2

devongovett:
A lot of import changes but this is just exporting `pressScale` (from `@react-spectrum/s2` since it is used at runtime) and `focusRing` from (`@react-spectrum/s2/style` since it is a macro). We could export additional style macro utils from here in the future.


packages/@react-spectrum/s2/src/style-utils.ts
```diff
@@ -13,16 +13,6 @@
 import {CSSProperties} from 'react';
 import {StyleString} from '../style/types';
```
ktabors, line 15: Should we also expose` getAllowedOverrides()` and `styleProps` from this file?

ktabors (approved):

snowystinger (approved):

packages/@react-spectrum/s2/src/style-utils.ts
devongovett: I think we'll keep those internal for now. It's pretty easy to pass in an array to the `style` macro, and different components may have different styles that they want to allow. I exported the StyleString type here so you can use that for now.

ktabors:
What about centerbaseline()? If logic button had an icon it would need that.

reidbarber (dismissed):

dannify:
> What about centerbaseline()? If logic button had an icon it would need that.

No icons in logic button that we know of.

devongovett:
hmm I'm not sure LogicButton supports icons. We can consider exporting `<CenterBaseline>` in the future potentially.

dannify (dismissed):

ktabors:
> hmm I'm not sure LogicButton supports icons. We can consider exporting `<CenterBaseline>` in the future potentially.

Sorry for the confusion, I meant to say it doesn't support icons, but another component might have both icon and text and we'd need `<CenterBaseline />` for it.

snowystinger (approved):

ktabors (approved):

PR #7085, opened 2024-09-25
fix(Card): fix undefined id in Card

reidbarber:
`undefined` was getting cast as a string and used as an id

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Check Card/CardView stories and ensure no undefined id's

### DIFF
--- a/packages/@react-spectrum/s2/src/Card.tsx
+++ b/packages/@react-spectrum/s2/src/Card.tsx
@@ -20,7 +20,7 @@ import {ContentContext, FooterContext, TextContext} from './Content';
 import {createContext, CSSProperties, forwardRef, ReactNode, useContext} from 'react';
 import {DividerContext} from './Divider';
 import {DOMProps, DOMRef, DOMRefValue} from '@react-types/shared';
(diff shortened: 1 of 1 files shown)


packages/@react-spectrum/s2/src/Card.tsx
```diff
@@ -382,6 +382,7 @@ export const Card = forwardRef(function Card(props: CardProps, ref: DOMRef<HTMLD
   [props] = useSpectrumContextProps(props, ref, CardContext);
   let domRef = useDOMRef(ref);
   let {density = 'regular', size = 'M', variant = 'primary', UNSAFE_className = '', UNSAFE_style, styles, id, ...otherProps} = props;
+  id = useId(props.id ? String(props.id) : undefined);
```
devongovett, line 385: Do we need the `useId`? Could just do `id={props.id != null ? String(props.id) : undefined}` (null check as well in case `id` is zero.

reidbarber: makes sense

devongovett (approved):

snowystinger (approved):

PR #7086, opened 2024-09-25
Fix Picker and ComboBox press state, Table types + docs

devongovett:
Noticed that Picker and ComboBox were getting stuck in the down state when open. That's due to RAC setting isPressed when they are open. Not sure what changed since the last release where this didn't happen, but it does now...

Fixed the storybook controls for Table not appearing correctly (it doesn't seem to like `forwardRefType`, which was unnecessary in this case. only needed when there are generics). Also exported Table components and props from the package and fixed the props for TableHeader to exclude className and style.


snowystinger (approved):
bummer about the pressed state :-/

reidbarber (approved):

PR #7087, opened 2024-09-25
S2 Table optimizations

devongovett:
* Removes extra element for cell background, which was needed for sticky cells due to transparency. The replacement is to use CSS `color-mix` to create an opaque version of the color we need by mixing it with the `gray-25` background color using the opacity as the percentage to mix. Lightning CSS will compile this down to an actual hex color for us.
* Only render the extra element for the cell focus ring when the cell is actually focused.
* Small fix for isQuiet: there should be an extra border on the top of the column header row in that case.


snowystinger (commented):
Table's props in the controls section seems to have disappeared?

reidbarber (dismissed):

devongovett:
They're already gone on main. I fixed it in https://github.com/adobe/react-spectrum/pull/7086

snowystinger (approved):
ahhh, i thought I'd checked main, but i just grabbed the comment from your PR which was why I thought it was working
sorry about that, thanks

snowystinger (approved):

reidbarber (approved):

PR #7089, opened 2024-09-26
Remove S2 prefix from docs storybook

yihuiliao:
Adds chromatic stories for all our S2 components and removes the S2 prefix so that our S2 stories don't fall under that heading.

We still need to improve our S2 chromatic story coverage since I mostly just copy and pasted from our docs storybook but figured we can get this in first and then work on that.

## ✅ Pull Request Checklist:


.chromatic/main.js
```diff
@@ -6,7 +6,6 @@ module.exports = {
   },
   stories: [
     '../packages/**/chromatic/**/*.stories.@(js|jsx|ts|tsx)',
-    '../packages/@react-spectrum/s2/stories/*.stories.@(js|jsx|mjs|ts|tsx)',
```
yihuiliao, line 9: removed this since we don't need to run chromatic on our s2 docs storybook anymore

yihuiliao:
question about this line:
https://github.com/adobe/react-spectrum/blob/246225b096442a39f1f91b351f4ff88d9707ca2d/.chromatic-fc/main.js#L9

now that we've gotten rid of the S2 prefix, all the S1 and S2 story's get jumbled together so it's hard to tell which is which. Do we want to create a separate chromatic-fc for S2 or run it on our S2 chromatic storybook or leave it like this for now?

packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
```diff
@@ -20,18 +20,17 @@ import './unsafe.css';
...
   component: ActionButton,
   parameters: {
-    layout: 'centered'
-    // TODO: uncomment when baseline for new S2 chromatic stories is accepted since these are resused in the chromatic stories
-    // chromatic: {
-    //   disableSnapshot: true
-    // }
+    layout: 'centered',
+    chromatic: {
+      disableSnapshot: true
```
yihuiliao, line 25: do we still need this `disableSnapshot` if we don't plan to run chromatic on the normal docs stories?

devongovett: I don't think we should need it anymore

snowystinger:
> question about this line:
> 
> https://github.com/adobe/react-spectrum/blob/246225b096442a39f1f91b351f4ff88d9707ca2d/.chromatic-fc/main.js#L9
> 
> now that we've gotten rid of the S2 prefix, all the S1 and S2 story's get jumbled together so it's hard to tell which is which. Do we want to create a separate chromatic-fc for S2 or run it on our S2 chromatic storybook or leave it like this for now?

I think we should change it to run against our chromatic stories instead?

devongovett:
Can we separate the docs and these other examples for style macro? I'd like to just have all the docs at the top without needing to open a sub-folder.

image

Well should probably just be a real component anyway. It's actually quite a bit of styles for everyone to paste. And since these examples don't show the code, you'd need to find the story source in order to copy it so not sure how useful it is to show here.

devongovett:
Oh, never mind, these stories are excluded from the docs anyway.

reidbarber:
@devongovett the Well story was just added for us to verify of the codemod styles used to replace it.

devongovett (approved):

reidbarber (approved):

PR #7113, opened 2024-09-27
Centralize S2 icon wrapper

devongovett:
Realized we had a bug in the S2 icon transformer: it generated imports for SkeletonWrapper and useSkeletonIcon internal utilities that weren't exported, so couldn't be used in custom icons. Instead of exporting these, I added a single `createIcon` helper function that centralizes all of the icon logic in one place. This has the benefit that we export fewer internals, and also reduces bundle size since we aren't duplicating the same code in every icon.

### DIFF
--- a/packages/@react-spectrum/s2/src/Icon.tsx
+++ b/packages/@react-spectrum/s2/src/Icon.tsx
@@ -11,10 +11,12 @@
  */
 
 import {AriaLabelingProps, DOMProps} from '@react-types/shared';
+import {ComponentType, Context, createContext, FunctionComponent, ReactNode, SVGProps, useRef} from 'react';
 import {ContextValue, SlotProps} from 'react-aria-components';
-import {createContext, ReactNode} from 'react';
+import {SkeletonWrapper, useSkeletonIcon} from './Skeleton';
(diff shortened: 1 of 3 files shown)


packages/@react-spectrum/s2/src/Icon.tsx
```diff
@@ -31,3 +33,45 @@ export interface IllustrationContextValue extends IconContextValue {
 
 export const IconContext = createContext<ContextValue<IconContextValue, SVGElement>>({});
 export const IllustrationContext = createContext<ContextValue<IllustrationContextValue, SVGElement>>({});
+
+export function createIcon(Component: ComponentType<SVGProps<SVGSVGElement>>, context: Context<ContextValue<IconContextValue, SVGElement>> = IconContext): FunctionComponent<IconProps> {
+  return (props: IconProps) => {
+    let ref = useRef<SVGElement>(null);
+    let ctx;
+    // TODO: remove this default once we release RAC and use DEFAULT_SLOT.
```
devongovett, line 41: Copied from old transformer. We should actually change this but not right before release.

reidbarber (approved):

dannify (approved):

PR #7117, opened 2024-09-27
feat: add TableView codemods

reidbarber:
- Updates `key` to be `id`
- Copies `columns` prop from `TableHeader` to  to `Row`
- Updates `Row` render function to pass column object, not just key
- Leaves a comment if nested columns are used (not supported yet)

## ✅ Pull Request Checklist:


devongovett:
gonna wait until after release for this one so we can test in more apps

packages/dev/codemods/src/s1-to-s2/__tests__/table.test.ts
```diff
@@ -6,9 +6,92 @@ const test = (name: string, input: string) => {
...
+test('Static TableView', `
+import {Cell, Column, Row, TableBody, TableHeader, TableView}  from '@adobe/react-spectrum';
 
-<div>
-</div>
+<TableView>
+  <TableHeader>
+    <Column key="test">Test</Column>
+    <Column title="Blah">
+      <Column title="Group 1">
```
devongovett, line 16: I think RAC/S2 doesn't support nested columns yet.

packages/dev/codemods/src/s1-to-s2/src/codemods/transforms.ts
```diff
@@ -946,6 +946,123 @@ function updateLegacyLink(
...
+
+        // Replace references to the old parameter name inside the function body
+        arrowFuncPath.get('body').traverse({
+          Identifier(innerPath) {
+            if (
+              innerPath.node.name === paramName &&
+              // Ensure we're not replacing the parameter declaration
+              innerPath.node !== params[0]
+            ) {
+              // Replace with 'column.id'
```
devongovett, line 1018: I guess this is sort of an assumption that columns have an `id` property? They may not, it could be `key` or something computed. Probably won't be able to be 100% accurate here. I guess if we wanted to get fancy we could try to look at the `id` (previously `key`) prop of the `Column` component? e.g. `<Column key={column.name}>` and then use `column.name` here too.

packages/dev/codemods/src/s1-to-s2/src/codemods/transforms.ts
```diff
@@ -946,6 +946,123 @@ function updateLegacyLink(
...
+        });
+      }
+    }
+  }
+}
+
+/**
+ * Updates the key prop to id. Keeps the key prop if it's used in an array.map function.
+ */
+function updateKeyToId(
```
devongovett, line 1033: Did we not already have something like this for other collection components like `Item`?

packages/dev/codemods/src/s1-to-s2/src/codemods/transforms.ts
reidbarber: Hopefully [0fc3555](https://github.com/adobe/react-spectrum/pull/7117/commits/0fc355579c2759f02f7eb2e8114a935150897b63) should suffice

packages/dev/codemods/src/s1-to-s2/src/codemods/transforms.ts
reidbarber: Good catch, updated the other usage of this to use this function [0d5c4ce](https://github.com/adobe/react-spectrum/pull/7117/commits/0d5c4ce4de271f2621a97f2d65519f519160635b)

packages/dev/codemods/src/s1-to-s2/__tests__/table.test.ts
reidbarber: Will leave a leave a comment when this is found for now: [a752dea](https://github.com/adobe/react-spectrum/pull/7117/commits/a752dea3b45c702e72e53e2b648aaf0344253556)

LFDanLu (commented):
spot checked the logic but currently running into some issues running the codemod locally sucessfully (using `node ../../packages/dev/codemods/dist/index.js s1-to-s2` after building the codemod locally in some of our example apps but nothing seems to get transformed)

.storybook-s2/docs/Migrating.jsx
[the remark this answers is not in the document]
reidbarber: Good idea, just added comments for those.

devongovett (approved):

yihuiliao (approved):
ran it on rsp-cra-18 and seemed to change the tableview correctly

PR #7118, opened 2024-09-28
Rename Table -> TableView, skip some codemods for now

devongovett:
Renames S2 Table back to TableView to match v3 and our other collection components.

Skips some codemods for components that aren't ready yet. These were getting their imports replaced even though the codemods aren't implemented yet.


devongovett:
no idea why docs-verdaccio decided to run. ignore it.

reidbarber (approved):

snowystinger (approved):

PR #7135, opened 2024-10-01
Temporary fix for S2 storybook crash when showing code in dark mode

LFDanLu:
Reported in slack

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Go to S2 storybook, switch to dark mode, and open "show code" for any component example in the docs

## 🧢 Your Project:

RSP

### DIFF
--- a/.storybook-s2/preview.tsx
+++ b/.storybook-s2/preview.tsx
@@ -27,7 +27,8 @@ const preview = {
           channel.on(DARK_MODE_EVENT_NAME, setDark);
           return () => channel.removeListener(DARK_MODE_EVENT_NAME, setDark);
         }, []);
-        return <DocsContainer {...props} theme={{...(dark ? themes.dark : themes.light), appContentBg: 'var(--s2-container-bg)'}} />;
+        var style = getComputedStyle(document.body)
+        return <DocsContainer {...props} theme={{...(dark ? themes.dark : themes.light), appContentBg: style.getPropertyValue('--s2-container-bg').trim()}} />;
       },
       source: {
         // code: null, // Will disable code button, and show "No code available"
@@ -68,7 +69,7 @@ const preview = {
       storySort: {
         order: ['Intro', 'Style Macro', 'Workflow Icons', 'Illustrations', 'Migrating', 'Release Notes'],
         method: 'alphabetical'
-      }  
+      }
     }
   },
   argTypes: {


.storybook-s2/preview.tsx
```diff
@@ -27,7 +27,8 @@ const preview = {
           channel.on(DARK_MODE_EVENT_NAME, setDark);
           return () => channel.removeListener(DARK_MODE_EVENT_NAME, setDark);
         }, []);
-        return <DocsContainer {...props} theme={{...(dark ? themes.dark : themes.light), appContentBg: 'var(--s2-container-bg)'}} />;
+        var style = getComputedStyle(document.body)
+        return <DocsContainer {...props} theme={{...(dark ? themes.dark : themes.light), appContentBg: style.getPropertyValue('--s2-container-bg').trim()}} />;
```
LFDanLu, line 31: Something in the Storybook DocsContainer doesn't parse the css variable properly, instead passing the whole `var(--s2-container-bg)` string to an RGB parser instead of a hex value/etc. This code path doesn't seem to be triggered at all when opening the code sample in light mode so there seems to be some intermediate code determining what color to set for the code sample itself

Alternatively we could revert to `return <DocsContainer {...props} theme={dark ? themes.dark : themes.light} />;` but this makes the example's background not quite the correct container background color.

reidbarber (approved):

.storybook-s2/preview.tsx
LFDanLu: Looks like this is why: https://github.com/storybookjs/storybook/blob/70d8f4d97069d9e901a5c0f6ddefa955d7bbd2b8/code/lib/blocks/src/components/Preview.tsx#L87-L91

Only in dark mode does it use the provided provided content background color

devongovett: won't this break if the variable changes value (e.g. when switching themes)?

LFDanLu: sorry must have missed the github notification for this comment. Just to clarify you mean when `--s2-container-bg` changes value due to the theme change, this appContentBG won't be up to date?

devongovett: Yeah. I guess it works because the other event handler that sets `document.documentElement.dataset.colorScheme` happens before the one that triggers this re-render which seems kinda brittle but I guess it's ok for now.

devongovett (approved):

PR #7163, opened 2024-10-09
Add optional shouldFocusOnHover prop to aria combobox

p1pah:
No issue to close, but solves this discussion! https://github.com/adobe/react-spectrum/discussions/7045

## ✅ Pull Request Checklist:


devongovett:
Wouldn't this result in there being multiple highlighted items at once? e.g. one that you focused with the keyboard and one that you hovered over. By focusing the item on hover we ensure that there is only ever one highlighted item at a time. That matches the behavior of native `<select>`, menus, etc. Is the behavior you're looking to disable actually the selecting of an item on Tab rather than focusing on hover?

p1pah:
@devongovett I'm looking to stop a use case where the user opens the menu, hovers over a item and then they decide tab out. Some of our users have found it confusing that if they are just hovering over something it's automatically selecting it for them :/. With the two highlighted items question, I think we would approach that with two different color shades (one to indicate keyboard selection and one to show mouse hover). This optional prop fixes all our woes our users were having with the mouse hover 🤭! Please let me know what you think

p1pah:
@devongovett Something like this :) https://ariakit.org/components/combobox

p1pah:
Any chance this could be considered/merged into next release 😄 ?

p1pah:
I hear you, but that is the exact thing I want haha "only preventing committing on tab if you previously hovered an item instead of using the keyboard.". With the approach of committing vs. not committing on tab, would one just pass the current focused element and detect whether it's focused/hovered via data attributes?

PR #7175, opened 2024-10-11
Docs: useListData.mdx: Fix `ListBoxItem` name

tordans:
This updates the docs which mentioned a `<List>` Component which appears to not exist. Instead it probably should be `ListBoxItem`.

I assume the Component was renamed at some point? The Docs  https://react-spectrum.adobe.com/react-aria/ListBox.html only mention `ListBoxItem`.

---

I would also recommend to add
```
import { ListBox, ListBoxItem } from 'react-aria-components';
import { useListData } from 'react-stately';
```

to the top of the code block, but since other code blocks did not have this, I left it out for now.

### DIFF
--- a/packages/@react-stately/data/docs/useListData.mdx
+++ b/packages/@react-stately/data/docs/useListData.mdx
@@ -75,7 +75,7 @@ let list = useListData({
   items={list.items}
   selectedKeys={list.selectedKeys}
   onSelectionChange={list.setSelectedKeys}>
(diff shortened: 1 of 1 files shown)


tordans:
I see now those imports are likely from https://react-spectrum.adobe.com/react-spectrum/ListView.html#listview
> `import {Item, ListView} from '@adobe/react-spectrum'`

I will close this. However, I would suggest to add the import lines to the example so this confusion does not come up for others. Let me know if I should create a PR for that.

Issue #7176, opened 2024-10-11
`useListData` fails with `Could not determine key for item`


tordans:
### Provide a general summary of the issue here

It looks like there is a regression in `useListData` which now requires the `key` to be part of the supplied values.

However, [the docs](https://react-spectrum.adobe.com/react-stately/useListData.html#options) suggest that the `key` can be generated dynamically with `getKey`.

The example given at https://react-spectrum.adobe.com/react-stately/collections.html#updating-data does not work.

Here is a test case: https://stackblitz.com/edit/vitejs-vite-vlh6df?file=src%2FApp.tsx

- Only when a `key` or `id` prop is provided, does the app render
- Adding `getKey` does not prevent the error 
    ```
    Uncaught Error: Could not determine key for item
    at react-aria-components.js?v=d5c21c8c:665:34
    at mountMemo (chunk-M324AGAM.js?v=4aa0b7af:12192:27)
    at Object.useMemo (chunk-M324AGAM.js?v=4aa0b7af:12516:24)
    at useMemo (chunk-I4MZPW7S.js?v=4aa0b7af:1116:29)
    ```
- It looks like the `(item) => <ListBoxItem>{item.name}</ListBoxItem>` does not require a `key` or `id` as suggested by other code examples in the docs; those should be updated

### 🤔 Expected Behavior?

It should be possible to use `getKey` on `useListData` to defined a key and provide items that do not have a `key` or `id` prop.

### 😯 Current Behavior

See above

### 💁 Possible Solution

The only reference to this I could find is https://github.com/adobe/react-spectrum/discussions/4431#discussioncomment-5714670 from April 2023 which hints that the current logic is flawed.

### 🖥️ Steps to Reproduce

1. open https://stackblitz.com/edit/vitejs-vite-vlh6df?file=src%2FApp.tsx
2. => see the app renders
3. remove the comment from `comment this in and the app fails:` 
4. => the app fails with the given error message
5. remove the comment from `getKey: (item) => item.name,`
6. => the app still fails with the same error message

### Version

    "react-aria-components": "^1.4.0",         "react-stately": "^3.33.0"

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

Mac

tordans:
I will close this.

It looks like the issue was, that I did not have an `id`/`key` prop on the Item as suggested in https://github.com/adobe/react-spectrum/pull/7177 
[...]

PR #7177, opened 2024-10-11
Doc: collections.mdx: Remove `key` prop

tordans:
My understanding from how those `{(item) => <Component>{item.name}</Component>}` work is, that the key prop is not required or recommended on this line but instead will be added automatically based on the item `key` or `id` prop or `getKey` callback.

This PR removes the key prop from the examples which makes them easier to read and understand.

See also  
- https://github.com/adobe/react-spectrum/issues/7176
- https://github.com/adobe/react-spectrum/discussions/4431#discussioncomment-5714670

---

~~Another thing I am confused about on this docs page is, what the `<Item>` component is. I assume it should be renamed, see https://github.com/adobe/react-spectrum/pull/7175~~

---

Another thing I am confused about is the mention of [`@react-stately/data`](https://www.npmjs.com/package/@react-stately/data) vs. [`react-stately`](https://www.npmjs.com/package/react-stately) especially in places like https://react-spectrum.adobe.com/react-stately/collections.html#asynchronous-loading: 
> image

### DIFF
--- a/packages/@react-stately/collections/docs/collections.mdx
+++ b/packages/@react-stately/collections/docs/collections.mdx
@@ -145,7 +145,7 @@ let [animals, setAnimals] = useState([
 ]);
 
 <ListBox items={animals}>
-  {item => <Item key={item.name}>{item.name}</Item>}
+  {item => <Item>{item.name}</Item>}
 </ListBox>
 ```
 
@@ -168,7 +168,7 @@ let [animals, setAnimals] = useState([
 
 <ListBox>
   {animals.map(item =>
-    <Item key={item.name}>{item.name}</Item>
+    <Item>{item.name}</Item>
   )}
 </ListBox>
 ```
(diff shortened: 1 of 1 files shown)


devongovett:
I don't think this is right. In those examples the item objects don't have a key or id property, so one does need to be supplied.

tordans:
@devongovett thanks. I might very well misunderstood how this works. From what you are saying I understand that the system works like this…

- the item component needs a key or id prop (one of them, not both)
- the key/id can be provided by the data, in which case they will be added automatically
- for this to work the item component needs to be a `Item` or `GridListItem` (or similar) which add those capabilities to add the key automatically
- if the data does not have key/id, it is enough to add it as props to the item component
- an alternative to both methods of providing a key is to use [`getKey`](https://react-spectrum.adobe.com/react-stately/useListData.html#options) (???)

This description sounds good to me but I am quite sure it is not how it is working ATM.  I have a demo in https://github.com/adobe/react-spectrum/issues/7176 
I will be happy to test it out more and look for a way to clarify the docs given some hints / guidance.

tordans:
I gave up getting this setup to work with my internal [nuqs](https://nuqs.47ng.com/) base state and switched to [Framer Motion Reorder](https://www.framer.com/motion/reorder/). 

Therefore, I will close this given I will not have the time to look into it more. Maybe someone will search for it and can share their results at some point …

PR #7179, opened 2024-10-11
Update S2 Accordion API

yihuiliao:
Closes 

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/s2/src/Disclosure.tsx
+++ b/packages/@react-spectrum/s2/src/Disclosure.tsx
@@ -93,7 +93,7 @@ function Disclosure(props: DisclosureProps, ref: DOMRef<HTMLDivElement>) {
 let _Disclosure = forwardRef(Disclosure);
 export {_Disclosure as Disclosure};
 
-export interface DisclosureHeaderProps extends UnsafeStyles, DOMProps {
+export interface DisclosureTriggerProps extends UnsafeStyles, DOMProps {
   /** The heading level of the disclosure header.
    * 
    * @default 3
@@ -103,8 +103,13 @@ export interface DisclosureHeaderProps extends UnsafeStyles, DOMProps {
   children: React.ReactNode
 }
 
+interface DisclosureHeaderProps extends UnsafeStyles, DOMProps {
+  children?: React.ReactNode
+}
+
 const headingStyle = style({
-  margin: 0
+  margin: 0,
+  flexGrow: 1
 });
 
 const buttonStyles = style({
@@ -195,7 +200,31 @@ const chevronStyles = style({
   flexShrink: 0
 });
 
+
 function DisclosureHeader(props: DisclosureHeaderProps, ref: DOMRef<HTMLDivElement>) {
+  let {
+    UNSAFE_className,
+    UNSAFE_style,
+    children
+  } = props
+  let domRef = useDOMRef(ref);
+  return (
+    <div
+      style={UNSAFE_style}
+      className={(UNSAFE_className ?? '') + style({display: 'flex', alignItems: 'center', gap: 4})}
+      ref={domRef}>
+      {children}
+    </div>
+  )
+}
+
+/**
+ * A header for a disclosure. Contains a heading and a trigger button to expand/collapse the panel.
+ */
+let _DisclosureHeader = forwardRef(DisclosureHeader);
+export {_DisclosureHeader as DisclosureHeader};
+
+function DisclosureTrigger(props: DisclosureTriggerProps, ref: DOMRef<HTMLDivElement>) {
   let {
     level = 3,
     UNSAFE_style,
@@ -228,8 +257,8 @@ function DisclosureHeader(props: DisclosureHeaderProps, ref: DOMRef<HTMLDivEleme
 /**
  * A header for a disclosure. Contains a heading and a trigger button to expand/collapse the panel.
  */
-let _DisclosureHeader = forwardRef(DisclosureHeader);
-export {_DisclosureHeader as DisclosureHeader};
+let _DisclosureTrigger = forwardRef(DisclosureTrigger);
+export {_DisclosureTrigger as DisclosureTrigger};
 
 export interface DisclosurePanelProps extends Omit<RACDisclosurePanelProps, 'className' | 'style' | 'children'>, UnsafeStyles, DOMProps, AriaLabelingProps {
   children: React.ReactNode
--- a/packages/@react-spectrum/s2/stories/Accordion.stories.tsx
+++ b/packages/@react-spectrum/s2/stories/Accordion.stories.tsx
@@ -10,10 +10,12 @@
  * governing permissions and limitations under the License.
  */
 
-import {Accordion, Disclosure, DisclosureHeader, DisclosurePanel, TextField} from '../src';
+import {Accordion, ActionButton, Disclosure, DisclosureHeader, DisclosurePanel, DisclosureTrigger, TextField} from '../src';
 import type {Meta, StoryObj} from '@storybook/react';
+import NewIcon from '../s2wf-icons/S2_Icon_New_20_N.svg';
 import React from 'react';
 import {style} from '../style' with { type: 'macro' };
+
 
 const meta: Meta<typeof Accordion> = {
   component: Accordion,
@@ -34,7 +36,9 @@ export const Example: Story = {
         <Accordion {...args}>
           <Disclosure id="files">
             <DisclosureHeader>
-              Files
+              <DisclosureTrigger>
+                Files
+              </DisclosureTrigger>
             </DisclosureHeader>
             <DisclosurePanel>
               Files content
@@ -42,7 +46,9 @@ export const Example: Story = {
           </Disclosure>
           <Disclosure id="people">
             <DisclosureHeader>
-              People
+              <DisclosureTrigger>
+                People
+              </DisclosureTrigger>
             </DisclosureHeader>
             <DisclosurePanel>
               <TextField label="Name" styles={style({maxWidth: 176})} />
@@ -61,7 +67,9 @@ export const WithLongTitle: Story = {
         <Accordion styles={style({maxWidth: 224})} {...args}>
           <Disclosure>
             <DisclosureHeader>
-              Files
+              <DisclosureTrigger>
+                Files
+              </DisclosureTrigger>
             </DisclosureHeader>
             <DisclosurePanel>
               Files content
@@ -69,7 +77,9 @@ export const WithLongTitle: Story = {
           </Disclosure>
           <Disclosure>
             <DisclosureHeader>
-              People
+              <DisclosureTrigger>
+                People
+              </DisclosureTrigger>
             </DisclosureHeader>
             <DisclosurePanel>
               People content
@@ -77,7 +87,9 @@ export const WithLongTitle: Story = {
           </Disclosure>
           <Disclosure>
             <DisclosureHeader>
-              Very very very very very long title that wraps
+              <DisclosureTrigger>
+                Very very very very very long title that wraps
+              </DisclosureTrigger>
             </DisclosureHeader>
             <DisclosurePanel>
               Accordion content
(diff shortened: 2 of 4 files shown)


packages/@react-spectrum/s2/src/Disclosure.tsx
```diff
@@ -195,7 +200,31 @@ const chevronStyles = style({
...
+    UNSAFE_style,
+    children
+  } = props;
+  let domRef = useDOMRef(ref);
+  return (
+    <div
+      style={UNSAFE_style}
+      className={(UNSAFE_className ?? '') + style({display: 'flex', alignItems: 'center', gap: 4})}
+      ref={domRef}>
+      {children}
```
reidbarber, line 216: An option for shorthand: if `typeof children === 'children'`, we could automatically wrap `children` in a `DisclosureTrigger`, which would let you use our existing API as well.

No need to implement this now, we could discuss it later

reidbarber (dismissed):
API and styles look good.

yihuiliao:
I'll update the chromatic stories as well once we settle on the API that we like

packages/@react-spectrum/s2/src/Disclosure.tsx
```diff
@@ -195,7 +200,31 @@ const chevronStyles = style({
...
+    </div>
+  );
+}
+
+/**
+ * A header for a disclosure. Contains a heading and a trigger button to expand/collapse the panel.
```
ktabors, line 222: Concerned that the header and trigger have the same description. I'm not sure we want to call this a wrapper for the trigger and an interactive element.

packages/@react-spectrum/s2/src/Disclosure.tsx
```diff
@@ -228,8 +257,8 @@ function DisclosureHeader(props: DisclosureHeaderProps, ref: DOMRef<HTMLDivEleme
 /**
  * A header for a disclosure. Contains a heading and a trigger button to expand/collapse the panel.
```
ktabors, line 258: Maybe "A disclosure header consisting of a..."

packages/@react-spectrum/s2/stories/Accordion.stories.tsx
```diff
@@ -128,3 +144,35 @@ WithDisabledDisclosure.parameters = {
...
+  render: (args) => {
+    return (
+      <div className={style({minHeight: 240})}>
+        <Accordion {...args}>
+          <Disclosure id="files">
+            <DisclosureHeader>
+              <DisclosureTrigger>
+                Files
+              </DisclosureTrigger>
+              <ActionButton><NewIcon /></ActionButton>
```
ktabors, line 157: aria-label?

packages/@react-spectrum/s2/src/Disclosure.tsx
ktabors, line 216: Do we need a context to disable, tshirt size, or quiet children?
image

ktabors (commented):
This is great! Just a few observations.

Density compact seems to have a focus ring issue. 
image

packages/@react-spectrum/s2/stories/Accordion.stories.tsx
```diff
@@ -34,15 +36,19 @@ export const Example: Story = {
         <Accordion {...args}>
           <Disclosure id="files">
             <DisclosureHeader>
-              Files
+              <DisclosureTrigger>
```
snowystinger, line 39: This pattern is opposite to our other triggers
`<DialogTrigger>` is the outermost, whereas here, `DisclosureTrigger` is the innermost

packages/@react-spectrum/s2/src/Disclosure.tsx
```diff
@@ -195,7 +200,31 @@ const chevronStyles = style({
...
+
+/**
+ * A header for a disclosure. Contains a heading and a trigger button to expand/collapse the panel.
+ */
+let _DisclosureHeader = forwardRef(DisclosureHeader);
+export {_DisclosureHeader as DisclosureHeader};
+
+function DisclosureTrigger(props: DisclosureTriggerProps, ref: DOMRef<HTMLDivElement>) {
   let {
     level = 3,
```
snowystinger, line 229: I find it a little confusing that the level is defined on the Trigger, not on the Header

packages/@react-spectrum/s2/src/Disclosure.tsx
```diff
@@ -103,8 +103,13 @@ export interface DisclosureHeaderProps extends UnsafeStyles, DOMProps {
   children: React.ReactNode
 }
 
+interface DisclosureHeaderProps extends UnsafeStyles, DOMProps {
+  children?: React.ReactNode
```
snowystinger, line 107: I don't think this is optional?

packages/@react-spectrum/s2/stories/Accordion.stories.tsx
devongovett: That's because the header can contain elements that are not part of the trigger (eg adjacent buttons).

packages/@react-spectrum/s2/src/Disclosure.tsx
snowystinger: I just caught up on standup, looks like y'all had some similar misgivings
I'm more in favor of `DisclosureTitle` or something more along those lines

packages/@react-spectrum/s2/src/Disclosure.tsx
yihuiliao: that's probably leftover from copy pasting. thanks for pointing it out! will update it

packages/@react-spectrum/s2/src/Disclosure.tsx
yihuiliao: oh yes, good catch!

packages/@react-spectrum/s2/stories/Accordion.stories.tsx
snowystinger: I understand that, but it also doesn't include the element that it opens like our other triggers
See my other concerns in my next comment down about `level`

devongovett:
My only question is whether we want to have a shortcut so you don't need to write `<DisclosureHeader><DisclosureTitle>Title</DisclosureTitle></DisclosureHeader>` in the common case where there are no sibling buttons. We could potentially have just `<DisclosureTitle>` and automatically wrap in a header if it isn't already in one, similar to what we do with `<Modal>` and `<ModalOverlay>`. There may be other options as well. Anyone else have input?

devongovett:
> It feels opposite to button and DisclosureHeader should wrap a string child in the DisclosureTitle automatically

But then as soon as you put a `<span>` or `<FormatMessage>` or something it wouldn't work right?

snowystinger:
> > It feels opposite to button and DisclosureHeader should wrap a string child in the DisclosureTitle automatically
> 
> But then as soon as you put a `<span>` or `<FormatMessage>` or something it wouldn't work right?

This is true. I guess I'm ok with it. I don't have a better idea.

packages/@react-spectrum/s2/src/ActionButton.tsx
```diff
@@ -181,6 +181,7 @@ function ActionButton(props: ActionButtonProps, ref: FocusableRef<HTMLButtonElem
   let domRef = useFocusableRef(ref);
   let overlayTriggerState = useContext(OverlayTriggerStateContext);
 
+  console.log(props.size);
```
snowystinger, line 184: remove?

packages/@react-spectrum/s2/src/Disclosure.tsx
```diff
@@ -195,7 +201,53 @@ const chevronStyles = style({
...
+
+  let mapSize = {
+    S: 'XS',
+    M: 'S',
+    L: 'M',
+    XL: 'L'
+  };
+
+  // maps to one size smaller in the compact density to ensure there is space between the top and bottom of the action button and container
+  // @ts-ignore
```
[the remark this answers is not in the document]
snowystinger: what was the error?

packages/@react-spectrum/s2/src/ActionButton.tsx
yihuiliao: yep i got it in a commit, just haven't pushed it up yet

snowystinger (commented):
**UPDATE** seems to have been a cache issue, I can't reproduce in incognito

Seeing an error when I try to visit the storybook build https://reactspectrum.blob.core.windows.net/reactspectrum/844be69102b36467a58c5be2e1aaee9e90d50c48/storybook/index.html?path=/story/accordion--default&providerSwitcher-express=false

Not seeing the error on Main, but this only have changes for s2... so I'm not sure what has changed.

snowystinger (dismissed):
are we doing the auto wrapping in a followup?

yihuiliao:
> are we doing the auto wrapping in a followup?

i'll just do it in this pr. i'd like to get the api figured out so that i can update the chromatic stories too

packages/@react-spectrum/s2/src/Disclosure.tsx
```diff
@@ -195,7 +201,52 @@ const chevronStyles = style({
...
+  // maps to one size smaller in the compact density to ensure there is space between the top and bottom of the action button and container
+  let newSize : 'XS' | 'S' | 'M' | 'L' | 'XL' | undefined = size;
+  if (density === 'compact') {
+    newSize = mapSize[size ?? 'M'] as 'XS' | 'S' | 'M' | 'L';
+  }
+
+  return (
+    <Provider
+      values={[
+        [ActionButtonContext, {size: newSize, isQuiet, isDisabled}]
```
ktabors, line 231: ActionButton isn't disabled for some reason. 
image

yihuiliao: ah yeah, that's because isDisabled doesn't get passed to the DisclosureContext from Accordion. that said, from testing yesterday, i believe we said that we don't want to disable the action button even when disclosure is disabled. if they want to disable it, they can pass disable directly to the component. 

im not sure if that's the same behavior we want to keep when the whole accordion is disabled though. thoughts?

packages/@react-spectrum/s2/src/Disclosure.tsx
```diff
@@ -208,28 +260,50 @@ function DisclosureHeader(props: DisclosureHeaderProps, ref: DOMRef<HTMLDivEleme
   let {isExpanded} = useContext(DisclosureStateContext)!;
   let {size, density, isQuiet} = useSlottedContext(DisclosureContext)!;
   let isRTL = direction === 'rtl';
+
+  let ctx = useContext(InternalDisclosureHeader);
+  if (ctx) {
```
ktabors, line 265: The <Heading... /> look the same. I've looked twice and not seen a difference. Couldn't we do the following?
```
let heading = (<Heading ...>...</Heading>);
if (ctx) {
  return heading;
}

return (<<DisclosureHeader>{heading}</Heading>);
```

ktabors (commented):
Out of scope. Shouldn't there be an example with an expanded disclosure?

packages/@react-spectrum/s2/src/Disclosure.tsx
yihuiliao: yeah for whatever reason i couldn't get it to work but i think i was just missing something. i'll push up the changes soon!

snowystinger (commented):
looks like a style got lost somewhere?
https://www.chromatic.com/build?appId=5f0dd5ad2b5fc10022a2e320&number=807

actually, looks like I can't interact with/open the disclosure anymore for some of these

yihuiliao:
> looks like a style got lost somewhere? https://www.chromatic.com/build?appId=5f0dd5ad2b5fc10022a2e320&number=807
> 

oops i think that's because i forgot to update the chromatic stories to use `DisclosureTitle`. 

> actually, looks like I can't interact with/open the disclosure anymore for some of these

which one's specifically can you not interact with/open and which browser? im not able to reproduce

snowystinger:
This is the chromatic story where i couldn't open it https://5f0dd5ad2b5fc10022a2e320-ykjpggsppc.chromatic.com/?path=/story/s2-chromatic-disclosure--example which is explained by not using the right elements. It's fixed now with your updates.

snowystinger (approved):

reidbarber (approved):

PR #7199, opened 2024-10-15
WIP: Support hidden="until-found" in DisclosureGroup

reidbarber:
Still testing and looking at options

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-aria/disclosure/src/useDisclosure.ts
+++ b/packages/@react-aria/disclosure/src/useDisclosure.ts
@@ -59,7 +59,7 @@ export function useDisclosure(
   let supportsBeforeMatch = !isSSR && 'onbeforematch' in document.body;
 
   let handleBeforeMatch = useCallback((e: Event) => {
-    if (groupState && id !== null) {
+    if (groupState && id !== undefined) {
       if (groupState.allowsMultipleExpanded) {
         groupState.toggleKey(id);
       } else {
--- a/packages/@react-aria/disclosure/src/useDisclosure.ts
+++ b/packages/@react-aria/disclosure/src/useDisclosure.ts
@@ -11,8 +11,9 @@
  */
 
 import {AriaButtonProps} from '@react-types/button';
-import {DisclosureState} from '@react-stately/disclosure';
-import {HTMLAttributes, RefObject, useEffect} from 'react';
+import {DisclosureGroupState, DisclosureState} from '@react-stately/disclosure';
+import {HTMLAttributes, RefObject, useCallback, useEffect} from 'react';
+import {Key} from '@react-types/shared';
 import {useEvent, useId} from '@react-aria/utils';
 import {useIsSSR} from '@react-aria/ssr';
 
@@ -24,7 +25,9 @@ export interface AriaDisclosureProps {
   /** Whether the disclosure is expanded (controlled). */
   isExpanded?: boolean,
   /** Whether the disclosure is expanded by default (uncontrolled). */
-  defaultExpanded?: boolean
+  defaultExpanded?: boolean,
+  /** Unique id for the disclosure. Useful if used in a disclosure group. */
+  id?: Key
 }
 
 export interface DisclosureAria {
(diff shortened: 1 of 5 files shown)


packages/@react-aria/disclosure/src/useDisclosure.ts
```diff
@@ -24,7 +25,9 @@ export interface AriaDisclosureProps {
   /** Whether the disclosure is expanded (controlled). */
   isExpanded?: boolean,
   /** Whether the disclosure is expanded by default (uncontrolled). */
-  defaultExpanded?: boolean
+  defaultExpanded?: boolean,
+  /** Unique id for the disclosure. Useful if used in a disclosure group. */
```
snowystinger, line 29:
```suggestion
  /** Unique id for the disclosure. Required in disclosure groups. */
```
Or "Used in..." if it's not required

if it's required in DisclosureGroups though, maybe it's a good idea to make it always required? or if the DisclosureGroup context is detected, throw and error if an id isn't provided.

packages/@react-aria/disclosure/src/useDisclosure.ts
```diff
@@ -40,29 +43,47 @@ export interface DisclosureAria {
  * @param state - State for the disclosure, as returned by `useDisclosureState`.
  * @param ref - A ref for the disclosure content.
  */
-export function useDisclosure(props: AriaDisclosureProps, state: DisclosureState, ref?: RefObject<Element | null>): DisclosureAria {
+export function useDisclosure(
+  props: AriaDisclosureProps,
+  state: DisclosureState,
+  ref: RefObject<Element | null>,
+  groupState?: DisclosureGroupState
```
snowystinger, line 50: Did we discuss doing it the same way as Checkbox and CheckboxGroup? https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/Checkbox.tsx#L182 since they can either be standalone or in a group as well

snowystinger:
Controlled doesn't seem to quite work right? https://reactspectrum.blob.core.windows.net/reactspectrum/f0457c575ce0bc023f31532191d0d5e809ccc052/storybook-s2/index.html?path=/story/accordion--controlled-open
Do we just need to add the rest of the controlled logic to this one? otherwise I get into this state and can't close the first one.
Screenshot 2024-10-17 at 7 36 24 am

packages/@react-spectrum/s2/stories/Accordion.stories.tsx
```diff
@@ -130,6 +131,68 @@ WithDisabledDisclosure.parameters = {
...
+function ControlledAccordion(props) {
+  let [expandedKeys, setExpandedKeys] = React.useState<Set<Key>>(new Set(['people']));
+  return (
+    <div className={style({font: 'body', display: 'flex', flexDirection: 'column', gap: 8})}>
+      <Accordion
+        onExpandedChange={setExpandedKeys}
+        expandedKeys={expandedKeys}
+        {...props}>
+        <Disclosure id="files">
+          <DisclosureHeader>
```
yihuiliao, line 143: these need to be `DisclosureTitle` instead of `DisclosureHeader` now. same thing in the Disclosure stories too

packages/@react-aria/disclosure/test/useDisclosure.test.ts
```diff
@@ -172,8 +172,9 @@ describe('useDisclosure', () => {
     });
   
     expect(result.current.state.isExpanded).toBe(false);
-    expect(ref.current.getAttribute('hidden')).toBeNull();
-    expect(onExpandedChange).not.toHaveBeenCalled();
+    expect(ref.current.getAttribute('hidden')).toBe('until-found');
```
snowystinger, line 175: I listed a lot of test cases in my review, can we add them either here or in followup?

snowystinger (approved):
Tested in Chrome, appears to be working as expected.
 - Uncontrolled, opens when text is found
 - Uncontrolled allows multiple expanded, all of them open sequentially as expected and stay open
 - Disabled disclosures do not open even if text is in them
 - Controlled, opens when text is found, key is updated. Can open other panel and redo my actions
 - Controlled works with allows multiple expanded
 - Controlled always open does not appear to ever open the accordion with the text we're looking for


Any other cases I should be trying?

I did notice that some of the Accordion s2 stories have disclosures without `id`'s, not part of this PR, but I thought those were required?

devongovett (approved):

PR #7216, opened 2024-10-21
fix: Accordion/Disclosure release prep

reidbarber:
- SSR tests and more v3 test coverage
- more v3 stories
- updated v3 dependencies
- types/docs udpates

## ✅ Pull Request Checklist:


packages/@react-stately/disclosure/src/useDisclosureGroupState.ts
```diff
@@ -10,15 +10,21 @@
...
 
-export interface DisclosureGroupProps extends Expandable {
+export interface DisclosureGroupProps {
   /** Whether multiple items can be expanded at the same time. */
   allowsMultipleExpanded?: boolean,
   /** Whether all items are disabled. */
-  isDisabled?: boolean
+  isDisabled?: boolean,
+  /** The currently expanded keys in the group (controlled). */
+  expandedKeys?: Iterable<Key>,
```
reidbarber, line 23: Just updated these to remove "collection" wording from docs

packages/react-aria-components/src/Disclosure.tsx
```diff
@@ -181,7 +181,15 @@ function Disclosure(props: DisclosureProps, ref: ForwardedRef<HTMLDivElement>) {
   );
 }
 
-export interface DisclosurePanelProps extends RenderProps<{}> {
+export interface DisclosurePanelRenderProps {
+  /**
+   * Whether keyboard focus is within the disclosure panel.
+   * @selector [data-focus-visible-within]
+   */
+  isFocusVisibleWithin: boolean
```
reidbarber, line 189: Should we have anything else in here? Maybe it would be redundant to include isExpanded, etc. since they're already in `DisclosureRenderProps`

packages/@react-spectrum/accordion/chromatic/Accordion.stories.tsx
```diff
@@ -89,5 +89,3 @@ export const WithDisabledDisclosure = {
       </Disclosure>
     </Accordion>)
 };
-
-// TODO: more stories
```
ktabors, line 93: Don't we still need more chromatic stories, like a button within a disclosure header and props liek isQuiet?

packages/react-aria-components/test/Disclosure.ssr.test.js
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2023 Adobe. All rights reserved.
```
ktabors, line 2:
```suggestion
 * Copyright 2024 Adobe. All rights reserved.
```

packages/react-aria-components/test/Disclosure.test.js
```diff
@@ -206,6 +210,92 @@ describe('Disclosure', () => {
...
+    expect(menu).not.toBeVisible();
+
+    await user.click(button);
+
+    expect(disclosure).not.toHaveAttribute('data-expanded');
+    expect(panel).not.toBeVisible();
+    expect(menu).not.toBeVisible();
+  });
+
+  it('should support nested Disclosures', async () => {
```
ktabors, line 265: Do RSP or S2 do anything where we would want to test nesting there too? Should this be a story in storybook or/and chromatic?

ktabors (commented):
Shouldn't we also have accordion/disclosure in the example apps?
Should this also include adding it to packages for release?

packages/@react-spectrum/accordion/chromatic/Accordion.stories.tsx
yihuiliao: not sure if we're supporting other buttons in the disclosure header for v3. i think that was just a change for s2

packages/react-aria-components/test/Disclosure.test.js
reidbarber: I would say not; I think Spectrum would discourage it. Mostly wanted to have this story in RAC to make sure the contexts work properly in this case. Happy to add them if we think there are legitimate use cases though.

packages/@react-spectrum/accordion/chromatic/Accordion.stories.tsx
reidbarber: We don't support `isQuiet` now in v3. Bringing that up with the team to see if we want that added.

packages/react-aria-components/src/Disclosure.tsx
LFDanLu: I think this is good for now since as you mentioned, the Disclosure's own render props can be used to style the panel itself

packages/react-aria-components/src/Disclosure.tsx
```diff
@@ -207,8 +215,10 @@ function DisclosurePanel(props: DisclosurePanelProps, ref: ForwardedRef<HTMLDivE
       isFocusVisibleWithin
     }
   });
+  let DOMProps = filterDOMProps(props);
```
LFDanLu, line 218: Do we need this? The `contentId` from context always wins right?

LFDanLu (commented):
Just one comment, but otherwise looks good

packages/react-aria-components/src/Disclosure.tsx
yihuiliao: i think this is fine for now. we could always open it up to more things in the future if we get any feedback for it

packages/@react-spectrum/accordion/test/Accordion.test.js
```diff
@@ -18,17 +18,17 @@ import {theme} from '@react-spectrum/theme-default';
...
+  {id: 'three', title: 'three title', children: <input type="text" />}
 ];
 
 function renderComponent(props) {
   return render(
     <Provider theme={theme}>
       <Accordion {...props}>
         {items.map(item => (
-          <Disclosure key={item.key}>
+          <Disclosure key={item.id} id={item.id}>
```
yihuiliao, line 31: i might just be missing some context here but do we still need `key` on Disclosure or is `id` sufficient enough?

reidbarber: The key prop is still there (for React map). We use id for our expanded keys, so I kept both.

yihuiliao: ohh right right

devongovett (dismissed):

LFDanLu (dismissed):
Approving for testing

LFDanLu (approved):

devongovett (approved):
approving for testing

PR #7244, opened 2024-10-23
docs: add new S2 components to example apps

reidbarber:
Adds Table, Accordion/Disclosure, SegmentedControl, and CardView to S2 Parcel and Webpack example apps.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

https://reactspectrum.blob.core.windows.net/reactspectrum/bfa313186ef6390c412e7fad981aee23a2b90e69/verdaccio/s2-parcel-example/index.html

https://reactspectrum.blob.core.windows.net/reactspectrum/bfa313186ef6390c412e7fad981aee23a2b90e69/verdaccio/s2-webpack-5-example/index.html


examples/s2-parcel-example/src/App.js
```diff
@@ -223,6 +301,26 @@ function App() {
         </Section>
 
         <Section title="Navigation">
+          <div className={style({ minHeight: 176 })}>
+            <Accordion>
+              <Disclosure id="files">
+                <DisclosureHeader>
```
yihuiliao, line 307: these need to be updated to be `<DisclosureTitle>`

also might be good to have an example of a disclosure with an action button

reidbarber: Ahh yeah, thanks!

examples/s2-parcel-example/src/App.js
```diff
@@ -236,6 +340,12 @@ function App() {
             The missing link.
           </Link>
           <Link href="/foo">Foo</Link>
+          <SegmentedControl styles={style({width: 384})}>
```
yihuiliao, line 343: Screenshot 2024-10-24 at 10 51 33 AM

It looks like segmented control is missing some styles for some reason. not sure what's going on here. seems to be happening in the parcel and webpack app

yihuiliao: or actually, it seems like all of them have `data-selected` which is why they look this way

yihuiliao: also we should probably add an aria label to the segmented control so that the warnings don't come up in the console

reidbarber: > or actually, it seems like all of them have data-selected which is why they look this way

I think I need to use `id` instead of `value`. Not sure why I have that, maybe copied from the old PR before we changed.

> also we should probably add an aria label to the segmented control so that the warnings don't come up in the console

Wonder why we didn't hit this in storybook. Maybe the above will fix it? We shouldn't need an aria-label if there is just text children.

reidbarber: We may want to delete the `value` prop to prevent this so it doesn't make it into the Radio

yihuiliao: i don't think `SegmentedControlItem` has a `value` prop so it should be fine. we do pass the `id` to the Radio underneath tho

> Wonder why we didn't hit this in storybook

the storybook examples have an aria-label on them so that's why there haven't been any warnings. i think we do need it because it uses RadioGroup under the hood. things might change with toolbar though

reidbarber: > the storybook examples have an aria-label

ahh ok I was looking at the items

LFDanLu (approved):

ktabors (approved):
All the new components in these examples are great!

PR #7264, opened 2024-10-25
feat: Add ToggleButtonGroup component to RAC and S2

devongovett:
This adds a new `ToggleButtonGroup` component to React Aria Components, along with corresponding hooks and S2 components. Similar to `CheckboxGroup` and `RadioGroup`, this component coordinates the selection state between multiple `ToggleButton` children, supporting both single and multiple selection. It also implements arrow key navigation in both horizontal and vertical orientations.

ActionGroup in S2 is split into two components: ActionButtonGroup (which takes ActionButtons as children), and ToggleButtonGroup (which takes ToggleButtons as children). This also swaps SegmentedControl to using ToggleButtonGroup instead of RadioGroup, which enables tooltips to work properly and makes keyboard navigation not affect selection until the user presses Enter.

Some API questions are noted inline.


packages/@react-aria/button/package.json
```diff
@@ -24,6 +24,7 @@
   "dependencies": {
     "@react-aria/focus": "^3.18.4",
     "@react-aria/interactions": "^3.22.4",
+    "@react-aria/toolbar": "3.0.0-beta.10",
```
devongovett, line 27: Not super ideal that the button package now depends on toolbar but I didn't want to make a whole new package for this...

packages/@react-spectrum/s2/src/ActionButtonGroup.tsx
```diff
@@ -0,0 +1,106 @@
...
+        }
+      }
+    }
+  }
+}, getAllowedOverrides({height: true}));
+
+
+export const ActionButtonGroupContext = createContext<ContextValue<ActionButtonGroupProps, HTMLDivElement>>(null);
+
+function ActionButtonGroup(props: ActionButtonGroupProps, ref: ForwardedRef<HTMLDivElement>) {
```
devongovett, line 76: I separated ActionButtonGroup and ToggleButtonGroup into their own components, rather than having a single ActionGroup like in S1. This matches what we did with ActionButton and ToggleButton since this is no longer a collection component. So you'd just use `onPress` directly on the ActionButton instead of an onAction on the root.

If we prefer shorter names we could go with ActionGroup and ToggleGroup potentially, but I kinda liked how clear it is to have ButtonGroup for Buttons, ActionButtonGroup for ActionButtons, and ToggleButtonGroup for ToggleButtons. 🤷

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -33,19 +34,19 @@ export interface SegmentedControlProps extends AriaLabelingProps, StyleProps, Sl
...
+  onSelectionChange?: (id: Key) => void
 }
 export interface SegmentedControlItemProps extends AriaLabelingProps, StyleProps {
   /**
    * The content to display in the segmented control item.
    */
   children: ReactNode,
   /** The id of the item, matching the value used in SegmentedControl's `selectedKey` prop. */
-  id: string,
+  id: Key,
```
devongovett, line 48: Do we want to keep `id` here or change to match ToggleButtonGroup? See below.

packages/@react-spectrum/s2/src/SegmentedControl.tsx
```diff
@@ -55,12 +56,14 @@ export const SegmentedControlContext = createContext<ContextValue<SegmentedContr
 const segmentedControl = style<{size: string}>({
   font: 'control',
   display: 'flex',
+  gap: 4,
   backgroundColor: 'gray-100',
-  borderRadius: 'lg',
-  width: 'full'
+  borderRadius: 'default',
+  width: 'fit'
```
devongovett, line 61: SegmentedControl now fits its contents by default rather than filling its parent

packages/@react-stately/toggle/src/useToggleGroupState.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  /** Replaces the set of expanded keys. */
+  setSelectedKeys(keys: Set<Key>): void
+}
+
+/**
+ * Manages state for a group of toggles.
+ * It supports both single and multiple selected items.
+ */
+export function useToggleGroupState(props: ToggleGroupProps): ToggleGroupState {
```
devongovett, line 56: This is called `useToggleGroupState` rather than `useToggleButtonGroupState` to make it non-button specific. Not sure what else we'd use it for yet though. Does this make sense?

packages/@react-types/button/src/index.d.ts
```diff
@@ -71,6 +71,10 @@ interface AriaBaseButtonProps extends FocusableDOMProps, AriaLabelingProps {
 
 export interface AriaButtonProps<T extends ElementType = 'button'> extends ButtonProps, LinkButtonProps<T>, AriaBaseButtonProps {}
 export interface AriaToggleButtonProps<T extends ElementType = 'button'> extends ToggleButtonProps, AriaBaseButtonProps, AriaButtonElementTypeProps<T> {}
+export interface AriaToggleButtonGroupItemProps<E extends ElementType = 'button'> extends AriaToggleButtonProps<E> {
+  /** An identifier for the item in the `selectedKeys` of a ToggleButtonGroup. */
+  value: Key
```
devongovett, line 76: This is really unfortunate. ToggleButton already supports an `id` prop, which gets passed through to the DOM. Overloading that to be a local id matching our other collection components feels weird. So I went with yet another prop name for this, `value`, which sorta matches RadioGroup and CheckboxGroup. But we still have `selectedKeys` at the top-level rather than `value` there. Thoughts?

snowystinger (commented):
Given the number of API hesitations we have, should we start with UNSTABLE prefixes?

packages/@react-spectrum/s2/src/SegmentedControl.tsx
yihuiliao: Did spectrum say anything about this in the meeting?

devongovett: They were ok with the change to fit the content width by default. I added an `isJustified` prop to SementedControl to allow the previous behavior as an option, which also matches the same option in ActionButtonGroup and ToggleButtonGroup.

packages/@react-aria/button/src/useToggleButtonGroup.ts
```diff
@@ -0,0 +1,95 @@
...
+export function useToggleButtonGroupItem(props: AriaToggleButtonGroupItemOptions<'input'>, state: ToggleGroupState, ref: RefObject<HTMLInputElement | null>): ToggleButtonAria<InputHTMLAttributes<HTMLInputElement>>;
+export function useToggleButtonGroupItem(props: AriaToggleButtonGroupItemOptions<'span'>, state: ToggleGroupState, ref: RefObject<HTMLSpanElement | null>): ToggleButtonAria<HTMLAttributes<HTMLSpanElement>>;
+export function useToggleButtonGroupItem(props: AriaToggleButtonGroupItemOptions<ElementType>, state: ToggleGroupState, ref: RefObject<Element | null>): ToggleButtonAria<DOMAttributes>;
+/**
+ * Provides the behavior and accessibility implementation for a toggle button component.
+ * ToggleButtons allow users to toggle a selection on or off, for example switching between two states or modes.
+ */
+export function useToggleButtonGroupItem(props: AriaToggleButtonGroupItemOptions<ElementType>, state: ToggleGroupState, ref: RefObject<any>): ToggleButtonAria<HTMLAttributes<any>> {
+  let toggleState: ToggleState = {
```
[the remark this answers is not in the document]
devongovett: I think it's not gonna matter because `useToggleButton` is also not memoed. Compiler will save us 😉

packages/@react-aria/button/src/useToggleButtonGroup.ts
```diff
@@ -0,0 +1,95 @@
...
+   */
+  groupProps: DOMAttributes
+}
+
+export function useToggleButtonGroup(props: AriaToggleButtonGroupProps, state: ToggleGroupState, ref: RefObject<HTMLElement | null>): ToggleButtonGroupAria {
+  let {isDisabled} = props;
+  let {toolbarProps} = useToolbar(props, ref);
+
+  return {
+    groupProps: mergeProps(toolbarProps, filterDOMProps(props, {labelable: true}), {
```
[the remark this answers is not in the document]
devongovett: seems like not. data attributes don't make it onto the element without this. Looks like in Toolbar we did the filterDOMProps in the RAC component instead of the hook. I forget which one we usually go with...

packages/react-aria-components/src/ToggleButton.tsx
```diff
@@ -29,18 +30,34 @@ export interface ToggleButtonRenderProps extends Omit<ButtonRenderProps, 'isPend
...
+    onChange(isSelected) {
+      groupState.setSelected(props.id!, isSelected);
+    }
+  } : props);
+
+  let {buttonProps, isPressed, isSelected, isDisabled} = groupState && props.id != null
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    ? useToggleButtonGroupItem({...props, id: props.id}, groupState, ref)
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    : useToggleButton({...props, id: props.id != null ? String(props.id) : undefined}, state, ref);
```
LFDanLu, line 54: Since we use the ToggleButtonGroup state to determine `isSelected`, should we also omit ToggleButton's `isSelected` and `defaultSelected` props from `AriaToggleButtonGroupItemProps`?

LFDanLu (approved):
Approving for testing, just some small questions and comments but otherwise the behavior looks good to me

snowystinger (dismissed):

LFDanLu (approved):

reidbarber (approved):
LGTM

PR #7271, opened 2024-10-28
fix: add `"use strict";` to style macro cjs output

reidbarber:
Should fix:

> When projects use size: 'control' via style macros, it's built as an arbitrary value, [2rem], which isn't valid css, because this is set as a global this instead of undefined. This is supposed to be undefined or a MacroContext. Adding "use strict"; prevents this from becoming the global this which causes values to be defined as arbitrary.

## ✅ Pull Request Checklist:


.parcelrc
```diff
@@ -23,6 +23,7 @@
     "*.html": "parcel-packager-ssg"
   },
   "optimizers": {
-    "*.js": ["@parcel/optimizer-terser"]
+    "*.js": ["@parcel/optimizer-terser"],
+    "**/spectrum-theme.cjs": ["parcel-optimizer-strict-mode"]
```
reidbarber, line 27: The exact path of `packages/@react-spectrum/s2/style/dist/spectrum-theme.cjs`doesn't work, but this works.

devongovett (approved):

snowystinger (approved):
I see it has been correclty added. Was there an audit of other files missing it potentially? or are we only applying it here because it'd be too unpredictable to add it elsewhere?

reidbarber:
@snowystinger Yeah we decided we can look at expanding this to more files later. It might introduce some unexpected bugs if we apply it everywhere at once.

Issue #5637, opened 2023-12-28
Group aria component is not disabling children button


federicocappellotto97:
### Provide a general summary of the issue here

 <Group isDisabled>
                <Input />
                <Button />
            </Group>

### 🤔 Expected Behavior?

Button should has disabled attribute

### 😯 Current Behavior

Now only the input get the attribute

### 🖥️ Steps to Reproduce

Create a component with the html specified above

### Version

"react-aria": "^3.31.0",         "react-aria-components": "^1.0.0",

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

Mac OS

yihuiliao:
Correct me if I am wrong, but I believe what you are trying to do is the following: https://codesandbox.io/s/peaceful-river-nys28q?file=/src/App.js where the expectation is that if `isDisabled` is applied to `<Group>` that it would also be passed to its children. 

I think `Group` was really more intended to support interactive states for styling so it doesn't currently pass down these states to its children. For now, if you wanted a disabled button within a group, you would also have to apply `isDisabled` directly onto the button. 

Maybe this is something we can consider implementing in the future since I can see how this would be confusing but I’d like to discuss with the team first.

Issue #5179, opened 2023-10-02
[RAC] Key navigation is broken for nested GridLists


piecyk:
### Provide a general summary of the issue here

After focusing on nested element, pressing arrow down/up will navigate to element in parent list.

### 🤔 Expected Behavior?

Key navigation should be trapped within nested list, till espaced with tab.

### 😯 Current Behavior

Pressing arrow down/up will navigate element in parent list.

### 🖥️ Steps to Reproduce

https://codesandbox.io/s/throbbing-wind-jflfv7?file=/src/App.js

### Version

"react-aria-components": "1.0.0-beta.0"

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

MacOS

piecyk:
It boils down that `useGridListItem` handles key events with `onKeyDownCapture` https://github.com/piecyk/react-spectrum/blame/main/packages/%40react-aria/gridlist/src/useGridListItem.ts#L184

Is there specific reason for that, would expect onKeyDown here 🤔

LFDanLu:
If I remember correctly, part of the reason for the `onKeyDownCapture` is so that the GridList always handles the keyboard events for grid navigation, regardless of what focusable element might have focus within the row (e.g. it would handle the ArrowUp/ArrowDown over say a menu button).

reidbarber:
I think supporting nested grid roles will take more to properly implement, and falls under: https://github.com/adobe/react-spectrum/issues/2328

piecyk:
> If I remember correctly, part of the reason for the onKeyDownCapture is so that the GridList always handles the keyboard events for grid navigation, regardless of what focusable element might have focus within the row (e.g. it would handle the ArrowUp/ArrowDown over say a menu button).

OK, that make sense, this should work the same for nested list.

devongovett:
Is your use case sections in the list? If so, I'm not sure if that should be implemented as nested grids. We were working on support for sections in our table and list components a few months back, and I believe they were all within the same list rather than using nested lists. @LFDanLu probably knows more about that.

piecyk:
My use case is that use can create list of comments ( it's reorderable ) and to each comment can include a list of some items ( this is also reorderable ). in theory can flatten it, just when reordering the comments need to keep track of it items, to move it together 🤔

LFDanLu:
> We were working on support for sections in our table and list components a few months back, and I believe they were all within the same list rather than using nested lists

[...]

Issue #5005, opened 2023-08-31
[RAC] - Table - fix keyboard navigation when using conditional render focusable cell


nirtamir2:
### Provide a general summary of the issue here

https://stackblitz.com/edit/stackblitz-starters-xgaei9?file=src%2FTestTable.tsx

When we have cells that are focusable (like elements with`tabIndex=0`, `inputs`) - the focus management of the table works well - and it move to the first element of the cell (and even if we have a cell with 2 inputs - we can navigate between them with the arrow keys). The problems starts when we render the input conditionally - depending on the `isFocus` / `isFocusVisible` props from react-aria. The element focus is lost in this case. The use case is for showing table cells as their view but when they are focusable change their look to their edit mode. In this case we lose focus. 

### 🤔 Expected Behavior?

The focus mode to the navigated element

### 😯 Current Behavior

Focus is lost

### 💁 Possible Solution

Allow controlling the keyboard management / add escape hatch / focus on the table td cell instead of the inner element

### 🔦 Context

I want to create a table that conditionally toggle between view mode of the data, and input mode when the cell is focusable. But the keyboard navigation breaks in this case

### 🖥️ Steps to Reproduce

Go go https://stackblitz.com/edit/stackblitz-starters-xgaei9?file=src%2FTestTable.tsx
focus the cells form 1&2 columns with the keyboard. Try to focus the conditional render column. See focus is lost

### Version

"react-aria-components": "1.0.0-alpha.6"

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

MacOS

reidbarber:
See https://github.com/adobe/react-spectrum/issues/2328 for grid navigation edit mode.

nirtamir2:
> See #2328 for grid navigation edit mode.

Thanks!
The focus behavior does not document well so I assume this is what happening.
Should I close this issue? It did not resolve but I hope #2328 will find a way / fix my issue (depending on the implementation)

PR #7277, opened 2024-10-28
feat: Add support for grid edit mode w/ nested interactive widgets

### LINKED ISSUE #6037: [RAC] Table typeahead captures all keyboard input, leaving none for nested Inputs

I've got a table like this, where the inputs can't capture a single keystroke, because the table hungrily swallows them all:

```jsx
import { Cell, Column, Input, Row, Table, TableBody, TableHeader } from "react-aria-components";

const items = "ABCDEFGHIJKLMNOPQRSTUVWXYZ".split("").map((letter) => ({id:letter}));

export default function Home() {
  return (
    <Table>
      <TableHeader>
        <Column isRowHeader>Name</Column>
        <Column>Value</Column>
      </TableHeader>
      <TableBody items={items}>
        {(row) => (
          <Row>
            <Cell>{row.id}</Cell>
            <Cell><Input type="text" /></Cell>
          </Row>
        )}
      </TableBody>
    </Table>
  );
}
```

### 🤔 Expected Behavior?

I'd love for the inputs to Just Work™ in a situation like this.

### 😯 Current Behavior

https://github.com/adobe/react-spectrum/assets/13100/c28fa89b-ba9d-4e3f-a881-ee5a7a367389

### 💁 Possible Solution

1. Is there a good reason for `useTypeSelect` to be using the capture phase? If it used the bubble phase, the inputs themselves could choose if/when to stop propagation.
2. This is a bit more blunt, but why not allow typeahead to be disabled on tables? `useSelectableCollection` already takes a `disallowTypeAhead` prop, but there's not currently a way to drill that all the way down through `<Table>` → `useTable` → `useGrid` → `useSelectableCollection`.

### 🔦 Context

#1214 is a 3-year-old ticket that mentions that tables swallow spaces, but somehow fails to mention that tables can also swallow **all keystrokes** as long as the input is matching the value of a row header.

### 🖥️ Steps to Reproduce

https://codesandbox.io/p/sandbox/old-voice-k5n9p4

### Version

"react-aria-components": "^1.1.1"

Chrome

### If other, please specify.

_No response_

macOS

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

nwidynski:
Closes #6037 #5005 #2328 #4957 #5637 #5179 #7121 #1214 #4674 #5003

This PR adds `edit mode` to `@react-aria/grid` and RAC `<Table />` in addition to generalized support for nested interactive (collection) components inside grid cells or items.

It also introduces a new `useGroup()` and `useGroupState()` hook meant to propagate validation and input state to nested interactive components, allowing for more convenience when controlling large sections of an application.

## ✅ Pull Request Checklist:


nwidynski:
Test suite is failing because of the auto-incrementing document key. Unfortunately this causes problems in scenarios where the document keys need to be persistent across (un-)mounts, like in Menu selection for example.

Not sure how to get around this for now, maybe someone from the core team has an idea?

packages/@react-aria/gridlist/src/useGridListItem.ts
```diff
@@ -240,6 +256,26 @@ export function useGridListItem<T>(props: AriaGridListItemOptions, state: ListSt
...
       return;
     }
+
+    if (e.relatedTarget && keyboardNavigationBehavior === 'tab') {
+      let comparedPosition = ref.current.compareDocumentPosition(e.relatedTarget);
+
+      let isFocusWithin = Boolean(comparedPosition & Node.DOCUMENT_POSITION_CONTAINED_BY);
+      let isShiftTab = isFocusVisible() && Boolean(comparedPosition & Node.DOCUMENT_POSITION_FOLLOWING);
+      let isVirtualizedSibling = isVirtualized && e.relatedTarget.id.startsWith(id) && e.relatedTarget.role === 'row';
+      let isSibling = e.relatedTarget.parentElement === ref.current.parentElement || isVirtualizedSibling;
```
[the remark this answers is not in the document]
LFDanLu: Eventually I believe we will want to add section support to grid lists (and tables). I had began work on that in https://github.com/adobe/react-spectrum/pull/4173 quite some time ago but we punted on it to wait for React Aria Component's implementation of the collections API. Here is a [build](https://reactspectrum.blob.core.windows.net/reactspectrum/2cf1054ded557db16c0e023865ada1d4c1edf354/storybook/index.html?path=/story/listview--static-sections&providerSwitcher-express=false) for the React Spectrum implementation if you are curious about the DOM structure, I imagine we'd still go for that structure but I'd have to revisit the work as a whole.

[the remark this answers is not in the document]
nwidynski: Got it, i think using the virtualized detection method is more reliable anyways - will use it for all scenarios.

packages/react-aria-components/src/TextField.tsx
```diff
@@ -53,6 +54,7 @@ export interface TextFieldProps extends Omit<AriaTextFieldProps, 'label' | 'plac
 export const TextFieldContext = createContext<ContextValue<TextFieldProps, HTMLDivElement>>(null);
 
 function TextField(props: TextFieldProps, ref: ForwardedRef<HTMLDivElement>) {
+  let groupState = useContext(GroupStateContext);
```
nwidynski, line 57: I just implemented this for `<TextField />` for now, but open to expand it to other field or action components.

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -205,18 +249,101 @@ export function useGridCell<T, C extends GridCollection<T>>(props: GridCellProps
...
+          // is handled by the browser and not by useSelectableCollection (which would take us out of the list).
+          let walker = getFocusableTreeWalker(ref.current, {tabbable: true});
+          walker.currentNode = document.activeElement;
+          let next = e.shiftKey ? walker.previousNode() : walker.nextNode();
+
+          if (next || isEditing) {
+            e.stopPropagation();
+
+            if (isEditable && document.activeElement === ref.current) {
+              state.setEditingKey(node.key); // Do we want this?
```
nwidynski, line 312: Not sure about this since it creates disparities between the two navigation methods.
I do find it nice to maintain the `Tab` to enter cell from grid list however.

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -53,14 +69,20 @@ export interface GridCellAria {
...
   } = props;
 
   let {direction} = useLocale();
-  let {keyboardDelegate, actions: {onCellAction}} = gridMap.get(state);
+  let {id, keyboardDelegate, keyboardNavigationBehavior, actions: {onCellAction}} = gridMap.get(state);
+
+  let isEditing = state.editingKey === node.key;
+  let isReadOnly = state.isReadOnly || props.isReadOnly;
+
+  focusMode ??= isEditable || keyboardNavigationBehavior === 'tab' ? 'cell' : 'child';
```
nwidynski, line 85: Do we want to move onto the first child automatically when in tab navigation mode?

packages/@react-stately/group/src/useGroupState.ts
```diff
@@ -0,0 +1,56 @@
...
+  isReadOnly: boolean
+}
+
+export const GroupStateContext = createContext<Partial<GroupState>>({});
+
+/**
+ * Provides the state for a group of fields.
+ * @param props - Props for the Group.
+ */
+export function useGroupState(props: GroupProps): GroupState  {
```
nwidynski, line 36: I think propagating the readonly state into the (editable) cell content makes a lot of sense. Since this was very similar to group behavior, I thought it might make sense to enhance it while we are at it.

nwidynski (commented):
As @devongovett suggested, i implemented this for grid cells and added support for edit mode into RAC table. 
Test suite is passing, but no new tests added just for now - i added a couple examples to showcase behavior.

Sorry for the large diff - most of it is boilerplate from the new package though 😓

packages/react-aria-components/src/Table.tsx
```diff
@@ -1175,14 +1197,21 @@ export interface CellRenderProps {
...
+  /**
+   * Whether the cell is read only.
+   * @selector [data-readonly]
+   */
+  isReadOnly: boolean
 }
 
 export interface CellProps extends RenderProps<CellRenderProps> {
-  /** The unique id of the cell. */
-  id?: Key,
```
nwidynski, line 1183: I removed the `id` because it wasn't being forwarded to the element anyways and it caused issues after adjusting the key structure. In case we want those back, some further debugging is required.

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -205,18 +249,101 @@ export function useGridCell<T, C extends GridCollection<T>>(props: GridCellProps
...
+            state.setEditingKey(node.key);
+            focusSafely(focusable);
+          }
         }
         break;
+      }
+    }
+  };
+
+  let onBlur = (e) => {
```
nwidynski, line 344: Known issue: Edit mode is not recognized when clicking into the editable cell.

It likely makes sense to enter edit mode `onFocus` instead of `onKeyDown`, similar to how we exit `onBlur`.

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -205,18 +249,101 @@ export function useGridCell<T, C extends GridCollection<T>>(props: GridCellProps
...
+        }
+        break;
+      }
+      case 'F2':
+      case 'Enter': {
+        if (isEditable) {
+          let focusable = walker.nextNode() as FocusableElement;
+
+          if (focusable && document.activeElement === ref.current) {
+            state.setEditingKey(node.key);
```
nwidynski, line 282: Do we want to announce entering and exiting?

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -205,18 +249,101 @@ export function useGridCell<T, C extends GridCollection<T>>(props: GridCellProps
...
+      }
+      case 'Escape': {
+        if (state.editingKey || keyboardNavigationBehavior === 'tab') {
+          state.setEditingKey(null);
+          focusSafely(ref.current);
+        }
+        break;
+      }
+      case 'Tab': {
+        if (keyboardNavigationBehavior === 'tab') {
```
nwidynski, line 301: Do we want to enable tab wrapping for edit mode in general, meaning also for `arrow` navigation behavior?

The W3C spec suggests this is acceptable behavior, and it's much easier to implement than marshaling to sibling content.

packages/react-aria-components/src/Table.tsx
```diff
@@ -1213,23 +1244,28 @@ export const Cell = /*#__PURE__*/ createLeafComponent('cell', (props: CellProps,
...
   let TD = useElementType('td');
+  let groupState = useMemo(() => ({isReadOnly}), [isReadOnly]);
 
   return (
     <TD
       {...mergeProps(filterDOMProps(props as any), gridCellProps, focusProps, hoverProps)}
       {...renderProps}
       ref={ref}
+      data-editing={isEditing || undefined}
       data-focused={isFocused || undefined}
```
nwidynski, line 1262: Is there a reason for `data-hovered` missing here?

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -288,3 +447,15 @@ function last(walker: TreeWalker) {
...
+
+function wrap(walker: TreeWalker) {
+  walker.currentNode = walker.root;
+  let start = walker.firstChild() as FocusableElement;
+  walker.currentNode = walker.root;
+  let end = last(walker);
+  return [start, end];
+}
+
+function isSingleLineTextInput(element: Element) {
```
nwidynski, line 459: Should this support all [text input types](https://github.com/facebook/react/blob/7c8e5e7ab8bb63de911637892392c5efd8ce1d0f/packages/react-dom-bindings/src/events/isTextInputElement.js)?

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -205,18 +249,101 @@ export function useGridCell<T, C extends GridCollection<T>>(props: GridCellProps
...
+            if (isEditing && e.shiftKey && document.activeElement === start) {
+              e.preventDefault();
+              focusSafely(end);
+            }
+          }
+        }
+        break;
+      }
+      default: {
+        if (document.activeElement === ref.current && /^[a-zA-Z0-9]$/.test(e.key)) {
```
nwidynski, line 331: Known issue: Entering an editable cell should be preferred over grid navigation, when node textValue is matched to a row.

packages/@react-aria/grid/src/useGridCell.ts
```diff
@@ -104,31 +126,47 @@ export function useGridCell<T, C extends GridCollection<T>>(props: GridCellProps
...
       return;
     }
 
     let walker = getFocusableTreeWalker(ref.current);
     walker.currentNode = document.activeElement;
 
+    let isFocused = ref.current === document.activeElement;
+    let isNavigation = NAVIGATION_KEYS.has(e.key) || (e.altKey && e.key === 'Tab');
+
+    if (keyboardNavigationBehavior === 'tab' && isNavigation && !isFocused) {
```
nwidynski, line 140: Known issue: When editing cells with a single focusable element, the arrow key events should be captured by the nested focusable element.

packages/react-aria-components/stories/Table.stories.tsx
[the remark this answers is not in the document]
LFDanLu: To me, the arrow navigation behavior is a bit non-intuitive for an keyboard user since it isn't perhaps readily apparent that the cell with the text fields is different from the cell that only has buttons, but the latter allows you to immediately focus buttons within via arrow keys while the former requires you to hit `Enter` to move focus within the cell.

I wonder if we could get away with only allowing `tab` mode rather than `arrow` mode haha

packages/@react-aria/grid/src/useGridCell.ts
LFDanLu: Personally I'd be fine with enabling tab wrapping for edit mode in general, feels more consistent that way. Happy for others to weigh in

packages/@react-aria/grid/src/useGridCell.ts
LFDanLu: I believe the idea was so that the user could use typeahead to quickly navigate to a new row regardless of the cell they were in, but its been a while so I'll have to dig for any discussions on this. Personally, I feel like enforcing that entering a cell via Tab/Enter is required for editing the cell's content is alright, but not too strong of a stance tbh

LFDanLu (commented):
Partial review, will see about wrapping it up soon. Looks promising so far!

packages/react-aria-components/stories/Table.stories.tsx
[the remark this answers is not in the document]
nwidynski: As per our issue discussion and Devon's [comment](https://github.com/adobe/react-spectrum/pull/7215#issuecomment-2436554014) I added `Tab` as the "intuitive" entering default. 

In cells, we can maintain `Enter`/`F2` as an alternative to improve fulfillment of the W3C `Grid` edit mode [spec](Enter). Contrarily to grid lists, there won't be collision issues with selection as only rows are selectable.

packages/@react-aria/collections/src/CollectionBuilder.tsx
```diff
@@ -51,6 +52,11 @@ export function CollectionBuilder<C extends BaseCollection<object>>(props: Colle
   // This is fine. CollectionDocumentContext never changes after mounting.
   // eslint-disable-next-line react-hooks/rules-of-hooks
   let {collection, document} = useCollectionDocument(props.createCollection);
+
+  if (props.id) {
+    document.key = props.id;
+  }
```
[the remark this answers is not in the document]
LFDanLu: I see, I was wondering where the `e.relatedTarget.id.startsWith(id)` came from, thanks for connecting the dots here for me! Will have to think on this some more, but I can't think of any negative ramifications

packages/react-aria-components/stories/Table.stories.tsx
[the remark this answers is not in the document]
LFDanLu: The spec link you added seems to be malformed but I assume you were talking about [this](https://www.w3.org/WAI/ARIA/apg/patterns/grid/#keyboardinteraction-settingfocusandnavigatinginsidecells)? If so, thanks for pointing this out, I don't remember seeing this spec when we first originally tried to implement edit mode into TableView but I'm happy to support `Enter`/`F2` to follow the spec.

packages/@react-aria/grid/src/useGridCell.ts
LFDanLu: gotcha, we can certainly play around with the announcements there then. Hopefully there is something simple we can do since in my experience using the LiveAnnouncer we have is a bit tricky due to the possibility of interrupting native announcements that the screen reader may be trying to issue/the inconsistent behavior of screenreaders across browser/screenreader combinations

packages/@react-aria/grid/src/useGridCell.ts
LFDanLu: thanks for the clarification! The "table edit mode with tab navigation example" story demonstrates the scenario you are describing right? If so, I think what you have now is fine good it always takes an explicit key press other than the arrow keys to enter a cell regardless of its edibility.

packages/@react-aria/grid/src/useGridCell.ts
LFDanLu: Could we check if the nested focusable element (aka `document.activeElement`) is contained by the current ref as well? Noticed that arrow keys don't move the text cursor when focused on the cell text inputs when in the tab navigation table story.

Related to the above behavior, it feels odd IMO that the arrow keys in the arrow navigation table story move between the focusable elements in an editable cell instead of moving the text cursor of your focused textfield. I do see that the spec mentioned that this is the correct behavior... cc @majornista any opinions here?

packages/@react-aria/grid/src/useGridCell.ts
LFDanLu: If I remember correctly, we did make use of `onFocus` for exactly this in our old edit mode explorations. That combined with some visual affordances via the `data-editing` that you added should hopefully be enough indication to the user that they are in edit mode and thus will need to exit it via Escape

packages/@react-aria/grid/src/useGridCell.ts
LFDanLu: I imagine so

packages/react-aria-components/src/Table.tsx
LFDanLu: What kind of issues did you run into? The `id` if provided is used by the CollectionBuilder as the key in the key map if provided, otherwise it autogenerates one (similar to Columns and Rows). Probably not super important for now but I imagine it will be if/when we support cell actions

packages/react-aria-components/src/Table.tsx
LFDanLu: Looks like something we missed, I recall any reason why not.

LFDanLu (commented):
Will see about getting this into our upcoming test session on Monday as well

packages/react-aria-components/src/Table.tsx
nwidynski: Seems to have sorted itself out with the latest commits so i added this back 👍

packages/@react-aria/collections/src/CollectionBuilder.tsx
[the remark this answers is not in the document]
nwidynski: Looking forward to your input on this 👍 

Just now also remembered to normalize the `disabledKeys` & `(default)SelectedKeys` after adding the prefix to them - I guess i should extend the test suite while im at this haha

packages/@react-aria/grid/src/useGridCell.ts
nwidynski: I finished laying the groundwork for this in `useGridEditAnnouncement()` until we have a proper idea of what and when to announce 👍 Would really appreciate guidance here, especially in terms of localization!

nwidynski:
Pushed the last the updates towards completing what I had in mind for this PR 👍 

As mentioned briefly, biggest change from the last review is the introduction of a `GridManager` which extends the `SelectionManager` by the grid editing and navigation state. This allows us to access the editing state to prevent features like `useTypeSelect()` from interrupting our navigation.

Another key insight was that we could streamline the current behavior of `isKeyboardNavigationDisabled` by just switching the navigation mode to `tab` whenever we need to disable arrow key navigation. This way we can ensure a great UX, since we only cater towards these two interaction models, giving the user full control at any times.

Lastly i extended support for the disabled state to rows and columns and adjusted the `GridKeyboardDelegate` accordingly.
There are a bunch of other small features and bug fixes along the way, but thats pretty much it.

I'm looking forward to your test session feedback to finalize the spec and get started on tests for this beast 😅 !

LFDanLu:
@nwidynski Thanks for all the work you've done here! The team took a closer look today and are a bit concerned about the extent of the changes made. The overall behavior seems promising but we would like to perhaps pare it down into smaller pieces so it is easier to incrementally update the behavior/understand the full ramifications of the impact of these changes.

With that in mind, we took some time to implement what we thought was the base amount of changes outlined in https://github.com/adobe/react-spectrum/pull/7215#issuecomment-2436554014 to support textfields in GridList, you can see those in [(WIP) Update GridList so it supports textfields](https://github.com/adobe/react-spectrum/pull/7360#top) with [this simple story](https://reactspectrum.blob.core.windows.net/reactspectrum/060bf7eb7aaaba7b2614bb41562944028b268009/storybook/index.html?path=/story/react-aria-components--grid-list-example&args=layout:grid;keyboardNavigationBehavior:tab;selectionMode:none&providerSwitcher-express=false). At the moment, that implementation doesn't have "memory" of the last focused element in the GridList and instead opts to focus the last child element of your last focused cell when shift tabbing back into the collection. Any thoughts? Some of the issues you noted in https://github.com/adobe/react-spectrum/pull/7215#issuecomment-2437512884 still apply but we are debating if saving the "memory" of the last focused element should still apply for non `Tab` cases or not and thus if it is important to save the "memory" of the last focused element at all.

LFDanLu:
@nwidynski Gotcha, thank you for the detailed response! As you mentioned, the team was moving towards having the children of the cell be responsible for stopping propagation but I can revisit this assumption to them in our meeting tomorrow and see if we may want to pivot there. As for "memory", we have prior precedence doing such tracking in [useToolbar](https://github.com/adobe/react-spectrum/blob/3949762f8911e736ef97bd703f7e642a2036f7fc/packages/%40react-aria/toolbar/src/useToolbar.ts#L101-L117), but there have been differing opinions as to what the behavior should be when tabbing back into the collection for specific cases like "blur happened due to change in browser window/mouse click/iframe" which I'll see about getting a consensus from the team about what behavior we want to go with.

As for edit mode as a whole, historically when we tried to implement it ourselves I recall that @devongovett still had concerns as to how it should work on mobile touch devices, where perhaps it wouldn't be clear enough to the end user when/how edit mode would be triggered and whether or not there would be confusion around that. I don't quite remember all the details but hopefully I'll be able to comeback with a more detailed answer after tomorrow's meeting.

Again, thank you so much for all the work done here, it is greatly appreciated! It quite evident that you've done a lot of good work providing a solution that also matches the ARIA spec, it is just that we need to be extra careful since these are quite hefty changes and as a library we need to make sure that we are ready to accept and maintain this behavior for the foreseeable future.

devongovett:
> this does not really capture the scope of our original issue https://github.com/adobe/react-spectrum/issues/7121, where we proposed preventing all navigation when in tab mode, not limited to having an input element as the activeElement

Could you explain why you want that? Obviously there are issues with using the arrow keys on certain components such as inputs conflicting with grid navigation, which Daniel's PR has solved, but I don't see any particular issue with still allowing arrow key navigation when focused on an element that _does not_ use the arrow keys. It seems nice to me that this approach allows us to support arbitrary elements such as inputs inside grid cells without needing edit mode at all, while not impacting the behavior of any existing applications.

> heading down a direction where we shift responsibility for nested interactive widgets down to children or possibly even the consumer to figure out himself

This is already the case in many ways throughout the library. For example, if you have a button inside a table row, pressing Enter while focused on the button triggers the press event only on that button and does not trigger row selection. That is because the button stops event propagation so the enter key event never reaches the row. The overall principle here is that once the deepest element in the DOM tree handles an event, it should stop propagation to prevent parent elements from also handling it. Without this, we would lose the ability to compose arbitrary components together – the parent would need to know about all of the possible children in order to decide whether to handle an event or not. You can see an example of that in Daniel's PR which does this for native input elements (as a nicety), but we couldn't possibly know all of the other arbitrary components that might handle arrow key navigation without leaving some responsibility to them to stop propagation.

devongovett:
btw, we definitely want to support a full edit mode in the future _in addition_ to what Daniel proposed, we're just looking to break up the work a bit and hopefully solve some immediate issues that might not require it first. As Daniel mentioned, many of the interactions for edit mode (especially on touch devices) have not yet been defined.

nwidynski:
Thanks for getting back on this so quickly 👍 

### TLDR; 
> [...] but I don't see any particular issue with still allowing arrow key navigation when focused on an element that _does not_ use the arrow keys. It seems nice to me that this approach allows us to support arbitrary elements such as inputs inside grid cells without needing edit mode at all [...]

The flag is called `keyboardNavigationBehavior === 'tab'`, why would it allow grid navigation with `arrow` still? To me, an implicit intention is phrased when enabling this setting: `Items will contain conflicting interactions`.

The only difference in our viewpoint is that you guys are trying to shift the decision of `when not` to navigate to user land, while I rather advocate for users to make the decision of `when` to navigate - a decision I find much easier to make.

--------
### Answer
As you mentioned, it's practically impossible for parent elements to predict what content they will render and shifting actions to children is a common pattern in react-aria. What I meant however, was that there is no slot we could use to carry default behavior to unknown, possibly custom, children. This means we shift all of the prevention responsibility down to the actual user land (not RAC, but actual users) to implement, which obviously isn't working out.

Users are attempting, and I think should be able to, render any arbitrary content inside a `<GridListItem />` or `<Cell />`, while expecting the content to work as if outside an item. While this is technically speaking part covered by the W3C edit mode spec, it's not related to an `edit` interaction per se. As an example, we were attempting to build **virtualized** grid list items with nested grid lists and tag groups inside ([see demo](https://codesandbox.io/p/sandbox/quirky-shannon-zccylw?file=%2Fsrc%2FApp.js)). The solution in #7360 is not helping here...

Leaving user land only an `isKeyboardNavigationDisabled` flag to control navigation is hardly a great experience, not only because half of the necessary props aren't even exposed, or the fact that it doesn't prevent type select, but because very deep knowledge of the underlying hooks is required to even come close to making an informed decision as to how and when to disable.

If you are concerned about interrupting existing behavior, I'm totally happy with introducing yet another `keyboardNavigationBehavior === 'isolate'` or something similar and make this work.

nwidynski:
@devongovett Maybe I should further clarify, after sleeping on this a night.

> That is because the button stops event propagation so the enter key event never reaches the row.

This is just not true at the moment, since event propagation isn't stopped by the button, but by `usePress()` in the `useSelectableItem()`. I understand that in an ideal scenario, each child component would stop propagation, leaving only the elements with native arrow key interactions (e.g. inputs) to deal with as a nicety.

Unfortunately though, pretty much all components leak keyboard events upwards ([demo](https://codesandbox.io/p/sandbox/loving-snow-zwsszm?file=%2Fsrc%2FApp.js%3A37%2C21)) when not at one point an `onKeyDown` handler is spread onto them, because `usePress()` or `useKeyboard()` otherwise never get to top propagation.

devongovett:
> The flag is called keyboardNavigationBehavior === 'tab', why would it allow grid navigation with arrow still?

It is for tabbing _within_ cells, not between them. You still use arrow keys to do that. There are a few different modes as well, for example in some components the cell itself is never actually focused - focus moves immediately to the first focusable child. If we disabled arrow key navigation in that case, the user would be stuck inside the cell.

> Unfortunately though, pretty much all components leak keyboard events upwards ([demo](https://codesandbox.io/p/sandbox/loving-snow-zwsszm?file=%2Fsrc%2FApp.js%3A37%2C21)) when not at one point an onKeyDown handler is spread onto them, because usePress() or useKeyboard() otherwise skip stopping propagation.

I'd say this is more of a bug in TagGroup and/or useSelectableCollection. Ideally those would stop propagation after handling the event. I think we just have not yet encountered nested collection components so this never came up before. We can certainly fix that.

I guess what we're discussing here is defaults: in what we proposed, the user can navigate between cells by default and if the developer wants to prevent that they can stopPropagation. In what you proposed, there's less work for the developer because they don't need to stop propagation to prevent arrow key navigation, but it's also not possible to enable the user to navigate between cells with the arrow keys at all (which might make the UI easier to use).

> If you are concerned about interrupting existing behavior, I'm totally happy with introducing yet another keyboardNavigationBehavior === 'isolate' or something similar and make this work.

I guess this still shifts the burden to the developer to know that this is something they need to use, the same way as stopPropagation would.

I think we should consider the ideal user behavior here first, e.g. whether arrow key navigation on elements such as buttons within cells should move focus between cells. Needing to shift tab back out in order to do that might not be expected. We will consult some internal accessibility folks on this before deciding on an API.

devongovett:
Btw, wanted to add that I see two separate but related features here:

1. Supporting interactive elements within grid cells at all times. A user could tab into the contents of a cell and interact with elements such as inputs that are already present.
2. A more explicit edit mode, where a cell might contain non-editable data such as text and a user would press Enter/F2 to swap the UI into an editing mode. For example, the text might turn into a textfield, date picker, or whatever other UI might be appropriate for editing that type of value. Then the user could press Enter again to commit their changes, or Escape to cancel. This is the more traditional approach taken by spreadsheets.

What we've mostly been discussing so far is (1), which is much simpler to implement. (2) is more complicated, because the interactions are not well specified for getting into and out of edit mode. For keyboard, the aria spec provides good guidelines, but you would need some way to do the same thing with a mouse (maybe double click? a visual edit button?), a touch screen (long press? a button?), and with a screen reader. You've made a good start here, but this will require further research and testing. I was hoping that we could land a solution to (1) first since it is much simpler, and follow up with (2) in the future once we figure out these interactions.

PR #7278, opened 2024-10-29
build: Update S2 storybook and example apps to optimize CSS bundling

devongovett:
This optimizes the S2 CSS in storybook and our example apps to bundle all CSS into one file instead of code splitting. Because the CSS is atomic, there is a ton of overlap between different components (e.g. `backgroundColor: 'gray-100'` is used in many components). With code splitting that leads to CSS rules being duplicated in multiple bundles. Because the degree of overlap is so high, it is smaller to just load everything up front than splitting.

This updates the parcel, webpack, and vite example apps to implement this and apply best practices. It also updates the documentation to discuss this and link to the examples.

Sorry for the large diff. Moved parts of the example apps to a new file to implement code splitting.


examples/s2-vite-project/src/App.tsx
LFDanLu, line 1: just an aside, noticed the the Vite app doesn't seem to have the CardView or Disclosure

examples/s2-vite-project/vite.config.ts
```diff
@@ -13,16 +13,28 @@
...
+        // Bundle all S2 and style-macro generated CSS into a single bundle instead of code splitting.
+        // Because atomic CSS has so much overlap between components, loading all CSS up front results in
+        // smaller bundles instead of producing duplication between pages.
+        manualChunks(id) {
+          if (/macro-(.*)\.css$/.test(id) || /@react-spectrum\/s2\/.*\.css$/.test(id)) {
+            return 's2-styles';
+          }
+        }
+      }
+    }
```
LFDanLu, line 38: Just to make sure, this configuration meant to only apply to the built version of the app and not to local dev? I noticed the styles were still in separate style sheets when running local dev but not in the built version unlike the webpack one.

LFDanLu (commented):
Tested locally and verified the improvements in deduping/bundling all the CSS into a single file. Just some things I noticed about the vite app in particular that I noted below

examples/s2-vite-project/vite.config.ts
devongovett: Yeah that's right. Vite does not bundle during development, so unfortunately there's no way to deduplicate there. This applies to the built version only.

examples/s2-vite-project/src/App.tsx
LFDanLu: Happy to approve this as is, up to you

devongovett: updated

snowystinger (dismissed):

LFDanLu (dismissed):

examples/s2-parcel-example/src/App.js
```diff
@@ -124,7 +66,7 @@ function App() {
         styles={style({ font: "heading-xl", textAlign: "center" })}
         level={1}
       >
-        Spectrum 2 + Parcel
+        Spectrum 2 + Webpack
```
reidbarber, line 69: typo

examples/s2-vite-project/src/App.tsx
```diff
@@ -10,20 +10,196 @@
...
   return (
     <main>
-      <Button
-        styles={style({
-          marginStart: 16,
+      <Heading
+        styles={style({ font: "heading-xl", textAlign: "center" })}
+        level={1}
+      >
+        Spectrum 2 + Webpack
```
reidbarber, line 69: same

examples/s2-parcel-example/src/App.js
devongovett:
```suggestion
        Spectrum 2 + Parcel
```

reidbarber (approved):

LFDanLu (approved):

PR #7304, opened 2024-11-01
WIP: prefix macro class names with hash unique to theme

reidbarber:
Fixes:

> **S1 and S2 style macros conflict with each other, breaking styling**
If a S1 component that uses macros (TreeView, ColorPicker) exists alongside a S2 component, the styling macros conflict with one another (aka same classnames are generated for different style sets). This causes styling to be completely broken.

Also should fix any issues if multiple versions of the S2 macro are used on the same page.

TODO:
- Iterate on hash function for collision resistance vs. class name length tradeoff
- Add to S1 macro

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/s2/style/style-macro.ts
+++ b/packages/@react-spectrum/s2/style/style-macro.ts
@@ -13,9 +13,9 @@
 import type {Condition, CSSProperties, CSSValue, CustomValue, PropertyFunction, PropertyValueDefinition, PropertyValueMap, RenderProps, ShorthandProperty, StyleFunction, StyleValue, Theme, ThemeProperties, Value} from './types';
 
 let defaultArbitraryProperty = <T extends Value>(value: T, property: string) => ({[property]: value} as CSSProperties);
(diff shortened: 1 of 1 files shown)


packages/@react-spectrum/s2/style/style-macro.ts
```diff
@@ -115,14 +115,40 @@ interface MacroContext {
...
+  for (let i = 0; i < v.length; i++) {
+    hash = ((hash << 5) + hash) + v.charCodeAt(i) >>> 0;
+  }
+  return hash;
+}
+
+function generateThemeHash(theme: Theme): string {
+  let themeString = JSON.stringify(theme, (_, value) => {
+    if (typeof value === 'function') {
+      return value.toString();
```
devongovett, line 131: I guess a problem here would be that many of the functions will be the same, just with different arguments. For example, `createColorProperty` always returns a function that will stringify to the same thing, but it uses different values from its closure scope...

reidbarber:
Closing for now in favor of simply preventing S1/S2 style macro collision. Will re-visit if needed later for preventing collisions between multiple S2 versions.

PR #7316, opened 2024-11-04
revert: radio animation

yihuiliao:
Closes 

## ✅ Pull Request Checklist:


reidbarber (approved):

snowystinger (approved):
out of curiosity what happened here?

yihuiliao:
> out of curiosity what happened here?

Basically, because we updated the transition to just animate the border width but that meant that the press scale and hover state were no longer animating. It's a subtle difference but it's there. Also, we had apparently discussed with design about what animation radio's should have and decided that animating the border width might be too much along with press scaling. 

However, in order to fix the flash, we'd also need to animate the border width but we don't want to do that so we're reverting this fix in favor of better animations.

PR #7329, opened 2024-11-06
chore: Update S2 icons

devongovett:
Adds many new S2 workflow icons, linear and gradient illustrations, and UI icons. Gradient illustrations now come in 3 sizes just like the linear ones, so we generate a component to swap between them depending on the t-shirt size. They are also now re-organized into subdirectories (generic1 and generic2), sorry for the huge diff. Design has actually renamed these categories to "allproduct1" and "allproduct2" but I'm not sure it's worth the breaking change...


reidbarber (approved):

LFDanLu:
Chromatic found some separate issues which I'm pretty sure aren't because of this PR, but just wanna make sure that this is expected:
image

aka darker fill.

LFDanLu:
nevermind, found the proper figma file

LFDanLu (approved):

snowystinger (approved):
Ooof, this is rough
Approving and we can figure it out after

PR #7352, opened 2024-11-11
Support custom dialog layouts in S2

devongovett:
This splits the S2 Dialog API into three components:

1. `StandardDialog` is what used to be called `Dialog`. It is now only used for modals (not popovers), and includes a standard layout with slots for the heading, content, hero image, etc.
2. `FullscreenDialog` is a fullscreen or takeover modal. It is similar to `StandardDialog` but has a different layout.
3. `Dialog` is now used for custom dialog layouts, such as popovers and other custom modals. It can have padding built in or allow its content to go edge-to-edge.

In addition, this enables menus to be embedded inside popover dialogs, for usecases where there are additional interactive elements adjacent to the menu within the popover. This required some refactors to the RAC submenu.

Finally, we also now support avatars inside S2 ActionButton, which work similar to icons.


packages/@react-aria/focus/src/FocusScope.tsx
```diff
@@ -608,7 +608,7 @@ function useRestoreFocus(scopeRef: RefObject<Element[] | null>, restoreFocus?: b
       }
 
       let focusedElement = ownerDocument.activeElement as FocusableElement;
-      if (!isElementInScope(focusedElement, scopeRef.current)) {
+      if (!isElementInChildScope(focusedElement, scopeRef) || !shouldRestoreFocus(scopeRef)) {
```
devongovett, line 611: FocusScope changes are so that focus restoration works across React portals. The difference here is that `isElementInChildScope` checks whether the element is either a descendant of this focus scope in the DOM, or any other child scope (which might be in a portal).

packages/@react-aria/overlays/src/usePopover.ts
```diff
@@ -82,7 +82,8 @@ export function usePopover(props: AriaPopoverProps, state: OverlayTriggerState):
 
   let {overlayProps, underlayProps} = useOverlay(
     {
-      isOpen: state.isOpen,
+      // If popover is in the top layer, it should not prevent other popovers from being dismissed.
+      isOpen: state.isOpen && !otherProps['data-react-aria-top-layer'],
```
devongovett, line 86: This is kinda not ideal. Should we have a more explicit prop for this? Basically we have to prevent submenus from being added to the `visibleOverlays` stack inside `useOverlay`, that way they don't prevent the click outside behavior of the root menu.

packages/@react-spectrum/menu/src/ContextualHelpTrigger.tsx
```diff
@@ -151,7 +151,7 @@ function ContextualHelpTrigger(props: InternalMenuDialogTriggerProps): ReactElem
         containerPadding={0}
         hideArrow
         enableBothDismissButtons>
-        <FocusScope restoreFocus>
+        <FocusScope>
```
devongovett, line 154: Due to the above FocusScope change, we don't want to restoreFocus from the unavailable menu dialog because that will move focus back to the parent menu item when tabbing out. There is a test that asserts that tabbing out moves to the item after the root menu trigger. Perhaps we might want to change this behavior at some point. For now, just retaining existing behavior.

packages/react-aria-components/src/Menu.tsx
```diff
@@ -233,7 +209,6 @@ function MenuInner<T extends object>({props, collection, menuRef: ref}: MenuInne
             scrollRef={ref} />
         </Provider>
       </div>
-      <div ref={setPopoverContainer} style={{width: '100vw', position: 'absolute', top: 0, ...leftOffset}} />
```
devongovett, line 236: The above focus scope changes mean we no longer need to mount the submenu inside the parent menu. This wasn't possible when the menu is embedded inside a larger popover because it would get clipped by overflow: auto. `data-react-aria-top-layer` prevents the parent overflow from making the submenu `aria-hidden`.

packages/react-aria-components/src/Dialog.tsx
```diff
@@ -39,7 +41,9 @@ export const OverlayTriggerStateContext = createContext<OverlayTriggerState | nu
  * A DialogTrigger opens a dialog when a trigger element is pressed.
  */
 export function DialogTrigger(props: DialogTriggerProps) {
-  let state = useOverlayTriggerState(props);
+  // Use useMenuTriggerState instead of useOverlayTriggerState in case a menu is embedded in the dialog.
+  // This is needed to handle submenus.
+  let state = useMenuTriggerState(props);
```
devongovett, line 46: Not super ideal...

snowystinger: I'm assuming we do this because submenus behave differently than say a picker inside a popover when it comes to closing? So we can't push it down in a generalized overlaytrigger?

Will this be an issue if we have a Dialog with a Menu inside it? will try to answer myself after review

snowystinger:
Should `/@react-spectrum/s2:CloseButtonProps` be exposing so many of the button props? like the ones related to forms?

packages/@react-spectrum/s2/src/ActionButton.tsx
```diff
@@ -57,7 +58,7 @@ export interface ActionButtonProps extends Omit<ButtonProps, 'className' | 'styl
 }
 
 // These styles handle both ActionButton and ToggleButton
-const iconOnly = ':has([slot=icon]):not(:has([data-rsp-slot=text]))';
+const iconOnly = ':has([slot=icon], [slot=avatar]):not(:has([data-rsp-slot=text]))';
```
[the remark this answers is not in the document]
devongovett: what if someone puts something other than text inside a button, e.g. a badge or something? I guess we wouldn't want it to be square then?

packages/@react-spectrum/s2/src/Dialog.tsx
```diff
@@ -10,128 +10,127 @@
...
-  switch (ctx.type) {
+  let dialog = (
+    <RACDialog
+      {...props}
+      ref={domRef}
+      style={props.UNSAFE_style}
+      className={(props.UNSAFE_className || '') + dialogStyle({type, isFullBleed}, props.styles)}>
+      {composeRenderProps(props.children, (children) => (
+        // Reset OverlayTriggerStateContext so the buttons inside the dialog don't retain their hover state.
+        <OverlayTriggerStateContext.Provider value={null}>
```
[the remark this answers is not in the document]
devongovett: This came from the previous implementation, but I think it's due to S2 buttons in particular consuming `OverlayTriggerStateContext` to apply the hover state. If we put this in RAC it would prevent people from accessing the state inside all overlays which might not be expected.

packages/@react-spectrum/s2/src/Menu.tsx
```diff
@@ -98,12 +98,18 @@ export let menu = style({
...
+  maxWidth: {
+    isPopover: 320
+  },
+  padding: {
+    isPopover: 8
+  },
   fontFamily: 'sans',
   fontSize: 'control'
-});
+}, getAllowedOverrides());
```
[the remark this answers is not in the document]
devongovett: I added it here so that you could embed a Menu inside another element (like a popover) and set the width to 100%

packages/@react-spectrum/s2/src/Popover.tsx
```diff
@@ -232,7 +236,7 @@ function Popover(props: PopoverProps, ref: DOMRef<HTMLDivElement>) {
         // Override default z-index from useOverlayPosition. We use isolation: isolate instead.
         zIndex: undefined
       }}
-      className={(renderProps) => UNSAFE_className + mergeStyles(popover({...renderProps, size, isArrowShown: !hideArrow, colorScheme}), styles)}>
+      className={(renderProps) => UNSAFE_className + mergeStyles(popover({...renderProps, size, isArrowShown: !hideArrow, colorScheme, isSubmenu: renderProps.trigger === 'SubmenuTrigger'}), styles)}>
```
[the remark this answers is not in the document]
devongovett: This comes from RAC. Our triggers set this prop

packages/@react-spectrum/s2/src/Dialog.tsx
```diff
@@ -10,128 +10,127 @@
...
   /**
    * Whether a popover type Dialog's arrow should be hidden.
    */
   hideArrow?: boolean,
-  /**
-   * Whether pressing the escape key to close the dialog should be disabled.
-   */
-  isKeyboardDismissDisabled?: boolean
+  /** Whether the content of the Dialog should extend edge-to-edge, without any padding. */
+  isFullBleed?: boolean
```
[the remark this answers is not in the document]
devongovett: padding="none | default"

packages/@react-spectrum/s2/src/ActionButton.tsx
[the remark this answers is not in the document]
snowystinger: ah, yeah, the notification?

packages/@react-aria/focus/src/FocusScope.tsx
snowystinger: do any of our tests cover the changes in this file indirectly?

packages/@react-aria/overlays/src/usePopover.ts
LFDanLu: For posterity, we discussed this and decided to hold off on making a separate prop until after release, will need to think more about the case where this will cause a submenu to still be visible/interactable when a separate modal opens (something that we do want for Toasts via top-layer but not for submenus)

packages/@react-spectrum/s2/src/FullscreenDialog.tsx
```diff
@@ -0,0 +1,153 @@
...
+  flexGrow: 1,
+  overflowY: {
+    default: 'auto',
+    // Make the whole dialog scroll rather than only the content when the height it small.
+    '@media (height < 400)': 'visible'
+  },
+  font: 'body'
+});
+
+const buttonGroup = style({
```
reidbarber, line 60: Should this have some padding? 

FullscreenDialog - Example ⋅ Storybook

packages/@react-spectrum/s2/src/Popover.tsx
```diff
@@ -221,6 +238,26 @@ function Popover(props: PopoverProps, ref: DOMRef<HTMLDivElement>) {
     }
   }, [locale, direction, domRef]);
 
+  // On small devices, show a modal (or eventually a tray) instead of a popover.
+  let isMobile = useIsMobileDevice();
```
LFDanLu, line 242: Ah, I had previously commented that my phone was showing the Menu as a modal now instead of a popover but just noticed this. I guess that's ok since we don't truly support mobile fully until we implement trays in S2 but it is a bit weird that the submenus appear as modals. 

However, S2 ComboBox and Picker doesn't even display any sort of dropdown or modal anymore but again, might be ok if our stance is that mobile isn't supported yet. Otherwise, if we want to keep them rendering as popovers, then we might have to do some temp API stuff to preserve that behavior for now.

packages/react-aria-components/src/Dialog.tsx
LFDanLu: > Will this be an issue if we have a Dialog with a Menu inside it? will try to answer myself after review

Theoretically should be fine I imagine since the MenuTrigger in the Menu provides its own menu state context and its wrapped menu with use that rather than the DialogTrigger's

LFDanLu (approved):
Approving for testing, just one comment about the mobile experience

devongovett:
Gonna merge this for testing. Feel free to add your comments to the testing sheet.

PR #7365, opened 2024-11-14
fix: prefix S1 style macro class names

reidbarber:
Prefixes class names produced by the S1 style macro with `s1-` to avoid class name collisions with S2 style macro when used on the same page.

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/style-macro-s1/src/style-macro.ts
+++ b/packages/@react-spectrum/style-macro-s1/src/style-macro.ts
@@ -11,6 +11,9 @@
  */
 import type {Condition, CSSProperties, CSSValue, CustomValue, PropertyFunction, PropertyValueDefinition, PropertyValueMap, StyleFunction, StyleValue, Theme, ThemeProperties, Value} from './types';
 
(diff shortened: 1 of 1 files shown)


packages/@react-spectrum/style-macro-s1/src/style-macro.ts
```diff
@@ -367,7 +370,7 @@ interface Rule {
 // This maps to an alphabet containing lower case letters, upper case letters, and numbers.
 // For numbers larger than 62, an underscore is prepended.
 // This encoding allows easy parsing to enable runtime merging by property.
-function generateName(index: number, atStart = false): string {
```
reidbarber, line 370: `atStart` no longer needed since it was being used to avoid class names starting with numbers, which can't happen anymore due to prefix.

LFDanLu (approved):
Chromatic looks good to me, do we happen to have a test app that demonstrates the previous issue?

reidbarber:
> Chromatic looks good to me, do we happen to have a test app that demonstrates the previous issue?

I don't think so, I think we just got reports of this from users.

PR #7376, opened 2024-11-15
feat: event leak demo

nwidynski:
Closes [comment](https://github.com/adobe/react-spectrum/pull/7277#issuecomment-2477755073)

> **NOT INTENDED FOR MERGE **  - I only needed access to Storybook examples for demo purposes.

TLDR; I did not test for every possible interaction for now and also limited the assertions to grid interaction keys only. It's very likely that additional components will also fail when tested more throughly. Point of this demo was to hightlight that many components, not only nested collections, will cause problems when used inside grids.

Also related are #7370 and my latest [comment](https://github.com/adobe/react-spectrum/issues/7321#issuecomment-2460557271) on #7321.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Open [localhost:9003](http://localhost:9003/?path=/story/react-aria-components--event-leak-grid&providerSwitcher-express=false&strict=true)
2. Navigate to `Event Leak Grid` story
2. Watch the Playwright test

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -7,6 +7,7 @@ module.exports = {
   ],
 
   addons: [
+    "@storybook/addon-interactions",
     "@storybook/addon-actions",
     "@storybook/addon-a11y",
     "@storybook/addon-controls",
--- a/packages/@react-aria/gridlist/src/useGridListItem.ts
+++ b/packages/@react-aria/gridlist/src/useGridListItem.ts
@@ -257,7 +257,7 @@ export function useGridListItem<T>(props: AriaGridListItemOptions, state: ListSt
   let rowProps: DOMAttributes = mergeProps(itemProps, linkProps, {
     role: 'row',
     onKeyDown: keyboardNavigationBehavior === 'tab' ? onKeyDown : undefined,
-    onKeyDownCapture:  keyboardNavigationBehavior === 'tab' ? undefined : onKeyDown,
+    onKeyDownCapture: keyboardNavigationBehavior === 'tab' ? undefined : onKeyDown,
     onFocus,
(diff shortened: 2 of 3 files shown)


LFDanLu:
Just chiming in, I believe I actually just ran into this event leak behavior myself in Autocomplete since I was dispatching fake keyboard events to the wrapped collection which bubbled up to the dialog and closed it prematurely. I'll make a duplicate of this branch since security dinged us on running builds on fork PRs and see about getting the team to take a look as soon as possible

LFDanLu:
For others looking at the PR, here is a build link: https://reactspectrum.blob.core.windows.net/reactspectrum/0b6cec88ba545ce9c6f8086c28bb33d07752ece3/storybook/index.html?path=/story/react-aria-components--event-leak-grid&providerSwitcher-express=false

LFDanLu:
@nwidynski Apologies for the delay, the team discussed this some more yesterday and moving forward we'll want to be more strict/consistent about stopping propagation on keyboard events. This usually happens by default when calling `useKeyboard` but as you've discovered we haven't been the best about consistently using that hook or making sure that a base level onKeyDown gets provided to the hook if none exists. We'll need to do an audit of all our components and make sure that keyboard events that get handled by a component have their propagation stopped and that keyboard events that aren't handled are still propagated. 

This change will hopefully make handing interactive components within a grid much easier as well as possibly allowing the typeahead listener to become a non-capturing listener.

nwidynski:
@LFDanLu Got it, I generally very much agree with the sentiment to have atomic building blocks. Let's hope this doesn't cause too many behavior issues in user applications 🙏 

As this effort will likely take a while (maybe next major?), we will begin work on open-sourcing a custom grid virtualizer which prevents these leaks from reaching the collection 👍

PR #7379, opened 2024-11-15
feat: nested collection support

nwidynski:
Closes [#7277 (comment)](https://github.com/adobe/react-spectrum/pull/7277#issuecomment-2479395031)

As discussed, this change focuses only on the bare minimum to get nested collections to work.

## ✅ Pull Request Checklist:


packages/@react-aria/gridlist/src/utils.ts
```diff
@@ -30,13 +30,18 @@ export function getRowId<T>(state: ListState<T>, key: Key) {
...
+  if (!id) {
+    throw new Error('Unknown list');
+  }
+
   if (typeof key === 'string') {
-    return key.replace(/\s*/g, '');
+    return key.replace(/\s*/g, '').replace(id, '');
   }
 
-  return '' + key;
```
nwidynski, line 41: Not sure, why were we casting here?

LFDanLu:
@nwidynski Heya, sorry about the delay, we were pretty busy with getting the release out. As mentioned in https://github.com/adobe/react-spectrum/pull/7277#issuecomment-2477762761 we are leaning towards supporting inputs in Gridlist first via something like https://github.com/adobe/react-spectrum/pull/7360, which you had already noted doesn't solve the typeahead collision problem. 

However, instead of making typeahead disable-able via a prop, we tossed around an idea of perhaps just doing some detection in the useTypeselect hook to see if the user was typing in an input and simply early return at that point. Any thoughts on that approach?

nwidynski:
@LFDanLu No worries about the timeline 👍 

I suppose you were exploring that idea to lighten the burden on the developer? If so, I don't see why these necessarily have to exclude each other. Preventing type select only when focus is currently on an input, will still cause issues with nested collections. Was there other reasoning for prohibiting the exposure of `disallowTypeAhead`?

As mentioned (see [comment](https://github.com/adobe/react-spectrum/pull/7277#issuecomment-2479395031)), this PR focuses only on the essentials of supporting **nested collections** and can be seen as independent of #7360. Only the `Document` key and type select are things we **have** to fix upstream to implement our use-case. Everything else we proposed, including #7360, we can temporarily implement ourselves through a custom virtualizer.

On that note, any feedback from the team on #7376?

LFDanLu:
There was also a concern about exposing `disallowTypeAhead` since it may not be clear to a user when/when not to disable it. That being said, at a glance it doesn't appear in the grid spec so maybe not as core of an interaction pattern as previously thought. No feedback on #7376 just yet, I hope to take a look at it soon as soon as I get to a good place with the SearchableMenu/Autocomplete support myself.

packages/@react-aria/collections/src/CollectionBuilder.tsx
```diff
@@ -51,6 +52,12 @@ export function CollectionBuilder<C extends BaseCollection<object>>(props: Colle
   // This is fine. CollectionDocumentContext never changes after mounting.
   // eslint-disable-next-line react-hooks/rules-of-hooks
   let {collection, document} = useCollectionDocument(props.createCollection);
+
+  if (props.id) {
+    // @ts-expect-error
+    document.key = props.id;
+  }
```
LFDanLu, line 59: This came up in the a team discussion yesterday, but this approach doesn't work if there is an `id` provided to the collection component children (e.g. `<GridItem id="1" />`) unfortunately. If said `id` is provided, that becomes the `data-key` and said `id`s need only be unique per collection, not globally. This is already a problem for us in https://github.com/adobe/react-spectrum/blob/b37f2790b77543a447fb0dc1833b5d0cfa8d7da5/packages/%40react-aria/selection/src/useSelectableCollection.ts#L143
as you may have noticed, so we are trying to think of a better way to go about this (always prefix the data keys with a collection id? Create a weak map that ties the generated child ids to the collection and use that for lookup?).

I believe this is what you were trying to fix with the above change, lemme know if not otherwise

packages/@react-aria/gridlist/src/useGridList.ts
```diff
@@ -51,6 +51,11 @@ export interface AriaGridListProps<T> extends GridListProps<T>, DOMProps, AriaLa
 export interface AriaGridListOptions<T> extends Omit<AriaGridListProps<T>, 'children'> {
   /** Whether the list uses virtual scrolling. */
   isVirtualized?: boolean,
+  /**
+   * Whether typeahead navigation is disabled.
+   * @default false
+   */
+  disallowTypeAhead?: boolean,
```
LFDanLu, line 58: With the eventual changes to stop propagation on handled keyboard events as mentioned in https://github.com/adobe/react-spectrum/pull/7376#issuecomment-2524214331, hopefully we would be able to make the typeahead listener not capturing which would help prevent typeahead from colliding with nested component/grid interactions. We would probably need to special case `Space` so as to not collide with row selection and type ahead, but that might be doable via checking if the `Space` in the middle of a string the user is typing (aka that would be typeahead so don't perform row selection) or not (aka that would be row selection so stop propagation at the row level). This would mean communicating from the grid level to the row level that typeahead is in progress or not though

If that was accomplished though, would you need `disallowTypeAhead` anymore?

packages/@react-aria/collections/src/CollectionBuilder.tsx
nwidynski: Puh, it's been a while now 😆 IIRC, I also encountered this but concluded this case could be ignored, since at that point the item keys of the nested collection would also have to be manually set to a colliding value. If you want to make sure though, I would probably prefix them always.

packages/@react-aria/gridlist/src/useGridList.ts
nwidynski: Hm, sounds like that will take a while, but yes I think that could work 👍 

Unfortunately, I think `Space` will not be the only special case to handle. I also see potential collisions with alphanumeric cell entry once the full W3C grid edit mode spec is adopted.

LFDanLu: > Maybe disallowTypeAhead could still be passed undocumented in the meantime?

potentially if only to get something out for you sooner rather than later, I'll ask the team

packages/@react-aria/collections/src/CollectionBuilder.tsx
LFDanLu: Said `node.key` is used in the generation of the whole collection key map setting up the next/prev relationship of each node in the collection, thus allowing us to do node look ups/return the key in selection or action/keyboard navigation etc. If the user provides their own `id`, we use that so that we can return the same `id` in `onSelectionChange`/`onAction` thus allowing the user to know which item is being acted upon.

nwidynski: Thanks for circling back on this 👍 I do understand the purpose of the `node.key` in general and most of its associated functionality, so maybe I should clarify further.

What I was wondering about is why a developer is able to override the `node.key` composition (e.g. `react-aria-<key>`) by passing the `id` property to the leaf component. I'm asking because I don't understand why this was necessary in the first place.

As of now, when not customizing the `id` property, the internal `node.key` is passed to both action and selection callbacks. In this mode, the user simply has to strip the internal `react-aria` key prefix to access the item. Thus the only purpose i see this behavior currently solving is making the reverse mapping slightly more convenient. 

Am I missing additional side-effects this is addressing? If not, I'm unsure whether the existence of this feature is even warranted in the first place (other than backwards compatibility).

LFDanLu: We could've possibly retained the `react-aria` prefix and simply appended the user's provided `id` and handled stripping the prefix when calling `onSelectionChange` and/or `onAction`, I believe it just felt easier to replace the `node.key` at the time, unsure if there were any other factors. It also matched what we had in the old collections api in React Spectrum where we accepted `key`s from the user and used those in the collection map generation.

packages/@react-aria/gridlist/src/useGridList.ts
LFDanLu: so the team is ok with exposing `disallowTypeAhead` since it isn't required behavior as per the grid spec and there is still the potential of collisions as you mentioned above. As for the collection id stuff that we chatted above above, are those changes required or could you ensure that the ids for each collection are globally unique (though I guess you can't if users are the ones providing those)?

nwidynski: Great 👍

In terms of the collection ids, yes those changes are required for the reason you mentioned.

How would the team view a removal of the `id` property from leaf components?

Ofc it could impact existing applications if they are relying on the`data-key` for querying/styling, but this would allow us to streamline callback behavior and solve the outstanding issue here at once. Impacted users would have to migrate to a custom data attribute.

Im out of ideas on how else to approach this issue 😬

nwidynski: I did want to maintain the extraction performed by `useCachedChildren` and only remove the `rendered.props.id` part of it. Was thinking we might be able to get away with using auto-incremental ids to maintain the ability to reverse map. Not a big fan of this myself however because of re-ordering. 

In the meantime, I thought of another possible angle which i pushed up. Let me know what you think 👍

LFDanLu: Thanks for the update! I think this is promising, we may need to think about moving the id unscoping to the hook level perhaps in case there are React Aria hook only users who've also leveraged the Collection implementation. I also noticed that the `idScope` isn't being applied for static collections that don't have `id`'s provided to their children (in your story the GridListItems still have `react-aria-1`/etc instead of being scoped), I assume that wasn't intentional? 
The condition below that you have in `useCachedChildren` will return true since the child doesn't have an id, perhaps it include that in the check?
```
if (!child || !child.props.id || !idScope) {return child;}
```

nwidynski: I added that intentionally for now until I figure out what causes the `Table.test.js` to fail when not early-returning. Unfortunately I only get an undescriptive `Should not already be working.` error message.

Also wanted to note that key structure will change slightly for these static nodes, since I don't have access to the auto-generated `node.key` at this point in time. Instead I think the index (e.g. `<idScope>:<index>`) will have to do.

I also thought about moving the unscoping into the hook layer, but wasn't sure whether we actually would want that. Since `idScope` is pretty implementation specific & set in the component layer I figured coupling it there might make more sense. I also felt like `useListState()` shouldn't have to be aware of a scope.

In case it shall still move further down, we would have to either have forward the `idScope` everywhere or maybe attach it to the `BaseCollection`, since we can't access the `CollectionContext` in the hidden tree unfortunately. 

Looking forward to additional feedback here 👍

nwidynski: Btw, another question popped up while fixing `Table.test.js`. Maybe its worth thinking about merging these together instead of picking either one (e.g. `<ctxIdScope>:<idScope>:<index>`).

https://github.com/adobe/react-spectrum/blob/fee532d6a2fc025177d96823d0b5e361b4b79934/packages/%40react-aria/collections/src/CollectionBuilder.tsx#L202

LFDanLu: > Also wanted to note that key structure will change slightly for these static nodes, since I don't have access to the auto-generated node.key at this point in time. Instead I think the index (e.g. <idScope>:<index>) will have to do.

I think that is ok, I feel like the end user mainly cares about the structure/format of the node key w/ respect with what gets returned by the various event handlers. I personally don't consider the `data-key` as public API but will discuss with the team further.

> I also thought about moving the unscoping into the hook layer, but wasn't sure whether we actually would want that. 
>  Since idScope is pretty implementation specific & set in the component layer I figured coupling it there might make more sense. I also felt like useListState() shouldn't have to be aware of a scope.
In case it shall still move further down, we would have to either forward the idScope or maybe attach it to the BaseCollection, since we can't access the CollectionContext in the hidden tree unfortunately.

That's true, mainly was thinking if hook users would want to implement nested collection support themselves and/or support it out of the box but that can be addressed down the line. 

> Maybe its worth thinking about merging these together instead of picking either one (e.g. <ctxIdScope>:<idScope>:<index>).

Interesting, so that would essentially append the parent collection's scope to the child collection, automatically ensuring each level would have a unique scopeId even if the user provided the same idScope to both the parent and the child. I think that is probably worthwhile, digging to see if there would any reason why not.

packages/react-aria-components/src/GridList.tsx
```diff
@@ -97,38 +104,57 @@ interface GridListInnerProps<T extends object> {
...
+  let selectedKeys = useMemo(() => !props.selectedKeys || props.selectedKeys === 'all' ? props.selectedKeys : [...props.selectedKeys].map(key => `${props.id}:${key}`), [props.selectedKeys, props.id]);
+  let defaultSelectedKeys = useMemo(() => !props.defaultSelectedKeys || props.defaultSelectedKeys === 'all' ? props.defaultSelectedKeys : [...props.defaultSelectedKeys].map(key => `${props.id}:${key}`), [props.defaultSelectedKeys, props.id]);
+  let onSelectionChange = useMemo(() => !props.onSelectionChange ? props.onSelectionChange : (keys: Selection) => {
+    props.onSelectionChange?.(new Set([...keys].map(key => typeof key === 'string' ? key.replace(`${props.id}:`, '') : key)));
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.onSelectionChange, props.id]);
+  let onAction = useMemo(() => !props.onAction ? props.onAction : (key: Key) => {
+    props.onAction?.(typeof key === 'string' ? key.replace(`${props.id}:`, '') : key);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.onAction, props.id]);
```
LFDanLu, line 119: One thing I just noticed is that this unscoping doesn't work with the Virtualizer wrapped GridList. It seems like the GridList's collection gets built with one instance of the autogenerated id, but then another render happens where the GridList generates another scope id but the collection itself doesn't get regenerated. This results in the above `props.id` mismatching with the collection's node key prefix. Poking around to see if I can figure out why that is

nwidynski: Sounds like strict mode maybe? Will habe to check in depth tmrw.

nwidynski: Good catch, just got around to take a proper look. I pushed a fix to remote 👍 

The issue wasn't with `useId` but rather with the drag and drop hooks generating an id without taking the `defaultId` into account. That also explains why you only observed it with `GridList` as all other virtualized examples simple didn't have drag and drop implemented 😄 

I opted to override the id in `mergeProps` of the component, but longterm might be wise to think about forwarding the `default` to the drag and drop hooks?

LFDanLu: Ah, nice catch! I completely forgot that the drop hooks generate an `id` of their own, was so confused how some autogenerated id was making it though. Unfortunately this means I've found even more problematic areas in the code 😅.

1. I think it is a good idea to forward the GridList's `id` to GridList's call o `useDroppableCollection`  so the weak map id [here](https://github.com/adobe/react-spectrum/blob/4b971490601c43f6fffb663cfdf7081ddc77c04d/packages/%40react-aria/dnd/src/useDroppableCollection.ts#L748-L749) is consistent. It actually already works at the moment due to `mergeProps` deduping the ids and causing all other instances of `useId` to use the last one provided ([aka via this call of mergeProps)](https://github.com/adobe/react-spectrum/blob/8e9dfb81d330aab2ce1e322443e3b4013c149003/packages/react-aria-components/src/GridList.tsx#L383), but it would be best not to rely on this IMO.

2. I noticed that the reordering in the Virtualized Gridlist story is broken because the selectionManager/collection has the prefixed `id` for the items (which is expected) and will use those prefixed `id` as [the dragged keys when drag starts](https://github.com/adobe/react-spectrum/blob/4b971490601c43f6fffb663cfdf7081ddc77c04d/packages/%40react-aria/dnd/src/useDraggableItem.ts#L77-L79) (see [here](https://github.com/adobe/react-spectrum/blob/4b971490601c43f6fffb663cfdf7081ddc77c04d/packages/%40react-stately/dnd/src/useDraggableCollectionState.ts#L73-L85) for how it is pulling them out of the `keyMap`). These are stored globally and used in [useDroppableCollection](https://github.com/adobe/react-spectrum/blob/4b971490601c43f6fffb663cfdf7081ddc77c04d/packages/%40react-aria/dnd/src/useDroppableCollection.ts#L99) in things like `onReorder`/etc, resulting in logic like https://github.com/adobe/react-spectrum/blob/4b971490601c43f6fffb663cfdf7081ddc77c04d/packages/react-aria-components/stories/GridList.stories.tsx#L130-L136 no longer working because the `list` has the user's original non prefixed `id` but `e.keys` has the prefixed `ids` and thus we have a mismatch. If we store the `prefix` and as part of the `collection`/`SelectionManager` via `BaseCollection` like you mentioned before, we could perhaps strip off the prefix and hopefully fix this

However, I'm now a bit more worried since this class of issue is due to our original implemented assumption that the `ids` stored in the `collection`/`SelectionManager` matches the original `ids` format provided by the user. This assumption is likely to exist in many other hooks, and I'm quite troubled by the fact that the tests ran clean...

LFDanLu: Perhaps we should step back and instead have the collection id scoping only applied to the generated DOM data-key rather than affecting the keys at collection/SelectionManger level.

nwidynski: haha, that's what i just had in mind aswell. pushed it to remote 👍

packages/@react-aria/selection/src/DOMLayoutDelegate.ts
```diff
@@ -24,7 +24,7 @@ export class DOMLayoutDelegate implements LayoutDelegate {
     if (!container) {
       return null;
     }
-    let item = key != null ? container.querySelector(`[data-key="${CSS.escape(key.toString())}"]`) : null;
+    let item = key != null ? container.querySelector(`[data-key$="${CSS.escape(key.toString())}"]`) : null;
```
nwidynski, line 27: Not really happy with this, but passing `idScope` down here will be a mess.

packages/react-aria-components/src/GridList.tsx
nwidynski: Btw, funnily enough your call about `useId` caused me to unearth a tricky little bug in it. 

When a component has 2 ids for some reason (e.g. because of drag & drop hooks), merging these ids in the order `id1 -> id2 -> id1` causes the component to infinitely re-render when in strict mode 😄. 

Not sure whether it's worth to fix that, but could be untangled by `mergeProps` perhaps.

packages/@react-aria/selection/src/DOMLayoutDelegate.ts
LFDanLu: what if you apply the collection's `idScope` to the collection node as its own `data-key` and look that up via the container ref? That way you can construct the expected node data key here. It might be good to make a util function that generalizes the construction of the nodeKey just so we can make sure we are consistent with the `${idScope}-${manager.focusedKey.toString()}` structure.

packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -369,7 +373,8 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
 
     if (manager.focusedKey != null && scrollRef.current) {
       // Refocus and scroll the focused item into view if it exists within the scrollable region.
-      let element = scrollRef.current.querySelector(`[data-key="${CSS.escape(manager.focusedKey.toString())}"]`) as HTMLElement;
+      let nodeKey = idScope ? `${idScope}-${manager.focusedKey.toString()}` : manager.focusedKey.toString();
```
LFDanLu, line 376: One other place I think you may need to apply this change as well is in this function: https://github.com/adobe/react-spectrum/blob/50c7ada5d1880a174b6b6d3f43e8d90ee9bd4ad8/packages/%40react-aria/dnd/src/ListDropTargetDelegate.ts#L102-L108. The elementMap uses the data-keys on the nodes which are now prefixed: https://github.com/adobe/react-spectrum/blob/50c7ada5d1880a174b6b6d3f43e8d90ee9bd4ad8/packages/%40react-aria/dnd/src/ListDropTargetDelegate.ts#L90-L97 but then we use the collection keys (aka non prefixed keys) to access the items in the elementMap on line 108.

nwidynski: Thanks, sometimes I wish test suite would cover all these 😓 .

As mentioned above, there is an edge case in here where a nested collection could look like this:

```tsx
<GridList id="grid" data-key="grid">
  <GridListItem id="dog" data-key="grid:dog"/>
  <GridListItem id="cat" data-key="grid:cat">
    <GridList id="grid:dog" data-key="grid:dog">
      <GridListItem id="dog" data-key="grid:dog:dog" />
    </GridList>
  </GridListItem>
</GridList>
```

causing the drop target to be incorrectly determined as the nested collection.

packages/@react-aria/selection/src/DOMLayoutDelegate.ts
LFDanLu: sure, I'll ask if anyone one remembers any other areas where we might have assumptions for the application of `data-key`. Does feel a bit iffy to pull in the `@react-aria/selection` package into the `@react-aria/dnd` package just for this util, I'll let others weight in there. If you'd like you could skip using that function for the dnd package for now, can be followup

packages/@react-aria/selection/src/useSelectableCollection.ts
LFDanLu: Yeahhhh, I think we are going to need to expand the coverage of the RAC tests, I think we were hoping the tests in RSP would be redundant enough since they leverage the same hooks, but the different Collections API is proving that wrong. 

As for the edge case checking for role could certainly help. I think in general that the ListDropTarget delegate and other drag and drop logic will need to be updated for nested collection support anyways. Would changing the `idScope` from matching the user provided `id` to an autogenerated one help reduce the likelihood of this edge case as well? Reusing the `id` as the `idScope` feels more of a convenience and is certainly helpful for the user if they wanted to use `data-key` as a targeting strategy, but I'd say its not really meant as public API.

packages/@react-aria/selection/src/DOMLayoutDelegate.ts
nwidynski: Ideally I would like to house the `getNodeKey` util inside `@react-aria/collections`. I feel like it fits best there and I would like to leverage it for the `idScope` of `useCachedChildren` too, since there are discrepancies currently.

I would argue the impact of that to be negligible as the vast majority of users will already have an implicit import of `@react-aria/collections` when consuming either `@react-aria/dnd` or `@react-aria/selection`, especially now that S2 is streamlining to this collection implementation as well. 

Let me know what the team thinks of that though 👍

packages/@react-aria/selection/src/useSelectableCollection.ts
LFDanLu: Thanks for the updates and hope you had an enjoyable break as well! I will try to take a look at the changes soon, but just wanted to let you know that we are gearing up for another release so responses may be slow. Due to the nature of the changes here as well as the revelation that our test suites didn't catch as much as I would've liked, I'm leaning towards getting this into a testing session next week and having it bake a bit so that we can be sure that no other regressions get introduced. Unfortunately, this probably means it will miss the coming release window, apologies about that.

As for the test suite enhancement/`useId` fix, I'd personally be happy for that to be a follow up. I'll see if I can get some resourcing/time to help out with that work as well.

LFDanLu:
Just wanted to give you an update, but the team tested this today and things looked pretty good. Only thing that we noticed was some of the [examples in the Table docs](https://reactspectrum.blob.core.windows.net/reactspectrum/dcae403153e11aa3a8c658d00d068b7c3a9f0557/docs/react-aria/Table.html#selection) here broke, I'll see if I can get a chance to dig soon. Still seeing about getting another pair of eyes on this. The team is also juggling a couple of other PRs alongside this one with widespread changes (usePress refactor/shadowDOM support) and is a bit worried about stacking them all into a single release. Ideally we'd give them all some time to bake before shipping, will keep you posted.

snowystinger:
Hey, just wanted to let you know we haven't forgotten about this. We're just moving a bit slowly right now as a few team members are on vacation. Thank you for your patience.

devongovett:
Hope you don't mind, I've pushed some changes to try to contain this within the selection package as much as possible so we don't need to handle this in each individual collection component. This way we generate a collection id in `useSelectableCollection` (attaching it to the actual Collection object via a WeakMap), and assign it to a `data-collection` attribute at the root and on each individual item. Then we can query for items using `[data-collection="..."]` to get all items in the collection, or `[data-collection="..."][data-key="..."]` to get a specific item.

packages/react-aria-components/src/GridList.tsx
```diff
@@ -218,7 +223,7 @@ function GridListInner<T extends object>({props, collection, gridListRef: ref}:
       <div
         {...filterDOMProps(props)}
         {...renderProps}
-        {...mergeProps(gridProps, focusProps, droppableCollection?.collectionProps, emptyStatePropOverrides)}
+        {...mergeProps(gridProps, focusProps, droppableCollection?.collectionProps, emptyStatePropOverrides, {id})}
```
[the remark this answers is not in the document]
nwidynski: This addresses two issues. The first is #7655, which we scheduled as a follow-up to this PR but has since landed.
Second is that drag & drop hooks generate their own id without a `defaultId`, causing them to override custom ids.

packages/react-aria-components/docs/Table.mdx
```diff
@@ -386,7 +386,7 @@ function MyRow<T extends object>({id, columns, children, ...otherProps}: RowProp
   let {selectionBehavior, allowsDragging} = useTableOptions();
 
   return (
-    <Row id={id} {...otherProps}>
+    <Row id={id} {...otherProps} columns={columns}>
```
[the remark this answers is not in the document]
nwidynski: Yes it is, unless we revert the [scope change](https://github.com/nwidynski/react-spectrum/blob/c0c33a8bb2f87e365deb418a329b7635d1393029/packages/react-aria-components/src/Table.tsx#L1182) here as well. That change was introduced to prevent unnecessary re-renders in child collections as the `idScope` was needlessly propagated.

packages/@react-aria/collections/src/CollectionBuilder.tsx
```diff
@@ -199,7 +200,7 @@ const CollectionContext = createContext<CachedChildrenOptions<unknown> | null>(n
 export function Collection<T extends object>(props: CollectionProps<T>): JSX.Element {
   let ctx = useContext(CollectionContext)!;
   let dependencies = (ctx?.dependencies || []).concat(props.dependencies);
-  let idScope = props.idScope || ctx?.idScope;
+  let idScope = props.idScope && ctx?.idScope ? getNodeKey(props.idScope, ctx.idScope) : (props.idScope || ctx?.idScope);
```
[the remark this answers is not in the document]
nwidynski: This was intended as a courtesy for users of the standalone package who may not pass unique values to `idScopes`.

packages/react-aria-components/docs/Table.mdx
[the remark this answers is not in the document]
nwidynski: No it isn't, but we would need to revert the [scope change](https://github.com/nwidynski/react-spectrum/blob/c0c33a8bb2f87e365deb418a329b7635d1393029/packages/react-aria-components/src/Table.tsx#L1182) here as well. I'm noticing we should probably also clear CollectionContext inside <CollectionRoot /> as it's currently propagating both idScope and dependencies through document boundaries.

packages/react-aria-components/src/GridList.tsx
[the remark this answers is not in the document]
devongovett: hmm does the latter case still occur? I think `mergeProps` should merge the ids

snowystinger (dismissed):

devongovett (approved):

snowystinger (approved):

devongovett (approved):

packages/react-aria-components/src/GridList.tsx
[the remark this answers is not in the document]
nwidynski: You are right. @LFDanLu and I just agreed it would be better to hand the id down to gridlists call to `useDroppableCollection` to avoid implicitly relying on mergeProps https://github.com/adobe/react-spectrum/pull/7379#discussion_r1891009032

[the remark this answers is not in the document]
nwidynski: I will just push that as follow-up once i get into the office tomorrow 👍

PR #7385, opened 2024-11-18
feat: Allow arbitrary pixel sizes in S2 style macro

devongovett:
Previously, both spacing and sizing values were limited to a 4px grid. This makes sense for spacing, which we want to be consistent, but we found many instances where we needed sizes that were not available, especially in the larger value range. It's less important that sizes be consistent than spacing since we are not designing on a fixed grid.

This updates our sizing properties to accept arbitrary pixel values, which are then converted to rems and scaled on touch devices. This is basically what the `size()` macro did before, but now it's built in and accepted by default.

Updated many instances of the size macro and arbitrary value bracket syntax to numbers instead. Also found bunch of places where `size()` was being used in a spacing property, so updated those to `space()`.


LFDanLu (approved):

snowystinger (dismissed):

reidbarber (commented):
Does the codemod need updating? Can be a follow-up if so.

yihuiliao (approved):

reidbarber (approved):

PR #7392, opened 2024-11-19
fix: Revert displaying popovers as modals on mobile for now

devongovett:
Until we implement trays, we won't display popovers as modals on mobile. Also includes some example improvements, and fixes https://github.com/adobe/react-spectrum/pull/7352#discussion_r1847081211

### DIFF
--- a/packages/@react-spectrum/s2/src/Popover.tsx
+++ b/packages/@react-spectrum/s2/src/Popover.tsx
@@ -16,22 +16,19 @@ import {
   composeRenderProps,
   Dialog,
   DialogProps,
-  ModalRenderProps,
   OverlayArrow,
   OverlayTriggerStateContext,
   useLocale
 } from 'react-aria-components';
 import {colorScheme, getAllowedOverrides, StyleProps, UnsafeStyles} from './style-utils' with {type: 'macro'};
 import {ColorSchemeContext} from './Provider';
-import {DismissButton} from 'react-aria';
 import {DOMRef} from '@react-types/shared';
 import {forwardRef, MutableRefObject, useCallback, useContext} from 'react';
 import {keyframes} from '../style/style-macro' with {type: 'macro'};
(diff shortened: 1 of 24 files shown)


snowystinger (dismissed):

packages/@react-spectrum/s2/src/Popover.tsx
```diff
@@ -16,22 +16,19 @@ import {
...
 import {forwardRef, MutableRefObject, useCallback, useContext} from 'react';
 import {keyframes} from '../style/style-macro' with {type: 'macro'};
 import {mergeStyles} from '../style/runtime';
-import {Modal} from './Modal';
 import {style} from '../style' with {type: 'macro'};
 import {StyleString} from '../style/types' with {type: 'macro'};
-import {useDOMRef, useIsMobileDevice} from '@react-spectrum/utils';
+import {useDOMRef} from '@react-spectrum/utils';
 
 export interface PopoverProps extends UnsafeStyles, Omit<AriaPopoverProps, 'arrowSize' | 'isNonModal' | 'arrowBoundaryOffset' | 'isKeyboardDismissDisabled' | 'shouldCloseOnInteractOutside' | 'shouldUpdatePosition'> {
```
yihuiliao, line 33: should we get rid of the mobileType prop for now if we're not showing popovers as modals on mobile?

yihuiliao (approved):

LFDanLu (approved):
Tested locally, verified that Combobox/Picker/ContextualHelp, etc are all rendering as popovers again now and that FullScreenDialog has additional padding when the buttons shift down now

ktabors (approved):
On my iPhone iOS17 I keep seeing it open and immediately close. It's intermittent and I can't figure out reproduction.

Probably an existing issue, but arrow to the right/left direction prop is showing a line between the arrow and the popover. 
image

PR #7401, opened 2024-11-19
fix: DatePicker.test fails with daylight savings change (#7399)

### LINKED ISSUE #7399: DatePicker.test throws error after deleting a defaultValue in DT/ST when current timeZone is ST/DT

The following test fails when the `defaultValue` timeZone's daylight savings/standard time does not match the current timeZone's standard/daylight savings time.

https://github.com/adobe/react-spectrum/blob/f33288c867e305341b63b4ec88e55bc566ae115e/packages/%40react-spectrum/datepicker/test/DatePicker.test.js#L1897-L1932

### 🤔 Expected Behavior?

If the `defaultValue` is on daylight savings time and the user clears the value while in a time zone on standard time, the DatePicker's time zone will change to standard time. I think this is the correct behavior, however, the test is written to always expect "PDT" Pacific Daylight Savings Time, which means it will fail when we're on standard time.

### 😯 Current Behavior

The test is written to always expect "PDT" Pacific Daylight Savings Time, which means it will fail when we're on standard time.

https://github.com/adobe/react-spectrum/blob/f33288c867e305341b63b4ec88e55bc566ae115e/packages/%40react-spectrum/datepicker/test/DatePicker.test.js#L1931

### 💁 Possible Solution

Use the `timeZoneName` for the current date time.

```ts
      let timeZoneName =
        new DateFormatter('en-US',
          {
            timeZone: 'America/Los_Angeles',
            timeZoneName: 'short'
          })
          .formatToParts(new Date())
          .find(p => p.type === 'timeZoneName')
          .value;

      expectPlaceholder(combobox, `mm/dd/yyyy, ––:–– AM ${timeZoneName}`);
```

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

Run `yarn test packages/@react-spectrum/datepicker/test/DatePicker.test.js` from the first Sunday in November to the second Sunday in March.

### Version

@react-spectrum/datepicker@3.10.3

### If other, please specify.

_No response_

MacOS

### 🧢 Your Company/Team

Adobe/Accessibility

### 🕷 Tracking Issue

_No response_

majornista:
If the defaultValue is on daylight savings time and the user clears the value while in a time zone on standard time, the DatePicker's time zone will change to standard time. I think this is the correct behavior, however, the test is written to always expect "PDT" Pacific Daylight Savings Time, which means it will fail when we're on standard time.

Closes #7399 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Run `yarn test packages/@react-spectrum/datepicker/test/DatePicker.test.js` from the first Sunday in November to the second Sunday in March, and all tests should pass.

## 🧢 Your Project:

Adobe/Accessibility

### DIFF
--- a/packages/@react-spectrum/datepicker/test/DatePicker.test.js
+++ b/packages/@react-spectrum/datepicker/test/DatePicker.test.js
@@ -12,7 +12,7 @@
 
 import {act, fireEvent, pointerMap, render as render_, waitFor, within} from '@react-spectrum/test-utils-internal';
 import {Button} from '@react-spectrum/button';
-import {CalendarDate, CalendarDateTime, EthiopicCalendar, getLocalTimeZone, JapaneseCalendar, parseZonedDateTime, toCalendarDateTime, today} from '@internationalized/date';
+import {CalendarDate, CalendarDateTime, DateFormatter, EthiopicCalendar, getLocalTimeZone, JapaneseCalendar, parseZonedDateTime, toCalendarDateTime, today} from '@internationalized/date';
 import {DatePicker} from '../';
 import {Form} from '@react-spectrum/form';
 import {Provider} from '@react-spectrum/provider';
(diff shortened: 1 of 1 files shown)


snowystinger (approved):
Interesting that this will fail even when `jest.setSystemTime` is in use

majornista:
> Interesting that this will fail even when `jest.setSystemTime` is in use
> 
> out of curiosity, how'd you find this one? it's not failing on main right now and I don't quite follow your comment in the issue description for how to see it.

I discovered it when running `yarn test` locally to make sure I wasn't breaking anything with #7403.

PR #7402, opened 2024-11-19
feat: Add RAC example of Menu nested in popover dialog

devongovett:
This documents the RAC change in #7352 to allow menus to be nested inside dialog popovers. It's basically the same account menu example from the S2 storybook recreated with RAC and tailwind and added to our examples page. This way we have something to point to when we are asked how to create menu-like patterns with additional interactive elements.


snowystinger (commented):
Screenshot 2024-11-20 at 11 03 45 am
not working so well in dark mode

snowystinger (approved):

yihuiliao (approved):

Issue #4711, opened 2023-06-26
Incorrect date / time segment order in several components, for RTL languages


lreis:
# 🐛 Bug Report 

The order of the date and time fields is incorrectly displayed by the date/time-related RSP components, for the RTL languages (Arabic, Hebrew).

## 🤔 Expected Behavior

The date and time field order should follow the same order observed with the values returned by **Intl.DateTimeFormat** / **DateFormatter** (@internationalized/date), when rendered in elements with the dir attribute set to"rtl".

The example below (using the options: day: "numeric", month: "numeric", year: "numeric", hour: "numeric", minute: "2-digit", timeZoneName: "short"), illustrates the order for all available fields currently used by all components:

| **ar-AE**  | **he-IL** | 
| ----------- | ----------- |
| timezone dayPeriod hour:minute year/month/day | timezone hour:minute ,day.month.year |  
| ![Screenshot 2023-06-25 at 8 42 50 AM](https://github.com/adobe/react-spectrum/assets/509658/8efb93fc-05f6-4c8b-ac4c-2e16e53c45ab) |  ![Screenshot 2023-06-25 at 8 42 46 AM](https://github.com/adobe/react-spectrum/assets/509658/5e52d2ec-c711-43e1-8b87-b258b1b2089d) |
 

## 😯 Current Behavior
<h3>Calendar</h3> 

_**with zoned time** story_

| **ar-AE**  | **he-IL** | 
| ----------- | ----------- |
| ![Screenshot 2023-06-26 at 10 27 27 AM](https://github.com/adobe/react-spectrum/assets/509658/01dbd1b2-3ade-4262-a54f-46755d9ec7ea)  |  ![Screenshot 2023-06-26 at 10 27 48 AM](https://github.com/adobe/react-spectrum/assets/509658/9a6cb019-6af2-429f-aef0-41820bff9052) |

<h3>DateField</h3> 

_**Date Time Value Zoned** story_

| **ar-AE**  | **he-IL** | 
| ----------- | ----------- |
|  ![Screenshot 2023-06-25 at 9 56 55 AM](https://github.com/adobe/react-spectrum/assets/509658/103f0c3a-8804-4ab7-b2ee-482ee9be2671) |  ![248872859-cb445c86-9268-41b7-a26b-47d16b80be3a](https://github.com/adobe/react-spectrum/assets/509658/b4970d77-7207-42d1-af51-5a59fffdabac) |

<h3>DatePicker</h3>

_**Default** story_

[...]

PR #7423, opened 2024-11-21
fix:  correctly format date/time in RTL

yihuiliao:
Close https://github.com/adobe/react-spectrum/issues/4711

Need to double check that nothing broke in https://github.com/adobe/react-spectrum/issues/6562 with these changes.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Arabic should render in the following order:

TimeZone DayPeriod Hour:Minute Year/Month/Day

Hebrew should render in the following order:

TimeZone DayPeriod Hour:Minute Day/Month/Year

Tab order should be:

Day -> Month -> Year -> Hour -> Minute -> DayPeriod -> TimeZone

Arabic Placeholders:

يوم / شهر / سنة

Hebrew Placeholders:

שנה.חודש.יום

Also be sure to test keyboard navigation


sadeghbarati:
Hey, Thanks for working on this PR

Is this also going to fix keyboard focus management on RTL too?

For example [`DateField`](https://reactspectrum.blob.core.windows.net/reactspectrum/0716ad52594aea1cd29e658eec389d942d6aa863/storybook/index.html?path=/story/date-and-time-datefield--default&providerSwitcher-express=false) component in the storybook link, if you pick `Farsi` locale, in order to focus `DateSegments` with arrow key you need to do it opposite direction to focus next date segment

---


- `DateField` with Farsi locale:
  - yyy/mm/dd, format is right ✅ 
  - focus management is not right ❌ 
  
  
- `DatePicker` with Farsi locale:
   - dd/mm/yyy, format is not right ❌ 
   - focus management is right ✅

yihuiliao:
@sadeghbarati Yes, the plan is to also correct the keyboard navigation for RTL as well. The PR is still in the early stages so there's a lot that has yet to be implemented.

packages/@react-aria/datepicker/src/useDateField.ts
```diff
@@ -181,6 +181,9 @@ export function useDateField<T extends DateValue>(props: AriaDateFieldOptions<T>
         if (props.onKeyUp) {
           props.onKeyUp(e);
         }
+      },
+      style: {
+        unicodeBidi: 'isolate'
```
yihuiliao, line 186: added for DateRangePicker because things were looking a little funky without it. wraps around each datefield. see [codepen](https://codepen.io/yihuiliao/pen/OPLQWvv) for reproduction

packages/react-aria-components/src/Popover.tsx
```diff
@@ -170,6 +173,7 @@ function PopoverInner({state, isExiting, UNSTABLE_portalContainer, ...props}: Po
         ref={ref}
         slot={props.slot || undefined}
         style={style}
+        dir={props.dir}
```
yihuiliao, line 176: popovers are portaled outside which means they don't get the direction of the locale causing the keyboard navigation to behave incorrectly for datepickers. in v3, we wrap the popovers in a provider so this doesn't happy to change the implementation if there's a better way to do this

packages/@react-aria/datepicker/src/useDatePickerGroup.ts
```diff
@@ -9,6 +9,8 @@ import {usePress} from '@react-aria/interactions';
 export function useDatePickerGroup(state: DatePickerState | DateRangePickerState | DateFieldState, ref: RefObject<Element | null>, disableArrowNavigation?: boolean) {
   let {direction} = useLocale();
   let focusManager = useMemo(() => createFocusManager(ref), [ref]);
+  let editableSegments: NodeListOf<Element> | undefined = ref.current?.querySelectorAll('span[role="spinbutton"], span[role="textbox"]');
```
snowystinger, line 12: this needs to be done in a useLayoutEffect and the result should be stored in a ref as other parts of state will determine if the effects needs to update

packages/@react-aria/datepicker/src/useDatePickerGroup.ts
```diff
@@ -104,3 +129,19 @@ export function useDatePickerGroup(state: DatePickerState | DateRangePickerState
...
   return mergeProps(pressProps, {onKeyDown});
 }
+
+function orderSegments(editableSegments: NodeListOf<Element> | undefined) {
+  if (editableSegments) {
+    let segments = Array.from(editableSegments);
+    let segmentArr = segments.map(node => {
+      return {
+        element: node as FocusableElement,
+        rectX: node?.getBoundingClientRect().left
```
snowystinger, line 139: if you run segments.filter(Boolean).map..., you may not need this `?`

might need to do it the line above so you can set the type correctly
```
let segments = Array.from(editableSegments).filter...
```

packages/@react-aria/datepicker/src/useDatePickerGroup.ts
```diff
@@ -31,7 +33,21 @@ export function useDatePickerGroup(state: DatePickerState | DateRangePickerState
         e.preventDefault();
         e.stopPropagation();
         if (direction === 'rtl') {
-          focusManager.focusNext();
+          if (orderedSegments) {
+            let button = ref.current?.querySelector('button');
```
snowystinger, line 37: is it possible someone could add another button to the Group? such as in RAC?
If so, it's be best to find the element by it's `id` if we can get that or some other more unique attribute

packages/@react-spectrum/datepicker/test/DateField.test.js
```diff
@@ -664,4 +664,75 @@ describe('DateField', function () {
       });
     });
   });
+
+  describe('style', () => {
```
snowystinger, line 668: I think all of these tests can be covered by chromatic

these unit tests are likely to break in unhelpful ways, as a change to the styles could result in the correct rendering and behavior, but break here

packages/@react-spectrum/datepicker/test/DatePicker.test.js
```diff
@@ -2334,4 +2334,70 @@ describe('DatePicker', function () {
       });
     });
   });
+
+  describe('style', () => {
```
snowystinger, line 2338: in general, testing styles in a unit test is not helpful since it doesn't actually render, these are all perfect for chromatic though

packages/@react-spectrum/datepicker/test/DatePickerBase.test.js
```diff
@@ -492,6 +462,74 @@ describe('DatePickerBase', function () {
...
     });
+
+    describe('RTL focus management', function () { 
+      beforeEach(() => {
+        jest.spyOn(HTMLElement.prototype, 'getBoundingClientRect').mockImplementation(function () {
+          let x = 0;
+          let el = this;
+  
+          if (el.getAttribute('role') === 'spinbutton') {
+            if (el.parentElement.getAttribute('data-testid') === 'end-date') {
```
snowystinger, line 473: i thought the issue with some of these was that the dom order was not in alignment with the rendered order?
this test is a good base, but should we include some tests where the order differs? I think you could use `el.getAttribute('data-type')` to order based on the day/month/hour/etc

packages/react-aria-components/stories/DatePicker.stories.tsx
```diff
@@ -92,11 +92,11 @@ export const DateRangePickerExample = () => (
     <Label style={{display: 'block'}}>Date</Label>
     <Group style={{display: 'inline-flex'}}>
       <div className={styles.field}>
-        <DateInput data-testid="date-range-picker-date-input" slot="start" style={{display: 'inline-flex'}}>
+        <DateInput data-testid="date-range-picker-date-input" slot="start" style={{display: 'inline'}}>
```
snowystinger, line 95: sort of a breaking change
what will happen to people who have used this as an example to build their DatePicker's in the past?

Also, why is the style change different from DateField.mdx?

yihuiliao: so i experimented a bit and found that for the rtl segments to display in the correct order, DateInput needs to have display inline, inline-block, or block. hence why one is display inline and the other is display block. it would probably be a good idea to make it consistent and stick with display inline though. 

and as for the sort of breaking change, i think folks can continue to use `display: inline-flex`. i don't believe the changes here will affect that. the only thing that they'll run into is that if they do choose to render their datepicker in a rtl locale, it will be in the incorrect order.

packages/@react-aria/datepicker/src/useDatePickerGroup.ts
yihuiliao: yeah you're right. https://github.com/adobe/react-spectrum/issues/7645 is a good case of how someone might add buttons in either through hooks or rac. if that's the case, maybe we should also query select all buttons? i'm assuming that they should be able to keyboard to any type of button so long as it's within the datefield?

packages/@react-spectrum/datepicker/test/DatePickerBase.test.js
yihuiliao: ah yes, it's more obvious with when using examples with time segments so i should update the examples to include those

packages/react-aria-components/stories/DatePicker.stories.tsx
snowystinger: ok, so long as inline-flex doesn't break anything, I'm fine with this
we should add a little blurb to the Accessibility section for the docs

packages/@react-aria/datepicker/src/useDatePickerGroup.ts
```diff
@@ -1,16 +1,48 @@
...
-  let editableSegments: NodeListOf<Element> | undefined = ref.current?.querySelectorAll('span[role="spinbutton"], span[role="textbox"]');
-  let orderedSegments = useMemo(() => orderSegments(editableSegments), [editableSegments]);
+  let segments = useRef<FocusableElement[]>(undefined);
+  useLayoutEffect(() => {
+    if (ref?.current) {
+
+      let update = () => {
+        if (ref.current) {
+          // TODO: For now, just querying this list of elements. However, it's possible that either through hooks or RAC that some users may include other focusable items that they would want to able to keyboard navigate to. In that case, we might want to utilize focusableElements in isFocusable.ts
+          let editableSegments: NodeListOf<Element> | undefined = ref.current?.querySelectorAll('span[role="spinbutton"], span[role="textbox"], button');
```
snowystinger, line 19: because this checks for `span` in all the selectors, will this break people who are using `div`s like we used to?

yihuiliao: oh good point. hmm in that case, we could update it so that it also checks for divs as well? 

although, my guess is that people aren't using this in rtl locales given that the time/date formats were so off.

packages/react-aria-components/stories/DatePicker.stories.tsx
yihuiliao: yeah, that's a good idea to call it out in our docs somewhere. i think maybe the Internationalization section instead?

packages/@react-spectrum/datepicker/test/DateField.test.js
```diff
@@ -167,14 +167,12 @@ describe('DateField', function () {
...
 
     it('should support format help text', function () {
-      let {getByRole, getByText, getAllByRole} = render(<DateField label="Date" showFormatHelpText />);
+      let {getByRole, getAllByRole} = render(<DateField label="Date" showFormatHelpText />);
 
       // Not needed in aria-described by because each segment has a label already, so this would be duplicative.
       let group = getByRole('group');
       expect(group).not.toHaveAttribute('aria-describedby');
 
-      expect(getByText('month / day / year')).toBeVisible();
```
yihuiliao, line 176: i removed this line since now each individual segment is in its own span and it would be hard to query those. we have a chromatic story to test this so it is covered. i kept the rest of the test since it would be good to continue checking the aria stuff.

snowystinger (approved):
Noticed that "Delete" likely needs the same logical movement treatment that the arrow keys got. Otherwise, in Hebrew I start deleting from the year -> month -> day and then it stops, it never moves to the time field element

I'm ok for this to be in follow up, otherwise looks good

yihuiliao:
> Noticed that "Delete" likely needs the same logical movement treatment that the arrow keys got. Otherwise, in Hebrew I start deleting from the year -> month -> day and then it stops, it never moves to the time field element

I think this is actually the correct behavior. "Delete" causes the focus to shift to the previous node. If you're deleting from year -> month -> day and then try deleting again, the focus never moves because there is no node before day. 

It's the same in LTR. In en-US, if you start deleting year -> day -> month, you'll also never move to the timefield. It just feels like you should in RTL because we're reading the entire field as LTR because that's what we're used to.

packages/@react-stately/datepicker/src/useDateFieldState.ts
```diff
@@ -423,6 +406,66 @@ export function useDateFieldState<T extends DateValue = DateValue>(props: DateFi
...
+    let dateSegment = {
+      type: TYPE_MAPPING[segment.type] || segment.type,
+      text: isPlaceholder ? placeholder : segment.value,
+      ...getSegmentLimits(displayValue, segment.type, resolvedOptions),
+      isPlaceholder,
+      placeholder,
+      isEditable
+    } as DateSegment;
+
+    if (segment.type === 'hour') {
```
devongovett, line 431: At some point we should add some comments to this function describing how it works and what the unicode characters do. Will be hard for future readers/contributors to understand what's going on if they aren't familiar with it. 😄

devongovett (approved):

packages/@react-aria/datepicker/src/useDatePickerGroup.ts
```diff
@@ -1,14 +1,48 @@
...
+          });
+      
+          let orderedSegments = segmentsArr.sort((a, b) => a.rectX - b.rectX).map((item => item.element));
+          segments.current = orderedSegments;
+        }
+      };
+
+      update();
+
+      let observer = new MutationObserver(update);
```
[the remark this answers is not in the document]
yihuiliao: hmm good point,,,i think it should be fine if were to get rid of it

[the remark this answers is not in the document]
snowystinger: it was for if users added more buttons to control the field, Yihui had an example with a `<<` and `<` button for jumping month and year I think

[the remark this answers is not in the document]
yihuiliao: Ah right, I think it was from the example in this issue https://github.com/adobe/react-spectrum/issues/7645

Issue #2506, opened 2021-10-28
onPress doesn't trigger on button pseudo-element area


tinleym:
# 🐛 Bug Report 

onPress doesn't trigger on button pseudo-element area, though onClick does.

I've seen this issue discussed for touch devices, but not for desktop/mouse.
https://github.com/adobe/react-spectrum/issues/1061

## 🤔 Expected Behavior

I'd expect onPress to trigger when button pseudo-element is pressed, like onClick.

## 😯 Current Behavior

onPress does not trigger when button pseudo-element area is pressed.

## 🔦 Context

Pseudo-elements can be useful appendages to tappable elements because they allow for a larger tappable area that won't affect the layout of other elements. They can enable you, for example, to enforce that any button or link's minimum clickable dimension be at least 44px (Apple's recommended minimum tap size). 

https://ishadeed.com/article/clickable-area/#using-pseudo-elements-to-increase-the-clickable-area

## 💻 Code Sample

Below, clicking the core of the button triggers onPress and onClick, but clicking in the pseudo-element area (inside the frame but outside the core of the button) only triggers onClick.

https://codesandbox.io/s/intelligent-cdn-t6iu9?file=/src/App.js

## 🌍 Your Environment

react-aria 3.10.0
Chrome 94.0.4606.81 (Official Build) (x86_64), Firefox 93.0, Safari 14.1.1
MacOS 11.4

LFDanLu:
This sounds like a problem with how we do our own hit testing in `usePress` but eventually we'd like to remove that hit testing entirely from `usePress` in the future (https://github.com/adobe/react-spectrum/issues/1720) which may rectify the issue. For now, perhaps we could use `document.elementFromPoint` with the press coordinates, that seems to return the button element when provided x,y coords of a click within the pseudo element area. I'd be curious what would be returned if there were overlapping pseudo elements

tinleym:
@LFDanLu yeah, it looks like it would...
https://codesandbox.io/s/modest-allen-6vxo0?file=/src/App.js

Georgegriff:
Is this issue still active? it looks like this is still an issue

universse:
I think this applies to not only pseudo element but also any absolutely positioned element that expand the tap area. Hope this gets addressed soon.

timoyan:
Event onPressUp is not working on pseudo element, either.

Do a quick review on source code and guess it does not get into this code execution

```
[...]

PR #7427, opened 2024-11-22
refactor: Remove custom hit testing in usePress

devongovett:
Related to #1720, #2506, #5290

Back in #62 we implemented our own hit testing in usePress in order to work around a bug in pointer events in iOS 13. Now that this version is long gone, we can probably remove this and use the `onPointerLeave` and `onPointerEnter` events instead. This depends on [releasePointerCapture](https://developer.mozilla.org/en-US/docs/Web/API/Element/releasePointerCapture) being called during `onPointerDown` on touch devices, so that the original target doesn't capture the pointer and cause the browser to retarget all events to it even outside the hit area.


packages/@react-aria/interactions/src/usePress.ts
```diff
@@ -433,7 +433,13 @@ export function usePress(props: PressHookProps): PressResult {
 
           shouldStopPropagation = triggerPressStart(e, state.pointerType);
 
-          addGlobalListener(getOwnerDocument(e.currentTarget), 'pointermove', onPointerMove, false);
+          // Release pointer capture so that touch interactions can leave the original target.
+          // This enables onPointerLeave and onPointerEnter to fire.
+          let target = e.target as Element;
+          if ('releasePointerCapture' in target) {
```
devongovett, line 439: not supported in jsdom

packages/@react-aria/interactions/src/usePress.ts
```diff
@@ -467,27 +473,20 @@ export function usePress(props: PressHookProps): PressResult {
         }
 
         // Only handle left clicks
-        // Safari on iOS sometimes fires pointerup events, even
-        // when the touch isn't over the target, so double check.
-        if (e.button === 0 && isOverTarget(e, e.currentTarget)) {
+        if (e.button === 0) {
```
devongovett, line 476: I am not certain yet if this bug still exists in Safari. This was added in #335. Will need to test.

snowystinger (approved):
Excited to get this into testing!

devongovett:
Updated to address two issues found in testing:

> On touch (ipad), open any sub menu and press down on a submenu item, without lifting, slide off the submenu and release. Menus all close now. The close behavior exists already with mouse.

This only happens in RSP v3, not in RAC, because we recently changed the DOM structure to no longer be nested there. In RSP we have an extra `useInteractOutside` in Menu, but it only checks if the interaction was outside the menu itself not any submenus. It checks that during `onInteractOutside` but not in the pointer down event. I moved this to the root menu trigger instead so submenus should be included.

> On ipad, starting a touch event on one item and then dragging and releasing of a different item selects that item. Doesn't happen on Android and doesn't happen if the dropdown is scrollable. Didn't happen on main previously, same issue as row 7. Also occurs for Menu (DL)

This works as intended with mouse (matching native behavior), but it is weird with touch since the user might be trying to scroll. Updated this logic to only select during `onPressUp` if `pointerType` is mouse, otherwise wait for `onPress` which requires pressing down on the same target your release over.

snowystinger (approved):
submenus appear to be working correctly in s1 and s2

LFDanLu (approved):
LGTM, verified the menu/pickers/etc don't auto close on touch when attempting to scroll. Also verified submenus don't close when dragging off of them in touch anymore

PR #7435, opened 2024-11-26
chore: Improve React Fast Refresh in storybook

devongovett:
I moved the Parcel plugins we use for storybook into an official Parcel repo: https://github.com/parcel-bundler/storybook. Updated them to handle React Fast Refresh much better than before. Now you can edit a component or a story and it'll hot update without losing state. In addition, updating the prop types in TypeScript will also hot reload without needing to restart Storybook. It also includes the original story source automatically instead of needing to manually copy and paste it into a string.

Fast Refresh also requires us to change the syntax we use for `forwardRef`, since it requires that component names must start with an upper case letter and not an underscore. I wrote a codemod to convert this syntax since it is so pervasive, and it now uses an inline style rather than declaring a separate variable. That is the majority of the diff.


snowystinger (approved):

snowystinger (commented):
Ran it locally, I can't get to a bunch of the s1 stories (ie.  ActionBar/ActionGroup/Breadcrumbs/Button/ButtonGroup/ColorField/ColorSlider).

`ActionBarStory is not defined`

Issue #6618, opened 2024-06-25
Unable to drag and drop links


plbstl:
### Provide a general summary of the issue here

When using the `Link` component or `useLink` hook, I am unable to drag and drop rendered links. 

### 🤔 Expected Behavior?

Links should be able to be dragged around and dropped even outside of the browser application, as that is the default behavior in browsers.

### 😯 Current Behavior

Links cannot be dragged.

### 🖥️ Steps to Reproduce

1. Go to https://react-spectrum.adobe.com/react-spectrum/Link.html#example
2. Attempt to drag the example link with a mouse or finger

### Version

3.0.0

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### What operating system are you using?

macOS 14.5

reidbarber:
Related discussion: https://github.com/adobe/react-spectrum/discussions/6010

PR #7448, opened 2024-11-28
fix: make `a` tag link draggable with usePress

mehm8128:
Closes https://github.com/adobe/react-spectrum/issues/6618

I'm not sure this solution is proper to solve the issue so if there is some missing consideration please let me know 🙏 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. open storybook Link Example http://localhost:9003/?path=/story/react-aria-components--link-example&providerSwitcher-express=false&strict=true
2. Attempt to drag the link with a mouse

### DIFF
--- a/packages/@react-aria/interactions/src/usePress.ts
+++ b/packages/@react-aria/interactions/src/usePress.ts
@@ -787,7 +787,7 @@ export function usePress(props: PressHookProps): PressResult {
   ]);
 
   // Remove user-select: none in case component unmounts immediately after pressStart
-   
+
   useEffect(() => {
     return () => {
       if (!allowTextSelectionOnPress) {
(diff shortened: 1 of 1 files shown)


packages/@react-aria/interactions/src/usePress.ts
```diff
@@ -935,7 +935,15 @@ function isOverTarget(point: EventPoint, target: Element) {
...
 function shouldPreventDefaultDown(target: Element) {
   // We cannot prevent default if the target is a draggable element.
-  return !(target instanceof HTMLElement) || !target.hasAttribute('draggable');
+  if (!(target instanceof HTMLElement)) {
+    return true;
+  }
+
+  if (target.tagName === 'A' && target.getAttribute('draggable') !== 'false') {
+    return false;
+  }
```
mehm8128, line 944: if target is `a` tag and doesn't have `draggable="false"` attr, we shouldn't prevent default to enable to drag the link

packages/@react-aria/interactions/src/usePress.ts
```diff
@@ -935,7 +935,15 @@ function isOverTarget(point: EventPoint, target: Element) {
...
+
+  return !target.hasAttribute('draggable');
```
mehm8128, line 946: Q: I think even if target element has `draggable="false"`, `target.hasAttribute('draggable')` is `true` (because `hasAttribute` ignore its value), is it ok? or should we fix it as `!(target.getsAttribute('draggable') === "true");`

PR #7464, opened 2024-12-04
feat: Add staticColor="auto" variant

devongovett:
This adds a new `staticColor="auto"` variant, alongside the existing white and black options. The auto variant automatically swaps between white and black based on the background color of the container it is placed within. The container must have the `--s2-container-bg` variable set for this to work, which we already set on the page, and in components like cards, modals, and popovers.

In the style macro theme, there is a new `transparent-overlay` series of colors at varying levels of opacity, which matches the `transparent-black` and `transparent-white` series but with a calc to automatically swap between them.

The `color` property also supports an `auto` option that does the same thing, but also includes `self(backgroundColor)` in addition to `var(--s2-container-bg)` so that it is based on the element's own background in addition to the ancestor container. We could potentially include `auto` as an alias for `transparent-overlay-1000` across all properties, but it would not make sense to include the element's own background color for most CSS properties other than `color` I think.

### DIFF
--- a/packages/@react-spectrum/s2/src/bar-utils.ts
+++ b/packages/@react-spectrum/s2/src/bar-utils.ts
@@ -13,6 +13,7 @@
 import {centerPadding, fieldInput, staticColor} from './style-utils' with {type: 'macro'};
 
 export const bar = () => ({
+  ...staticColor(),
   position: 'relative',
   display: 'grid',
   gridTemplateColumns: {
@@ -64,7 +65,7 @@ export const track = () => ({
   borderRadius: 'full',
   backgroundColor: {
     default: 'gray-300',
-    staticColor: staticColor('transparent-overlay-300'),
+    isStaticColor: 'transparent-overlay-300',
     forcedColors: 'ButtonFace'
   },
   outlineWidth: {
--- a/packages/@react-spectrum/s2/src/bar-utils.ts
+++ b/packages/@react-spectrum/s2/src/bar-utils.ts
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
-import {centerPadding, fieldInput} from './style-utils' with {type: 'macro'};
+import {centerPadding, fieldInput, staticColor} from './style-utils' with {type: 'macro'};
 
 export const bar = () => ({
   position: 'relative',
@@ -64,15 +64,7 @@ export const track = () => ({
   borderRadius: 'full',
   backgroundColor: {
     default: 'gray-300',
-    staticColor: {
-      white: {
-        default: 'transparent-white-100'
-      },
-      // TODO: Is there a black static color in S2?
-      black: {
-        default: 'transparent-black-400'
-      }
-    },
+    staticColor: staticColor('transparent-overlay-300'),
     forcedColors: 'ButtonFace'
   },
   outlineWidth: {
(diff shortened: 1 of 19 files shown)


packages/@react-spectrum/s2/src/style-utils.ts
```diff
@@ -83,14 +83,7 @@ export const fieldLabel = () => ({
...
     isDisabled: 'disabled',
-    staticColor: {
-      white: {
-        default: 'transparent-white-700'
-      },
-      black: {
-        default: 'transparent-black-900'
-      }
-    },
+    staticColor: staticColor('transparent-overlay-1000'),
```
devongovett, line 86: Design updated the static color labels of progress bar and meter to be fully opaque

snowystinger (commented):
This is so cool!!!!
Lets run chromatic? don't forget to run it on main first

reidbarber (approved):
This is awesome!

packages/@react-spectrum/s2/src/bar-utils.ts
```diff
@@ -64,15 +65,7 @@ export const track = () => ({
...
-    staticColor: {
-      white: {
-        default: 'transparent-white-100'
-      },
-      // TODO: Is there a black static color in S2?
-      black: {
-        default: 'transparent-black-400'
-      }
-    },
+    isStaticColor: 'transparent-overlay-300',
```
LFDanLu, line 68: just to verify, this value is essentially the transparent white 300 that the figma token references for the progress bar right? And we'd ditched the black color token because it doesn't have an actual design token definition?

packages/@react-spectrum/s2/stories/utils.tsx
```diff
@@ -23,8 +24,38 @@ function getBackgroundColor(staticColor: StaticColor) {
...
-  return <div style={{padding: 8, background: getBackgroundColor(props.staticColor)}}>{props.children}</div>;
+  let [autoBg, setAutoBg] = useState('#5131c4');
+  return (
+    <>
+      <div
+        style={{
+          padding: 8,
+          // @ts-ignore
+          '--s2-container-bg': props.staticColor === 'auto' ? autoBg : undefined,
+          background: props.staticColor === 'auto' ? autoBg : getBackgroundColor(props.staticColor)
```
LFDanLu, line 37: Will we be documenting setting `--s2-container-bg` in the future (aka if a user has a custom container and wants custom background + `staticColor: auto` support) or are S2 users expected to mainly use the preset ones in Provider aka background layer color presets?

LFDanLu (approved):
tested the components with staticColor="auto" and things look good to me. I support this was pulled out of your actionBar PR (was looking to test the divider changes)? Otherwise just a couple of clarifying comments

packages/@react-spectrum/s2/src/bar-utils.ts
devongovett: it's either transparent-white-400 or transparent-black-400 depending on the background color. According to the tokens figma file they updated it to be consistent.

packages/@react-spectrum/s2/stories/utils.tsx
devongovett: Yeah we can document that. I was also thinking maybe we'd create a `container` macro for people to use that could automatically set it.

LFDanLu:
not sure why chromatic is failing, digging

LFDanLu:
looks like the chromatic story that was erroring out is actually fixed in main

PR #7475, opened 2024-12-04
feat: RAC Autocomplete audit for release

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP


packages/react-aria-components/src/index.ts
```diff
@@ -16,7 +16,7 @@ import 'client-only';
 
 export {CheckboxContext, ColorAreaContext, ColorFieldContext, ColorSliderContext, ColorWheelContext, HeadingContext} from './RSPContexts';
 
-export {Autocomplete, AutocompleteContext, AutocompleteStateContext, InternalAutocompleteContext} from './Autocomplete';
+export {UNSTABLE_Autocomplete, UNSTABLE_AutocompleteContext, UNSTABLE_AutocompleteStateContext, UNSTABLE_InternalAutocompleteContext} from './Autocomplete';
```
devongovett, line 19: If it's named **Internal**AutocompleteContext, is it meant to be exported? I think eventually we'd want to export it but probably with a different name in that case.

LFDanLu: I'll remove it from the exports from now and add it to the followup ticket. Perhaps a name like `AutocompleteCollectionContext` or something might appropriate

packages/@react-aria/autocomplete/src/useAutocomplete.ts
```diff
@@ -130,20 +130,18 @@ export function useAutocomplete(props: AriaAutocompleteOptions, state: Autocompl
...
-  useEffect(() => {
-    if (state.inputValue != null) {
-      if (lastInputValue.current != null && lastInputValue.current !== state.inputValue && lastInputValue.current?.length <= state.inputValue.length) {
-        focusFirstItem();
-      } else {
-        clearVirtualFocus();
-      }
-
-      lastInputValue.current = state.inputValue;
+  // TODO: update to see if we can tell what kind of event (paste vs backspace vs typing) is happening instead
```
LFDanLu, line 133: To be handled in follow up

packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -391,6 +391,10 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
     }
   };
 
+  // Ref to track whether the first item in the collection should be automatically focused. Specifically used for autocomplete when user types
+  // to focus the first key AFTER the collection updates.
+  // TODO: potentially expand the usage of this
+  let shouldVirtualFocusFirst = useRef(false);
```
LFDanLu, line 397: A flag has been added to let the collection know if it should focus the first item in the list after various user interactions. This may have other potential use cases (perhaps fixes the delay rendering RAC menu only until the collection is populated) but to be determined later, for now keeping changes scoped to Autocomplete

packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -401,21 +405,52 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
...
+
+    lastCollection.current = manager.collection;
+  }, [manager.collection, updateActiveDescendant]);
+
+  let resetFocusFirstFlag = useEffectEvent(() => {
+    // If user causes the focused key to change in any other way, clear shouldVirtualFocusFirst so we don't
+    // accidentally move focus from under them. Skip this if the collection was empty because we might be in a load
+    // state and will still want to focus the first item after load
+    if (manager.collection.size > 0) {
+      shouldVirtualFocusFirst.current = false;
```
LFDanLu, line 446: This is a bit interesting since there are cases where the collection is temporarily empty due to a loading state after a user types in the field. We still want focus the first item after the load so we don't want to reset the flag. I believe the check against size should be correct since loading state nodes don't count towards the size of the collection  (at least for table/menu/etc) but will need to make sure that remains true moving forward

packages/react-aria-components/stories/Autocomplete.stories.tsx
```diff
@@ -235,7 +249,10 @@ export const AutocompleteAsyncLoadingExample = {
   render: (args) => {
     return <AsyncExample {...args} />;
   },
-  name: 'Autocomplete, useAsync level filtering'
+  name: 'Autocomplete, useAsync level filtering with load state',
+  args: {
+    includeLoadState: true
+  }
```
LFDanLu, line 255: When testing the async story, note this control. This will toggle the existance of a intermediate loading state as the list is filtered

packages/react-aria-components/test/Autocomplete.test.tsx
```diff
@@ -98,29 +100,77 @@ let ListBoxWithSections = (props) => (
...
+        <Label style={{display: 'block'}}>Test</Label>
+        <Input />
+        <Text style={{display: 'block'}} slot="description">Please select an option below.</Text>
+      </SearchField>
+      <Menu
+        items={list.items}
+        onAction={onAction}
+        onSelectionChange={onSelectionChange}>
+        {item => <MenuItem id={item.id}>{item.name}</MenuItem>}
+      </Menu>
```
LFDanLu, line 172: We may want to look into giving Menu/Listbox a loading indicator component like Table has. Menu itself doesn't even have the notion of an empty state

packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -401,21 +405,52 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
...
     }
   });
 
+  useEffect(() => {
+    resetFocusFirstFlag();
```
devongovett, line 451: Won't this also run on the initial render too? We might want to use `useUpdateEffect` here?

packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -401,21 +405,52 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
...
       manager.setFocusedKey(keyToFocus);
+      // Only set shouldVirtualFocusFirst to false if we've successfully set the first key as the focused key
+      // If there wasn't a key to focus, we might be in a temporary loading state so we'll want to still focus the first key
+      // after the collection updates after load
+      shouldVirtualFocusFirst.current = false;
+    }
+  });
+
+  let lastCollection = useRef(manager.collection);
+  useEffect(() => {
```
devongovett, line 433: You might be able to use `useUpdateEffect` here as well as it will do the comparison of collection for you (so you don't need to store it in a ref yourself).

packages/react-aria-components/test/Autocomplete.test.tsx
devongovett: yeah idk if the loader should be on the menu or inside the searchfield like ComboBox/SearchAutocomplete. We might need to talk to design about that one

packages/@react-aria/selection/src/useSelectableCollection.ts
LFDanLu: ah thanks, updated. I forgot all about that haha

packages/react-aria-components/test/Autocomplete.test.tsx
LFDanLu: I'll make a note to ask them

devongovett (approved):

devongovett:
totally tiny nit I just noticed when testing, can be fixed later: there seems like there is a slight delay sometimes between when the filtered list updates and when the first item is focused (can be seen in the virtualized list most easily). we may need a `useUpdateLayoutEffect` or something, not sure. Maybe I am seeing things 😂

LFDanLu:
@devongovett I'm not quite sure I'm seeing the delay you are mentioning 😅. I tried the virtualized story and a couple of the other ones, was there a particular sequence of interactions/browsers/etc?

romansndlr:
Can't wait for this to be released! Y'all are amazing! 🙌🏻

reidbarber (dismissed):

packages/@react-aria/autocomplete/src/useAutocomplete.ts
```diff
@@ -152,11 +157,21 @@ export function useAutocomplete(props: AriaAutocompleteOptions, state: Autocompl
...
         // Early return for Escape here so it doesn't leak the Escape event from the simulated collection event below and
         // close the dialog prematurely. Ideally that should be up to the discretion of the input element hence the check
         // for isPropagationStopped
+        // Also set the inputValue to '' to cover Firefox case where Esc doesn't actually clear searchfields. Normally we already
+        // handle this in useSearchField, but we are directly setting the inputValue on the input element in RAC Autocomplete instead of
+        // passing it to the SearchField via props. This means that a controlled value set on the Autocomplete isn't synced up with the
+        // SearchField until the user makes a change to the field's value via typing
         if (e.isPropagationStopped()) {
+          state.setInputValue('');
           return;
```
LFDanLu, line 175: As noted in the comment, we currently apply the tracked `inputValue` directly on to the Autocomplete wrapped input element rather than providing it via props to TextField/SearchField so that we can just consume the `inputValue` from the input context once rather than needing to add it to every input field component in RAC. This means that the input value tracked here in `useAutocomplete` isn't synced with the input value tracked in `useSearchField` until the user types into the field to cause the chained `onChange` to fire. Since Firefox doesn't also clear the native search input fields on `Esc` we won't get a `onChange` fired that clears the field so we'll need to do this here. Kinda gross, maybe reason enough to pass `value` to the input components via props (aka via TextFieldContext/SearchFieldContext) instead of applying it directly to the input element (aka via InputContext like we are doing right now)?

packages/@react-aria/searchfield/src/useSearchField.ts
```diff
@@ -73,7 +73,9 @@ export function useSearchField(
     }
 
     if (key === 'Escape') {
-      if (state.value === '') {
+      // Also check the inputRef value for the case where the value was set directly on the input element instead of going through
+      // the hook
+      if (state.value === '' && (!inputRef.current || inputRef.current.value === '')) {
```
LFDanLu, line 78: As mentioned above, currently there can be a situation where the value of the input field isn't passed in via props but is instead applied directly onto the input element and thus `state.value` here could be `''` but the value applied on the input is `"blah"` or something and thus we'd erroneously would continuePropagation. Check the ref's value as a fallback

packages/@react-aria/selection/src/useSelectableItem.ts
```diff
@@ -160,6 +160,9 @@ export function useSelectableItem(options: SelectableItemOptions): SelectableIte
   };
 
   // Focus the associated DOM node when this item becomes the focusedKey
+  // TODO: can't make this useLayoutEffect bacause it breaks menus inside dialogs
+  // However, if this is a useEffect, it runs twice and dispatches two UPDATE_ACTIVEDESCENDANT and immediately sets
+  // aria-activeDescendant in useAutocomplete... I've worked around this for now
   useEffect(() => {
```
LFDanLu, line 166: Bit of an interesting thing I discovered. If this is a useLayoutEffect then it won't run twice when auto focusing the first item of a newly filtered collection presumably because it runs in the same phase as the `useUpdateLayoutEffect` in useSelectableCollection. However, this breaks menus in dialogs where keyboard opening the menu causes the dialog itself to close...

packages/@react-aria/utils/src/useUpdateLayoutEffect.ts
```diff
@@ -0,0 +1,37 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import {EffectCallback, useRef} from 'react';
+import {useLayoutEffect} from './useLayoutEffect';
+
+// Like useLayoutEffect, but only called for updates after the initial render.
+export function useUpdateLayoutEffect(effect: EffectCallback, dependencies: any[]) {
```
LFDanLu, line 17: Added to get rid of some flickering when auto focusing the first element in the list

romansndlr:
Hey friends 👋🏻

Any rough estimation on when this feature might get released? 

Thanks again for all your amazing work!

LFDanLu:
@romansndlr Hopefully within this month with the next release!

reidbarber (approved):
LGTM. One behavior we may want to consider revisiting is when you press Left/Right arrow while an item is focused and the input is empty. Focus goes back up to the input, but we might want to keep it on the item in this case. Not a huge deal, just something to consider.

packages/@react-aria/autocomplete/src/useAutocomplete.ts
devongovett: Does this mean it now happens for textfields in addition to search field? Might be unexpected?

> maybe reason enough to pass value to the input components via props (aka via TextFieldContext/SearchFieldContext) instead of applying it directly to the input element (aka via InputContext like we are doing right now)?

yeah maybe. I guess there might be more too - anything that uses `<Input>`? or would we limit it to those two?

devongovett (approved):

packages/@react-aria/autocomplete/src/useAutocomplete.ts
LFDanLu: > Does this mean it now happens for textfields in addition to search field? Might be unexpected?

The value is still only cleared for search field for now because we only trigger this if the event coming from the input field has `isPropagationStopped`. 

> yeah maybe. I guess there might be more too - anything that uses <Input>? or would we limit it to those two?

Yep, it would need to be anything that uses `Input` that we could reasonably expect to be used within the Autocomplete which is a bit annoying since we'd need to remember to update if for any new components in the future. For now I'm leaning towards keeping it as is for now unless we find any edge cases/bugs, but will file something to track this

PR #7491, opened 2024-12-06
fix(S2): ContextualHelp in Field wrapping to new line

reidbarber:
In small width fields, the ContextualHelp would wrap to a new line if the widest checkbox item was smaller than the Label + ContextualHelp. 

Before:

Contextual help wrapping to new line

After:

Contextual help on same line as last line for label

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/s2/src/Field.tsx
+++ b/packages/@react-spectrum/s2/src/Field.tsx
@@ -76,7 +76,6 @@ export const FieldLabel = forwardRef(function FieldLabel(props: FieldLabelProps,
       className={style({
         gridArea: 'label',
         display: 'inline',
-        whiteSpace: 'nowrap',
         textAlign: {
           labelAlign: {
             start: 'start',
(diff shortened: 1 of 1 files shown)


packages/@react-spectrum/s2/src/Field.tsx
```diff
@@ -76,6 +76,7 @@ export const FieldLabel = forwardRef(function FieldLabel(props: FieldLabelProps,
       className={style({
         gridArea: 'label',
         display: 'inline',
+        whiteSpace: 'nowrap',
```
devongovett, line 79: Doesn't this break the long label example in the form story? Looks like it overflows now. I think we do want the label to wrap, just not right before the contextual help. We could potentially use the same solution we used for the necessity indicator, where we wrapped the contextual help in a `<span>` with `white-space: nowrap` and put a `&nbsp;` before it (instead of the margin). That should allow the label to wrap, but prevent the contextual help from being on a line by itself.

reidbarber: Makes sense, updated

LFDanLu (dismissed):
Tested the contextual help in the stories with labels of various length, confirmed that the contextual help icon no longer appears on a line by itself.

packages/@react-spectrum/s2/src/Field.tsx
```diff
@@ -127,21 +127,24 @@ export const FieldLabel = forwardRef(function FieldLabel(props: FieldLabelProps,
...
-            {contextualHelp}
-          </ContextualHelpContext.Provider>
-        </CenterBaseline>
+        <span className={style({whiteSpace: 'nowrap'})}>
+          &nbsp;
+          <CenterBaseline
+            styles={style({
+              display: 'inline-flex',
+              height: 0,
+              marginStart: 4
```
ktabors, line 136: Do we still want this margin since we have the `&nbsp;` now? Seems like too much space.
image
image

ktabors (commented):
I think we're okay, but the second line can become longer than the first.
image

SearchField's and TagGroup's ContextualHelp seem to be one pixel off when compared to Slider's ContextualHelp.

packages/@react-spectrum/s2/src/Field.tsx
devongovett: I think we should remove the margin

reidbarber: @ktabors Good catch! updated

yihuiliao:
I feel like there should be like 1px or 2px of space because this is what it looks like now. I agree that the margin before was probably too much before though with the addition of `&nbsp;`
Screenshot 2024-12-17 at 2 45 48 PM

For reference, this is what it looks like currently on main. 
Screenshot 2024-12-17 at 2 45 53 PM

snowystinger:
Could possibly use a zero width character instead and keep the margin https://stackoverflow.com/questions/11392312/zero-width-non-breaking-space otherwise it'll change with fontsize

devongovett:
I think it _should_ change with font size. Design doesn't have contextual help yet but the spacing with the asterisk does change so makes sense for this to as well.

devongovett (approved):

snowystinger (approved):

PR #7514, opened 2024-12-12
feat: add gradient variants to Button

jluyau:
Closes 

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/s2/src/Button.tsx
+++ b/packages/@react-spectrum/s2/src/Button.tsx
@@ -267,7 +267,7 @@ const button = style<ButtonRenderProps & ButtonStyleProps>({
         fillStyle: {
           fill: {
             variant: {
-              primary: 'wghite',
+              primary: 'white',
               secondary: baseColor('transparent-white-800'),
               premium: 'white',
               genai: 'white'
--- a/packages/@react-spectrum/s2/src/Button.tsx
+++ b/packages/@react-spectrum/s2/src/Button.tsx
@@ -185,20 +185,22 @@ const button = style<ButtonRenderProps & ButtonStyleProps & {isStaticColor: bool
         fill: {
           variant: {
             primary: baseColor('transparent-overlay-800'),
-            secondary: baseColor('transparent-overlay-100')
+            secondary: baseColor('transparent-overlay-100'),
+            premium: 'gray-800',
+            genai: 'gray-800'
           },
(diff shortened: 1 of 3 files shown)


packages/@react-spectrum/s2/src/Button.tsx
```diff
@@ -161,13 +178,18 @@ const button = style<ButtonRenderProps & ButtonStyleProps>({
         isDisabled: 'transparent'
       }
     },
+    variant: {
```
snowystinger, line 181: the variant should probably be moved up into the fillStyle with all the other variants

snowystinger: disabled styling is not working
Screenshot 2024-12-18 at 8 40 03 am
because variant is winning over the fillStyle

snowystinger (commented):
Looks like the default button styles were affected
Before
Screenshot 2024-12-18 at 8 37 48 am

After
Screenshot 2024-12-18 at 8 38 00 am

packages/@react-spectrum/s2/src/Button.tsx
```diff
@@ -142,6 +146,19 @@ const button = style<ButtonRenderProps & ButtonStyleProps>({
...
   },
+  backgroundImage: {
+    variant: {
+      premium: {
+        default: linearGradient('96deg', ['fuchsia-900', 0], ['indigo-900', 66], ['blue-900', 100]),
+        isHovered: linearGradient('96deg', ['fuchsia-1000', 0], ['indigo-1000', 66], ['blue-1000', 100])
+      },
+      genai: {
+        default: linearGradient('96deg', ['red-900', 0], ['magenta-900', 33], ['indigo-900', 100]),
+        isHovered: linearGradient('96deg', ['red-1000', 0], ['magenta-1000', 33], ['indigo-1000', 100])
```
devongovett, line 157: This should probably also apply in the `isFocusVisible` and `isPressed` states.

snowystinger (approved):
I fixed up HCM and disabled, please test them out some more in case I missed anything

reidbarber (approved):
Looks great!

devongovett (approved):
We should probably find a way to make the gradients transition on hover. The other variants smoothly fade between background colors whereas these jump on hover. I think it's possible now with `@property`. We can do this later. https://dev.to/afif/we-can-finally-animate-css-gradient-kdk

PR #7523, opened 2024-12-16
fix: prod docs

devongovett:
Two issues in verdaccio docs build on main:

* Due to an update in some very popular npm packages that included invalid values in their package.json (`"main": false`), parcel was no longer parsing the package and failing to resolve it. This is addressed in Parcel v2.13.3.
* Due to React 19 release, yarn started installing multiple copies of React. That is because some packages depend on "^18 || ^19" but the root only depends on "^18". Yarn tries to resolve to the highest possible version for each package. This is solved by making most dependencies peerDependencies instead so only the root one will be installed.

The root issue is that we don't have a lock file in the prod build, which we should definitely fix at some point.


packages/dev/codemods/package.json
```diff
@@ -32,14 +32,12 @@
...
     "execa": "^5.1.1",
     "jscodeshift": "^0.15.2",
-    "react": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0 || ^19.0.0",
     "recast": "^0.23.9",
     "ts-node": "^10.9.2",
     "uuid": "^9.0.1"
   },
   "devDependencies": {
     "@types/jscodeshift": "^0.11.11",
-    "react-dom": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0 || ^19.0.0",
```
devongovett, line 42: These appear to be unused. React is only used for types.

reidbarber (approved):

LFDanLu (approved):

PR #7542, opened 2024-12-20
refactor: No longer preventDefault in usePress and allow browser to manage focus

devongovett:
Related: #1720 

This is a major refactor to `usePress` to improve compatibility with other libraries, improve the way focus is managed, and fix many issues caused by our current approach. 🎄 

This is mainly possible because of a change in Safari 17, which enabled buttons and other native inputs to be focused (both via mouse/touch and keyboard tabbing) only when they have an explicit `tabIndex` set. https://github.com/WebKit/WebKit/pull/12743 We previously worked around this by calling `preventDefault` and manually managing focus. However, because `preventDefault` is not granular this had many unintended side effects that we needed to work around.

...

### DIFF
--- a/packages/@react-aria/focus/src/FocusScope.tsx
+++ b/packages/@react-aria/focus/src/FocusScope.tsx
@@ -12,7 +12,7 @@
 
 import {FocusableElement, RefObject} from '@react-types/shared';
 import {focusSafely} from './focusSafely';
-import {getOwnerDocument, useLayoutEffect} from '@react-aria/utils';
+import {getOwnerDocument, isFocusable, isTabbable, useLayoutEffect} from '@react-aria/utils';
 import {isElementVisible} from './isElementVisible';
 import React, {ReactNode, useContext, useEffect, useMemo, useRef} from 'react';
 
@@ -265,31 +265,6 @@ function createFocusManagerForScope(scopeRef: React.RefObject<Element[] | null>)
       return previousNode;
     }
   };
-}
-
-const focusableElements = [
-  'input:not([disabled]):not([type=hidden])',
-  'select:not([disabled])',
-  'textarea:not([disabled])',
-  'button:not([disabled])',
-  'a[href]',
-  'area[href]',
-  'summary',
-  'iframe',
-  'object',
-  'embed',
-  'audio[controls]',
-  'video[controls]',
-  '[contenteditable]:not([contenteditable^="false"])'
-];
-
-const FOCUSABLE_ELEMENT_SELECTOR = focusableElements.join(':not([hidden]),') + ',[tabindex]:not([disabled]):not([hidden])';
-
-focusableElements.push('[tabindex]:not([tabindex="-1"]):not([disabled])');
-const TABBABLE_ELEMENT_SELECTOR = focusableElements.join(':not([hidden]):not([tabindex="-1"]),');
-
-export function isFocusable(element: HTMLElement) {
-  return element.matches(FOCUSABLE_ELEMENT_SELECTOR);
 }
 
 function getScopeRoot(scope: Element[]) {
@@ -742,7 +717,7 @@ function restoreFocusToElement(node: FocusableElement) {
  * that matches all focusable/tabbable elements.
  */
 export function getFocusableTreeWalker(root: Element, opts?: FocusManagerOptions, scope?: Element[]) {
-  let selector = opts?.tabbable ? TABBABLE_ELEMENT_SELECTOR : FOCUSABLE_ELEMENT_SELECTOR;
+  let filter = opts?.tabbable ? isTabbable : isFocusable;
   let walker = getOwnerDocument(root).createTreeWalker(
     root,
     NodeFilter.SHOW_ELEMENT,
@@ -753,7 +728,7 @@ export function getFocusableTreeWalker(root: Element, opts?: FocusManagerOptions
           return NodeFilter.FILTER_REJECT;
         }
 
-        if ((node as Element).matches(selector)
+        if (filter(node as Element)
           && isElementVisible(node as Element)
           && (!scope || isElementInScope(node as Element, scope))
           && (!opts?.accept || opts.accept(node as Element))
--- a/packages/@react-aria/grid/src/useGridSelectionAnnouncement.ts
+++ b/packages/@react-aria/grid/src/useGridSelectionAnnouncement.ts
@@ -15,9 +15,9 @@ import {Collection, Key, Node, Selection} from '@react-types/shared';
 // @ts-ignore
 import intlMessages from '../intl/*.json';
 import {SelectionManager} from '@react-stately/selection';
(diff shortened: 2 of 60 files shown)


packages/@react-aria/focus/src/FocusScope.tsx
```diff
@@ -267,31 +267,6 @@ function createFocusManagerForScope(scopeRef: React.RefObject<Element[] | null>)
...
-
-const FOCUSABLE_ELEMENT_SELECTOR = focusableElements.join(':not([hidden]),') + ',[tabindex]:not([disabled]):not([hidden])';
-
-focusableElements.push('[tabindex]:not([tabindex="-1"]):not([disabled])');
-const TABBABLE_ELEMENT_SELECTOR = focusableElements.join(':not([hidden]):not([tabindex="-1"]),');
-
-export function isFocusable(element: HTMLElement) {
-  return element.matches(FOCUSABLE_ELEMENT_SELECTOR);
-}
-
```
devongovett, line 294: Moved to `@react-aria/utils` to avoid creating a circular dependency when I import it in `@react-aria/interactions`.

packages/@react-aria/interactions/src/usePress.ts
```diff
@@ -334,23 +340,22 @@ export function usePress(props: PressHookProps): PressResult {
           if (isDisabled) {
             e.preventDefault();
           }
-
+          
           // If triggered from a screen reader or by using element.click(),
           // trigger as if it were a keyboard click.
-          if (!state.ignoreClickAfterPress && !state.ignoreEmulatedMouseEvents && !state.isPressed && (state.pointerType === 'virtual' || isVirtualClick(e.nativeEvent))) {
-            // Ensure the element receives focus (VoiceOver on iOS does not do this)
```
devongovett, line 341: I can't reproduce the original issue anymore. iOS VO uses DOM focus now. Not sure if we need to worry about this elsewhere. If so, it should probably be done e.g. in dialog trigger instead of here so it doesn't affect other things like submitting a form.

packages/@react-spectrum/table/test/Table.test.js
```diff
@@ -3191,22 +3193,21 @@ export let tableTests = () => {
...
         let tree = renderTable({onSelectionChange, selectionStyle: 'highlight', onAction});
 
-        fireEvent.keyDown(getCell(tree, 'Baz 10'), {key: ' '});
-        fireEvent.keyUp(getCell(tree, 'Baz 10'), {key: ' '});
+        act(() => getCell(tree, 'Baz 10').focus());
+        await user.keyboard(' ');
         checkSelection(onSelectionChange, ['Foo 10']);
-        // screen reader automatically handles this one
-        expect(announce).not.toHaveBeenCalled();
+        expect(announce).toHaveBeenCalledWith('Foo 10 selected.');
```
devongovett, line 3204: This test was wrong before. The announcement was not fired because the table never had focus.

packages/@react-aria/grid/src/useGridSelectionAnnouncement.ts
```diff
@@ -96,7 +96,17 @@ export function useGridSelectionAnnouncement<T>(props: GridSelectionAnnouncement
...
+
+  useUpdateEffect(() => {
+    if (state.selectionManager.isFocused) {
+      announceSelectionChange();
+    } else {
+      // Wait a frame in case the collection is about to become focused (e.g. on mouse down).
+      let raf = requestAnimationFrame(announceSelectionChange);
+      return () => cancelAnimationFrame(raf);
+    }
+  }, [selection, state.selectionManager.isFocused]);
```
LFDanLu, line 109: Not necessarily related to this change (but spurred due to the comment of mouse down here) but now that the Table's cell is focused on click rather than the row, the announcement of row selection seems to be interrupted a lot more, at least on Safari 17 + VoiceOver.

packages/@react-aria/interactions/src/useLongPress.ts
```diff
@@ -81,6 +81,12 @@ export function useLongPress(props: LongPressProps): LongPressResult {
         timeRef.current = setTimeout(() => {
           // Prevent other usePress handlers from also handling this event.
           e.target.dispatchEvent(new PointerEvent('pointercancel', {bubbles: true}));
+
+          // Ensure target is focused. On touch devices, browsers typically focus on pointer up.
+          if (getOwnerDocument(e.target).activeElement !== e.target) {
+            focusWithoutScrolling(e.target as FocusableElement);
+          }
```
LFDanLu, line 88: Perhaps not a blocker, but just noticed that long pressing on "Two" in ["TableView: static" story w/ highlight selection and multiple selection](https://reactspectrum.blob.core.windows.net/reactspectrum/f7e4d9e220b5afbeecd1be5498070e39a77a6db5/storybook/index.html?path=/story/tableview--static&args=selectionMode:multiple;selectionStyle:highlight&providerSwitcher-express=false) via Chrome Android causes a focus ring to appear on the "One" cell instead of "Two". Doesn't seem to happen on my iPad though

packages/@react-aria/interactions/src/usePress.ts
LFDanLu: Unable to reproduce as well

packages/@react-aria/interactions/src/usePress.ts
```diff
@@ -455,11 +450,11 @@ export function usePress(props: PressHookProps): PressResult {
         }
 
         if (e.button === 0) {
-          // Chrome and Firefox on touch Windows devices require mouse down events
-          // to be canceled in addition to pointer events, or an extra asynchronous
-          // focus event will be fired.
-          if (shouldPreventDefaultDown(e.currentTarget as Element)) {
-            e.preventDefault();
```
LFDanLu, line 462: note: was unable to confirm this case still, will need someone with a touch Windows device

snowystinger (approved):

LFDanLu (approved):

haukurmar:
Will we see this fix anytime soon in react-aria-components without a nightly version?

subvertallchris:
Glad to see this. Since it depends on a WebKit change, does it mean that Mobile Safari < 17 will still have problems?

devongovett:
@haukurmar will be in our next release in a couple weeks or so. Please feel free to try out the nightly and report any issues you find. Since this is a significant change, we are letting it bake for a bit to ensure we find any issues.

@subvertallchris as far as I know, the difference in Safari < 17 is that it will not focus button elements at all when you tap on them. That is the default native behavior. Adding an explicit `tabindex` fixes this in 17+, in which case Mobile Safari will focus the element when the tap is released (different from our prior behavior of focusing on press start which caused the unexpected keyboard behavior). We're not currently planning on adding any workarounds for older Safari versions, but please let us know if you see any significant issues.

subvertallchris:
@devongovett great explanation, thank you. Matching the native behavior is 100% good.

PR #7580, opened 2025-01-07
feat: S2 Menu design updates

devongovett:
* Flips the link out icon in menu items in RTL locales. Found in testing and checked with design.
* Adds a prop to hide the link out icons in menu items, requested by shell. This is useful when all of the items in a menu open in a new tab, and an icon on every item looks silly. This prop is applied at the global menu level rather than on individual menu items.
* Updated the alignment of menu items in multi-section menus to reserve space for icons, selection indicators, etc. on a per-section basis rather than across the entire menu. For example in the selection group example the section without selection is now no longer indented due to the other sections. Same thing if one section has icons and another does not. This was also requested by shell.

### DIFF
--- a/packages/@react-spectrum/s2/src/Menu.tsx
+++ b/packages/@react-spectrum/s2/src/Menu.tsx
@@ -88,17 +88,19 @@ export interface MenuProps<T> extends Omit<AriaMenuProps<T>, 'children' | 'style
 
 export const MenuContext = createContext<ContextValue<MenuProps<any>, DOMRefValue<HTMLDivElement>>>(null);
 
+const menuItemGrid = {
+  size: {
+    S: [edgeToText(24), 'auto', 'auto', 'minmax(0, 1fr)', 'auto', 'auto', 'auto', edgeToText(24)],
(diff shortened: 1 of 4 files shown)


packages/@react-spectrum/s2/src/Menu.tsx
```diff
@@ -81,22 +81,26 @@ export interface MenuProps<T> extends Omit<AriaMenuProps<T>, 'children' | 'style
   /**
    * The contents of the collection.
    */
-  children?: ReactNode | ((item: T) => ReactNode)
+  children?: ReactNode | ((item: T) => ReactNode),
+  /** Hides the default link out icons on menu items that open links in a new tab. */
+  hideLinkOutIcon?: boolean
```
devongovett, line 86: name?

LFDanLu: I think this name is fine actually haha

LFDanLu (approved):
LGTM

packages/@react-spectrum/s2/src/Menu.tsx
ktabors: Does it need the `icon` part of the name?

ktabors (approved):

PR #7584, opened 2025-01-08
fix: Autocomplete event leak and erroneous item focus after backspacing 

LFDanLu:
Found when testing Autocomplete with S2 components

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Tests should be sufficient, tested before and after the code changes to confirm that they caught the issues. If you want to test manually, the event leak can be tested by constructing a story in the S2 storybook using Autocomplete and seeing if the Storybook keyboard shortcuts like "O" aren't triggered when typing those letters in the Autocomplete. 

The backspacing issue can be tested in the "Autocomplete disabled key" RAC story. Type "r", backspace, and press ArrowDown. The first item should be focused at this point

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-aria/autocomplete/src/useAutocomplete.ts
+++ b/packages/@react-aria/autocomplete/src/useAutocomplete.ts
@@ -223,7 +223,7 @@ export function UNSTABLE_useAutocomplete(props: AriaAutocompleteOptions, state: 
     }
   };
 
-  let onKeyUp = useEffectEvent((e) => {
+  let onKeyUpCapture = useEffectEvent((e) => {
     // Dispatch simulated key up events for things like triggering links in listbox
     // Make sure to stop the propagation of the input keyup event so that the simulated keyup/down pair
     // is detected by usePress instead of the original keyup originating from the input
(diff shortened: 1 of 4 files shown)


reidbarber (approved):

devongovett (approved):

estenp:
@LFDanLu Looks like this merged last year, but I'm seeing this behavior in Storybook with RAC 1.14

Maybe a possible regression? Is there something I can do to help verify the legitimacy?

I'd make a sandbox repro but as noted in the original issue description a good way to test this is in storybook and typing "s" or "a". The storybook hotkeys are enacted to toggle sidebar/nav visibilty.

Removing the Autocomplete component wrapper will cause this behavior to stop while typing into the same TextField

LFDanLu:
@estenp A sandbox reproduction would still be helpful since storybook for us was just the simplest way to reproduce it. I currently can't reproduce it in our storybook but the hotkeys seem to not work at all so I'm digging to see what changed there.

LFDanLu:
I've reproduced locally in our Storybook now, looks like we turned off the hotkeys shortly after this pull went in :/. Not sure why the test didn't fail, will have to investigate. Mind filing a new issue so we can track this?

estenp:
> I've reproduced locally in our Storybook now, looks like we turned off the hotkeys shortly after this pull went in :/. Not sure why the test didn't fail, will have to investigate. Mind filing a new issue so we can track this?

Ok got it, yes I can create an issue. For the repro, I've got a minimal code example which is 95% pulled from the Autocomplete docs. Is there an existing codesandbox template for React Aria + Storybook (or even just RA, for future reference)?

LFDanLu:
Thanks! For a template, I'd recommend using the docs example to grab a stackblitz that is fully setup with the attached example like so: https://stackblitz.com/run?file=src%2FExample.tsx
That can be found under any example via the menu shown below in the screenshot. Otherwise, https://react-aria.adobe.com/getting-started#storybook-starter-kits has some storybooks with RAC premade components that could be helpful as well.

image

PR #7574, opened 2025-01-06
fix(s2): update button gradients to use static colors

jluyau:
update Button gradients (genai & premium) to use static colors as to not change for light/dark theme

## ✅ Pull Request Checklist:

- [ ] Included link to corresponding [React Spectrum GitHub Issue](https://github.com/adobe/react-spectrum/issues).
- [ ] Added/updated unit tests and storybook for this change (for new code or code which already has tests).
- [ ] Filled out test instructions.
- [ ] Updated documentation (if it already exists for this component).
- [ ] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/WAI/ARIA/apg/)

PR #7585, opened 2025-01-08
fix(s2): Make gradient buttons have an animated transition

devongovett:
This enables animated transitions on gradient buttons in the hover state like the other variants. It works by using registered custom properties via the [@property](https://developer.mozilla.org/en-US/docs/Web/CSS/@property) rule, which allows defining the syntax that variables should use (`<color>` in this case). Then, you can add these custom properties to `transition-property` and they'll animate. The `linearGradient` macro now generates these variable definitions on demand.

Unfortunately, due to yet another [Safari bug](https://bugs.webkit.org/show_bug.cgi?id=285622) the gradient must be on a separate element from the button itself otherwise it causes layout flickering ([reproduction](https://codepen.io/devongovett/pen/ByBYQXd)).

Also fixes the static color issue (closes #7574) by simply setting the `color-scheme` to `light` so it doesn't change in dark mode. I'd rather not add a bunch of new tokens to the theme if they are only going to be used here.


reidbarber (approved):

yihuiliao:
Are the gradient button's supposed to retain their background when pending or should it go gray like the other buttons?

devongovett:
oops yeah, updated

yihuiliao (approved):

LFDanLu (approved):

PR #7606, opened 2025-01-14
fix: Properly clear Autocomplete wrapped SearchField when "x" button is pressed

LFDanLu:
From testing

## ✅ Pull Request Checklist:

## 🧢 Your Project:

RSP

### DIFF
--- a/packages/@react-aria/autocomplete/src/useAutocomplete.ts
+++ b/packages/@react-aria/autocomplete/src/useAutocomplete.ts
@@ -11,12 +11,12 @@
  */
 
 import {AriaLabelingProps, BaseEvent, DOMProps, RefObject} from '@react-types/shared';
+import {AriaTextFieldProps} from '@react-aria/textfield';
 import {AutocompleteProps, AutocompleteState} from '@react-stately/autocomplete';
-import {ChangeEvent, InputHTMLAttributes, KeyboardEvent as ReactKeyboardEvent, useCallback, useEffect, useMemo, useRef} from 'react';
 import {CLEAR_FOCUS_EVENT, FOCUS_EVENT, isCtrlKeyPressed, mergeProps, mergeRefs, UPDATE_ACTIVEDESCENDANT, useEffectEvent, useId, useLabels, useObjectRef} from '@react-aria/utils';
+import {InputHTMLAttributes, KeyboardEvent as ReactKeyboardEvent, useCallback, useEffect, useMemo, useRef} from 'react';
 // @ts-ignore
 import intlMessages from '../intl/*.json';
-import {useKeyboard} from '@react-aria/interactions';
 import {useLocalizedStringFormatter} from '@react-aria/i18n';
 
 export interface CollectionOptions extends DOMProps, AriaLabelingProps {
@@ -43,6 +43,8 @@ export interface AriaAutocompleteOptions extends Omit<AriaAutocompleteProps, 'ch
 export interface AutocompleteAria {
   /** Props for the autocomplete input element. */
   inputProps: InputHTMLAttributes<HTMLInputElement>,
+  /** Props for the autocomplete textfield/searchfield element. These should be passed to the textfield/searchfield aria hooks respectively. */
+  textFieldProps: AriaTextFieldProps,
   /** Props for the collection, to be passed to collection's respective aria hook (e.g. useMenu). */
   collectionProps: CollectionOptions,
   /** Ref to attach to the wrapped collection. */
@@ -138,16 +140,16 @@ export function UNSTABLE_useAutocomplete(props: AriaAutocompleteOptions, state: 
   });
 
   // TODO: update to see if we can tell what kind of event (paste vs backspace vs typing) is happening instead
-  let onChange = (e: ChangeEvent<HTMLInputElement>) => {
+  let onChange = (value: string) => {
     // Tell wrapped collection to focus the first element in the list when typing forward and to clear focused key when deleting text
     // for screen reader announcements
-    if (state.inputValue !== e.target.value && state.inputValue.length <= e.target.value.length) {
+    if (state.inputValue !== value && state.inputValue.length <= value.length) {
       focusFirstItem();
     } else {
       clearVirtualFocus();
     }
 
-    state.setInputValue(e.target.value);
+    state.setInputValue(value);
   };
 
   // For textfield specific keydown operations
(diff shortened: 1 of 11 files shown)


packages/react-aria-components/test/Autocomplete.test.tsx
```diff
@@ -199,6 +200,26 @@ describe('Autocomplete', () => {
     expect(onKeyDown).not.toHaveBeenCalled();
     onKeyDown.mockReset();
   });
+
+  it('should prevent clear the input field when clicking on the clear button', async () => {
```
snowystinger, line 204: prevent clear?

LFDanLu: lol whoops

packages/@react-aria/autocomplete/src/useAutocomplete.ts
```diff
@@ -43,6 +43,8 @@ export interface AriaAutocompleteOptions extends Omit<AriaAutocompleteProps, 'ch
 export interface AutocompleteAria {
   /** Props for the autocomplete input element. */
   inputProps: InputHTMLAttributes<HTMLInputElement>,
+  /** Props for the autocomplete textfield/searchfield element. These should be passed to the textfield/searchfield aria hooks respectively. */
+  textFieldProps: AriaTextFieldProps,
```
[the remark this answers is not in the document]
snowystinger: Wouldn't SearchField/TextField interfere with InputProps anyways? so really there are three components we're potentially targeting, two of them just happen to use the same set of props?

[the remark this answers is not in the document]
devongovett: hmm maybe we should add those props to AriaTextfieldProps. Then we could remove the weird merging with InputContext too. Looks like just `aria-controls`, `autoCorrect`, and `spellCheck` are missing? Those seem like reasonable props for TextField to support.

packages/@react-aria/autocomplete/src/useAutocomplete.ts
```diff
@@ -11,12 +11,12 @@
  */
 
 import {AriaLabelingProps, BaseEvent, DOMProps, RefObject} from '@react-types/shared';
+import {AriaTextFieldProps} from '@react-aria/textfield';
```
devongovett, line 14: Can we get this through the `react-aria` monopackage instead of adding a new dependency?

packages/react-aria-components/src/Autocomplete.tsx
```diff
@@ -58,6 +61,8 @@ export function UNSTABLE_Autocomplete(props: AutocompleteProps) {
     <Provider
       values={[
         [UNSTABLE_AutocompleteStateContext, state],
+        [SearchFieldContext, {...textFieldProps}],
+        [TextFieldContext, {...textFieldProps}],
```
devongovett, line 65: shouldn't need the spread?

packages/@react-aria/autocomplete/src/useAutocomplete.ts
LFDanLu: wait, do we really want to add `react-aria` as a dependency here? Wouldn't that pull in a lot more?

ktabors (dismissed):

packages/react-aria-components/src/Autocomplete.tsx
```diff
@@ -58,7 +60,9 @@ export function UNSTABLE_Autocomplete(props: AutocompleteProps) {
     <Provider
       values={[
         [UNSTABLE_AutocompleteStateContext, state],
-        [InputContext, {...inputProps, ref: inputRef}],
+        [SearchFieldContext, textFieldProps],
+        [TextFieldContext, textFieldProps],
+        [InputContext, {ref: inputRef}],
```
snowystinger, line 65: can't we pass the inputRef to SearchField or TextField Context now?

packages/@react-aria/autocomplete/src/useAutocomplete.ts
devongovett: Oh sorry I thought this was in RAC

packages/@react-aria/autocomplete/src/useAutocomplete.ts
```diff
@@ -166,12 +166,7 @@ export function UNSTABLE_useAutocomplete(props: AriaAutocompleteOptions, state:
         // Early return for Escape here so it doesn't leak the Escape event from the simulated collection event below and
         // close the dialog prematurely. Ideally that should be up to the discretion of the input element hence the check
         // for isPropagationStopped
-        // Also set the inputValue to '' to cover Firefox case where Esc doesn't actually clear searchfields. Normally we already
-        // handle this in useSearchField, but we are directly setting the inputValue on the input element in RAC Autocomplete instead of
-        // passing it to the SearchField via props. This means that a controlled value set on the Autocomplete isn't synced up with the
-        // SearchField until the user makes a change to the field's value via typing
         if (e.isPropagationStopped()) {
```
devongovett, line 169: Is this still needed?

packages/react-aria-components/src/Autocomplete.tsx
devongovett: if so can we remove the InputContext merging?

devongovett:
We might want to omit `autoCorrect` and `spellCheck` from ColorField props.

packages/react-aria-components/src/Autocomplete.tsx
LFDanLu: if I pass the ref down through context, it will be attached to the wrapping div rather than the input itself so I still need the InputContext. If I can get rid of the checks against `inputRef` in useAutocomplete then I can get rid of it but digging for alternatives right now

devongovett: remind me why we need a document-level capturing listener?

LFDanLu: We need to dispatch a keyup event onto the ListBox link item so that we can properly simulate a Enter press and trigger a the Listbox's link navigation

snowystinger (approved):

packages/@react-aria/autocomplete/src/useAutocomplete.ts
```diff
@@ -35,14 +35,12 @@ export interface AriaAutocompleteProps extends AutocompleteProps {
 
 export interface AriaAutocompleteOptions extends Omit<AriaAutocompleteProps, 'children'> {
   /** The ref for the wrapped collection element. */
-  collectionRef: RefObject<HTMLElement | null>,
-  /** The ref for the wrapped input element. */
-  inputRef: RefObject<HTMLInputElement | null>
+  collectionRef: RefObject<HTMLElement | null>
```
devongovett, line 38: wondering if this could move to a third parameter now that there is only one, but fine for this release

devongovett (approved):

packages/@react-aria/autocomplete/src/useAutocomplete.ts
LFDanLu: sounds good, I'll make a note and handle this in a follow up

PR #7607, opened 2025-01-14
fix(s2): Improve ActionBar transition when scrollbars are visible

devongovett:
Small fix from testing. Transitioning only the `transform` property prevents animating the width when scrollbars appear. Removed opacity animation because it isn't necessary since the table/cardview clips it.


ktabors (approved):
Transitions looks good.

reidbarber (approved):

PR #7608, opened 2025-01-14
fix(TagGroup): ensure Tags with and without icons are aligned vertically

reidbarber:
Previously if you tried to use Tags with icons and without icons in the same TagGroup, they would not be vertically aligned. This seems to be a valid use case within Spectrum guidelines, so we want to ensure they can be used together without being misaligned.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Check new story and chromatic.

### DIFF
--- a/packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
+++ b/packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
@@ -122,20 +122,6 @@ export const WithIcons: TagGroupStory = {
   )
 };
 
-export const WithSomeIcons: TagGroupStory = {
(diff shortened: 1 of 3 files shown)


snowystinger (approved):

packages/@react-spectrum/tag/stories/TagGroup.stories.tsx
```diff
@@ -122,6 +122,20 @@ export const WithIcons: TagGroupStory = {
   )
 };
 
+export const WithSomeIcons: TagGroupStory = {
```
dannify, line 125: instead of a new one, can you change an existing story and just remove/add an avatar? That way we can also check chromatic.

snowystinger (approved):

devongovett (approved):

PR #7626, opened 2025-01-17
fix: (WIP) Fix autocomplete searchfield focus ring

LFDanLu:
Closes 

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/react-aria-components/src/Group.tsx
+++ b/packages/react-aria-components/src/Group.tsx
@@ -69,8 +69,7 @@ export const Group = /*#__PURE__*/ (forwardRef as forwardRefType)(function Group
 
   let {hoverProps, isHovered} = useHover({onHoverStart, onHoverChange, onHoverEnd, isDisabled});
   let {isFocused, isFocusVisible, focusProps} = useFocusRing({
-    within: true,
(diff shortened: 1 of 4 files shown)


packages/react-aria-components/src/Group.tsx
```diff
@@ -69,7 +69,8 @@ export const Group = /*#__PURE__*/ (forwardRef as forwardRefType)(function Group
 
   let {hoverProps, isHovered} = useHover({onHoverStart, onHoverChange, onHoverEnd, isDisabled});
   let {isFocused, isFocusVisible, focusProps} = useFocusRing({
-    within: true
+    within: true,
+    isTextInput: true
```
LFDanLu, line 73: The reason for the focus ring always reappearing as you type in the search field is because the Group itself has `focusRing` applied to it but doesn't have `isTextInput`. When `useFocusVisible` detects the native keyboards we dispatch on the autocomplete collection/collection items, it believes that it needs to update the modality for the Group. Normally this would be correct, however, the Group is still focused despite those events happening on different elements and thus its focus visibility get updated. Still trying to think of a good way to handle this

Issue #7447, opened 2024-11-28
Select Component Fails to Pass autoComplete Prop to HiddenSelect Component


nabanitabania:
### Provide a general summary of the issue here

Browser autofill functionality is unavailable in forms because the Select component does not pass the autoComplete attribute to the underlying [HiddenSelect](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/Select.tsx#L204) component.

### 🤔 Expected Behavior?

The `Select` component should properly pass down the `autoComplete` attribute to the underlying HiddenSelect component, enabling browser autofill functionality in forms where the `Select` component is used.

### 😯 Current Behavior

The browser autofill is unable to populate dropdown fields, such as those for state, country, or other similar form inputs. This limitation prevents seamless autofill functionality, disrupting the user experience when filling out forms that rely on browser-managed autofill data.

### 💁 Possible Solution

Send `autoComplete` prop to `HiddenSelect` component from `Select`.

### 🖥️ Steps to Reproduce
https://codesandbox.io/p/sandbox/hungry-dan-sv87ql?file=%2Fsrc%2FApp.js

Use the link above to access the form and attempt to fill it using the browser's autofill feature.  
You'll notice that all fields, except the "Country" field, are successfully autofilled.

### Version

	1.5.0

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

macOS Sonoma (14.6)

PR #7670, opened 2025-01-25
fix(Select): HiddenSelect Component Auto-Completion Issue

### LINKED ISSUE #7660: Browser autofill for React Aria Select is not working.

The issue involves browser autofill not functioning correctly with dropdowns in React Aria. While the library's `useHiddenSelect` hook uses a visually hidden `<select>` element for autofill compatibility, the `onChange` handler in its `selectProps` does not receive the correct `e.target.value` when autofill is triggered. This is likely due to how the `useEffectEvent` hook manages event handlers, potentially causing synchronization issues between React's synthetic events and native browser autofill events. As a result, the dropdown's state fails to update with the autofilled value, disrupting expected behavior.

### 🤔 Expected Behavior?

When browser autofill is triggered on a dropdown using React Aria, the hidden `<select>` element should seamlessly integrate with the autofill functionality. The `onChange` handler within the `useHiddenSelect` hook's `selectProps` should capture the event and receive the correct value in `e.target.value`, reflecting the autofilled selection. This value should then update the component's state (`state.selectedKey`) to ensure the dropdown displays the autofilled choice accurately. The entire process should be smooth and consistent, aligning with native autofill behavior across all supported browsers.

### 😯 Current Behavior

Currently, when browser autofill is used on dropdowns in React Aria, the interaction breaks down at the `onChange` handler of the hidden `<select>` element. Instead of receiving the expected autofilled value, `e.target.value` is empty (`""`). This prevents the component's state (`state.selectedKey`) from updating to reflect the autofilled input. While the library's implementation attempts to handle autofill through the visually hidden `<select>`, it appears that native autofill events are not being processed correctly, leading to a disconnect between the browser's autofill mechanism and the component's event handling.

### 💁 Possible Solution

_No response_

### 🔦 Context

_No response_

### 🖥️ Steps to Reproduce

-https://codesandbox.io/p/sandbox/focused-water-lhqtty
 
- Open the sandbox link and try autofilling the form using the browser's saved address.  
- Open Developer Tools and navigate to the **Sources** tab.  
- Locate `HiddenSelect.mjs` and `useEffectEvent.mjs`.  
- Set breakpoints on:  
  - The `onChange` function in `HiddenSelect.mjs`.  
  - The line `const f = ref.current!;` in `useEffectEvent.mjs`.  
- Trigger autofill and observe the behavior.  
- Notice how the native event's value (`e.target.value`) is lost in the `onChange` handler, causing the dropdown state not to update.

Image
Image

### Version

3.39.0

Chrome

### If other, please specify.

_No response_

Windows

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

igorlima:
Closes #7660

## Summary

This pull request addresses the issue #7660, focusing on fixing the value assignment in the `HiddenSelect` component within the `@react-aria/select` package. The changes ensure that the `value` attribute now defaults to the current value of `selectRef` when `state.selectedKey` isn't set. This should enhance the native browser autocomplete.

## Changes Made

- **File:** `packages/@react-aria/select/src/HiddenSelect.tsx`
  - Updated the `value` property in the `useHiddenSelect` function to default to `props.selectRef?.current?.value` if `state.selectedKey` is unavailable.

## Verification

- Checked that the `HiddenSelect` component now assigns the `value` attribute, and the fallback mechanism works as intended!
- Ensured that all existing functionality and behaviors are still intact and not affected by these changes.

## Additional Context

When the browser triggers autocomplete, it first fills out the hidden select component before passing the value to the visible React Aria Select component. Unfortunately, there was a hiccup where the values weren't being passed down properly, which meant the select form field wasn't being filled out correctly by the native browser autocomplete.
- @nabanitabania has shared a lot of helpful details in the links below:
  - https://github.com/adobe/react-spectrum/issues/7660
  - https://github.com/adobe/react-spectrum/issues/7447
  - https://github.com/adobe/react-spectrum/pull/7452
- Below is a demo link to see the difference before and after the fix:
  - [Demo Link](https://igorlima.github.io/unapologetic-snippets/docs/languages/js/playground/react/20250124-1451-react-aria-components-autocomplete-issue.html) - Inspired by the [sandbox](https://codesandbox.io/p/sandbox/focused-water-lhqtty) @nabanitabania shared in [ISSUE-7660](https://github.com/adobe/react-spectrum/issues/7660).

    - | Before Fix                             | After Fix                                      |
      | :---: | :---: |
      | image | image |
      | [_demo link_](https://igorlima.github.io/unapologetic-snippets/docs/languages/js/playground/react/20250124-1451-react-aria-components-autocomplete-issue.html) | [_demo link_](https://igorlima.github.io/unapologetic-snippets/docs/languages/js/playground/react/20250124-1451-react-aria-components-autocomplete-issue.html) |
      | native browser behavior wasn't working | native browser behavior works as intended |

Feel free to share thoughts!

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

There isn't a specific React Aria form story to test this issue, but I've included a [demo link](https://igorlima.github.io/unapologetic-snippets/docs/languages/js/playground/react/20250124-1451-react-aria-components-autocomplete-issue.html) to check it out. It was inspired by the [sandbox](https://codesandbox.io/p/sandbox/hungry-dan-sv87ql) that @nabanitabania shared in [a previous PR](https://github.com/adobe/react-spectrum/pull/7452).

### DIFF
--- a/packages/@react-aria/select/src/HiddenSelect.tsx
+++ b/packages/@react-aria/select/src/HiddenSelect.tsx
@@ -114,7 +114,6 @@ export function HiddenSelect<T>(props: HiddenSelectProps<T>) {
   let {state, triggerRef, label, name, isDisabled} = props;
   let selectRef = useRef(null);
   let {containerProps, selectProps} = useHiddenSelect({...props, selectRef}, state, triggerRef);
-  let selectRefValue = selectRef?.current?.value;
 
   // If used in a <form>, use a hidden input so the value can be submitted to a server.
   // If the collection isn't too big, use a hidden <select> element for this so that browser
@@ -149,7 +148,7 @@ export function HiddenSelect<T>(props: HiddenSelectProps<T>) {
         autoComplete={selectProps.autoComplete}
         name={name}
         disabled={isDisabled}
-        value={state.selectedKey ?? (selectRefValue || '')} />
+        value={state.selectedKey ?? ''} />
     );
   }
 
--- a/packages/@react-aria/select/src/HiddenSelect.tsx
+++ b/packages/@react-aria/select/src/HiddenSelect.tsx
@@ -68,7 +68,6 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectOptions, state: Select
   let {autoComplete, name = data.name, isDisabled = data.isDisabled} = props;
   let {validationBehavior, isRequired} = data;
   let {visuallyHiddenProps} = useVisuallyHidden();
-  let selectRefValue = props.selectRef?.current?.value;
 
   useFormReset(props.selectRef, state.selectedKey, state.setSelectedKey);
   useFormValidation({
--- a/packages/@react-aria/select/src/HiddenSelect.tsx
+++ b/packages/@react-aria/select/src/HiddenSelect.tsx
@@ -100,7 +100,7 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectOptions, state: Select
       disabled: isDisabled,
       required: validationBehavior === 'native' && isRequired,
       name,
-      value: state.selectedKey ?? (selectRefValue || ''),
+      value: state.selectedKey ?? undefined,
       onChange: (e: React.ChangeEvent<HTMLSelectElement>) => state.setSelectedKey(e.target.value)
     }
   };
--- a/packages/@react-aria/select/src/HiddenSelect.tsx
+++ b/packages/@react-aria/select/src/HiddenSelect.tsx
@@ -68,6 +68,7 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectOptions, state: Select
   let {autoComplete, name = data.name, isDisabled = data.isDisabled} = props;
   let {validationBehavior, isRequired} = data;
   let {visuallyHiddenProps} = useVisuallyHidden();
+  let selectRefValue = props.selectRef?.current?.value;
 
   useFormReset(props.selectRef, state.selectedKey, state.setSelectedKey);
   useFormValidation({
@@ -99,7 +100,7 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectOptions, state: Select
       disabled: isDisabled,
       required: validationBehavior === 'native' && isRequired,
       name,
-      value: state.selectedKey ?? (props.selectRef?.current?.value || ''),
+      value: state.selectedKey ?? (selectRefValue || ''),
       onChange: (e: React.ChangeEvent<HTMLSelectElement>) => state.setSelectedKey(e.target.value)
     }
   };
@@ -113,6 +114,7 @@ export function HiddenSelect<T>(props: HiddenSelectProps<T>) {
   let {state, triggerRef, label, name, isDisabled} = props;
   let selectRef = useRef(null);
   let {containerProps, selectProps} = useHiddenSelect({...props, selectRef}, state, triggerRef);
+  let selectRefValue = selectRef?.current?.value;
 
   // If used in a <form>, use a hidden input so the value can be submitted to a server.
   // If the collection isn't too big, use a hidden <select> element for this so that browser
@@ -147,7 +149,7 @@ export function HiddenSelect<T>(props: HiddenSelectProps<T>) {
         autoComplete={selectProps.autoComplete}
         name={name}
         disabled={isDisabled}
-        value={state.selectedKey ?? (selectRef?.current?.value || '')} />
+        value={state.selectedKey ?? (selectRefValue || '')} />
     );
   }
 
--- a/packages/@react-aria/select/src/HiddenSelect.tsx
+++ b/packages/@react-aria/select/src/HiddenSelect.tsx
@@ -99,7 +99,7 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectOptions, state: Select
       disabled: isDisabled,
       required: validationBehavior === 'native' && isRequired,
       name,
-      value: state.selectedKey ?? '',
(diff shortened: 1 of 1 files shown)


samuelmd (approved):

samuelmd (approved):

leandrolimasi (approved):

felipeac:
LGTM

reidbarber (approved):
LGTM, and thanks for the demo link!

packages/@react-aria/select/src/HiddenSelect.tsx
```diff
@@ -68,6 +68,7 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectOptions, state: Select
   let {autoComplete, name = data.name, isDisabled = data.isDisabled} = props;
   let {validationBehavior, isRequired} = data;
   let {visuallyHiddenProps} = useVisuallyHidden();
+  let selectRefValue = props.selectRef?.current?.value;
```
snowystinger, line 71: Reading from a ref during render is typically unsafe and unfortunately it's hard to tell when it potentially breaks React's concurrent mode since they don't expose any of that for tests.

We should try to find another way to handle this.

does the hidden select fire an onChange that we aren't listening to? I'm worried that the state doesn't know about this value.

igorlima: Thank you for raising this concern!  

This repository already has a [lint rule](https://github.com/adobe/react-spectrum/blob/fd7075c5f0e06998ea8ac6a341dee165919fb6c1/eslint.config.mjs#L227C10-L227C30) to detect issues related to reading from refs during rendering. In my initial proposal, the [lint failed](https://app.circleci.com/pipelines/github/adobe/react-spectrum/24452/workflows/4ffb7f30-23eb-43fd-a40b-92438801db89/jobs/395890?invite=true#step-102-4882_111) because it did indeed read from a ref during render. However, I [addressed this](https://github.com/adobe/react-spectrum/pull/7670/commits/4a2615e8ed16529194a8e58763a52aee526321de), and the lint has now passed.  
- <details><summary><i>screenshot <sup>lint fixed</sup></i></summary>
   
   image
   </details<


Regarding the use of `selectRef`, there are existing references to it in this function:  
- Lines [73](https://github.com/adobe/react-spectrum/blob/4a2615e8ed16529194a8e58763a52aee526321de/packages/%40react-aria/select/src/HiddenSelect.tsx#L73), [76](https://github.com/adobe/react-spectrum/blob/4a2615e8ed16529194a8e58763a52aee526321de/packages/%40react-aria/select/src/HiddenSelect.tsx#L76), and [77](https://github.com/adobe/react-spectrum/blob/4a2615e8ed16529194a8e58763a52aee526321de/packages/%40react-aria/select/src/HiddenSelect.tsx#L77).  
- Additionally, line [74 calls `useFormValidation`](https://github.com/adobe/react-spectrum/blob/4a2615e8ed16529194a8e58763a52aee526321de/packages/%40react-aria/select/src/HiddenSelect.tsx#L74) and passes `selectRef` as a parameter. Within [`useFormValidation`](https://github.com/adobe/react-spectrum/blob/fd7075c5f0e06998ea8ac6a341dee165919fb6c1/packages/%40react-aria/form/src/useFormValidation.ts#L25), [`ref`](https://github.com/adobe/react-spectrum/blob/fd7075c5f0e06998ea8ac6a341dee165919fb6c1/packages/%40react-aria/form/src/useFormValidation.ts#L41) is heavily used to access the ref's value.  

@snowystinger given that the **lint passes** for this case and considering the **function's existing use of refs**, I believe **it’s safe to use `selectRef` here**. 🙏

igorlima: The concern that the state might not be aware of the value is valid. I added a `debugger` to the demo link to check everything is still working as intended. By opening the DevTools and triggering the native browser autocomplete, we can see that the value is set. Additionally, we can check all values when we submit the form.
- <details><summary><i>screenshots</i></summary>
   
   image
   
   ------
   image
   
   ------
   image
   </details>

packages/@react-aria/select/src/HiddenSelect.tsx
```diff
@@ -99,7 +100,7 @@ export function useHiddenSelect<T>(props: AriaHiddenSelectOptions, state: Select
       disabled: isDisabled,
       required: validationBehavior === 'native' && isRequired,
       name,
-      value: state.selectedKey ?? '',
+      value: state.selectedKey ?? (selectRefValue || ''),
```
snowystinger, line 103: I propose this instead
```suggestion
      value: state.selectedKey ?? undefined,
```
I'm not sure how I'm avoiding the uncontrolled -> controlled issue, but I get no warnings for it.

packages/@react-aria/select/src/HiddenSelect.tsx
snowystinger, line 71:
```suggestion
```

igorlima: Everything seems to work as intended, and the code looks much more straightforward and cleaner now. Thanks for the help!

snowystinger (approved):

igorlima:
@reidbarber, the previous approval was dismissed after I applied one reviewer's suggestion. If you have a moment, could you take a look and let me know if anything else needs attention? The library requires at least two approvals before moving forward. Thanks. 🙏 
- <details><summary><sup><i>approval was dismissed</i></sup></summary>
   
   😥 
   image
   😞 
   </details>

filipw01:
It breaks for me the case where I want to unset the value by using `null`. Can we remove the `??` altogether since all it does is casting null and undefined to undefined?

PR #7677, opened 2025-01-28
fix: Fix focusing radio, checkbox, and switch on press

devongovett:
Fixes two issues introduced in usePress refactor:

1. Clicking on a RAC Radio, Checkbox, or Switch no longer focused the element. This is because usePress no longer programmatically focuses, and `useRadio`/`useToggle` include a `preventDefault` in the onClick handler on the label. RSP was not affected because there the native input is positioned on top of the label. Removing the preventDefault introduced the second problem...
2. On iOS, tapping the radio or toggle caused it to be triggered twice. Additionally, in testing we noticed that unexpected focus rings were appearing on listbox items in Tailwind examples on iOS. This was because our own click simulation that we trigger during `onPointerUp` in case the browser's click event is not fired was triggering in addition to the native `click` event. After some spelunking into the WebKit source code, I found that iOS has a [32ms delay](https://github.com/WebKit/WebKit/blob/dccfae42bb29bd4bdef052e469f604a9387241c0/Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm#L875-L892) on the click event for non-form elements without a [specific ARIA role](https://github.com/WebKit/WebKit/blob/dccfae42bb29bd4bdef052e469f604a9387241c0/Source/WebCore/accessibility/AccessibilityObject.cpp#L1186). For radios and toggles, you are tapping on the `<label>` element not the actual input so the fast path exception in WebKit does not apply. So our timer of 5ms was firing before the browser's click event was fired, ending up with two clicks. Raising this timeout fixes the issue.

### DIFF
--- a/packages/@react-aria/interactions/src/usePress.ts
+++ b/packages/@react-aria/interactions/src/usePress.ts
@@ -498,7 +498,9 @@ export function usePress(props: PressHookProps): PressResult {
             // However, iOS and Android do not focus or fire onClick after a long press.
             // We work around this by triggering a click ourselves after a timeout.
             // This timeout is canceled during the click event in case the real one fires first.
-            // In testing, a 0ms delay is too short. 5ms seems long enough for the browser to fire the real events.
+            // The timeout must be at least 32ms, because Safari on iOS delays the click event on
+            // non-form elements without certain ARIA roles (for hover emulation).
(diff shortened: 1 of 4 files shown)


reidbarber (approved):
LGTM, tested Chrome+Safari Mac and iOS Safari.

snowystinger (approved):

PR #7692, opened 2025-01-30
[WIP] Tree drag and drop

rob-clayburn:
I'm looking to get some feedback on this PR to add drag and drop to the Tree.

I am really really new to this codebase, and there are probably a ton of things I've not understood and have messed up

I've mainly just copied the logic from the listbox and applied it to the tree component.

There's a storybook story which I've added at http://localhost:9003/?path=/story/treeview--tree-example-dynamic-drag-n-drop&providerSwitcher-express=false&strict=true

Some things that would be great to get some feedback/help on are:

* I'm unclear as to whether the dragAndDropHooks should be created by the user each time (e.g. in the story). Some things i can see being useful, but it feels like the reorder function shouldn't have to be implemented here. Is there a better place to put this?

* Re-ordering works, but at some point it stops working. I can't seem to find a repeatable pattern of drag / drops to consistently replicate this though.

* In storybook the hot reload gives an error: 
image

* Although I've added the keyboard code - I've not yet tested that - was hoping to get the other issues ironed out first. 
* The way I've implemented the drop insert indicator causes the tree items to move which in turn can cause the drop to switch from before to after. The droppable listbox story renders items with a gap in between but I'm not sure if that is what is wanted in a tree view.

### DIFF
--- a/packages/@react-stately/data/src/useTreeData.ts
+++ b/packages/@react-stately/data/src/useTreeData.ts
@@ -116,10 +116,10 @@ export interface TreeData<T extends object> {
   moveBefore(key: Key, toParentKey: Key | null, index: number): void,
 
   /**
-   * Moves an item aftera node within the tree.
+   * Moves an item after ia node within the tree.
    * @param key - The key of the item to move.
    * @param toParentKey - The key of the new parent to insert into. `null` for the root.
-   * @param index - The index within the new parent to insert before.
+   * @param index - The index within the new parent to insert after.
    */
   moveAfter(key: Key, toParentKey: Key | null, index: number): void,
 
--- a/packages/@react-stately/data/src/useTreeData.ts
+++ b/packages/@react-stately/data/src/useTreeData.ts
@@ -116,7 +116,7 @@ export interface TreeData<T extends object> {
   moveBefore(key: Key, toParentKey: Key | null, index: number): void,
 
   /**
-   * Moves an item after ia node within the tree.
+   * Moves an item after a node within the tree.
    * @param key - The key of the item to move.
    * @param toParentKey - The key of the new parent to insert into. `null` for the root.
    * @param index - The index within the new parent to insert after.
@@ -419,7 +419,6 @@ export function useTreeData<T extends object>(options: TreeOptions<T>): TreeData
 
         // Otherwise, update the parent node and its ancestors.
         return updateTree(newItems, toParentKey, parentNode => {
-          console.log('parent node children ', parentNode.children);
           const c = [
             ...parentNode.children!.slice(0, afterIndex),
             movedNode,
--- a/packages/@react-stately/data/src/useTreeData.ts
+++ b/packages/@react-stately/data/src/useTreeData.ts
@@ -108,6 +108,22 @@ export interface TreeData<T extends object> {
   move(key: Key, toParentKey: Key | null, index: number): void,
 
   /**
+   * Moves an item before a node within the tree.
+   * @param key - The key of the item to move.
+   * @param toParentKey - The key of the new parent to insert into. `null` for the root.
+   * @param index - The index within the new parent to insert before.
+   */
+  moveBefore(key: Key, toParentKey: Key | null, index: number): void,
+
+  /**
+   * Moves an item aftera node within the tree.
+   * @param key - The key of the item to move.
+   * @param toParentKey - The key of the new parent to insert into. `null` for the root.
+   * @param index - The index within the new parent to insert before.
+   */
+  moveAfter(key: Key, toParentKey: Key | null, index: number): void,
+
+  /**
    * Updates an item in the tree.
    * @param key - The key of the item to update.
    * @param newValue - The new value for the item.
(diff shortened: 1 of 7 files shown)


reidbarber:
Thanks for the PR! I'm going to be taking a closer look when I have some time soon, just wanted to leave an update.

rob-clayburn:
thanks, have you managed to carve out some time to look at this?

reidbarber:
Sorry for the delay, we've been busy with the latest release! I just pushed a few small updates and a story that has DnD working in RAC Tree. I still need to debug RSP TreeView and fix some merge conflicts. The state management in the story should get simplified once we merge https://github.com/adobe/react-spectrum/pull/7854.

reidbarber:
Pushed a few changes so it should be in a decent state to test out. Some things to note:
- We're going to scope this to Tree in React Aria Components for now (adding to React Spectrum TreeView can be a follow-up). I made a new story for this and removed the old one.
- We'll probably need a way expand/collapse items while keyboard navigating during a drop.

packages/@react-stately/data/src/useTreeData.ts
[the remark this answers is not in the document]
LFDanLu: Another odd case that I discovered:
1. Drag "Reports 1" into "Projects 2" 
2. Drag "Project 2" down below "Project 3"
3. Drag "Project 2" down into "Reports" (Aka between "Reports" and "Reports 1" so that it becomes a child of "Reports"

Project 2 actually becomes a root level folder after this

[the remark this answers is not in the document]
LFDanLu: It might be that the drop operation thinks the drop position is "After Reports" which technically isn't wrong but it should really be "Before Reports 1" or have a greater degree of granularity (aka "between Reports and Reports 1"). This actually works with keyboard DnD (and has the proper "insert between Reports and Reports 1" drop target) but for some reason is targeting the wrong drop insertion target I think for mouse DnD

packages/react-aria-components/src/Tree.tsx
```diff
@@ -194,11 +219,128 @@ function TreeInner<T extends object>({props, collection, treeRef: ref}: TreeInne
...
+        onDropActivate: (e) => {
+          // Expand collapsed item when dragging over
+          if (e.target.type === 'item') {
+            let key = e.target.key;
+            let item = state.collection.getItem(key);
+            if (item && item.hasChildNodes && expandedKeys !== 'all' && !expandedKeys.has(key)) {
+              state.toggleKey(key);
+            }
+          }
+        },
```
LFDanLu, line 272: I can't seem to get this to trigger, not sure if regression or known?

packages/react-aria-components/src/Tree.tsx
```diff
@@ -194,11 +219,128 @@ function TreeInner<T extends object>({props, collection, treeRef: ref}: TreeInne
...
+        onKeyDown: e => {
+          if (e.key === 'ArrowRight' || e.key === 'ArrowLeft') {
+            // Toggle expansion when drop target is after expandable item
+            let target = dropState?.target;
+            if (target && target.type === 'item') {
```
LFDanLu, line 277: We may want to update this behavior in the future and make it possible for the user to keyboard navigate onto the row when drag and dropping and then accept "ArrowRight/ArrowLeft". Feels a bit confusing at the moment

packages/react-aria-components/src/Tree.tsx
```diff
@@ -397,6 +567,14 @@ export const TreeItem = /*#__PURE__*/ createBranchComponent('item', <T extends o
   // eslint-disable-next-line
   }, []);
 
+  let dragButtonRef = useRef<HTMLButtonElement>(null);
+  useEffect(() => {
+    if (dragState && !dragButtonRef.current && process.env.NODE_ENV !== 'production') {
+      console.warn('Draggable items in a Table must contain a <Button slot="drag"> element so that keyboard and screen reader users can drag them.');
```
LFDanLu, line 573:
```suggestion
      console.warn('Draggable items in a Tree must contain a <Button slot="drag"> element so that keyboard and screen reader users can drag them.');
```

LFDanLu (approved):
Approving for further testing. Issues can be investigated separately in a followup PR (I'll need to dive into the useTreeData for one of the aforementioned cases, unsure if it is a drop hooks issue or if the tree data is getting malformed)

packages/react-aria-components/src/Tree.tsx
```diff
@@ -128,7 +130,15 @@ export interface TreeProps<T> extends Omit<AriaTreeProps<T>, 'children'>, Multip
...
-  disabledBehavior?: DisabledBehavior
+  disabledBehavior?: DisabledBehavior,
+  /** The drag and drop hooks returned by `useDragAndDrop` used to enable drag and drop behavior for the Tree. */
+  dragAndDropHooks?: DragAndDropHooks,
+
+  /**
+   * An optional keyboard delegate implementation for type to select,
+   * to override the default.
+   */
+  keyboardDelegate?: KeyboardDelegate
```
devongovett, line 141: Do we need to expose this as a public prop? Doesn't look like we do in other collection components e.g. GridList

devongovett (approved):
let's get this in for testing

marchingband:
I am eagerly awaiting the next release to try this, as I was not able to locate any demo code for tree DND. Will some be made available with the release, and/or is there any available that I am missing? thank you!

reidbarber:
@marchingband We had a few outstanding bugs, so we ended up reverting this before the latest release. It should be in the next release though, and you can try it out in the nightlies now.

PR #7695, opened 2025-01-31
use useHref'ed href for router based navigation

jprosevear:
Closes https://github.com/adobe/react-spectrum/issues/5476#issuecomment-2296522896

This make require a lot of setup for testing, but generally speaking relative links in routers don't work because the useHref'ed value is not used for navigation.

Originally investigated in heroui-inc/heroui

## ✅ Pull Request Checklist:


packages/@react-aria/link/src/useLink.ts
```diff
@@ -87,10 +87,10 @@ export function useLink(props: AriaLinkOptions, ref: RefObject<FocusableElement
           // If props are applied to a router Link component, it may have already prevented default.
           !e.isDefaultPrevented() &&
           shouldClientNavigate(e.currentTarget, e) &&
-          props.href
+          routerLinkProps.href
         ) {
           e.preventDefault();
-          router.open(e.currentTarget, e, props.href, props.routerOptions);
+          router.open(e.currentTarget, e, routerLinkProps.href, props.routerOptions);
```
jprosevear, line 93: routerLinkProps.href is the useHref value, so that relative links are computed correctly.  For instance if you are under /blogs in React Router, useHref will compute "/blogs/1" if href="1"

jprosevear:
The test failures make sense broadly in that its not expecting to use the full path from useHref for navigation.

```
expect(navigate).toHaveBeenCalledWith('/base/foo', {foo: 'bar'});
```

would be the fix

packages/@react-aria/link/src/useLink.ts
jprosevear: An alternative might be to call it with e.currentTarget.href

jprosevear:
@devongovett i think you wrote the related code originally?

snowystinger:
Thanks for the PR, you'll need to sign the CLA https://react-spectrum.adobe.com/contribute.html#contributor-license-agreement
Also, if you could update the tests so they pass, it'd help with the review.

Thanks again!

jprosevear:
> Thanks for the PR, you'll need to sign the CLA https://react-spectrum.adobe.com/contribute.html#contributor-license-agreement Also, if you could update the tests so they pass, it'd help with the review.

That's not a problem, the key question is if the updated behaviour is correct or is the current behaviour "as designed" or something.

Also how to potentially handle host based links like https://example.com which are quasi-broken even now afaict.

jprosevear:
Ok, I went back to first creating stories that I believe demonstrate the problem for better discussion.  

First, you can see that the href prop on the link is incorrect (useHref constructs ).

image

https://github.com/adobe/react-spectrum/blob/a861ca4d8ecd6c5d3e5b42abf752a2a16c97f3f2/packages/%40react-aria/link/src/useLink.ts#L70 merges the useHref outcome into the props and that becomes the href prop on the link.

Second, the next_nested link has the correct href attribute because routerProps.href is merge, but the onclick handler uses the original href (props.href == "next_nested" at https://github.com/adobe/react-spectrum/blob/a861ca4d8ecd6c5d3e5b42abf752a2a16c97f3f2/packages/%40react-aria/link/src/useLink.ts#L93) instead of the instead of the useHref value ("nested/next_nested")

image

So the correct behavior is:

* if the navigation will be done by the router use routerProps.href as the path to router.open AND pass  routerProps.href as href from useLink
* if the navigation will not be done by the router, pass props.href back as href from useLink

the trick is how to detect navigation will be done by the router onside the onClick handler since shouldClientNavigate requires the dom element (however, perhaps parsing props.href and seeing if location.origin exists would be enough?)

@snowystinger @devongovett

devongovett:
This is a breaking change. Shouldn't the router be handling this in the `navigate` function? That should do the same resolution as `useHref`.

devongovett:
I think it might depend on where the `RouterProvider` is rendered in the tree. In the examples here, `useNavigate()` is called outside the `Route` component, so it won't know the nested route context. However, `useHref` is passed down and is called _inside_ the link component, so it _does_ know the nested route context. Hence the potential mismatch.

You could fix this by calling `useNavigate` and rendering `RouterProvider` inside each `<Route>`, i.e. inside your page component, instead of outside the whole router.

jprosevear:
Some findings:

1) React Router checks for absolute (external) vs non-absolute hrefs and handles them differently; https://github.com/remix-run/react-router/blob/3b45394fc7bd7c4aa58b802fda2558093217baca/packages/react-router/lib/dom/lib.tsx#L577

2) Both Tanstack Router and React Router are looking to prefetch links , Tanstack has a React Aria guide for building custom links https://tanstack.com/router/latest/docs/framework/react/guide/custom-link 

3) Tanstack Router requires explicitness around relative routes (need a "from" https://tanstack.com/router/v1/docs/framework/react/guide/navigation#everything-is-relative - and Aria doesn't have that)

4) Tanstack Router does not support external refs

One thought is to separate router links and non-router links in React Aria.

jprosevear:
So the Tanstack Link component does actually look for "external" 

```
  const type: 'internal' | 'external' = React.useMemo(() => {
    try {
      new URL(`${to}`)
      return 'external'
    } catch {}
    return 'internal'
  }, [to])
```

The buildLocation (https://tanstack.com/router/latest/docs/framework/react/api/router/RouterType#buildlocation-method)  router function in Tanstack Router however needs more than just "to" in order to navigate (ie  params for Dynamic links - https://tanstack.com/router/latest/docs/framework/react/guide/navigation#dynamic-links ) 

this would roughly be:

```
declare module 'react-aria-components' {
  interface RouterConfig {
    href: ToOptions['to'];
    routerOptions: Omit<NavigateOptions, 'to'>;
  }
}
```

but also requires router.useHref in React Aria to pass the routerOptions into useHref as well (not just the original href as per `useHref={(to) => router.buildLocation({ to }).href}`

packages/react-aria-components/stories/LinkTanstackRouter.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+  history
+});
+
+declare module '@tanstack/react-router' {
+  interface Register {
+    router: typeof router
+  }
+}
+
+declare module '@adobe/react-spectrum' {
```
jprosevear, line 57: The typing of this is not fully correct - build the buildLocation and navigate options do overlap quite a bit.  Also this module extension breaks the types in Link.stories.tsx because its global - however this is useful to temporarily check type safety.

packages/react-aria-components/stories/LinkTanstackRouter.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+
+const LinkTanstackRouterAriaProvider = () => {
+  const router = useRouter();
+
+  return (
+    <RouterProvider
+      navigate={(to, options) => router.navigate({to, ...options})}
+      useHref={(to, options) => router.buildLocation({to, ...options}).href}>
```
jprosevear, line 73: This is the change so that $dynamic gets replaced by React Aria.

jprosevear:
@devongovett updated approach, and possibly @schiller-manuel can help

jprosevear:
Next step might be to test Tabs or other components that support hrefs?

devongovett:
We're rethinking some routing-related things at the moment and will revisit this.

PR #7720, opened 2025-02-04
fix: prevent focus ring from appearing when selecting RAC Listbox on iPad

yihuiliao:
Found some issues on iPad in the Tailwind Examples on our docs and RAC Tailwind Storybook during testing.

Decided to increase the timeout to 80ms. It's not perfect by any means but the focus ring is less persistent than when the timeout is 40ms. I opted not to do 60ms because I could still reproduce the issue pretty frequently, and when I tried the same steps with 80ms, I found it much harder to reproduce the issue. 

I don't think people will notice a change with 80ms and it will significantly reduce the focus ring from appearing on iPad. 

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test using RAC examples:
https://reactspectrum.blob.core.windows.net/reactspectrum/9ef2f75ae1d6e0085304f7520ce34215dca605d3/docs/index.html
https://tinyurl.com/34d5kmeh
The RAC examples to test are "Contact List" and "Image Grid"


snowystinger (approved):

devongovett (approved):
Tested on my iPad running 17 in the docs and it looks good to me

PR #7727, opened 2025-02-05
feat: implemented details prop to the getItems function in the useClipboard hook

### LINKED ISSUE #7678: [Feature request] useClipboard: Add information to getItems so that the receiving onPaste can tell whether the user has cut or copied

I am using useClipboard inside our app to manage copying/pasting of custom components. In order to move our components properly around the app, we need to know whether the user is copying or cutting in the destination component, rather than in the source.

Currently, when using useClipboard, the DragItem (I think) provided by getItems does not contain any information about whether the user has copied or cut. You can access onCopy and onCut through useClipboard, and modify some external variable like a context or other state change, but you cannot use those event handlers to modify the actual data that gets sent. I could onCut delete the component and its various states, but due to how our state system is built, that requires an even more complex system.

It would be a lot simpler if that information was available to the receiving onPaste directly.

### 🤔 Expected Behavior?

useClipboard should provide a simple way for the receiving component to know whether the user initiated a copy or a cut. That way developers would have a simple solution for when they can't or don't want to immediately delete the source component on cut.

### 😯 Current Behavior

useClipboard provides onCopy and onCut, but these do not allow you to modify getItems. As a result, in order for the receiving component to know what action the user took, you currently must implement an external system to store that action somewhere else.

### 💁 Possible Solution

I posted a question about this [here](https://github.com/adobe/react-spectrum/discussions/7673), and a suggested solution from @reidbarber  was:

> We could probably support this by doing something like:

  `getItems?: (details: {type: 'cut' | 'copy'}) => DragItem[],`

> and passing the type for each when getItems is called.

### 🔦 Context

Our app has a bunch of custom components that can be moved around a board/map type interface. Copying and cutting each require independent functions, so we need to know which action the user took. onCopy and onCut inside useClipboard work, but I then have to store the result elsewhere.

To do this, I am simply accessing a context (copyContext in the example below) and storing it as a useRef value to prevent every component from rerendering when the user copies/cuts.

```
let copyContext = useContext(CopyContext);

let { clipboardProps } = useClipboard({
    getItems() {
        return [
          {
            "my-variable": "a random id string",
          },
        ];
    },
    onCopy() {
      if (copyContext) copyContext.current = "copy";
    },
    onCut() {
      if (copyContext) copyContext.current = "cut";
    }
  })

```

This is working fine but it is a band aid on an otherwise quite functional hook. A small change here would result in an improved developer experience.

### 💻 Examples

_No response_

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

Nevnet99:
Closes #7678

## ✅ Pull Request Checklist:

- [n/a] Looked at the Accessibility Practices for this feature - [Aria Practices](https://www.w3.org/WAI/ARIA/apg/)

## 📝 Test Instructions:

onItems change for the useClipboard hook now offers type in the onItems function this can be console.log()'d when the hook is used 

## 🧢 Your Project:

Nothing just looking to work on some open source :)

### DIFF
--- a/packages/@react-aria/dnd/src/useClipboard.ts
+++ b/packages/@react-aria/dnd/src/useClipboard.ts
@@ -18,7 +18,7 @@ import {useFocus} from '@react-aria/interactions';
 
 export interface ClipboardProps {
   /** A function that returns the items to copy. */
-  getItems?: () => DragItem[],
+  getItems?: (details: {type: "cut" | "copy"}) => DragItem[],
   /** Handler that is called when the user triggers a copy interaction. */
   onCopy?: () => void,
   /** Handler that is called when the user triggers a cut interaction. */
@@ -88,7 +88,7 @@ export function useClipboard(options: ClipboardProps): ClipboardResult {
 
     e.preventDefault();
     if (e.clipboardData) {
-      writeToDataTransfer(e.clipboardData, options.getItems());
+      writeToDataTransfer(e.clipboardData, options.getItems({type: 'copy'}));
       options.onCopy?.();
     }
   });
@@ -106,7 +106,7 @@ export function useClipboard(options: ClipboardProps): ClipboardResult {
 
     e.preventDefault();
     if (e.clipboardData) {
-      writeToDataTransfer(e.clipboardData, options.getItems());
+      writeToDataTransfer(e.clipboardData, options.getItems({type: 'cut'}));
       options.onCut();
     }
   });
--- a/packages/@react-aria/dnd/src/useClipboard.ts
+++ b/packages/@react-aria/dnd/src/useClipboard.ts
@@ -18,7 +18,7 @@ import {useFocus} from '@react-aria/interactions';
 
 export interface ClipboardProps {
   /** A function that returns the items to copy. */
-  getItems?: (details: {type: "cut" | "copy"}) => DragItem[],
+  getItems?: (details: {type: 'cut' | 'copy'}) => DragItem[],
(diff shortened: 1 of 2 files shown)


Nevnet99:
Any suggestions for the Storybook story ? looking into where to update the docs now ! :)

Nevnet99:
I've also signed the CLA after the workflow failed :)

snowystinger:
trying to get the cla to take

snowystinger:
I'm not sure why, maybe github is having some issues. Looks like you have a lint issue though, so maybe when that's addressed, the cla check will go through.

Nevnet99:
for the documentation it seems to auto gen from the types so I don't think I need to update any of the docs 
image

and ill take a look at the linting issue now :)

Nevnet99:
Should be all good I signed the CLA again incase I messed up any of the info might need to retrigger the workflow again in 10 or so mins

snowystinger:
Odd it's still not passing green. I know you double checked, but all I can think is that maybe the wrong e-mail was used for this PR compared to what you signed the CLA with? or your name/nickname maybe?

Nevnet99:
that'll be it my git config is using my works config damn, ill have a look at changing that

Nevnet99:
pushed an empty commit to trigger with the correct git user apologies

Nevnet99:
🎉  we got there in the end :)

snowystinger (approved):
Looks good, thanks for the tests

reidbarber (approved):
LGTM, thanks!

PR #7731, opened 2025-02-06
Do not build list of available slots when no error

jorrit:
An error was logged on my website because `Intl.ListFormat` was not available:

```
TypeError: undefined is not a constructor (evaluating 'new Intl.ListFormat()')
  at c(./node_modules/react-aria-components/dist/utils.mjs:76:1)
  at m(./node_modules/react-aria-components/dist/utils.mjs:87:1)
```

It seemed the site was accessed using Playwright, an automated browser. (https://github.com/microsoft/playwright, https://github.com/microsoft/playwright/issues/23978).

Out of curiousity, I checked which react-aria code was using `Intl.ListFormat` and it seemed that the code in utils that caused this should not have run, as the list is only needed in case of errors.

I refactored the code so the list of available slots is only compiled in case the requested slot is not available. This is a small speed optimization, as there is no longer a new instance of `Intl.ListFormat` on every render.

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

1. Find a browser with no support for `Intl.ListFormat`.
2. Create a component with subcomponents that use slots.
3. Use an valid slot name.
4. Notice that there is no error because missing `Intl.ListFormat`.

### DIFF
--- a/packages/react-aria-components/src/utils.tsx
+++ b/packages/react-aria-components/src/utils.tsx
@@ -160,15 +160,12 @@ export function useSlottedContext<T>(context: Context<SlottedContextValue<T>>, s
     return null;
   }
   if (ctx && typeof ctx === 'object' && 'slots' in ctx && ctx.slots) {
-    let availableSlots = new Intl.ListFormat().format(Object.keys(ctx.slots).map(p => `"${p}"`));
-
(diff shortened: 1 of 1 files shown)


snowystinger (approved):
thanks for the optimization!

reidbarber (approved):

PR #7741, opened 2025-02-07
chore: correct Table/Collection array types

### LINKED ISSUE #7656: Types are generally incompatible with ReadonlyArray

Props that take array's generally use the mutable form e.g. here the columns

https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/Table.tsx#L529

### 🤔 Expected Behavior?

They instead be typed to take ReadonlyArray where the array is not mutated - that then allows the consumer to pass a mutable or a readonly array.

### 😯 Current Behavior

An error trying to assign a readonly array to a prop that takes a mutable array

Image

### 💁 Possible Solution

Updating alot of `T[]` -> `ReadonlyArray<T>`

### 🔦 Context

We use readonly arrays frequently to make sure no-one is mutating a value they shouldn't and causing a unintended side effect. The current design means we would need to add a typescript expect error to use readonly arrays or a cast.

### 🖥️ Steps to Reproduce

N/A

### Version

N/A

Other

### If other, please specify.

_No response_

N/A

### 🧢 Your Company/Team

_No response_

### 🕷 Tracking Issue

_No response_

lukeapage:
Closes #7656

I used Iterable on the columns collection. For dependencies its used in a few places that require arrays specifically, so I updated these to ReadonlyArray (which means Array or ReadonlyArray is accepted - if it is Array then only Array is accepted because you are saying you might modify the array if it is not Readonly).

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

n/a - Fixing Types.

## 🧢 Your Project:

Saxo Bank (Though this account and contribution is on my own time)

### DIFF
--- a/packages/@react-aria/collections/src/useCachedChildren.ts
+++ b/packages/@react-aria/collections/src/useCachedChildren.ts
@@ -19,7 +19,7 @@ export interface CachedChildrenOptions<T> {
   /** The contents of the collection. */
   children?: ReactNode | ((item: T) => ReactNode),
   /** Values that should invalidate the item cache when using dynamic collections. */
-  dependencies?: any[],
+  dependencies?: ReadonlyArray<any>,
   /** A scope to prepend to all child item ids to ensure they are unique. */
   idScope?: Key,
   /** Whether to add `id` and `value` props to all child items. */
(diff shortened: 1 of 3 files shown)


reidbarber (approved):
Thanks for this!

PR #7786, opened 2025-02-19
fix: Avoid triggering collection updates when hovering over S2 TableView headers

devongovett:
Noticed while working on #7785 that hovering over the column headers in an S2 TableView caused the collection to update. That's because `TableColumn` was consuming a context that was set when the `TableHeader` was hovered. We can avoid this by showing/hiding the column dividers in CSS instead of via context.


snowystinger (approved):

yihuiliao (approved):

yihuiliao:
might be worth excluding the onResize control from the storybook bc it slows it down a lot. i did a similar thing in Slider since it made the movements very choppy otherwise

PR #7803, opened 2025-02-20
feat: custom calendar support

### LINKED ISSUE #3270: Allow Custom Business Calendars to the Datepicker/Calendar
# 🙋 Feature Request

There are a number of custom calendars supported by Analytics Workspace that are not supported by the react spectrum calendar controls.  These include:  
- 445
- 454
- 544
- NRF
- QRS
- Modified Gregorian 

Here's a wikipedia link to one of them that explains more:  https://en.wikipedia.org/wiki/4%E2%80%934%E2%80%935_calendar

Right now, the spectrum calendar only supports the Gregorian calendar, but we cannot use it in our app until it supports all the special calendars.

## 🤔 Expected Behavior

The most flexible way would be to allow the developer to specify the calendar details to the date range picker.  The second best way would be to have something like `calendar={"NRF"}` as one of the parameters.

## 😯 Current Behavior

Currently the calendar only supports the Gregorian calendar, which is not sufficient for a business analytics app. 

## 💁 Possible Solution

Analytics Workspace (and Report Builder) have written custom calendar controls that that the calendar type as an input and produce the correct calendar display.  You can see it in the report builder code base.

## 🔦 Context

Report builder allows customers to create reports within Microsoft Excel using their data from Analytics Workspace.  Some companies use custom calendars and we need to also use those calendars for determining the start and end dates of when we should gather the data.  We cannot do that using the react spectrum components.

## 💻 Examples

Here is an example of a custom calendar for the month of August
Screen Shot 2022-06-24 at 2 02 09 PM

You can see this calendar in action in Analytics Workspace by using the company OBU ENG SC and the report suite Geometrixx A4T

## 🧢 Your Company/Team

Adobe/Report Builder under the Client Apps team.

ToyWalrus:
Closes #3270 

# Custom calendar support

This PR...
- Exposes the `createCalendar` prop on `Calendar`, `RangeCalendar`, `DatePicker`, and `DateRangePicker` to allow for a custom calendar implementation
- Updates the `Calendar` interface with a new optional method to implement:
  - `getFormattableMonth()`
    - Used to change the label for month name and year in a calendar component
- Adds new stories for `Calendar`, `RangeCalendar`, `DatePicker`, and `DateRangePicker`
- Updates documentation for the changed components explaining how to create and use a custom calendar implementation

**This change is backwards compatible and will not break any existing calendars.**

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

I added a custom 4-5-4 calendar implementation that gets used in the added stories and test cases, at `@internationalized/date/tests/customCalendarImpl.js`. (_Side note: I don't know if there's a better place for this file to live, since I didn't see any sort of "test utils" files that were commonly shared_) This calendar was originally added to the Typescript calendar stories file in this PR's [commit history](d880b79e93d5d597afb5acecaf8494f9314d5f1f) but was later moved to a JS file in the test folder to be shared from there.

This custom calendar's correctness can be verified by using this [link to a custom 4-5-4 calendar](https://www5.an.adobe.com/sc15/settings/customize_calendar_preview.html?&type=1), or by comparing it to other such calendars found online.

## 🧢 Your Project:

Adobe, Analysis Workspace


devongovett:
Hey @ToyWalrus, thanks so much for working on this! I was wondering if we could possibly simplify the logic a little, and make it more similar to the other calendars. As you found, you needed to adjust the logic in quite a few places: arithmetic on months, comparing dates, etc.

Another potential way we could implement this is to number the dates as if they started from 1 on each month. So, you'd have February 1 to 28 as month number 1, March 1 to 35 as month number 2, etc. Then the logic for adding and subtracting months, and performing comparisons would work the same way as in other calendars. However, when _displaying_ the dates, and emitting the `onChange` event, they'd be converted to regular Gregorian dates. So, everything should appear the same to the user, but just be represented differently internally.

To implement that, I think you just need to implement `fromJulianDay` and `toJulianDay`, to convert to and from this numbering scheme. The remaining part is month naming, which will require a new method. I made a first pass at implementing this in [this branch](https://github.com/adobe/react-spectrum/compare/80dc84c248fe72b7c12f45e1236fe92021c5768c...custom-calendar-support) if you want to take a look. I'm on PTO the rest of this week, but I can pick this up again next week. Let me know what you think.

ToyWalrus:
@devongovett That's a different way of thinking about it that I hadn't considered, thanks for the insight! Yes, looking into it further it looks like this will be possible. I'll take your PoC branch work and update my code to follow that pattern. Doing things this way will definitely be less intrusive and touch fewer parts of the codebase, which is a plus.

devongovett:
I updated my branch to clean it up a bit and fix the tests. I am working on updating the documentation as well. Would be great if you could test this out and verify the correctness. Also whether it will work for your real use-cases.

packages/@internationalized/date/tests/customCalendarImpl.ts
```diff
@@ -0,0 +1,84 @@
+import {AnyCalendarDate, CalendarDate, GregorianCalendar} from '../src';
```
ToyWalrus, line 1: @devongovett This file is used throughout multiple stories and tests in several packages, I'm wondering if there's a better place to put it?

devongovett: I think this is ok for now

devongovett:
@ToyWalrus I see you pulled in some of my changes but not all. Could you summarize the differences and why you chose them?

ToyWalrus:
> @ToyWalrus I see you pulled in some of my changes but not all. Could you summarize the differences and why you chose them?

@devongovett Sure, so the main difference was that I had already made some of the same changes that you had by the time I saw that you had created your own PR with commits. These mainly had to do with cleaning up no-longer-used code and reverting previous changes, like the Calendar interface and alternate paths in the queries and manipulations file.
The other change I didn't pull in was the commit d74c6d6b697b577ed97d76a28a63fa5a5a065338, because while I like the idea of typing known string values, with this whole change being about adding support for new calendar "systems" it didn't seem like a good idea to lock in those existing identifier values without allowing the chance for a Calendar to have their own. Also I agree with @snowystinger's [comment](https://github.com/adobe/react-spectrum/pull/7949#discussion_r2000029855) where comparing the prototypes of the Calendar object could prove error-prone when there is more than one version of RSP on the page. Comparing string identifiers seemed much more straightforward.

Having said that, I'm not familiar with `Intl.DateTimeFormat` and how the identifier strings are used for it and where that might cause problems. Will it be a problem if an unknown string gets used for wherever `Intl.DateTimeFormat` is created?

I did use all your other commits though, for the documentation and examples.

ToyWalrus:
Also, for purely personal reasons, I'd like to have a contribution to an open-source library under my own name :)

devongovett:
Yeah, so my reasoning for https://github.com/adobe/react-spectrum/commit/d74c6d6b697b577ed97d76a28a63fa5a5a065338 is that right now we're relying on fallback to get the right formatting. For example, `useCalendarCell` calls `useDateFormatter({calendar: date.calendar.identifier})`, which means `calendar` gets set to `'custom-454'`. That is not a valid value for `Intl.DateTimeFormat`, so it falls back to the locale's default calendar (in our case Gregorian). But if this component were loaded in another country where a different calendar than Gregorian is the default, it would format in that calendar instead. That would result in unexpected behavior, because `'custom-454'` is supposed to be a derivative of Gregorian but would be displayed in a different calendar.

So the idea there was to keep `identifier` as a valid calendar identifier that can be formatted natively, so `date.calendar.identifier` would be `'gregory'` instead of `'custom-454'`. But then we need another way to compare calendar _objects_. I thought it would be pretty unlikely to call something like `isEqualDay` with two dates from _different @internationalized/date_ versions. There's already other `instanceof` checks in the library as well. But open to other ideas if you have them.

packages/@react-spectrum/datepicker/src/DatePicker.tsx
```diff
@@ -64,6 +64,9 @@ export const DatePicker = React.forwardRef(function DatePicker<T extends DateVal
   let {direction} = useLocale();
   let domRef = useFocusManagerRef(ref);
   let stringFormatter = useLocalizedStringFormatter(intlMessages, '@react-spectrum/datepicker');
+  if (props.createCalendar) {
+    (calendarProps as SpectrumCalendarProps<DateValue>).createCalendar = props.createCalendar;
+  }
```
devongovett, line 69: Would we ever want to pass `createCalendar` into the `DateField` as well? If so we might need different props for them. In the 4-5-4 case, the behavior is pretty weird if you pass it into DateField. But in other cases, maybe someone would want that?

ToyWalrus:
> I also added back the docs for `useCalendar` and `useRangeCalendar` (not sure why you removed those. Would be great if you could pull in those commits so we can merge this.

I removed the docs for the `useCalendar()` and `useRangeCalendar()` because those interfaces don't have the `createCalendar` prop on them

devongovett:
I believe `useCalendarState` and `useRangeCalendarState` already have the `createCalendar` option, and it is required there. https://github.com/adobe/react-spectrum/blob/d7511d723e697dae537728556a2f7caf14c59783/packages/%40react-stately/calendar/src/useCalendarState.ts#L39-L45

devongovett:
Yeah, React Aria depends on React Stately. In the [docs example](https://reactspectrum.blob.core.windows.net/reactspectrum/c960800373ad8cc68b26d0de7778996887e322f1/docs/react-aria/useCalendar.html#example) the component passes its props through to `useCalendarState`. Here's the working [custom calendar system](https://reactspectrum.blob.core.windows.net/reactspectrum/c960800373ad8cc68b26d0de7778996887e322f1/docs/react-aria/useCalendar.html#custom-calendar-systems) example.

devongovett:
The react aria docs serve as an example of actually building components using aria and stately. The stately docs just have prop tables, so not terribly useful on their own. The code examples are in aria. So I think it's important to show.

devongovett (approved):
Thanks for your work on this!

ToyWalrus:
Thanks!

snowystinger (approved):

LFDanLu (approved):

PR #7829, opened 2025-02-25
fix: Re-show focus ring on ComboBox input when closed

devongovett:
Autocomplete introduced a change for virtual focus collections where we fire focus/blur events to move the focus ring. This also affects ComboBox - now when the virtual focus is on an item, the input loses its focus ring. However, when closing the ComboBox, the focus ring never shifted back to the input. We also need to better account for this state in the styles for the v3 ComboBox.

### DIFF
--- a/packages/@react-aria/utils/src/useUpdateEffect.ts
+++ b/packages/@react-aria/utils/src/useUpdateEffect.ts
@@ -13,7 +13,7 @@
 import {EffectCallback, useEffect, useRef} from 'react';
 
 // Like useEffect, but only called for updates after the initial render.
-export function useUpdateEffect(effect: EffectCallback, dependencies: any[]): (() => void) | void {
+export function useUpdateEffect(effect: EffectCallback, dependencies: any[]): void {
   const isInitialMount = useRef(true);
   const lastDeps = useRef<any[] | null>(null);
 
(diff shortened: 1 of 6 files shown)


packages/@react-aria/utils/src/useUpdateEffect.ts
```diff
@@ -25,10 +25,11 @@ export function useUpdateEffect(effect: EffectCallback, dependencies: any[]): ((
...
 
   useEffect(() => {
+    let prevDeps = lastDeps.current;
     if (isInitialMount.current) {
       isInitialMount.current = false;
-    } else if (!lastDeps.current || dependencies.some((dep, i) => !Object.is(dep, lastDeps[i]))) {
-      return effect();
+    } else if (!prevDeps || dependencies.some((dep, i) => !Object.is(dep, prevDeps[i]))) {
+      effect();
     }
```
devongovett, line 33: This hook seemed to be entirely broken? `lastDeps[i]` never exists (should be `lastDeps.current[I]`). And the early return prevented `lastDeps.current` from being updated.

snowystinger (approved):
Looks pretty good, the one oddity I found was that clicking to open the combobox in v3 causes a thin (not keyboard focus) blue border around the input, but using the keyboard to open it, does not include either the thick or thin border, which seems a little counterintuitive

s2 is working great

devongovett:
hmm I think the `moveVirtualFocus` in `useSelectableItem` happens "too late" in ComboBox, after the `aria-activedescendant` has already updated. So we never fire blur/focus events because the previous and next focused element are already the same. Autocomplete doesn't have this problem because the active descendant is set _after_ we move focus instead of concurrently. 🤔

PR #7886, opened 2025-03-07
fix: Fix virtualizer drag and drop with layout gaps

devongovett:
When using drag and drop reorder items in a grid or list with gaps between the items, the drop target would disappear when over the gap because we were looking for items under the mouse and there were none. This was broken by the recent refactor. Now instead of using `keyAtPoint`, we get the items in a small rectangle that's as wide as the gap, and then find the closest item to the mouse position within that rectangle.

### DIFF
--- a/packages/@react-stately/layout/src/GridLayout.ts
+++ b/packages/@react-stately/layout/src/GridLayout.ts
@@ -102,6 +102,7 @@ export class GridLayout<T, O extends GridLayoutOptions = GridLayoutOptions> exte
     // Compute the number of rows and columns needed to display the content
     let columns = Math.floor(visibleWidth / (minItemSize.width + minSpace.width));
     let numColumns = Math.max(1, Math.min(maxColumns, columns));
+    this.numColumns = numColumns;
 
     // Compute the available width (minus the space between items)
     let width = visibleWidth - (minSpace.width * Math.max(0, numColumns));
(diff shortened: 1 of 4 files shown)


snowystinger (approved):

reidbarber (approved):
LGTM

packages/@react-stately/layout/src/GridLayout.ts
```diff
@@ -223,7 +224,46 @@ export class GridLayout<T, O extends GridLayoutOptions = GridLayoutOptions> exte
...
+      let candidates = this.getVisibleLayoutInfos(searchRect);
+      let minDistance = Infinity;
+      for (let candidate of candidates) {
+        // Ignore items outside the search rect, e.g. persisted keys.
+        if (!candidate.rect.intersects(searchRect)) {
+          continue;
+        }
+
+        let yDist = Math.abs(candidate.rect.y - x);
+        let maxYDist = Math.abs(candidate.rect.maxY - x);
```
LFDanLu, line 240: maybe I'm having a brain fart and not visualizing this quite right, but shouldn't this be `candidate.rect.y - y` instead? Like comparing the candidate's top/bottom edge's Y position with the current point to find the closest edge ->  reiterating on the next candidate to find the candidate that is closest to the drop point?

devongovett: https://github.com/adobe/react-spectrum/pull/7893

PR #7894, opened 2025-03-07
fix: Prevent submenus from closing on scroll

devongovett:
Fixes the "Complex content" submenu example in the docs on iPad. This was closing due to a scroll event when the virtual keyboard appeared. Close on scroll is enabled for `isNonModal` popovers, but we should exclude submenus from this. Another reason to eventually split up `isNonModal` into multiple props.

Also fixed the search field overflowing the popover on iPad in that example.


reidbarber (approved):
Fixed for me, but the menu briefly flashes as the keyboard opens.

devongovett:
I think that's an existing issue with usePreventScroll when there are text inputs in an overlay. In that case we apply a transform to the page to avoid browser scrolling, and I guess that causes the flash...

LFDanLu (approved):
LGTM, tested on iPad, Android Chrome, and desktop browsers

PR #7930, opened 2025-03-13
wip: S2 notification badge

yihuiliao:
Closes 

## ✅ Pull Request Checklist:


packages/@react-spectrum/s2/intl/en-US.json
```diff
@@ -17,6 +17,7 @@
   "label.(required)": "(required)",
   "label.(optional)": "(optional)",
   "menu.moreActions": "More actions",
+  "notificationbadge.plus": "+",
```
snowystinger, line 20: you'll need to do something like the line below for `tag.showAllButtonLabel`
you should be able to use https://formatjs.github.io/docs/intl#formatmessage as a reference for the API

But the end goal should be that a string comes out of this with a number and the plus sign integrated together.

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,146 @@
...
+  alignItems: 'center',
+  backgroundColor: 'accent',
+  height: {
+    size: {
+      S: {
+        default: 12,
+        isEmpty: 8
+      },
+      M: {
+        default: '[14px]',
```
snowystinger, line 65: these values won't scale, use `fontRelative`

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,146 @@
...
+  [props, ref] = useSpectrumContextProps(props, ref, NotificationBadgeContext);
+  let {
+    size = 'S',
+    value,
+    ...otherProps
+  } = props;
+  let domRef = useDOMRef(ref);
+  let {locale} = useLocale();
+
+  if (value && value < 0) {
```
snowystinger, line 103: can value be `0`? or is that synonymous with empty?

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,146 @@
...
+  if (value && value > 99) {
+    truncatedValue = 99;
+  }
+
+  let formattedValue = '';
+  if (truncatedValue) {
+    formattedValue = new NumberFormatter(locale).format(truncatedValue);
+  } else if (value) {
+    formattedValue = new NumberFormatter(locale).format(value);
+  }
```
snowystinger, line 117:
```suggestion
  let formattedValue = new NumberFormatter(locale).format(Math.min(value, 99));
```

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  let length = formattedValue.length;
```
snowystinger, line 119: Might be better to use math to figure out number of digits just in case a number is multibyte character at some point
https://stackoverflow.com/questions/14879691/get-number-of-digits-with-javascript

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,146 @@
...
+  let isEmpty = false;
+  let isSingleDigit = false;
+  let isDoubleDigit = false;
+  let isMaxDigit = false;
```
snowystinger, line 123: this is always false?

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,146 @@
...
+  if (!value) {
```
snowystinger, line 124:
```suggestion
  if (value == undefined) {
```

snowystinger: good habit to be in since 0 is falsy and everything else is truth
could also use isNaN, though search the codebase for which one to use, there's two of them

packages/@react-spectrum/s2/stories/NotificationBadge.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+  parameters: {
+    layout: 'centered'
+  },
+  tags: ['autodocs'],
+  title: 'NotificationBadge'
+};
+
+export default meta;
+
+export const Example = (args: any) => (
```
snowystinger, line 27: new stories should be written in story object format with a render property like in ActionButton.stories

Check ActionButton.stories for how to type the story object so that args doesn't need to be `any`

packages/@react-spectrum/s2/src/NotificationBadge.tsx
yihuiliao: based on the information in the figma design, i assumed that the value can't be 0. the way design described it is that it is used to indicate SINGLE or multiple unread notifications. in my interpretation, single means at least 1. 

i don't think that a value of 0 is synonymous with empty. 0 assumes that there are no notifications but a indicator with no value is used to imply that there are notifications (it might just be that there is not enough room to display the number)

packages/@react-spectrum/s2/src/NotificationBadge.tsx
yihuiliao: oh yeah i meant to get rid of that! leftover from when i trying some previous stuff

packages/@react-spectrum/s2/src/NotificationBadge.tsx
yihuiliao: maybe im just overthinking this but im not sure if this would work for characters like 四 (4). it's technically a number but it's also more so a character so a formula like `Math.log(number) * Math.LOG10E + 1 | 0` (as suggested in the post) wouldn't work.

i was thinking we use the length of the value after it's been formatted, otherwise, we make the assumption that something like the number 20 in another locale will always convert to a number that is also 2 in length

yihuiliao: maybe im overthinking this but im not sure if this would work with a character like 四 (4) with the formula suggested in the post `var length = Math.log(number) * Math.LOG10E + 1 | 0`

i was thinking we should use the length of the formatted value, otherwise, are we assuming that a number like 20 in another locale will always equal a formatted string with a length of 2?

snowystinger: We only support base 10 systems, so I believe that's a true statement. 20 requires one tens digit and one ones digit.

But maybe best to ask in the slack channel for g11n about the best way to determine this.

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,154 @@
...
+   * @default 'S'
+   */
+  size?: 'S' | 'M' | 'L' | 'XL'
+}
+
+export interface NotificationBadgeProps extends DOMProps, AriaLabelingProps, StyleProps, NotificationBadgeStyleProps, SlotProps {
+  /**
+   * The value to be displayed in the notification badge.
+   */
+  value?: number
```
snowystinger, line 39: `undefined` is a teeny bit weird, maybe it's better to make `value: number | null` so that it's explicitly empty? i could be persuaded either way. thoughts from the team?

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,154 @@
...
+    let length = Math.log(value <= 99 ? value : 99) * Math.LOG10E + 1 | 0;  // for positive integers (https://stackoverflow.com/questions/14879691/get-number-of-digits-with-javascript)
+    // or alternatively let length = formattedValue.length? depends on if we want to rely on the formatted value which is a string
+
+    if (length === 1) {
+      isSingleDigit = true;
+    } else if (length === 2) {
+      isDoubleDigit = true;
+    }
+
+    if (value > 99) {
```
snowystinger, line 139: does > 99 fall into the `isDoubleDigit` styling bucket?

snowystinger (commented):
Noticed VO isn't ready the notification count or the blank notification for the first two buttons in the story, only the second two

yihuiliao:
> Noticed VO isn't reading the notification count or the blank notification for the first two buttons in the story, only the second two

I think it's because the first two examples have an aria-label with "press me". I can update the aria-label

packages/@react-spectrum/s2/src/NotificationBadge.tsx
yihuiliao: hm im fine either way. i think it would be nice not having to explicitly type null if you want an empty notification, it's a little less work, but also i wouldn't be opposed to being have to be explicit about it

packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
```diff
@@ -198,3 +200,17 @@ export const Avatars: Story = {
     );
   }
 };
+
+export const NotificationBadges: Story = {
+  render: (args) => {
+    return (
+      <div style={{display: 'flex', gap: 8, padding: 8, justifyContent: 'center'}}>
+        <ActionButton aria-label="Press me" {...args}><NewIcon /><NotificationBadge /></ActionButton>
```
snowystinger, line 208: ah, i see. we still want the notification to be announced I'm pretty sure. so we should add some logic in ActionButton, if an aria-label is provided, then we need to add an aria-labelledby= self + notificationbadge

`useLabel` or `useLabels` should handle the logic for you, I forget which one off the top of my head

snowystinger (commented):
After playing with the announcements some more, I'm thinking we need to add something to them just like we want for the null/empty case, because reading "Messages 5" doesn't really tell me much.
Maybe it makes the most sense for a user to supply it and have aria-label be required?
Then in our story we'd do something like
```
<ActionButton {...args}><CommentIcon /><Text>Messages</Text><NotificationBadge value={5} aria-label="new" /></ActionButton>
```
So that it'd read "Messages 5 new"

Can you bring this up in standup?

Also, right now it doesn't actually read "Messages 5", it just reads "5", so we'll need to fix that. You can probably use the context to pass an id to Text as well if that's the reason for it. Check in the Chrome accessibility tab for more information.

yihuiliao:
> Can you bring this up in standup?

Seems like we're okay with how it is now (once I get it to actually announce the text inside the button). Just make sure we have a default label on the visual indicator one.

> Also, right now it doesn't actually read "Messages 5", it just reads "5", so we'll need to fix that. You can probably use the context to pass an id to Text as well if that's the reason for it. Check in the Chrome accessibility tab for more information.

i swear it worked...but will look into it. must've changed something...

yihuiliao:
yeah, happy to take a look at other options and see if we can improve the VO support. i'll explore the options a bit more tomorrow hopefully

packages/@react-spectrum/s2/intl/en-US.json
```diff
@@ -17,6 +17,8 @@
   "label.(required)": "(required)",
   "label.(optional)": "(optional)",
   "menu.moreActions": "More actions",
+  "notificationbadge.plus": "{notifications}+",
+  "notificationbadge.indicatorOnly": "New activity",
```
yihuiliao, line 21: based off of designs suggestion. they did mention that it is very context specific but for general guidance they suggested:
- New notification
- New activity
- New update available

i just chose one of them but we can discuss which one we like best as a default label

packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
```diff
@@ -198,3 +201,19 @@ export const Avatars: Story = {
...
+
+export const NotificationBadges: Story = {
+  render: (args) => {
+    let badgeValue = 10;
+    let formattedValue = new NumberFormatter('en-US').format(badgeValue);
+
+    return (
+      <div style={{display: 'flex', gap: 8, padding: 8, justifyContent: 'center'}}>
+        <ActionButton aria-label="Messages has new activity" {...args}><CommentIcon /><NotificationBadge /></ActionButton>
+        <ActionButton aria-label={`${formattedValue} notifications`} {...args}><BellIcon /><NotificationBadge value={badgeValue} /></ActionButton>
```
yihuiliao, line 213: are people okay with this story? tried to show how you might want to format the number when using it in an aria-label. the hard coding of the locale was because i ran into some storybook stuff and figured it would be easier this way.

packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
```diff
@@ -203,14 +204,16 @@ export const Avatars: Story = {
 
 export const NotificationBadges: Story = {
   render: (args) => {
+    let badgeValue = 10;
+    let formattedValue = new NumberFormatter('en-US').format(badgeValue);
```
snowystinger, line 208: we should use https://react-spectrum.adobe.com/react-aria/useNumberFormatter.html#usenumberformatter so it picks up the locale automatically

packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
snowystinger: general idea, yes, but let's figure out why the hook didn't work, happy to look on monday

packages/@react-spectrum/s2/stories/NotificationBadge.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <NotificationBadge {...args} />
+        <NotificationBadge {...args} value={1} />
+        <NotificationBadge {...args} value={24} />
+        <NotificationBadge {...args} value={100} />
+      </div>
+    );
+  }
+};
+
+export const AllValues: Story = {
```
snowystinger, line 41: I think we can omit this story
if it's useful, then please add a description of what I should be looking at/for

packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
yihuiliao: oh it's a long story but basically, ran into linting issues with useLocale because it was inside the render function. so then i moved it outside of that but that messed up the "show code". tried to fix that but then it wasn't really happy with the `${formattedValue} notifications` part which is the exact part that i'm trying to show. i would have to change it to `{formattedValue} notifications` but since people might copy our code and expect it to work, that wouldn't be too great because it wouldn't work

yihuiliao: nvm...i've figured it out

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+export const NotificationBadgeContext = createContext<ContextValue<Partial<NotificationBadgeProps>, DOMRefValue<HTMLDivElement>>>(null);
+
+const badge = style({
+  display: 'flex',
+  font: 'control',
+  color: {
+    default: 'white',
+    forcedColors: 'ButtonText'
+  },
```
LFDanLu, line 50: thinking out loud, wonder if the text and/or the background color could/should follow the same kind of static background color logic that the action buttons have since it can be hard to see the number against certain backgrounds

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -0,0 +1,160 @@
...
+      L: {
+        default: 16,
+        isIndicatorOnly: fontRelative(12)
+      },
+      XL: {
+        default: 18,
+        isIndicatorOnly: fontRelative(12)
+      }
+    }
+  },
```
LFDanLu, line 93: I'm guessing there weren't tokens for these? Tried looking for the designs myself but couldn't find anything other than the very high level badge ones haha

packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
```diff
@@ -198,3 +201,41 @@ export const Avatars: Story = {
...
+        let formattedValue = useNumberFormatter().format(badgeValue);
+      
+        return (
+          <div style={{display: 'flex', gap: 8, padding: 8, justifyContent: 'center'}}>
+            <ActionButton aria-label="Messages has new activity" {...props}><CommentIcon /><NotificationBadge /></ActionButton>
+            <ActionButton aria-label={\`\${formattedValue} notifications\`}  {...props}><BellIcon /><NotificationBadge value={badgeValue} /></ActionButton>
+            <ActionButton {...props}><CommentIcon /><Text>Messages</Text><NotificationBadge value={5} /></ActionButton>
+            <ActionButton {...props}><Text>Notifications</Text><NotificationBadge value={105} /></ActionButton>
+          </div>
+        )`;
```
LFDanLu, line 237: you might need to new line the `let`? Getting this
image

packages/@react-spectrum/s2/src/ActionButton.tsx
```diff
@@ -281,6 +311,10 @@ export const ActionButton = forwardRef(function ActionButton(props: ActionButton
           [AvatarContext, {
             size: avatarSize[size],
             styles: style({marginStart: '--iconMargin', flexShrink: 0, order: 0})
+          }],
+          [NotificationBadgeContext, {
+            size: props.size === 'XS' ? undefined : props.size,
+            styles: style({position: 'absolute', top: '--badgeTop', insetStart: '[var(--badgePosition)]', marginTop: '[calc((self(height) * -1)/2)]', marginStart: '[calc((self(height) * -1)/2)]'})
```
LFDanLu, line 317: I wonder if there should be disabled styles as well if the ActionButton is disabled but has a notification. Maybe the notification shouldn't even appear? Also if the ActionButton is quiet and only has text, the badge feels a bit far from the text IMO (though maybe it is obvious that it is still associated in real world cases), might be worth bringing up with design

LFDanLu (approved):
LGTM, the announcements sound good to me in Talkback/NVDA/VO. Overall styling looks good, but couldn't find any super detailed designs/tokens so will want to watch out for those if they become a thing.

Will also need translations, so we'll need to remember to request those in advance of release when this goes in

packages/@react-spectrum/s2/intl/en-US.json
LFDanLu: What you have feel like fine defaults, wonder how `+` gets localized though

packages/@react-spectrum/s2/src/ActionButton.tsx
snowystinger: good catch on quiet, that should definitely be brought up with design

snowystinger (approved):
Happy to have the comments be follow up. Given we're aiming for this to make it to next release, if you decide to merge now, please make a ticket with all remaining tasks so we can all help out if need be

packages/@react-spectrum/s2/intl/en-US.json
devongovett: We should make sure to specifically ask about how this `+` string should be localized, and make sure it's clear for the translators.

packages/@react-spectrum/s2/src/ActionButton.tsx
devongovett: Might be a little simpler to do `translate: '[-50%']` here instead of using margins. Percentages in translate are relative to the size.

packages/@react-spectrum/s2/src/ActionButton.tsx
```diff
@@ -216,7 +218,34 @@ export const btnStyles = style<ButtonRenderProps & ActionButtonStyleProps & Togg
   zIndex: {
     isFocusVisible: 2
   },
-  disableTapHighlight: true
+  disableTapHighlight: true,
+  '--badgeTop': {
+    type: 'top',
+    value: {
+      default: '[calc(self(height)/2 - var(--iconWidth)/2)]',
+      [iconOnly]: '[calc(self(height)/2 - var(--iconWidth)/2)]',
```
devongovett, line 226: Same as default? Is this case needed?

devongovett (commented):
We might need to ask design how these should look in other button states:

* In quiet, they are just floating - not attached to the button because there is no background
* When disabled, they remain bright blue
* In staticColor variants they don't change color and may not be visible

packages/@react-spectrum/s2/src/ActionButton.tsx
yihuiliao: hm we could maybe. i think the issue is mainly:
1) translate doesn't exist on styles. we'd have to use UNSAFE_style which i'd like to avoid 

2) probably the bigger issue of the two. the way that the indicator is positioned is reliant on its height because its constant. the width of the badge is variable. the way i think of it, each badge has an invisible circle along the rounded corner, and we want the corner of the svg to align with the center of this circle. if we were to use translateX(-50%), it'd position itself such that the corner of the svg is aligned with the middle of the whatever the width of the badge is. maybe that'd help with the visual collisions though so i guess we can see what design says about that

PR #7938, opened 2025-03-14
feat: Async loading support for S2 ComboBox/Picker

LFDanLu:
Closes 

- Adds loading support for S2 ComboBox/Picker and their RAC equivalents (and RAC ListBox too since that is what ComboBox and Picker uses 
- refactors useLoadMore to accept collections instead of items

## ✅ Pull Request Checklist:

## 📝 Test Instructions:

Test S2 Combobox/Picker async loading stories and make sure it works as expected (aka infinite scrolling). Also test the RAC equivalents as well as S2 CardView/Table/other components that already had async loading support.

## 🧢 Your Project:

RSP


packages/@react-aria/utils/src/useLoadMore.ts
LFDanLu, line 1: Make a new hook since util package version might not match up with pinned version of tableview/etc, which makes this a breaking change

packages/react-aria-components/src/ComboBox.tsx
```diff
@@ -47,10 +47,16 @@ export interface ComboBoxRenderProps {
...
    * @selector [data-required]
    */
-  isRequired: boolean
+  isRequired: boolean,
+  // TODO: do we want loadingState for RAC Combobox or just S2
+  /**
+   * Whether the combobox is currently loading items.
+   * @selector [data-loading]
+   */
+  isLoading?: boolean
```
LFDanLu, line 56: For now see about moving `isLoading` to the loading spinner? Otherwise, just make sure the followup refactor goes in before next release

packages/react-aria-components/src/ListBox.tsx
```diff
@@ -230,6 +235,21 @@ function ListBoxInner<T extends object>({state: inputState, props, listBoxRef}:
     );
   }
 
+  let sentinelRef = useRef(null);
+  // TODO: Should scrollOffset for useLoadMore should be configurable by the user
```
LFDanLu, line 239: expose on sentinel

packages/react-aria-components/src/ListBox.tsx
```diff
@@ -230,6 +235,21 @@ function ListBoxInner<T extends object>({state: inputState, props, listBoxRef}:
...
+    collection,
+    sentinelRef
+  }), [props.isLoading, props.onLoadMore, collection]);
+  // TODO: maybe this should be called at the ListBox level and the StandaloneListBox level. At its current place it is only called
+  // when the Listbox in the dropdown is rendered. The benefit to this would be that useLoadMore would trigger a load for the user before the
+  // dropdown opens but the current state gives the user more freedom as to whether they would like to pre-fetch or not
+  useLoadMore(memoedLoadMoreProps, listBoxRef);
+
+  // TODO: Think about if completely empty state. Do we leave it up to the user to setup the two states for empty and empty + loading?
+  // Do we add a data attibute/prop/renderprop to ListBox for isLoading
```
LFDanLu, line 252: Keep both, will be clearer with refactor to loadingSentinel

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -147,186 +159,222 @@ const iconStyles = style({
...
+      L: 'ui-lg',
+      XL: 'ui-xl'
+    }
+  },
+  display: 'flex',
+  alignItems: 'center',
+  paddingStart: 'edge-to-text'
+});
+
+export let listbox = style<{size: 'S' | 'M' | 'L' | 'XL'}>({
```
LFDanLu, line 207: Here to LOADER_ROW_HEIGHTS is from https://github.com/adobe/react-spectrum/pull/8110/files

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -383,11 +431,333 @@ export function ComboBoxSection<T extends object>(props: ComboBoxSectionProps<T>
...
         {props.children}
       </AriaListBoxSection>
-      <Divider />
+      <Divider size={size} />
+    </>
+  );
+}
+
+// TODO: not quite sure why typescript is complaining when I types this as T extends object
+const ComboboxInner = forwardRef(function ComboboxInner(props: ComboBoxProps<any> & {isOpen: boolean}, ref: ForwardedRef<TextFieldRef | null>) {
```
LFDanLu, line 443: Mainly a reorg of code + adding loader logic/spinner rendering, for the most part most of this is the same as before

packages/react-aria-components/src/ListBox.tsx
LFDanLu: At the moment, `renderEmptyState` is meant for rendering the empty state AND the initial load spinner, whereas the LoadingSentinel is more for the "loadingMore" spinner

packages/@react-aria/utils/src/useLoadMoreSentinel.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface LoadMoreSentinelProps extends AsyncLoadable {
+  /**
+   * The amount of offset from the bottom of your scrollable region that should trigger load more.
+   * Uses a percentage value relative to the scroll body's client height. Load more is then triggered
+   * when your current scroll position's distance from the bottom of the currently loaded list of items is less than
+   * or equal to the provided value. (e.g. 1 = 100% of the scroll region's height).
+   * @default 1
+   */
+  scrollOffset?: number
+  // TODO: Maybe include a scrollRef option so the user can provide the scrollParent to compare against instead of having us look it up
```
LFDanLu, line 28: leaning towards not adding it for now, the hook is unstable as of now so we could always add it later. Open to opinions if anyone feels strongly about the approach

packages/@react-aria/utils/src/useLoadMoreSentinel.ts
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  useLayoutEffect(() => {
+    if (ref.current) {
+      // TODO: problem with S2's Table loading spinner. Now that we use the isLoading provided to the sentinel in the layout to adjust the height of the loader,
+      // we are getting space reserved for the loadMore spinner when doing initial loading and rendering empty state at the same time. We can somewhat fix this by providing isLoading={loadingState === 'loadingMore'}
+      // which will mean the layout won't reserve space for the loader for initial loads, but that breaks the load more behavior (specifically, auto load more to fill scrollOffset. Scroll to load more seems broken to after initial load).
+      // We need to tear down and set up a new IntersectionObserver to force a check if the sentinel is "in view", see  https://codesandbox.io/p/sandbox/magical-swanson-dhgp89?file=%2Fsrc%2FApp.js%3A21%2C21
+      // I've actually fixed this via a ListLayout change (TableLayout extends this) where I check "collection?.size === 0 || (collection.size === 1 && collection.getItem(collection.getFirstKey()!)!.type === 'loader')"
+      // as well as isLoading, but it feels pretty opinionated/implementation specifc
```
LFDanLu, line 53: A bit unfortunate that fixing this involved layout level changes. Also a bit unfortunate that we can't just create single instance of a IntersectionObserver upfront, but we need to do this setup/teardown every time isLoading changes so that we can properly trigger the loadMore if the sentinel remains within the rootMargin after the items are loaded.

packages/@react-spectrum/s2/intl/ar-AE.json
```diff
@@ -6,6 +6,7 @@
   "actionbar.selectedAll": "تم تحديد الكل",
   "breadcrumbs.more": "المزيد من العناصر",
   "button.pending": "قيد الانتظار",
+  "combobox.noResults": "لا توجد نتائج",
```
LFDanLu, line 9: From S1

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -383,11 +431,331 @@ export function ComboBoxSection<T extends object>(props: ComboBoxSectionProps<T>
...
+  }, []);
+
+  let renderer;
+  let listBoxLoadingCircle = (
+    <UNSTABLE_ListBoxLoadingSentinel
+      // TODO: we only want the listbox loading sentinel row to have a height when performing a loadingMore
+      // Unlike table, this works because the listbox's initial load isn't triggerd by the sentintel and thus the first
+      // observation occurs when we've already loaded our first set of items rather than starting from empty, therefor flipping
+      // isLoading here from true to false and triggering the creation of a new InterserctionObserver
+      isLoading={loadingState === 'loadingMore'}
```
LFDanLu, line 565: An interesting facet about how useLoadMoreSentinel works. Prior to some additional changes in layout, virtualized TableView ran into issues triggering additional `loadMore` calls when it was passing `isLoading={loadingState === 'loadingMore'}` to its own sentinel. This is because the sentinel never got a prop change before and after the initial load finished because the `loadingState` at that point was `loading` and thus we never caused a second `shouldLoadMore` check to happen in `useLoadMoreSentinel`

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -383,11 +431,331 @@ export function ComboBoxSection<T extends object>(props: ComboBoxSectionProps<T>
...
+              }]
+            ]}>
+            <Virtualizer
+              layout={ListLayout}
+              layoutOptions={{
+                estimatedRowHeight: 32,
+                // TODO: can we get rid of this and instead handle padding else where other than the layout options? Kinda gross that we need to do this check.
+                // Perhaps can consider only applying padding if the collection actually has content since the renderEmptyState content is outside the virtualizer
+                // Otherwise could consider moving renderEmptyState into the collection...
+                padding: isEmptyOrLoading ? 0 : 8,
```
LFDanLu, line 700: we actually don't want any extra padding (or have the virtualizer div render anything really) so that the "No results"/isLoading spinner dictate the height of the the popover. This check is a bit gross but not really sure there is a good way to do this otherwise.

packages/@react-spectrum/s2/src/TableView.tsx
```diff
@@ -197,7 +196,9 @@ export class S2TableLayout<T> extends TableLayout<T> {
     let {children, layoutInfo} = body;
     // TableLayout's buildCollection always sets the body width to the max width between the header width, but
     // we want the body to be sticky and only as wide as the table so it is always in view if loading/empty
-    if (children?.length === 0) {
+    // TODO: we may want to adjust RAC layouts to do something simlar? Current users of RAC table will probably run into something similar
+    let isEmptyOrLoading = children?.length === 0 || (children?.length === 1 && children[0].layoutInfo.type === 'loader');
+    if (isEmptyOrLoading) {
       layoutInfo.rect.width = this.virtualizer!.visibleRect.width - 80;
     }
```
LFDanLu, line 203: One of the several places that I had to change from our prior assumption of "empty = collection doesn't have children". Quite specific to a virtualized table but might be useful to document or include by default?

packages/@react-spectrum/s2/src/TableView.tsx
```diff
@@ -377,18 +374,24 @@ export interface TableBodyProps<T> extends Omit<RACTableBodyProps<T>, 'style' |
...
   let domRef = useDOMRef(ref);
-  let {loadingState} = useContext(InternalTableContext);
+  let {loadingState, onLoadMore} = useContext(InternalTableContext);
+  let isLoading = loadingState === 'loading' || loadingState === 'loadingMore';
   let emptyRender;
   let renderer = children;
   let stringFormatter = useLocalizedStringFormatter(intlMessages, '@react-spectrum/s2');
+  // TODO: still is offset strangely if loadingMore when there aren't any items in the table, see http://localhost:6006/?path=/story/tableview--empty-state&args=loadingState:loadingMore
+  // This is because we don't distinguish between loadingMore and loading in the layout, resulting in a different rect being used to build the body. Perhaps can be considered as a user error
+  // if they pass loadingMore without having any other items in the table. Alternatively, could update the layout so it knows the current loading state.
```
LFDanLu, line 384: Specifically, this is a case where you have a empty table and pass `isLoadingMore` as your loading state. The `loadingMore` spinner row won't be built by the layout since the layout thinks you are performing initial load so in the story linked the body is actually just blank. IMO user error but otherwise we need more information in the layout to differentiate the two cases

packages/@react-stately/layout/src/ListLayout.ts
```diff
@@ -315,16 +329,23 @@ export class ListLayout<T, O extends ListLayoutOptions = ListLayoutOptions> exte
...
+    let collection = this.virtualizer!.collection;
+    let isEmptyOrLoading = collection?.size === 0 || (collection.size === 1 && collection.getItem(collection.getFirstKey()!)!.type === 'loader');
     let rect = new Rect(x, y, this.padding, 0);
     let layoutInfo = new LayoutInfo('loader', node.key, rect);
     rect.width = this.virtualizer!.contentSize.width - this.padding - x;
-    rect.height = this.loaderHeight || this.rowHeight || this.estimatedRowHeight || DEFAULT_HEIGHT;
+    // TODO: Kinda gross but we also have to differentiate between isLoading and isLoadingMore so that we dont'reserve room
+    // for the loadMore loader row when we are performing initial load. Is this too opinionated? Note that users creating their own layouts
+    // may need to perform similar logic
+    rect.height = node.props.isLoading && !isEmptyOrLoading ? this.loaderHeight ?? this.rowHeight ?? this.estimatedRowHeight ?? DEFAULT_HEIGHT : 0;
```
LFDanLu, line 348: As mentioned previously, we don't want to render the `loadingMore` progress spinner row when performing initial load otherwise we get two spinners and/or extra room rendered above the `isLoading` spinner

packages/react-aria-components/src/GridList.tsx
```diff
@@ -194,9 +194,11 @@ function GridListInner<T extends object>({props, collection, gridListRef: ref}:
   }
 
   let {focusProps, isFocused, isFocusVisible} = useFocusRing();
+  // TODO: What do we think about this check? Ideally we could just query the collection and see if ALL node are loaders and thus have it return that it is empty
+  let isEmpty = state.collection.size === 0 || (state.collection.size === 1 && state.collection.getItem(state.collection.getFirstKey()!)?.type === 'loader');
```
LFDanLu, line 198: Will need to update this logic (and other areas where I've made the assumption that the loader is the last element in the collection) when we move to supporting tree multi section loading

packages/react-aria-components/src/Virtualizer.tsx
```diff
@@ -102,7 +102,9 @@ function CollectionRoot({collection, persistedKeys, scrollRef, renderDropIndicat
     onScrollEnd: state.endScrolling
   }, scrollRef!);
 
-  if (state.contentSize.area === 0) {
+  // TODO: wull have to update this when multi section loading is implemented, will need to check if all items in a collection are loaders instead perhaps
+  let hasLoadingSentinel = collection.size === 1 && collection.getItem(collection.getFirstKey()!)!.type === 'loader';
+  if (state.contentSize.area === 0 && !hasLoadingSentinel) {
```
LFDanLu, line 107: This will change with https://github.com/adobe/react-spectrum/pull/8110/files#diff-6026043a183382ab1099d5cc5c47d7028cef57057d81b9dfc55acfba173e6d0f, can remove this logic then

packages/@react-aria/collections/src/Document.ts
```diff
@@ -103,7 +103,7 @@ export class BaseNode<T> {
   }
 
   private invalidateChildIndices(child: ElementNode<T>): void {
-    if (this._minInvalidChildIndex == null || child.index < this._minInvalidChildIndex.index) {
+    if (this._minInvalidChildIndex == null || !this._minInvalidChildIndex.isConnected || child.index < this._minInvalidChildIndex.index) {
```
LFDanLu, line 106: I found cases where the minInvalidChildIndex set on the Document was pointing to nodes that didn't exist in the collection any more so added this check so we'd always have an up to date one

packages/@react-aria/collections/src/Document.ts
```diff
@@ -470,6 +473,12 @@ export class Document<T, C extends BaseCollection<T> = BaseCollection<T>> extend
     }
 
     // Next, update dirty collection nodes.
+    // TODO: when updateCollection is called here, shouldn't we be calling this.updateChildIndicies as well? Theoretically it should only update
+    // nodes from _minInvalidChildIndex onwards so the increase in dirtyNodes should be minimal.
+    // Is element.updateNode supposed to handle that (it currently assumes the index stored on the node is correct already).
+    // At the moment, without this call to updateChildIndicies, filtering an async combobox doesn't actually update the index values of the
+    // updated collection...
+    this.updateChildIndices();
```
LFDanLu, line 481: This change along with the above were mainly to fix an issue I noticed where the collection indexes weren't updating when filtering the combobox. Above, we properly update the `childIndices` of the Document's dirty nodes (aka the children's children) but not for the main children. Not sure if the call to `element.updateNode` below is supposed to handle this but this felt the most correct

packages/react-aria-components/test/ListBox.test.js
```diff
@@ -1321,11 +1324,225 @@ describe('ListBox', () => {
...
+
+        let sentinel = within(loaderRow.parentElement).getByTestId('loadMoreSentinel');
+        expect(sentinel.parentElement).toHaveAttribute('inert');
+      });
+
+      // TODO: for some reason this tree renders empty if ran with the above test...
+      // Even if the above test doesn't do anything within it, the below tree won't render with content until the above test
+      // is fully commented out (aka even the it(...))
+      // It thinks that the contextSize is 0 and never updates
+      it.skip('should not reserve room for the loader if isLoading is false', () => {
```
LFDanLu, line 1506: Not quite sure why this is happening... As stated, the below test only renders the listbox items if the above test is completely commented out. Even if the above test doesn't have any contents (aka just an empty `it()`) this test still fails to render the ListBox content

packages/@react-stately/select/src/useSelectState.ts
```diff
@@ -15,7 +15,8 @@ import {FormValidationState, useFormValidationState} from '@react-stately/form';
 import {OverlayTriggerState, useOverlayTriggerState} from '@react-stately/overlays';
 import {SelectProps} from '@react-types/select';
 import {SingleSelectListState, useSingleSelectListState} from '@react-stately/list';
-import {useState} from 'react';
+import {useEffectEvent} from '@react-aria/utils';
```
snowystinger, line 18: can't import from react-aria in a react-stately package

packages/@react-aria/collections/src/Document.ts
```diff
@@ -154,8 +154,11 @@ export class BaseNode<T> {
 
     newNode.nextSibling = referenceNode;
     newNode.previousSibling = referenceNode.previousSibling;
-    newNode.index = referenceNode.index;
-
+    // Ensure that the newNode's index is less than that of the reference node so that
+    // invalidateChildIndices will properly use the newNode as the _minInvalidChildIndex, thus making sure
+    // we will properly update the indexes of all sibiling nodes after the newNode. The value here doesn't matter
+    // since updateChildIndices should calculate the proper indexes.
+    newNode.index = referenceNode.index - 1;
```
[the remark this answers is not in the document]
LFDanLu: Said previous sibling may not exist and either way updateChildIndicies will perform the calculation to generate the proper index I believe.

packages/@react-aria/virtualizer/src/ScrollView.tsx
```diff
@@ -209,7 +209,11 @@ export function useScrollView(props: ScrollViewProps, ref: RefObject<HTMLElement
       // https://github.com/reactwg/react-18/discussions/102
       // @ts-ignore
       if (typeof IS_REACT_ACT_ENVIRONMENT === 'boolean' ? IS_REACT_ACT_ENVIRONMENT : typeof jest !== 'undefined') {
-        updateSize(fn => fn());
+        // This is so we update size in a separate render but within the same act. Needs to be setState instead of refs
```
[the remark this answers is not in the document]
LFDanLu: This is to specifically fix the tests where the order in which things run (such as when the ref is connected to the ScrollView and when we perform rect measurements) vs how focus is moved by simulated clicks/presses from userEvent was causing odd behaviors (i.e. focus would move to the trigger button -> the dropdown -> back to the button specifically due to the order in which the dropdown becomes available/focusable + `preventFocus`'s logic thinking it should then restore focus to the button vs how it happens in the browser with the dropdown becoming available after that `preventFocus` logic runs). 

The only way to fix this was to make sure the size of the ScrollView updates in a separate render but within the same act, but ideally the high level flow should be that focus moves to the Picker trigger on click -> all press handling finishes -> collection is fully formed -> dropdown opens and focus moves into it.

packages/@react-spectrum/s2/chromatic/Combobox.stories.tsx
```diff
@@ -69,3 +71,90 @@ export const WithCustomWidth = {
...
+  play: async ({canvasElement}) => {
+    await userEvent.tab();
+    await userEvent.keyboard('{ArrowDown}');
+    let body = canvasElement.ownerDocument.body;
+    let listbox = await within(body).findByRole('listbox');
+    await within(listbox).findByText('No results');
+  }
+};
+
+// TODO: this one is probably not great for chromatic since it has the spinner, check if ignoreSelectors works for it
```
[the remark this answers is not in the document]
LFDanLu: ah, I forgot to actually run it against chromatic, just checked against the local chromatic build, thanks for the reminder

packages/@react-aria/utils/src/useLoadMoreSentinel.ts
LFDanLu: the ref that is provided is the sentinel ref so not quite the same, but the general idea would be for the case you mentioned. Figured we could omit it until we actually try that use case

packages/@react-stately/select/src/useSelectState.ts
```diff
@@ -62,25 +63,28 @@ export function useSelectState<T extends object>(props: SelectStateOptions<T>):
   });
 
   let [isFocused, setFocused] = useState(false);
+  let isEmpty = useMemo(() => listState.collection.size === 0 || (listState.collection.size === 1 && listState.collection.getItem(listState.collection.getFirstKey()!)?.type === 'loader'), [listState.collection]);
+  let open = useEffectEvent((focusStrategy: FocusStrategy | null = null) => {
```
devongovett, line 67: why?

packages/react-aria-components/src/GridList.tsx
devongovett: could remove if initial loading state was handled in renderEmptyState

devongovett (commented):
I do think we should use the sentinel's specifically for load more, and use `renderEmptyState` for initial loading. That would possibly simplify some of the logic here.

packages/@react-aria/collections/src/Document.ts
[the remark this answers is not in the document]
snowystinger: if the previous sibling doesn't exist, then is index - 1 accurate?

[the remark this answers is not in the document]
LFDanLu: it isn't, in that case we would want the index of the reference node to change by +1 as well as every following node. However, instead of trying to figure out all of that inside here, I think it is more appropriate to make sure that `newNode` will properly be set as the `_minInvalidChildIndex` if it is indeed the "top" most item that will need its index updated which will be checked by the subsequent `invalidateChildIndices` call shortly after this. 

By simply setting index to `referenceNode.index - 1` we ensure that `newNode` will be the `_minInvalidChildIndex` if the `referenceNode` it is being inserted before happens to also be the `_minInvalidChildIndex`. If `_minInvalidChildIndex` is some other node, then that is fine I believe since these are the other possible cases:


-  `_minInvalidChildIndex` is a node that comes after the `referenceNode`: 

In that case the index of `newNode` is already less than the index and will become the new `_minInvalidChildIndex`

- `_minInvalidChildIndex` is a node that comes before the `referenceNode`:

Only the case of the `referenceNode.previousSibling` being the `_minInvalidChildIndex`  is potential interesting since then `referenceNode.previousSibling.index` is equal to `newNode.index`. In that case we are still ok since `invalidateChildIndices` will prefer `referenceNode.previousSibling` over `newNode` as the `_minInvalidChildIndex` which makes sense since this operation is a "insert newNode between referenceNode and its previous sibiling"

packages/@react-spectrum/s2/src/TableView.tsx
```diff
@@ -197,7 +196,9 @@ export class S2TableLayout<T> extends TableLayout<T> {
     let {children, layoutInfo} = body;
     // TableLayout's buildCollection always sets the body width to the max width between the header width, but
     // we want the body to be sticky and only as wide as the table so it is always in view if loading/empty
-    if (children?.length === 0) {
+    // TODO: we may want to adjust RAC layouts to do something simlar? Current users of RAC table will probably run into something similar
```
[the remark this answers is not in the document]
LFDanLu: Is it too opinionated though?

packages/@react-stately/combobox/package.json
```diff
@@ -33,7 +33,8 @@
     "@swc/helpers": "^0.5.0"
   },
   "peerDependencies": {
-    "react": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0 || ^19.0.0-rc.1"
+    "react": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0 || ^19.0.0-rc.1",
+    "react-dom": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0 || ^19.0.0-rc.1"
```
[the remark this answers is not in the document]
LFDanLu: I'm pretty sure lint complained about this and is why I added it, but lemme check again

packages/@react-spectrum/s2/src/TableView.tsx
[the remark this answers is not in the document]
snowystinger: just as a data attribute? or was this about the max width?

packages/@react-stately/layout/src/GridLayout.ts
```diff
@@ -177,6 +186,14 @@ export class GridLayout<T, O extends GridLayoutOptions = GridLayoutOptions> exte
       }
     }
 
+    // Always add the loader sentinel if present in the collection so we can make sure it is never virtualized out.
```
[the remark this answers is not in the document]
LFDanLu: This logic will definitely need to be revisited when trying multiple loader sentinels but adding it here should persist it just fine since it will always be in the collection

packages/@react-stately/select/src/useSelectState.ts
LFDanLu: even if this is just useEffectEvent?

packages/@react-stately/combobox/package.json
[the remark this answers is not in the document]
snowystinger: it's probably because of the dependency you had on react-aria at one point, now that it's removed you can remove this

packages/@react-stately/select/src/useSelectState.ts
snowystinger: it should be moved to react-stately/utils then and re-exported by react-aria/utils
otherwise you'll get other weird dependencies

packages/react-aria-components/src/Table.tsx
```diff
@@ -1378,15 +1397,23 @@ export const UNSTABLE_TableLoadingIndicator = createLeafComponent('loader', func
...
     <>
-      <TR
-        role="row"
-        ref={ref}
-        {...mergeProps(filterDOMProps(props as any), rowProps)}
-        {...renderProps}>
-        <TD role="rowheader" {...rowHeaderProps} style={style}>
-          {renderProps.children}
-        </TD>
+      {/* TODO weird structure? Renders a extra row but we hide via inert so maybe ok? */}
```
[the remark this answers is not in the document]
LFDanLu: we could perhaps skip it via checking some attribute or something right?

packages/@react-spectrum/s2/src/ComboBox.tsx
LFDanLu: I'll defer to @yihuiliao for this one

yarn.lock
```diff
@@ -29,6 +29,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@adobe/css-tools@npm:^4.4.0":
```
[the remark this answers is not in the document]
LFDanLu: yeah, from `@storybook/jest` it seems

packages/react-aria-components/src/ComboBox.tsx
```diff
@@ -103,7 +110,9 @@ function ComboBoxInner<T extends object>({props, collection, comboBoxRef: ref}:
   let {
     name,
     formValue = 'key',
-    allowsCustomValue
+    allowsCustomValue,
+    isLoading,
+    onLoadMore
```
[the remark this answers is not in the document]
LFDanLu: Maybe a old diff? Combobox shouldn't have loading props, they should all be on the sentinel now

yarn.lock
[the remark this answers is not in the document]
snowystinger: haha, even @adobe/css-tools? that's pretty funny

packages/react-aria-components/src/GridList.tsx
LFDanLu: so renderEmptyState does handle initial loading state, this is more because of how the LoadingSentinel collection item is always included:
```
      <GridList
        renderEmptyState={() => renderEmptyState({isLoading: list.isLoading})}>
        <Collection items={list.items}>
          {item => <GridListItem id={item.name}>{item.name}</GridListItem>}
        </Collection>
        <UNSTABLE_GridListLoadingSentinel isLoading={list.loadingState === 'loadingMore'} onLoadMore={list.loadMore} />
      </GridList>
```
It won't render the spinner due to `isLoading={list.loadingState === 'loadingMore'}` but the sentinel will be there to check if it should call onLoadMore. If we want the user to instead conditionally render that sentinel based on if there are items in the list then it might be a bit more complicated on their end

packages/@react-aria/collections/src/Document.ts
LFDanLu: yeah a bit, definitely solves a part of that issue

packages/@react-stately/layout/src/GridLayout.ts
```diff
@@ -120,8 +120,12 @@ export class GridLayout<T, O extends GridLayoutOptions = GridLayoutOptions> exte
     let horizontalSpacing = Math.floor((visibleWidth - numColumns * itemWidth) / (numColumns + 1));
     this.gap = new Size(horizontalSpacing, minSpace.height);
 
-    let rows = Math.ceil(this.virtualizer!.collection.size / numColumns);
-    let iterator = this.virtualizer!.collection[Symbol.iterator]();
+    let collection = this.virtualizer!.collection;
+    // Make sure to set rows to 0 if we performing a first time load or are rendering the empty state so that Virtualizer
+    // won't try to render its body
+    let isEmptyOrLoading = collection?.size === 0 || (collection.size === 1 && collection.getItem(collection.getFirstKey()!)!.type === 'loader');
```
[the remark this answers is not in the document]
LFDanLu: as mentioned above, `renderEmptyState` is used for initial loading but the LoadingSentinel is expected to always be included as a child

packages/@react-stately/layout/src/GridLayout.ts
```diff
@@ -177,6 +186,14 @@ export class GridLayout<T, O extends GridLayoutOptions = GridLayoutOptions> exte
       }
     }
 
+    // Always add the loader sentinel if present in the collection so we can make sure it is never virtualized out.
+    let lastNode = collection.getItem(collection.getLastKey()!);
+    if (lastNode?.type === 'loader') {
+      let rect = new Rect(horizontalSpacing, y, itemWidth, 0);
```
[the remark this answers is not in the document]
LFDanLu: For the CardView we don't have a spinner so I've set it to 0. Can definitely revisit if we expect multi section loading there or for a spinner to be used instead of Skeleton loading but for now I opted for the simpler change

packages/@react-stately/select/src/useSelectState.ts
LFDanLu: the main change here is to prevent the select from opening if it only has the loading sentinel in its collection. Otherwise it was just to make it a bit nicer but I can discard the effect event and revert it to define open/toggle in the return

packages/@react-aria/collections/src/Document.ts
LFDanLu: The todo is part of the comment above, basically just noting that I'm not 100% sure about the change. From what I observed the document never marks itself as dirty when performing `insertBefore` operations, only when calling `appendChild` or if `setFirstChild` gets hit in `insertBefore`. Instead of this change, I could add `  this.ownerDocument.markDirty(this);` to BaseNode's `insertBefore` and `removeChild` but it felt safer to isolate it here to `updateCollection` as a "end step" of sorts

packages/@react-spectrum/s2/src/TableView.tsx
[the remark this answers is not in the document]
LFDanLu: This is about what the calculated virtualizer body width should be if the user's table is either empty or performing initial load. At the moment we make it essentially centered + offset on both sides by 80 pixels for our own table but users may want something different. We could certainly apply a data-attribute to the actual DOM node but this here is the virtualizer layout level calculation which is a bit different and could be overwritten by a user's implementation of TableLayout. However, it might be worthwhile to apply this as a default of sorts if we feel like it is common enough, but I'm not too sure if I'd want to do that

packages/@react-stately/combobox/package.json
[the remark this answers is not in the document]
LFDanLu: hmm, my lint still seems to complain about it even after removing the dependency on react-aria :/

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -384,11 +423,321 @@ export function ComboBoxSection<T extends object>(props: ComboBoxSectionProps<T>
...
+      className={mergeStyles(
+        divider({
+          size: 'M',
+          orientation: 'horizontal',
+          isStaticColor: false
+        }, style({alignSelf: 'center', width: 'full'})))} />
+  );
 }
+
+const Separator = /*#__PURE__*/ createLeafComponent('separator', function Separator(props: SeparatorProps & {size?: 'S' | 'M' | 'L' | 'XL'}, ref: ForwardedRef<HTMLElement>) {
```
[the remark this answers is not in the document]
LFDanLu: deferring to @yihuiliao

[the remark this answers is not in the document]
yihuiliao: updated in the other PR, explanation in this comment https://github.com/adobe/react-spectrum/pull/8110#discussion_r2078312774

packages/@react-spectrum/s2/src/ComboBox.tsx
yihuiliao: actually nvm, we'll need to keep `listboxItem` as well. in `menuitem`, we define `gridTemplateColumns: subgrid` but we can't do that with the virtualized listbox for everything to layout correctly so there are differences between the two

packages/@react-stately/layout/src/GridLayout.ts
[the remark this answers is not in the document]
devongovett: probably will need to be revisited for RAC users too. there might be a spinner there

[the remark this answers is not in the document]
LFDanLu: right good point, I'll add a todo so I don't forget

packages/@react-stately/select/src/useSelectState.ts
```diff
@@ -62,6 +62,7 @@ export function useSelectState<T extends object>(props: SelectStateOptions<T>):
   });
 
   let [isFocused, setFocused] = useState(false);
+  let isEmpty = useMemo(() => listState.collection.size === 0 || (listState.collection.size === 1 && listState.collection.getItem(listState.collection.getFirstKey()!)?.type === 'loader'), [listState.collection]);
```
devongovett, line 65: not great that stately has to know about this... perhaps we can refactor the isEmpty checks to centralize this logic

reidbarber (approved):
LGTM

devongovett (approved):

devongovett (approved):

dannify (approved):

PR #7955, opened 2025-03-18
fix: pass `loadingState` to `useAsyncList`'s `load` function

joshuajaco:
The `loadingState` was typed but never passed


joshuajaco:
It looks like it was removed in https://github.com/adobe/react-spectrum/pull/1829

devongovett (approved):
@LFDanLu I know it was a long time ago, but any memory why we removed it?

joshuajaco:
> devongovett it seems to be because of combobox bugs, specifically when triggering multiple `loadMores` and/or if a `loadMore` was triggered -> canceled via selecting an option -> triggered again when opening the ComboBox.

What does that mean for this PR?
If the intention is to never pass it, it should be removed from the type definition.

LFDanLu (approved):
Seems fine to me, the aforementioned issues I mentioned were specifically around how we used to track/update the loadingState so this change shouldn't be a problem.

PR #7956, opened 2025-03-18
chore: Upgrade Parcel to v2.14.0

devongovett:
2.14 supports server components, which we will need for the new website. https://parceljs.org/blog/v2-14-0/


LFDanLu (approved):
built storybooks and docs seem fine, anything else I should check?

snowystinger (approved):

PR #7961, opened 2025-03-19
fix: exclude isEditable from DateSegment render prop values

Persists:
Found in testing. This Prop is Omited from `DateSegmentRenderProps`, but it was still being passed to the `useRenderProps` hook.

## ✅ Pull Request Checklist:


snowystinger (approved):
Looks like `isEditable` was intentionally excluded, so I *think* this makes sense

not that it really makes a difference, the types would have told someone not to access it regardless

PR #7975, opened 2025-03-21
feat: S2 toast

devongovett:
This implements the Toast component in Spectrum 2, with the new toast stacking animation implemented with CSS view transitions. Style macros don't have an API for view transitions yet, and I haven't come up with a good API for that yet. In the meantime, I used a CSS module file for the toast animations.

A couple RAC Toast changes were needed:

* I split out a new `ToastList` component from `ToastRegion` so that I could add some additional elements adjacent to the list, namely the clear/collapse buttons when the stack is expanded. If you pass a function directly to the `ToastRegion` as before then we'll render a `ToastList` for you. Not sure I'm totally happy with this API.
* I needed to add a hover state to the `ToastRegion` so I could implement an animation.
* I added an `action` parameter to the toast queue subscription function, which tells you what happened (e.g. a toast was added, removed, etc.). This allows us to change the animations accordingly.

Some stuff that's specific to Spectrum:

* To implement clicking on the region to expand it, I attached a click event to the ref for now. Wasn't sure if we wanted to add press events to the region or not since it's kind of an unusual pattern.
* In the expanded state, the whole toast lists acts kind of like a modal. There's a transparent backdrop that covers the rest of the page, we aria hide outside, prevent page scrolling, and contain focus.

### Test instructions

Make sure to test the animations with reduce motion settings enabled as well. In this mode, the toasts fade in/out instead of animating their positions.


packages/@react-spectrum/s2/src/index.ts
```diff
@@ -73,6 +73,7 @@ export {TableView, TableHeader, TableBody, Row, Cell, Column, TableContext} from
 export {Tabs, TabList, Tab, TabPanel, TabsContext} from './Tabs';
 export {TagGroup, Tag, TagGroupContext} from './TagGroup';
 export {TextArea, TextField, TextAreaContext, TextFieldContext} from './TextField';
+export {ToastContainer, ToastQueue} from './Toast';
```
snowystinger, line 76: ToastContainerContext

snowystinger (commented):
Looks really really good, only have a couple comments so far.

When keyboard navigating, it's not always clear that focus has gone to the region. Especially when the region is the same focus scope, just behind all the others. So it looks like there is a bug taking two Tabs to get back into the newest toast after dismissing one. Other times, the region focus ring doesn't show at all.
First one:
```
Open a few toasts
Keyboard navigate to them
Expand all
Esc
Focus is on the region but kind of looks like it's on the top toast, and it takes two tabs to get back inside the toast
```

Second:
```
Open a few toasts
Keyboard navigate to them
Keyboard navigate to the close button
Top toast dismisses
Focus goes to the toast region, but is not visible
```

packages/@react-spectrum/s2/src/Toast.tsx
```diff
@@ -0,0 +1,579 @@
...
+  let queue = getGlobalToastQueue();
+  let key = queue.add(value, {timeout, onClose: options.onClose});
+  return () => queue.close(key);
+}
+
+type CloseFunction = () => void;
+
+const SpectrumToastQueue = {
+  /** Queues a neutral toast. */
+  neutral(children: string, options: ToastOptions = {}): CloseFunction {
```
snowystinger, line 124: We should consider this for S2: [Add support for string formatting in React Spectrum v3 toast](https://github.com/adobe/react-spectrum/issues/6038#top)

packages/@react-spectrum/s2/src/index.ts
devongovett: No, that one is internal.

packages/@react-spectrum/s2/src/Toast.tsx
devongovett: perhaps, but maybe post this release

snowystinger (commented):
Just a heads up that I'm seeing a content jump depending on the toast I open/close. This is present on both my monitors and likely because my scrollbars are always on. I'm not sure why though. I can't find anything about the viewTransition hiding the scrollbars, so I'm not sure why they flicker briefly.
https://github.com/user-attachments/assets/0bd678c3-0788-488c-a9d0-6785f2521be0

packages/react-aria-components/src/Toast.tsx
```diff
@@ -94,12 +104,28 @@ export const ToastRegion = /*#__PURE__*/ (forwardRef as forwardRefType)(function
...
 });
 
-// TODO: possibly export this so additional children can be added to the region, outside the list.
-const ToastList = /*#__PURE__*/ (forwardRef as forwardRefType)(function ToastList<T>(props: ToastRegionProps<T>, ref: ForwardedRef<HTMLOListElement>) {
+export interface ToastListProps<T> extends Omit<ToastRegionProps<T>, 'queue' | 'children'> {
+  /** A function to render each toast. */
+  children: (renderProps: {toast: QueuedToast<T>}) => ReactElement
+}
+
+export const ToastList = /*#__PURE__*/ (forwardRef as forwardRefType)(function ToastList<T>(props: ToastListProps<T>, ref: ForwardedRef<HTMLOListElement>) {
```
snowystinger, line 112: Can we have a test for the newly split out ToastList setup? It could be a chromatic of the s2 Toast, so long as there is something so we know if S2 Toasts stopped working catastrophically. We should get some actual s2 tests as well.

packages/@react-spectrum/s2/src/Toast.tsx
```diff
@@ -0,0 +1,579 @@
...
+          }
+          <Text slot="title">{toast.content.children}</Text>
+        </ToastContent>
+        {!isExpanded && visibleToasts.length > 1 && 
+          <ActionButton
+            isQuiet
+            staticColor="white"
+            styles={style({gridArea: 'expand'})}
+            // Make the chevron line up with the toast text, even though there is padding within the button.
+            UNSAFE_style={{marginInlineStart: variant === 'neutral' ? -10 : 14}}
```
snowystinger, line 545: Roughly lined up? 
Screenshot 2025-04-09 at 4 08 57 pm
Looks farther off on iphone mobile for the Neutral toast

Instead of doing this as an UNSAFE_style, we could wrap the ActionButton in a div, use that for the grid area and the negative margin, it'd then get all the right scaling, and we could do away with the unsafe's?

snowystinger (dismissed):
Nothing I see blocking, the reduced motion looks good. Only after removing one in an expanded state does it still motion animate a bit. But I don't think we can realistically do much about that. If we instantly change the content I think that would be confusing.

snowystinger (approved):

packages/@react-spectrum/s2/src/Toast.tsx
```diff
@@ -0,0 +1,579 @@
...
+}
+
+export interface ToastOptions extends Omit<RACToastOptions, 'priority'>, DOMProps {
+  /** A label for the action button within the toast. */
+  actionLabel?: string,
+  /** Handler that is called when the action button is pressed. */
+  onAction?: () => void,
+  /** Whether the toast should automatically close when an action is performed. */
+  shouldCloseOnAction?: boolean
+}
```
LFDanLu, line 53: Would be nice to mention that the minimum time that the toast stays open is 5 seconds in the prop description, otherwise people might get confused why their provided timeout doesn't seem to work. Would also be nice to tie the storybook onAction to a control but that can be followup

packages/@react-spectrum/s2/src/Toast.module.css
LFDanLu, line 1: I noticed that the view transition prevents subsequent clicks from actually happening while the transition is in effect which seems to be an intentional behavior of view transitions. Do we want to hack around it and keep the page interactive while transitioning or do we deem it a unlikely case that someone might try to click on a different element while a Toast is appearing/exiting?

LFDanLu (approved):
just a few comments from local testing but otherwise seems good enough for an alpha

packages/@react-spectrum/s2/src/Toast.module.css
snowystinger: Yeah, see
https://github.com/adobe/react-spectrum/pull/7838#issuecomment-2686805724

LFDanLu: ah derp, forgot it was mentioned there. Seems we could do something like this though: https://www.bram.us/2025/01/29/view-transitions-page-interactivity/ perhaps?

reidbarber (approved):

packages/react-aria-components/src/Toast.tsx
devongovett: yeah I guess we wanted to get this into a release as unstable for now, and I'll definitely be following up with a bunch more stuff including tests.

packages/@react-spectrum/s2/src/Toast.module.css
devongovett: yeah I'll try that in a followup

PR #7982, opened 2025-03-24
docs: remove BoxAlignmentStyleProps unset "@default" from types

dsmmcken:
`Grid` and `Flex` share the defaults doc strings set for `justifyContent`, `alignContent` and `alignItems`, from `BoxAlignmentStyleProps`. However, no such defaults are actually set by the components, they are just passthroughStyle props for grid and normalize flex-start/start flex-end/end for flex, but neither explicitly set the stated defaults. They resolve in as just `normal` computed styles by the browser.

The doc string is misleading, and should be removed. My guess was the defaults were written as such to reference the fast that a browsers "normal" setting behaves similar to the defaults mentioned, so maybe that's why they were added? Or maybe at one point those were the defaults? However, claiming to set a default in the doc string, when it is not in the component is confusing.

Doc strings shown as default: 
image

Defaults not actually set:
image

PR removes the docstring only, component behaviour is as current (no default).

### DIFF
--- a/packages/@react-types/shared/src/style.d.ts
+++ b/packages/@react-types/shared/src/style.d.ts
@@ -209,20 +209,11 @@ export interface ViewStyleProps<C extends ColorVersion> extends StyleProps {
 }
 
 export interface BoxAlignmentStyleProps {
-  /**
-   * The distribution of space around items along the main axis. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content).
-   * @default 'stretch'
-   */
+  /** The distribution of space around items along the main axis. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content). */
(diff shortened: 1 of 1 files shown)


snowystinger (approved):
This change makes sense.

> My guess was the defaults were written as such to reference the fast that a browsers "normal" setting behaves similar to the defaults mentioned

This is correct, though, as you noted, 'normal' doesn't exactly equal these values and so they are currently a little misleading.

Note, in the next iteration of our React Spectrum library, S2, Flex and Grid won't exist anymore. As knowledge and comfort using flex and grid has grown, these have become unnecessary. Access to variables to set spacing, etc, will be useable through our style macros theme.
https://react-spectrum.adobe.com/s2/index.html?path=/docs/migrating--docs#flex

devongovett (approved):

PR #8003, opened 2025-03-28
fix: S2 notification badge follow-ups

yihuiliao:
notes from design:

1. Notification badge cannot be used on a label-only quiet button. they are updating their guidelines to include this information
2. Remove the badge when an action button is disabled
3. No answers for static colors, they will workshop
4. Badges should be swapped with an indicator to avoid visual collision but they are exploring other options (probably something the app needs to do)

## ✅ Pull Request Checklist:

### DIFF
--- a/packages/@react-spectrum/s2/src/NotificationBadge.tsx
+++ b/packages/@react-spectrum/s2/src/NotificationBadge.tsx
@@ -39,10 +39,17 @@ export interface NotificationBadgeProps extends DOMProps, AriaLabelingProps, Sty
   value?: number | null
 }
 
-export const NotificationBadgeContext = createContext<ContextValue<Partial<NotificationBadgeProps>, DOMRefValue<HTMLDivElement>>>(null);
+interface NotificationBadgeContextProps extends Partial<NotificationBadgeProps> {
+  isDisabled?: boolean
+}
+
+export const NotificationBadgeContext = createContext<ContextValue<Partial<NotificationBadgeContextProps>, DOMRefValue<HTMLDivElement>>>(null);
 
 const badge = style({
-  display: 'flex',
+  display: {
+    default: 'flex',
+    isDisabled: 'none'
+  },
   font: 'control',
   color: {
     default: 'white',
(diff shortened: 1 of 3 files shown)


packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
```diff
@@ -224,16 +224,18 @@ NotificationBadges.parameters = {
...
+        return `
+let badgeValue = 10;
+let formattedValue = useNumberFormatter().format(badgeValue);
+
+return (
+  <div style={{display: 'flex', gap: 8, padding: 8, justifyContent: 'center'}}>
+    <ActionButton aria-label="Messages has new activity"><CommentIcon /><NotificationBadge /></ActionButton>
+    <ActionButton aria-label={\`\${formattedValue} notifications\`} ><BellIcon /><NotificationBadge value={badgeValue} /></ActionButton>
+    <ActionButton><CommentIcon /><Text>Messages</Text><NotificationBadge value={5} /></ActionButton>
+    {/* cannot have an isQuiet label-only Action Button with a Notification Badge */}
```
yihuiliao, line 236: i added a comment here about the label-only quiet action button but i can also make it so that the story example doesn't change to isQuiet when it is label-only

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -39,10 +39,17 @@ export interface NotificationBadgeProps extends DOMProps, AriaLabelingProps, Sty
...
+  isDisabled?: boolean
+}
+
+export const NotificationBadgeContext = createContext<ContextValue<Partial<NotificationBadgeContextProps>, DOMRefValue<HTMLDivElement>>>(null);
 
 const badge = style({
-  display: 'flex',
+  display: {
+    default: 'flex',
+    isDisabled: 'none'
```
devongovett, line 51: Do we need `isDisabled` as a prop on a standalone notification badge, or could we just set `display: none` in ActionButton in the `styles`?

packages/@react-spectrum/s2/stories/ActionButton.stories.tsx
devongovett: or hide that button when isQuiet?

packages/@react-spectrum/s2/src/NotificationBadge.tsx
yihuiliao: well...`display` isn't part of StyleProps (or I don't think it is?). we could also opt for UNSAFE_style

devongovett: missed that this was only on the context.

devongovett (approved):

snowystinger (approved):

PR #8055, opened 2025-04-08
fix: add static color to s2 notification badge

yihuiliao:
Closes 

## ✅ Pull Request Checklist:


reidbarber (approved):
LGTM

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -40,7 +40,8 @@ export interface NotificationBadgeProps extends DOMProps, AriaLabelingProps, Sty
 }
 
 interface NotificationBadgeContextProps extends Partial<NotificationBadgeProps> {
-  isDisabled?: boolean
+  isDisabled?: boolean,
+  staticColor?: 'black' | 'white' | 'auto'
```
LFDanLu, line 44: Since in practice the NotificationBadge either has `color: auto` applied or not, I wonder whether the prop set here should accept `black`/`white` or if it should just be a boolean. Maybe it could always be auto (but would need to assume it would always get its color from a parent)?

LFDanLu (commented):
LGTM, verified the static color behavior of the NotificationBadge in the ActionButton stories. Just one API question, mainly thinking about the standalone NotificationBadge case which I assume the MenuItem case will run into

LFDanLu (approved):

packages/@react-spectrum/s2/src/NotificationBadge.tsx
```diff
@@ -102,7 +106,7 @@ const badge = style({
     isIndicatorOnly: 'square',
     isSingleDigit: 'square'
   },
-  width: 'fit',
+  width: 'max',
```
yihuiliao, line 109: noticed that having width set to `fit` caused the text to wrap in locales like `'zh-CN-u-nu-hanidec'` so i changed it

snowystinger (approved):

LFDanLu (approved):

PR #8087, opened 2025-04-12
update optimize-locales plugin type

stefanprobst:
this updates the optimize-locales plugin type:

the second generic in `UnpluginInstance` is for nested plugins, see https://github.com/unjs/unplugin/blob/466f4d5be7ba048d81613d853c34d7eea2518507/src/types.ts#L162

setting to `false` instead of `boolean` means that the optimize-locales plugin will be typed as just `Plugin` instead of `Array<Plugin> | Plugin>`.

### DIFF
--- a/packages/dev/optimize-locales-plugin/LocalesPlugin.d.ts
+++ b/packages/dev/optimize-locales-plugin/LocalesPlugin.d.ts
@@ -4,5 +4,5 @@ type Options = {
   locales: readonly string[]
 };
 
-declare const plugin: UnpluginInstance<Options, boolean>;
+declare const plugin: UnpluginInstance<Options, false>;
 export = plugin;


snowystinger:
Thanks for the PR, it's interesting, I can't seem to reproduce the error here https://github.com/adobe/react-spectrum/blob/main/examples/remix/vite.config.ts though the change makes sense.
Do you have a reproduction handy?

I copied our remix example out to a different directory so the workspaces wouldn't exist and built it from there. Found a couple updates that needed to be made https://github.com/adobe/react-spectrum/pull/8090 and removed the workspaces/resolutions entry in the package.json. But no error in my vite.config.ts like this one.

stefanprobst:
i ran into this trying our tanstack/start - tsc seems to be fine with it but i did get this eslint error:

eslint-rac

here's a repo where you can see the lint error: https://github.com/stefanprobst/issue-rac-optimize-locales-types/blob/main/app.config.ts

snowystinger (approved):
o, because technically you can do it, just some eslint rule thinks it's better if you don't

haha, ok, approved, for now you can ts-expect-error

reidbarber (approved):
Thanks!

PR #8110, opened 2025-04-17
feat: add virtualization to S2 combobox and picker

yihuiliao:
Closes 

## ✅ Pull Request Checklist:


packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -147,6 +148,130 @@ const iconStyles = style({
...
+  gridColumnEnd: -2
+});
+
+// not sure why edgeToText won't work...
+const separatorWrapper = style({
+  display: {
+    ':is(:last-child > &)': 'none',
+    default: 'flex'
+  },
+  // marginX: {
```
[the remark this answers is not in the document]
yihuiliao: alternatively, i could do something like this which seems to work and wouldn't define the space using pixels:

```
  marginX: {
    size: {
      S: '[calc(var(--marginSpace) * 3 / 8)]',
      M: '[calc(var(--marginSpace) * 3 / 8)]',
      L: '[calc(var(--marginSpace) * 3 / 8)]',
      XL: '[calc(var(--marginSpace) * 3 / 8)]'
    }
  },
  '--marginSpace': {
    type: 'marginStart',
    value: {
      size: {
        S: 24,
        M: 32,
        L: 40,
        XL: 48
      }
    }
```

packages/@react-spectrum/s2/stories/Picker.stories.tsx
```diff
@@ -131,6 +131,28 @@ export const WithIcons: Story = {
...
+  }
+
+  return (
+    <Picker {...props} items={items}>
+      {(item) => <PickerItem id={(item as IExampleItem).id} textValue={(item as IExampleItem).label}>{(item as IExampleItem).label}</PickerItem>}
+    </Picker>
+  );
+}
+
+export const ManyItems: Story = {
```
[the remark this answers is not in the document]
LFDanLu: For posterity, this was due to how the render function of the Picker had a button that would rerender on every press start/end. Fixed by moving the button to its own components so only the button would rerender on press and not the entire render function of the Picker, see https://github.com/adobe/react-spectrum/pull/8110/commits/77b4b62ff441c077ef8ab6610759d77d09fca16f

packages/@react-spectrum/s2/package.json
```diff
@@ -136,6 +136,7 @@
     "@react-aria/interactions": "^3.25.0",
     "@react-aria/live-announcer": "^3.4.2",
     "@react-aria/overlays": "^3.27.0",
+    "@react-aria/separator": "^3.4.8",
```
LFDanLu, line 139: don't think we need this anymore?

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -148,6 +149,114 @@ const iconStyles = style({
...
+  ...focusRing(),
+  boxSizing: 'border-box',
+  borderRadius: 'control',
+  font: 'control',
+  '--labelPadding': {
+    type: 'paddingTop',
+    value: centerPadding()
+  },
+  paddingBottom: '--labelPadding',
+  backgroundColor: { // TODO: revisit color when I have access to dev mode again
```
LFDanLu, line 173: need help checking this?

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -274,20 +278,7 @@ export const Picker = /*#__PURE__*/ (forwardRef as forwardRefType)(function Pick
...
-  let onPressStart = (e: PressEvent) => {
-    if (e.pointerType !== 'mouse') {
-      return;
-    }
-    setPressed(true);
-    addGlobalListener(document, 'pointerup', () => {
-      setPressed(false);
-    }, {once: true, capture: true});
-  };
+  let layout = new ListLayout({estimatedRowHeight: 32, estimatedHeadingHeight: 50, padding: 8});
```
LFDanLu, line 281: Interesting that this had to be pulled out of the Virtualizer `layout` and `layoutOptions` since now this means we generate a new ListLayout every render. I assume this was to fix the scroll into view behavior?

packages/@react-spectrum/s2/src/ComboBox.tsx
yihuiliao: oh yeah that's probably from copy paste. it's been there prior to starting this PR but i can look into it

packages/@react-spectrum/s2/src/Picker.tsx
yihuiliao: Yeah, this was to fix the scroll into view. Without the change, the view would only scroll when we were closing the popover which isn't very useful.  

It seemed like what was happening prior to this change was that when we initially open the picker without any selected item, we call `layoutIfNeeded` as we scroll to select an item which sets `this.requestedRect`. We want to preserve that value even when we close the popover but was happening was that we were basically resetting this value every time we closed the popover so this.requestedRect didn't have a size (everything was just zero). Then when called buildCollection, at [this line](https://github.com/adobe/react-spectrum/blob/d1c558744d14a608a9b232a2f472fbb411ebabf7/packages/%40react-stately/layout/src/ListLayout.ts#L272-L275) we would exit the loop early so we only had one node when we called [getVisibleLayoutInfos](https://github.com/adobe/react-spectrum/blob/d1c558744d14a608a9b232a2f472fbb411ebabf7/packages/%40react-stately/layout/src/ListLayout.ts#L150)

Might have some gaps in my understanding but that's what I found

packages/@react-spectrum/s2/src/ComboBox.tsx
yihuiliao: took a look at the figma and it seems correct. gonna remove the comment

packages/@react-spectrum/autocomplete/test/SearchAutocomplete.test.js
```diff
@@ -143,6 +143,7 @@ describe('SearchAutocomplete', function () {
     user = userEvent.setup({delay: null, pointerMap});
     jest.spyOn(window.HTMLElement.prototype, 'clientWidth', 'get').mockImplementation(() => 1000);
     jest.spyOn(window.HTMLElement.prototype, 'clientHeight', 'get').mockImplementation(() => 1000);
+    jest.spyOn(window.HTMLElement.prototype, 'scrollHeight', 'get').mockImplementation(() => 50);
```
snowystinger, line 146: is this going to be problematic for users of v3 in their own tests?
will we need to call out in release?

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -148,6 +149,114 @@ const iconStyles = style({
   }
 });
 
+export let listbox = style<{size: 'S' | 'M' | 'L' | 'XL'}>({
+  width: 'full',
+  boxSizing: 'border-box',
+  maxHeight: '[inherit]',
+  // TODO: Might help with horizontal scrolling happening on Windows, will need to check somehow. Otherwise, revert back to overflow: auto
```
snowystinger, line 156: what is the todo here? was something else being tried?

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -148,6 +149,114 @@ const iconStyles = style({
...
+  },
+  color: {
+    default: 'neutral',
+    isDisabled: {
+      default: 'disabled',
+      forcedColors: 'GrayText'
+    }
+  },
+  position: 'relative',
+  // each menu item should take up the entire width, the subgrid will handle within the item
```
snowystinger, line 185: no longer relevant comment?

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -148,6 +149,114 @@ const iconStyles = style({
...
+  gridColumnStart: 1,
+  gridColumnEnd: -1,
+  display: 'grid',
+  gridTemplateAreas: [
+    '. checkmark icon label  .',
+    '.      .      .  description .'
```
snowystinger, line 191:
```suggestion
    '. checkmark icon label       .',
    '. .         .    description .'
```

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -148,6 +149,114 @@ const iconStyles = style({
...
+      S: [edgeToText(24), 'auto', 'auto', 'minmax(0, 1fr)', edgeToText(24)],
+      M: [edgeToText(32), 'auto', 'auto', 'minmax(0, 1fr)', edgeToText(32)],
+      L: [edgeToText(40), 'auto', 'auto', 'minmax(0, 1fr)', edgeToText(40)],
+      XL: [edgeToText(48), 'auto', 'auto', 'minmax(0, 1fr)', edgeToText(48)]
+    }
+  },
+  gridTemplateRows: {
+    // min-content prevents second row from 'auto'ing to a size larger then 0 when empty
+    default: 'auto minmax(0, min-content)',
+    ':has([slot=description])': 'auto auto'
```
snowystinger, line 204: pull out into constant? seems to be the common pattern we're adopting

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -148,6 +149,114 @@ const iconStyles = style({
...
+});
+
+export let listboxHeading = style({
+  fontSize: 'ui',
+  fontWeight: 'bold',
+  lineHeight: 'ui',
+  margin: 0
+});
+
+// not sure why edgeToText won't work...
```
snowystinger, line 242: different message than what's on line 248?

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -148,6 +149,114 @@ const iconStyles = style({
...
+const separatorWrapper = style({
+  display: {
+    ':is(:last-child > *)': 'none',
+    default: 'flex'
+  },
+  // A workaround since edgeToText() returns undefined for some reason
```
snowystinger, line 248: you seem to be using the function now? dead comment?

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -383,11 +515,52 @@ export function ComboBoxSection<T extends object>(props: ComboBoxSectionProps<T>
...
+      className={mergeStyles(
+        divider({
+          size: 'M',
+          orientation: 'horizontal',
+          isStaticColor: false
+        }, style({alignSelf: 'center', width: 'full'})))} />
+  );
+}
+
+const Separator = /*#__PURE__*/ createLeafComponent('separator', function Separator(props: SeparatorProps & {size?: 'S' | 'M' | 'L' | 'XL'}, ref: ForwardedRef<HTMLElement>) {
```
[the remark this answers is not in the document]
snowystinger: What is the reason we can't use the S2 Divider which already renders the RAC Separator which is a LeafComponent?
Is it just the extra size requirement?

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -384,11 +500,52 @@ export function ComboBoxSection<T extends object>(props: ComboBoxSectionProps<T>
...
     </>
   );
 }
+
+export function Divider(props: SeparatorProps & {size?: 'S' | 'M' | 'L' | 'XL' | undefined}): ReactNode {
+  return (
+    <Separator
+      {...props}
+      className={mergeStyles(
+        divider({
```
snowystinger, line 516: `divider` only supports `'S' | 'M' | 'L'`, if XL is sent in, it'll treat it like M

packages/@react-spectrum/s2/src/Picker.tsx
snowystinger: Can we memo the ListLayout? looks like it doesn't have any dependencies

snowystinger: hmm this also goes against our common pattern for Virtualizer now
Could we instead move the Virtualizer up to wrap the PopoverBase?

packages/@react-spectrum/s2/src/TableView.tsx
```diff
@@ -194,6 +194,9 @@ export class S2TableLayout<T> extends TableLayout<T> {
 
   protected buildCollection(): LayoutNode[] {
     let [header, body] = super.buildCollection();
+    if (!header) {
```
snowystinger, line 197: @LFDanLu can you explain when this happens?

LFDanLu: If I remember correctly, now that we are rendering the contents of the virtualized content even when we haven't fully processed the collection (aka the changes in Virtualizer and rect size check changes) there is a moment where  we haven't processed the collection and thus can't perform build the header/body yet

packages/@react-spectrum/autocomplete/test/SearchAutocomplete.test.js
yihuiliao: Probably something we'd like to call out but we also we have this section in are docs where we do mock the scrollHeight for virtualized components

packages/@react-spectrum/s2/src/ComboBox.tsx
yihuiliao: It was something that I thought might fix the issue with the horizontal scrollbar but I'm unable to see if it actually works since I can't reproduce the original issue

packages/@react-spectrum/s2/src/ComboBox.tsx
yihuiliao: yeah this is just leftover that i forgot to delete

packages/@react-spectrum/s2/src/ComboBox.tsx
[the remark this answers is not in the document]
yihuiliao: No, it was that it wasn't measuring the height correctly. We had some additional padding on the divider that wasn't being accounted for when measuring the layout. By copying over the S2 Divider, wrapping it in a div, we could account for the height correctly. The reason I had to copy it over was because if I just wrapped the S2 Divider in a div, it would throw an error. I had to do it inside the `createLeafComponent`. 

I tried updating the ListLayout to have a separatorHeight but the reason that didn't work out was because even if we had display: 'none' set on the last separator, the collection would still account for the separatorHeight so there was additional empty space on the bottom that shouldn't have been there.  In that case, we would need it to be an estimatedSeparatorHeight and involve slightly different logic similar to buildNode but I didn't test that out. 

Overall, we decided that it was probably not worth our time spending too much time on just a separator so this is what we came up with.

packages/@react-spectrum/s2/src/ComboBox.tsx
yihuiliao: Hm...actually looking back on it, we explicitly set the size to always be 'M' regardless of the ComboBox or  Picker's size so I'll update it to follow that instead

yihuiliao: Hm...actually looking back on it, we never passed down the size prop to the Divider so it always defaulted to 'M'. In that case, I'll just update it so that it's always M regardless of the ComboBox or Picker's size.

yihuiliao: I think it should be fine? Since we are passing the size as 'M', it should always use the 'M' styles regardless of ComboBox or Picker

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -148,6 +149,114 @@ const iconStyles = style({
   }
 });
 
+export let listbox = style<{size: 'S' | 'M' | 'L' | 'XL'}>({
```
devongovett, line 152: What are the differences between the styles here and the ones in menu? A bit unfortunate to duplicate them

packages/@react-spectrum/s2/src/ComboBox.tsx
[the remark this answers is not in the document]
devongovett: what was the reason for the extra div? where was the extra padding coming from?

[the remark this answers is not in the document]
devongovett: actually I think ListBox doesn't even allow separators. aXe currently fails on S2 prod storybook. In v3 we overrode the role to presentation, but RAC doesn't support that. Any reason it needs to be a separate element at all? Can it just be a border-bottom on the section?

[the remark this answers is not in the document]
yihuiliao: Or actually it was margins, not padding. It was coming from [here](https://github.com/adobe/react-spectrum/blob/0548d304a968078fd5aec2bff71a6563ffe216dc/packages/%40react-spectrum/s2/src/Menu.tsx#L419)

We could potentially use border-bottom instead, but in order to align the separator with the margins we have on the menu items + heading on the left + right, we'd probably need something like border-image. For example, if we added a border (this just added it to the top, very basic, no styling, for demonstration purposes), understandably, the length of it is the width of the section. However, the separator has a length that is less than the width of the section. Right now, I'm able to account for that by adding a margin with edgeToText(). I suppose we could just calculate it with a percentage tho

Screenshot 2025-05-07 at 4 58 35 PM

packages/@react-spectrum/s2/src/ComboBox.tsx
yihuiliao: Mainly that menu has some additional styles that Combobox and Picker didn't need but it wasn't actually affecting anything. I could probably revert the change and just use menu but I'd need to double check

packages/@react-spectrum/s2/src/ComboBox.tsx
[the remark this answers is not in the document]
devongovett: Since it doesn't need to be an actual divider, can we make a stripped down version like this?

```tsx
const dividerStyle = style({
  backgroundColor: {
    default: 'gray-200',
    forcedColors: 'ButtonBorder'
  },
  borderRadius: 'full',
  height: '[2px]',
  width: 'full'
});

export const Divider = /*#__PURE__*/ createLeafComponent('separator', function Divider({size}: {size?: 'S' | 'M' | 'L' | 'XL'}, ref: ForwardedRef<HTMLDivElement>) {
  return (
    <div className={separatorWrapper({size})}>
      <div ref={ref} className={dividerStyle} />
    </div>
  );
});
```

needs `alignItems: 'center'` in `separatorWrapper` too

LFDanLu:
@yihuiliao digging right now, but just in case you might have an idea: https://www.chromatic.com/test?appId=5f0dd5ad2b5fc10022a2e320&id=681bf2b5cef784c52dc39ae5. Noticed that the section divider after the last item here in my PR, doesn't seem to happen in this PR though. I remember that there were discussions about using css to detect if the divider should render after the item, I suppose I would need to handle the loader sentinel...

packages/@react-spectrum/autocomplete/test/SearchAutocomplete.test.js
LFDanLu: just my 2 cents, would be good to call out in the release notes. Just spent some time trying to figure out why one of my Picker tests were failing and it turned out it also needed this scrollHeight mock lol

packages/@react-spectrum/s2/src/ComboBox.tsx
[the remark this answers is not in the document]
yihuiliao: yep, happy to simplify it

packages/@react-aria/selection/src/useSelectableCollection.ts
```diff
@@ -516,8 +516,11 @@ export function useSelectableCollection(options: AriaSelectableCollectionOptions
       }
 
       if (modality === 'keyboard' || didAutoFocusRef.current) {
-        scrollIntoView(scrollRef.current, element);
-
+        requestAnimationFrame(() => {
```
LFDanLu, line 519: should this wrap the below scrollIntoViewport as well? Not sure if the order is going to matter. Might be a good idea to clean up the raf as well

packages/@react-spectrum/s2/src/ComboBox.tsx
LFDanLu: I'll make a note to test this tmrw on my windows machine

packages/@react-spectrum/s2/src/Picker.tsx
```diff
@@ -480,11 +519,10 @@ export function PickerSection<T extends object>(props: PickerSectionProps<T>): R
   return (
     <>
       <AriaListBoxSection
-        {...props}
-        className={section({size})}>
```
LFDanLu, line 484: We don't need the section styles anymore?

yihuiliao: no we don't need those styles anymore. it took advantage of grid layouts which we couldn't take advantage of with the way virtualizer works

packages/@react-aria/selection/src/useSelectableCollection.ts
yihuiliao: maybe it would be good to test it without and see if it needs it?

yihuiliao:
> @yihuiliao digging right now, but just in case you might have an idea: https://www.chromatic.com/test?appId=5f0dd5ad2b5fc10022a2e320&id=681bf2b5cef784c52dc39ae5. Noticed that the section divider after the last item here in my PR, doesn't seem to happen in this PR though. I remember that there were discussions about using css to detect if the divider should render after the item, I suppose I would need to handle the loader sentinel...

ah right. currently this is handled like this 
```
const separatorWrapper = style({
  display: {
    ':is(:last-child > *)': 'none',
    default: 'flex'
  },
});
```

and this works because we are looking for the last child. maybe we can change it to look for the 2nd to last child (n-1?) as long as the loading indicator is always rendered as the last child regardless. otherwise, might need to add some additional logic

packages/@react-spectrum/s2/src/ComboBox.tsx
```diff
@@ -146,6 +143,111 @@ const iconStyles = style({
...
+  gridTemplateRows: {
+    // min-content prevents second row from 'auto'ing to a size larger then 0 when empty
+    default: 'auto minmax(0, min-content)',
+    ':has([slot=description])': 'auto auto'
+  },
+  rowGap: {
+    ':has([slot=description])': space(1)
+  },
+  alignItems: 'baseline',
+  minHeight: controlSize(),
```
devongovett, line 199: fwiw, I think `control` sets alignItems and minHeight

devongovett (approved):

packages/@react-spectrum/s2/src/ComboBox.tsx
LFDanLu: I'm unable to reproduce the original issue as well on Chrome/FF, probably ok for now then

LFDanLu (approved):

packages/@react-aria/selection/src/useSelectableCollection.ts
LFDanLu: did some quick testing and haven't found any weirdness so ok for now. We can test more during the testing session

PR #8155, opened 2025-04-28
refactor: Theme updates

devongovett:
* Reduced the default spacing scale so it only goes up to 96px. Larger values are unlikely.
* Switched to rems for media queries. Since component sizes scale with rems, breakpoints need to match.
* Switch to px instead of rems for padding and absolute positioning. This avoids adding extra whitespace that causes additional text wrapping.
* Used rems and touch scaling for icon sizes, and added t-shirt `size` prop
* Used rems for checkbox and radio borders, even though normally borders don't scale. These feel more like icons that happen to be implemented with borders.
* Added support for percentages and viewport relative units (e.g. `vw`) without square bracket syntax
* Added support for arbitrary aspect ratio values without bracket syntax
* Added support for `calc` and other math functions without bracket syntax
* Added support for css-wide keywords like `inherit` without bracket syntax


packages/@react-spectrum/s2/style/spectrum-theme.ts
```diff
@@ -74,41 +74,138 @@ const color = {
...
+    };
+  }
+
+  if (token.type === 'color') {
+    return token;
+  }
+
+  let lightToken = baseColors[token.light];
+  if (!lightToken) {
+    throw new Error(`${token.light} is not a valid color reference`);
```
snowystinger, line 93:
```
throw new Error(`${token.light} is not a valid light color reference`);
```
might be hard to know which one is failing if there isn't 1:1 parity

packages/@react-spectrum/s2/style/spectrum-theme.ts
```diff
@@ -74,41 +74,138 @@ const color = {
...
+  }
+  let darkToken = baseColors[token.dark];
+  if (!darkToken) {
+    throw new Error(`${token.dark} is not a valid color reference`);
```
snowystinger, line 97:
```
throw new Error(`${token.dark} is not a valid dark color reference`);
```

packages/@react-spectrum/s2/style/spectrum-theme.ts
```diff
@@ -74,41 +74,138 @@ const color = {
...
+
+  let index = colorList.indexOf(name);
   if (index === -1) {
-    throw new Error('Invalid base color ' + base);
+    throw new Error(`${name} does not support states`);
+  }
+
+  let key = colorList[index + 1];
+  if (key.split('-')[0] !== name.split('-')[0]) {
+    throw new Error(`${name} does not support states`);
```
snowystinger, line 139: difference between this error and the previous one? is the first one actually that the color doesn't exist and this one is that there isn't a state?

packages/@react-spectrum/s2/src/Checkbox.tsx
```diff
@@ -66,8 +66,8 @@ const wrapper = style({
 
 export const box = style<RenderProps>({
   ...focusRing(),
-  size: 'control-sm',
-  borderRadius: 'control-sm',
+  ...controlBorderRadius('sm'),
+  size: controlSize('sm'),
```
snowystinger, line 70: a little confused so far on when it's a spreadable property vs an individual property, the name looks like they should both be one or the other

packages/@react-spectrum/s2/src/Dialog.tsx
```diff
@@ -130,14 +130,14 @@ export const Dialog = forwardRef(function Dialog(props: DialogProps, ref: DOMRef
                 display: 'flex',
                 alignItems: 'start',
                 columnGap: 12,
-                marginStart: {
+                paddingStart: {
```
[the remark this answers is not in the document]
snowystinger: Looks like dialog broke compared to main, I can't see the Content anymore in a small screen. Probably missing a flex: 0, I think we discussed a couple weeks ago
Screenshot 2025-05-07 at 1 37 45 pm

Large fonts look good though 👍🏻

snowystinger (approved):
Did a quick sweep of components on chrome/ff and with chrome font size changes. I did find one change on NumberField in the Chromatic you ran that I'm not sure about. Otherwise, lets get this in for testing

reidbarber (approved):

packages/@react-spectrum/s2/style/properties.json
```diff
@@ -907,73 +466,47 @@
...
+    "fontWeight": {
+      "300": "e",
+      "400": "a",
+      "500": "d",
+      "700": "b",
+      "800": "c"
+    },
+    "columnGap": {
+      "0.75rem": "a",
+      "0.42857142857142855em": "b",
```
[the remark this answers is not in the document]
devongovett: These are the fontRelative values, e.g. `edge-to-text`.

packages/@react-spectrum/s2/src/AvatarGroup.tsx
```diff
@@ -39,7 +39,7 @@ export const AvatarGroupContext = createContext<ContextValue<Partial<AvatarGroup
 
 const avatar = style({
   marginStart: {
-    default: '[calc(var(--size) / -4)]',
+    default: 'calc(var(--size) / -4)',
```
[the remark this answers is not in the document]
devongovett: but you can't put anything inside an Avatar so it doesn't matter.

packages/@react-spectrum/s2/style/spectrum-theme.ts
devongovett: If this throws it's a bug in the Spectrum tokens. It won't be caused by user error.

packages/@react-spectrum/s2/style/spectrum-theme.ts
devongovett: e.g. if you have `gray-1000`, you can't apply a hover state to that because `gray-1100` does not exist. The next color after it would be in a different scale (e.g. `blue-100` or something). So this checks if they are in the same scale.

packages/@react-spectrum/s2/src/Checkbox.tsx
devongovett: just the return type of the function. If it returns multiple properties it has to be spread.

packages/@react-spectrum/s2/src/Dialog.tsx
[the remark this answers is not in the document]
devongovett: I'll fix this in a separate PR.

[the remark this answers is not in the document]
devongovett: I can't reproduce this. Where did you see this?

